PR #171, opened 2022-07-11
Changes to eth-bridge-integration branch

james-chf:
* Add `BlockHeight` to `MultiSignedEthEvent` as in https://github.com/anoma/specs/commit/2198767818a56bff6632b4df3d47f2af3a358a99
* resync `Cargo.lock` so that only dependencies related to changes in `eth-bridge-integration` branch are updated

### DIFF
--- a/wasm/tx_template/Cargo.lock
+++ b/wasm/tx_template/Cargo.lock
@@ -59,7 +59,7 @@ dependencies = [
  "ics23",
  "itertools",
  "loupe",
- "num-rational",
+ "num",
  "parity-wasm",
  "proptest",
  "prost",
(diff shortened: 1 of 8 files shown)


wasm/tx_template/Cargo.lock
```diff
@@ -343,9 +343,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitvec"
-version = "1.0.0"
+version = "1.0.1"
```
james-chf, line 346: When I resynced to match `namada` branch's dependencies, it also updated any dependencies which are only in `eth-bridge-integration` (eg this `bitvec` dependency, haven't seen any others)

batconjurer (approved):

PR #177, opened 2022-07-11
Add ethereum events to vote extensions

batconjurer:
This PR adds the following:

 * Add ethereum events to vote extensions that are sent to the ledger from the ethereum oracle
 * Vereifies these vote extensions
 
 This PR does not address:
 * Including the vote extensions in the next proposed block as part of `PrepareProposal`
 * Verify the included vote extensions in `ProcessProposal`
 * The data in the vote extensions are not persisted in any way and make no state changes.


apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,460 @@
...
+                        FractionalVotingPower::new(power, total_voting_power)
+                            .unwrap();
+                    if power == voting_power {
+                        public_keys.push(pk);
+                    }
+                }
+            }
+            // check that we found all the public keys and
+            // check that the signatures are valid
+            public_keys.len() == event.number_of_signers()
```
james-chf, line 137: if we don't find all public keys, this is the validator's error so we should return an error rather than reject the event as invalid. Even if functionally the behaviour will be the same, it'd be clearer in the code I think

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,460 @@
...
+            // inaccurately stated their voting power at a given block height
+            let mut public_keys = vec![];
+            for EpochPower {
+                validator,
+                voting_power,
+                block_height,
+            } in event.get_voting_powers().into_iter()
+            {
+                if block_height != height {
+                    continue;
```
james-chf, line 122: when would we reach this code path, is this unusual? we could log (info or warning?) if we reach here

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,460 @@
...
+                }
+                if let Some((power, pk)) =
+                    self.get_validator_from_address(&validator, epoch)
```
james-chf, line 125: similary - if `self.get_validator_from_address` is `None`, we could handle that case and log, at a glance I'm not sure when that would happen

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,460 @@
...
+                {
+                    let power =
+                        FractionalVotingPower::new(power, total_voting_power)
+                            .unwrap();
+                    if power == voting_power {
```
james-chf, line 130: should we log a warning if `power != voting_power`? could that mean a validator tried to lie about their voting power and we caught it?

james-chf (commented):
I think it'd be clearer to return `Result`s rather than `Option`s for our fns in `queries.rs` (not just the ones in this PR), handle the unhappy paths explicitly with an error message rather than a `None`. Then log the errors where appropriate. In general, it'd be good to add some logging as well (at `tracing::debug!` maybe), perhaps we could do some of these changes in a later PR though.

apps/src/lib/node/ledger/shell/vote_extensions.rs
batconjurer: I have no idea if it is unusual. Depends on how many times we get replay attacks (recycling old vote extensions).

batconjurer: I guess, in the future we may want to use this to detect a slashable offense.

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -299,16 +301,17 @@ where
...
             .expect("Serializing public key should not fail");
         // get the current epoch
-        let (current_epoch, _) = self.storage.get_current_epoch();
+        let epoch = epoch.unwrap_or_else(|| self.storage.get_current_epoch().0);
         // get the active validator set
         self.storage
             .read_validator_set()
-            .get(current_epoch)
+            .get(epoch)
             .expect("Validators for the next epoch should be known")
```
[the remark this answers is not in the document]
james-chf: I have it in commit 7d69aaf67f9521fe565304172ed16ec39533948b but this comment might have got out of date

apps/src/lib/node/ledger/shell/vote_extensions.rs
batconjurer: This method will actually be called by the validator proposing the block to make sure they got everything right as well as others checking the proposer's work. This is also used to weed out malformed vote extensions as a result. I could put an error here, but it would require enlarging the error enum and I would never print the message out anyway.

apps/src/lib/node/ledger/shell/vote_extensions.rs
batconjurer: Only if the signature is valid. Otherwise, it could be someone impersonating that validator

james-chf: yeah I see we could extend this method in the future so that we can detect slashable behaviour, if we log for now in the negative spaces we could do this in some later PR down the line

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,460 @@
...
+                    response::VerifyVoteExtension {
+                        status: VerifyStatus::Accept.into(),
+                    }
+                } else {
+                    response::VerifyVoteExtension {
+                        status: VerifyStatus::Reject.into(),
+                    }
+                };
+            }
+            Default::default()
```
james-chf, line 60: Been looking more at how vote extensions work - the `Default::default()` here will return [`VerifyStatus::Unknown`](https://docs.tendermint.com/master/spec/abci++/abci++_methods.html#verifystatus) which will panic Tendermint (we can hit this code path if a validator's vote extension doesn't serialize to a `VoteExtension { ethereum_events }`). We should explicitly accept or reject in the case where that happens.

I think we should reject the validator's bad vote now although I'm not 100% on that.

Although Tendermint docs says we SHOULD accept invalid vote extensions at this stage and deal with them in our logic later, we could do this change later on (accept syntactically invalid vote extensions and the block proposer can flag it as invalid in `PrepareProposal`).

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -341,5 +365,83 @@ where
...
+    /// address from storage using this hash.
+    pub fn get_validator_from_tm_address(
+        &self,
+        tm_address: &[u8],
+        epoch: Option<Epoch>,
+    ) -> std::result::Result<Address, Error> {
+        // get the current epoch
+        let epoch = epoch.unwrap_or_else(|| self.storage.get_current_epoch().0);
+        let validator_raw_hash = core::str::from_utf8(tm_address)
+            .map_err(|_| Error::InvalidTMAddress)?;
```
sug0, line 437: Maybe add a TODO here to sync with https://github.com/anoma/namada/issues/200

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -167,12 +169,55 @@ pub(super) enum ShellMode {
...
+            queue: vec![],
+        }
+    }
+
+    /// Pull messages from the channel and add to queue
+    /// Since vote extensions require ordering of ethereum
+    /// events, we do that here. We also de-duplicate events
+    pub fn fill_queue(&mut self) {
+        while let Ok(eth_event) = self.channel.try_recv() {
+            self.queue.push(eth_event);
```
sug0, line 201: [`BTreeSet`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html) should keep everything sorted and de-duplicated.

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -167,12 +169,55 @@ pub(super) enum ShellMode {
...
+        }
+        self.queue.sort();
+        self.queue.dedup();
+    }
+
+    /// Get a copy of the queue
+    pub fn get_events(&self) -> Vec<EthereumEvent> {
+        self.queue.clone()
```
sug0, line 209: Here we would call `self.queue.iter().cloned().collect()`

james-chf (approved):

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,381 @@
...
+        D: DB + for<'iter> DBIter<'iter> + Sync + 'static,
+        H: StorageHasher + Sync + 'static,
+    {
+        /// INVARIANT: This method must be stateless.
+        pub fn extend_vote(
+            &mut self,
+            _req: request::ExtendVote,
+        ) -> response::ExtendVote {
+            let ext = VoteExtension {
+                block_height: self.storage.last_height + 1,
```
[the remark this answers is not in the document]
batconjurer: We could use tendermint here, but since we are planning on replacing it someday with Typhon, it's more future proof to use our internal code logic.

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -167,12 +169,55 @@ pub(super) enum ShellMode {
...
+        }
+    }
+
+    /// Pull messages from the channel and add to queue
+    /// Since vote extensions require ordering of ethereum
+    /// events, we do that here. We also de-duplicate events
+    pub fn fill_queue(&mut self) {
+        while let Ok(eth_event) = self.channel.try_recv() {
+            self.queue.push(eth_event);
+        }
```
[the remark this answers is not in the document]
batconjurer: :+1:

sug0 (approved):

PR #185, opened 2022-07-13
Masp should not incentivize "destroying the disabled"

batconjurer:
At least that's my opinion. This PR is just to enable highlighting parts of the spec I would change and enable comments.


cwgoes:
@batconjurer please rebase and retarget on `main`

cwgoes (approved):
ACK

PR #192, opened 2022-07-15
Add new types related with subsequent vote extensions PRs

sug0:
This PR will add new types to be used by the ethereum bridge integration branch.

The current spec was derived from a meeting me, James and Jacob had, so it's still not finalized, but the current version can be found [here](https://hackmd.io/RTtsCl_RRBu2acISHL0-2g).


shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub signatures: Vec<(Signature, Address)>,
+    /// The events that were reported
+    pub events: Vec<MultiSignedEthEvent>,
+    /// The validators who saw no events
+    pub nulls: HashSet<Address>
+}
+
+impl VoteExtensionDigest {
+    /// Decompresses a set of signed `VoteExtension` instances.
+    pub fn decompress<D, H>(self, storage: &Storage<D, H>) -> Vec<Signed<VoteExtension>>
```
batconjurer, line 133: instead of passing in the storage here, lets just pass in the block height.

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Test error handling on the FractionalVotingPower type
+    #[test]
+    fn test_fractional_voting_power_valid_fractions() {
+        assert!(FractionalVotingPower::new(0, 0).is_err());
+        assert!(FractionalVotingPower::new(1, 0).is_err());
+        assert!(FractionalVotingPower::new(0, 1).is_ok());
+        assert!(FractionalVotingPower::new(1, 1).is_ok());
+        assert!(FractionalVotingPower::new(1, 2).is_ok());
+        assert!(FractionalVotingPower::new(3, 2).is_err());
+    }
```
batconjurer, line 228: I would love a unit test or two for the decompress algorithm if that's not too much trouble.

james-chf (approved):
lgtm. During [VerifyVoteExtension](https://docs.tendermint.com/master/spec/abci++/abci++_methods.html#verifyvoteextension) we get access to the voting validator's Tendermint address, I'm thinking ultimately we will want to check that matches up with the application-level signature of the vote extension being verified, to prevent replays of stale vote extensions from a previous failed round for the the current round which is at the same block height.

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -0,0 +1,299 @@
...
+use super::EthereumEvent;
+use crate::proto::Signed;
+use crate::types::address::Address;
+use crate::types::key::common::Signature;
+use crate::types::storage::BlockHeight;
+
+/// This struct will be created and signed over by each
+/// validator as their vote extension.
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct VoteExtension {
```
batconjurer, line 19: This is fine for now, but I will move this struct to a different place later since vote extesnions have uses other than just the ethereum bridge.

batconjurer: But I will have to think about how to decompress digests in that case, so I'm not going to spend brain power on it now.

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -0,0 +1,299 @@
...
+        } = self;
+
+        let mut extensions = vec![];
+
+        for (sig, addr) in signatures.into_iter() {
+            let mut ext = VoteExtension::empty(last_height);
+
+            // TODO: perhaps remove the `nulls` field,
+            // as this code will behave much the same without it
+            if !nulls.contains(&addr) {
```
batconjurer, line 144: You're completely right. We should remove the nulls field. I would rather have slightly longer execution times if it means a smaller digest.

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -0,0 +1,299 @@
...
+            events,
+            signatures,
+            nulls: HashSet::new(),
+        };
+
+        // finally, decompress the `VoteExtensionDigest` back into a
+        // `Vec<Signed<VoteExtension>>`
+        let decompressed = digest.decompress(last_block_height);
+
+        assert_eq!(ext, decompressed);
```
batconjurer, line 297: For complete sanity's sake, can we test that that signatures for each `Signed<VoteExtension>` from the decompressed verifies correctly?
```
assert!(decompressed[0].verify(&sk_1.ref_to()).is_ok());
assert!(decompressed[1].verify(&sk_2.ref_to()).is_ok());
```

batconjurer (changes requested):
Two small requests. Otherwise looks good.

batconjurer (approved):

PR #202, opened 2022-07-20
Include vote extensions in PrepareProposal

sug0:
This PR implements the inclusion of vote extensions in PrepareProposal.


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,230 @@ mod prepare_block {
...
+            &self,
+            vote_extensions: Vec<ExtendedVoteInfo>,
+        ) -> Option<VoteExtensionDigest> {
+            let events_epoch = self
+                .storage
+                .block
+                .pred_epochs
+                .get_epoch(self.storage.last_height)
+                // TODO: is this `unwrap()` fine?
+                .unwrap();
```
batconjurer, line 157: You can even put an `expect` here.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,231 @@ mod prepare_block {
...
+                        .map_err(|err| {
+                            tracing::error!(
+                                "Failed to deserialize signed vote extension: \
+                                 {}",
+                                err
+                            );
+                        })
+                        .ok()
+                        .and_then(|ext| {
+                            let ext_height = ext.data.block_height;
```
batconjurer, line 175: This block can be simplified now that you have this https://github.com/anoma/namada/blob/5e6ea5ee57cdcc9acfc9e7a9a5119dd7f27a58f7/apps/src/lib/node/ledger/shell/vote_extensions.rs#L83

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,231 @@ mod prepare_block {
...
+                        .ok()?;
+
+                    Some((validator_addr, vote_extension))
+                });
+
+            let mut event_observers = BTreeMap::new();
+            let mut signatures = Vec::new();
+
+            let total_voting_power =
+                self.get_total_voting_power(Some(events_epoch)).into();
```
batconjurer, line 241: Can you remove the `#[allow(dead_code)]` from this function?

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,231 @@ mod prepare_block {
...
+                let sig = vote_extension.sig;
+
+                signatures.push((sig, addr));
+            }
+
+            let voting_power =
+                FractionalVotingPower::new(voting_power, total_voting_power)
+                    .unwrap();
+
+            if voting_power <= FractionalVotingPower::TWO_THIRDS {
```
batconjurer, line 277: If this ever happens, we are in the shit.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,231 @@ mod prepare_block {
...
+                    .to_bytes()
+                })
+                .map(record::add)
+                .collect()
+        }
+
+        /// Compresses a set of vote extensions into a single
+        /// [`VoteExtensionDigest`], whilst filtering invalid
+        /// `Signed<VoteExtension>` instances in the process
+        fn compress_vote_extensions(
```
batconjurer, line 147: I would like to have a few unit tests for this function. 
 - Happy flow
 - Deduplicate multiple events from same validator
 - Filter out non-validator vote extensions
 - Filter out bad signatures from validators

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +64,231 @@ mod prepare_block {
...
+                    tracing::error!(
+                        "The genesis block should not contain vote extensions"
+                    );
+                    return vec![];
+                }
+                Some(d) => d,
+                // if no vote extensions were found, we return an empty
+                // `Vec` of protocol
+                // transactions
+                _ => return vec![],
```
batconjurer, line 95: Apart from genesis, this is a tendermint consensus failure. I think we should put an `unreachable!()` and crash out because something very bad has happened.

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
sug0: how do we handle this scenario? right now I'm just returning an empty vec of protocol transactions

sug0: actually this is already covered by the `unreachable!()` I added recently. I'll mark this as solved

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +66,204 @@ mod prepare_block {
...
+                    } else {
+                        record::remove(tx_bytes)
+                    }
+                })
+                .collect()
+        }
+
+        /// Builds a batch of DKG decrypted transactions
+        fn build_decrypted_txs(&mut self) -> Vec<TxRecord> {
+            // TODO: This should not be hardcoded
```
james-chf, line 143: This isn't related to the PR but since we're refactoring here. I think this TODO should be rewritten and maybe link to an issue explaining what the status is with the DKG/front running protection. As I understand for Namada, this encryption step doesn't provide any real protection to transactions right now but Namada specifically won't have DKG/front running protection as a feature.

james-chf (commented):
I've only glanced so far as I guess this PR will change a bit but will come back to review again.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +66,204 @@ mod prepare_block {
...
+
+            let vote_extension_digest =
+                local_last_commit.and_then(|local_last_commit| {
+                    let votes = local_last_commit.votes;
+                    self.compress_vote_extensions(votes)
+                });
+            let vote_extension_digest =
+                match (vote_extension_digest, self.storage.last_height) {
+                    // handle genesis block
+                    (None, BlockHeight(0)) => return vec![],
```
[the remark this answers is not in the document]
batconjurer: Genesis is 0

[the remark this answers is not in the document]
james-chf: Ah right makes sense, the blocks start from height=1 so when proposing the genesis block the last height will be 0

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -77,6 +66,204 @@ mod prepare_block {
...
+                    (Some(_), BlockHeight(0)) => {
+                        tracing::error!(
+                            "The genesis block should not contain vote \
+                             extensions"
+                        );
+                        // TODO: maybe slash validators who claim to have
+                        // seen vote extensions at H=0
+                        return vec![];
```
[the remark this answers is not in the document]
sug0: this scenario is possible if byzantine validators claim they have vote extensions to act upon, when the last height is 0 (ie. the genesis block). since vote extensions are included at the end of a round in the precommit, they couldn't have seen any vote extensions then.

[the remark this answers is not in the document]
james-chf: When the last height is 0, there has been no previous round (or previous precommit stage), so there definitely shouldn't be any vote extensions? ie byzantine validators haven't had a chance to call `ExtendVote` yet. I would have thought Tendermint could never be handing us vote extensions during `PrepareProposal` for the genesis block. i.e. `self.storage.last_height == 0` -> `req.local_last_commit.votes.is_empty()` as block proposer.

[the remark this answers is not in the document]
sug0: ah actually good point, but I don't think anything else gets called at height 0. definitely worth double checking though.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -606,6 +609,87 @@ mod prepare_block {
...
+                    .map(|v| WeightedValidator {
+                        voting_power: VotingPower::from(0u64),
+                        ..v
+                    })
+                    .collect();
+
+                epochs.set(data, events_epoch, &params);
+                epochs
+            };
+            shell.storage.write_validator_set(&validator_set);
```
sug0, line 656: I'm not sure why this is not working as expected (the one default validator would get a voting power of 0). This would provoke the `prepare_proposal` call to reach `unreachable!()` code, as expected (since we would get <= 2/3 voting power).

Any ideas @james-chf @batconjurer ??

apps/src/lib/node/ledger/shell/prepare_proposal.rs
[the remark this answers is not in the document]
sug0: I don't see anything related with this in the [specs](https://docs.tendermint.com/master/spec/abci++/abci++_methods.html), so I'm marking the issue as resolved

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -144,6 +341,357 @@ mod prepare_block {
...
+        fn test_prepare_proposal_filter_out_bad_vext_validators() {
+            const LAST_HEIGHT: BlockHeight = BlockHeight(2);
+
+            let (mut shell, _, _) = test_utils::setup();
+
+            // artificially change the block height
+            shell.storage.last_height = LAST_HEIGHT;
+
+            let (validator_addr, protocol_key) = {
+                let bertha_key = wallet::defaults::bertha_keypair();
```
batconjurer, line 437: Excellent choice of non-validator

sug0: pwned by bertha

apps/src/lib/node/ledger/shell/prepare_proposal.rs
batconjurer:
```
let mut req = FinalizeBlock::default();
            req.header.time = namada::types::time::DateTimeUtc::now();
            shell.storage.last_height = BlockHeight(11);
            shell.finalize_block(req).expect("Test failed");
            shell.commit();
            ``

batconjurer: validator set changes don't take effect unless the epoch changes. You've only queued up the change

batconjurer (changes requested):

batconjurer (approved):

james-chf (approved):
lgtm, I ran the ledger manually and can see in the logs that the protocol transaction is being created and reaching `FinalizeBlock`

Issue #144, opened 2022-06-28
Store newly seen Ethereum events


james-chf:
During `FinalizeBlock`, when we are applying a `ProtocolTxType::EthereumEvents(_)`, we should store any newly seen Ethereum events under the `/eth_msgs` storage subspace.

Issue #104, opened 2022-04-28
Test runner exits suddenly if `wasm/checksums.json` is not up-to-date


james-chf:
Running `make test-unit` for the first time may result in an error like:

```
test ledger::storage::write_log::tests::verifiers_changed_key_tx_all_key ... ok

test result: ok. 95 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.83s

     Running unittests (target/debug/deps/anoma_apps-94c0e34acd15bcd8)

running 31 tests
test config::genesis::tests::gen_genesis_validator ... ok
error: test failed, to rerun pass '-p anoma_apps --lib'
make: *** [test-unit] Error 1
```

This happens when needed WASMs aren't present under `wasm` or `wasm/checksums.json` isn't up to date. `wasm_loader::read_wasm` terminates the process when it can't read a WASM, so further tests don't run and the output stops abruptly.

The workaround when this happens is the following:

```shell
make -C wasm/wasm_source
make checksum-wasm
```

james-chf:
I feel like this issue is hit very often. After https://github.com/anoma/anoma/pull/1099, we could change the Makefile targets to require `checksum-wasm` to be run e.g.

```Makefile
test-unit: checksum-wasm
	$(cargo) test --no-default-features \
		--features "wasm-runtime ABCI ibc-mocks-abci eth-fullnode" \
		-- --skip e2e
```

We could then update the `wasm/checksums.py` script to give a friendlier error message, e.g. fail and prompt the user to build wasms if there are none present, rather than outputting an empty JSON `{}` for `checksums.json`. So on first run of `make test-unit`, the user should realize they need to build some other things (but don't necessarily need to build those other things every run)

james-chf:
The root cause of this issue is that some of our unit tests depend on the presence of `.wasm` files on disk, we should fix our unit tests not to do this (unit tests should be pure and not depend on the disk environment imo). If we have some tests that expect a healthy `wasm/checksums.json`, they could be put under a different Makefile target.

PR #205, opened 2022-07-21
Add `tx_eth_bridge` that is executed by all full nodes during FinalizeBlock

james-chf:
Relates to https://github.com/anoma/namada/issues/144

This PR adds a no-op `tx_eth_bridge` wasm and executes it whenever we apply a `ProtocolTransactionType::EthereumEvents` during `FinalizeBlock`. Logic for actually looking at the protocol transaction and making changes to storage will be in later PRs.


james-chf:
Current dependencies on/for this PR:
* eth-bridge-integration
  * **PR #205** <a href="https://app.graphite.dev/github/pr/anoma/namada/205" target="_blank">Graphite</a>  👈
    * **PR #210** <a href="https://app.graphite.dev/github/pr/anoma/namada/210" target="_blank">Graphite</a> 
      * **PR #211** <a href="https://app.graphite.dev/github/pr/anoma/namada/211" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/anoma/namada/205?utm_source=stack-comment).

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -181,6 +212,40 @@ where
...
+const TX_ETH_BRIDGE_WASM_NAME: &str = "tx_eth_bridge";
+
+fn construct_tx_eth_bridge(wasm_dir: &Path) -> Result<Tx> {
+    let tx_data = vec![];
+    tracing::debug!(
+        bytes = tx_data.len(),
+        "serialized tx_data for state update transaction"
+    );
+    let tx_code = {
+        let checksums = crate::wasm_loader::Checksums::read_checksums(wasm_dir);
```
james-chf, line 224: NB: we `safe_exit` the process here if a `checksums.json` can't be found, so we can't write a test for an empty wasm_dir yet (as the test runner exits similar to https://github.com/anoma/namada/issues/104)

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -466,3 +531,65 @@ fn merge_vp_results(
...
+
+    use serde_json::json;
+
+    use super::*;
+
+    // constructs a temporary fake wasm_dir with one wasm and a checksums.json
+    fn fake_wasm_dir(
+        wasm_name: impl AsRef<str>,
+        wasm_contents: impl AsRef<[u8]>,
+    ) -> PathBuf {
```
james-chf, line 548: There is probably a better place for this test helper, I took this from an old branch - https://github.dev/anoma/anoma/blob/f390706b6018e1ce46d76cc51c2d8671591b5925/tests/src/e2e/helpers/wasms.rs

apps/src/lib/node/ledger/protocol/mod.rs
sug0: if we panicked instead of exiting the process, we would be able to catch the panic... bah but at that point, it would be better to return some `Result` type from `wasm_loader::Checksums::read_checksums`

batconjurer: I'm not sure how I feel about doing this checksums check every block. It's a lot of io and a validator who uses the wrong wasm should be considered byzantine. Is there a way to refactor this to do the check once at start up?

batconjurer: Then e2e tests would cover the fail to start up when there is an empty / nonexistent wasm dir

batconjurer (changes requested):

apps/src/lib/node/ledger/protocol/mod.rs
james-chf: Yeah it's not great that it reads and hashes every block. Possible solutions:

- we could bundle the wasm into the binary at compile time using `include_bytes!` (along with a precalculated hardcoded hash to check against once at start up)
- instead of `wasm_dir` we could use a `preloaded_wasms` (`HashMap<String, Vec<u8>>`) that is constructed once at start up time. It would only contain `tx_eth_bridge.wasm` for now.

james-chf: I think we can have something similar to the `Checksums(pub HashMap<String, String>)` struct that keeps `.wasm` files it has read and checked in memory. Then we can refactor `Shell`/`AbcippShim` slightly to have one of these components and pass it down to `apply_tx`.

batconjurer: Can't we just use the wasm cache? It gets initialized in the `Shell::new` method and we can dump the wasm blob in there.

james-chf: `tx_wasm_cache` lets us look up a compiled wasm module by wasm code. We could warm it up using `tx_wasm_cache.fetch_or_compile(code)` in `Shell::new` though we'll still need the wasm code itself during `apply_tx` to get the compiled wasm out of the cache.

batconjurer: We could write a fetch method that would return an error if the wasm code isn't in there.  Also in the Cache code, it uses a hash of the code internally for lookups. So our new fetch method could just take a hash as input

james-chf: :+1: I'll try implementing this, we should either pass the hash down to `apply_tx` somehow or look up the compiled wasm by some preknown name e.g. `tx_wasm_cache.fetch_by_name("tx_eth_bridge")`

Issue #200, opened 2022-07-19
Some Byzantine validators may not be getting slashed


sug0:
I believe the [`PosBase`](https://github.com/anoma/namada/blob/3a65fed50c3f63f435b2e8223044729b26dc70b1/proof_of_stake/src/lib.rs#L474=) API is being misused in the [`Shell`](https://github.com/anoma/namada/blob/3a65fed50c3f63f435b2e8223044729b26dc70b1/apps/src/lib/node/ledger/shell/mod.rs#L488=).

This trait contains a method [`read_validator_address_raw_hash`](https://github.com/anoma/namada/blob/3a65fed50c3f63f435b2e8223044729b26dc70b1/proof_of_stake/src/lib.rs#L543=), which takes a `raw_hash` parameter of type `impl AsRef<str>` (the Tendermint address of some validator), and returns the [`Address`](https://github.com/anoma/namada/blob/3a65fed50c3f63f435b2e8223044729b26dc70b1/shared/src/types/address.rs#L117=) of the equivalent node on Anoma's side, from what I've gathered.

ABCI++ hands us an [`Evidence`](https://github.com/heliaxdev/tendermint-rs/blob/master/proto/src/prost/tendermint.abci.rs#L481=) struct for each Byzantine validator detected during a consensus round, which internally contains a [`Validator`](https://github.com/heliaxdev/tendermint-rs/blob/master/proto/src/prost/tendermint.abci.rs#L486=') struct reflecting the [`address`](https://github.com/heliaxdev/tendermint-rs/blob/master/proto/src/prost/tendermint.abci.rs#L457=) of the Byzantine node. This `address` is the 20-byte prefix of the SHA256 hash of the public key of a Tendermint validator.

When processing individual `Evidence` structs, we try to decode the `address` handed us via ABCI++ as a UTF-8 string, which may not always succeed (since we are dealing with a raw hash). In case this operation fails, we skip Byzantine validators which would have otherwise been slashed.

tzemanovic:
nice catch! I'm not sure if it's currently broken because we also give tendermint validator's address, which isn't derived from consensus key and it seems to be fine with it, but it's not following it's convention, so I think it really should be changed to use the consensus key instead.

Let's check this with #14.

PR #206, opened 2022-07-21
Log reason for vote extension validation failures

james-chf:
This PR is an idea for adding more logging around the reason for vote extension validation failures. There is an issue currently where we try to parse the Tendermint address as UTF-8 (https://github.com/anoma/namada/issues/200) which means we might not be able to verify a vote extension.

A single genesis validator fails to start up on a new chain. Binaries built with these feature flags.
```
namada/ABCI-plus-plus namada/ibc-mocks namada/testing namada/wasm-runtime namada_apps/ABCI-plus-plus namada_apps/eth-fullnode
```
then copied into image `ghcr.io/james-chf/devchain-container/abcipp:sha-5be04b7` and the container run.

The error is (with the reason logged)
```
compose-ledger-1  | 2022-07-21T12:16:45.717092Z DEBUG namada_apps::node::ledger::shell: received Ethereum events n=0
compose-ledger-1  | 2022-07-21T12:16:45.719641Z  WARN namada_apps::node::ledger::shell::vote_extensions::extend_votes: received vote extension that didn't validate req.validator_address=[215, 207, 7, 44, 27, 0, 113, 64, 148, 192, 1, 153, 225, 116, 180, 176, 128, 69, 173, 194] req.hash=[216, 162, 130, 64, 137, 37, 1, 40, 205, 170, 68, 138, 155, 82, 206, 143, 106, 1, 186, 148, 170, 249, 40, 161, 243, 196, 191, 146, 195, 82, 100, 230] reason="couldn't get validator address: InvalidTMAddress(Utf8Error { valid_up_to: 0, error_len: Some(1) })" req.height=1
compose-ledger-1  | 2022-07-21T12:16:45Z ERROR failed to process message err="error adding vote" height=1 module=consensus msg_type=*consensus.VoteMessage peer= round=0
```

### DIFF
--- a/apps/src/lib/node/ledger/shell/queries.rs
+++ b/apps/src/lib/node/ledger/shell/queries.rs
@@ -1,5 +1,6 @@
 //! Shell methods for querying state
 use std::cmp::max;
+use std::str::Utf8Error;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use ferveo_common::TendermintValidator;
@@ -47,7 +48,7 @@ pub enum Error {
     )]
     NotValidatorKeyHash(String, Epoch),
     #[error("Invalid validator tendermint address")]
-    InvalidTMAddress,
+    InvalidTMAddress(#[from] Utf8Error),
 }
 
 impl<D, H> Shell<D, H>
@@ -436,7 +437,7 @@ where
         // get the current epoch
         let epoch = epoch.unwrap_or_else(|| self.storage.get_current_epoch().0);
         let validator_raw_hash = core::str::from_utf8(tm_address)
-            .map_err(|_| Error::InvalidTMAddress)?;
+            .map_err(|err| Error::InvalidTMAddress(err))?;
         self.storage
             .read_validator_address_raw_hash(&validator_raw_hash)
             .ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -47,7 +48,7 @@ pub enum Error {
     )]
     NotValidatorKeyHash(String, Epoch),
     #[error("Invalid validator tendermint address")]
-    InvalidTMAddress,
+    InvalidTMAddress(#[from] Utf8Error),
```
james-chf, line 51: This shouldn't land as is, just added this to figure out what was going on. I think we will need to fix https://github.com/anoma/namada/issues/200 then remove this error

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -8,6 +8,11 @@ mod extend_votes {
 
     type SignedExt = Signed<VoteExtension>;
 
+    pub enum ValidateVoteExtensionResult {
+        Success,
+        Failure { reason: String },
+    }
```
james-chf, line 14: This is crude but thinking something like this we can log (maybe a better variant for each possible reason?). I thought maybe `Result<()>` and logging the `Err(_)` but if `validate_vote_extension` is passed an invalid vote extension and we verify it as not valid, that isn't an error case but an expected execution path.

james-chf:
More logging has been/is being added around this code since this PR was opened

Issue #208, opened 2022-07-21
Temporarily add Anoma address or public key to vote extensions


james-chf:
Until https://github.com/anoma/namada/issues/200 is fixed, we should work around by directly including the validator's Anoma address or public key in vote extensions, since we won't be able to look this up by their Tendermint address for the time being.

PR #209, opened 2022-07-21
Temporarily add validator address to vote extensions

james-chf:
Resolves https://github.com/anoma/namada/issues/208

`make test-unit-abci-plus-plus` passes, and I manually verified running a chain with `namadan` doesn't encounter the error from https://github.com/anoma/namada/pull/206 anymore

### DIFF
--- a/apps/src/lib/node/ledger/shell/vote_extensions.rs
+++ b/apps/src/lib/node/ledger/shell/vote_extensions.rs
@@ -18,16 +18,11 @@ mod extend_votes {
             &mut self,
             _req: request::ExtendVote,
         ) -> response::ExtendVote {
-            let validator_addr = match self.mode.get_validator_address() {
-                Some(validator_addr) => validator_addr.to_owned(),
-                None => {
-                    tracing::warn!(
-                        "couldn't get validator address, returning empty vote \
-                         extension"
-                    );
-                    return response::ExtendVote::default();
-                }
-            };
+            let validator_addr = self
+                .mode
+                .get_validator_address()
+                .expect("only validators should receive this method call")
+                .to_owned();
             let ext = VoteExtension {
                 block_height: self.storage.last_height + 1,
                 ethereum_events: self.new_ethereum_events(),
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -18,9 +18,20 @@ mod extend_votes {
             &mut self,
             _req: request::ExtendVote,
         ) -> response::ExtendVote {
+            let validator_addr = match self.mode.get_validator_address() {
```
batconjurer, line 21: We should put an expect here I think. Only validators should receive this method call.

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -19,17 +19,22 @@ use crate::types::storage::BlockHeight;
 pub struct VoteExtension {
     /// The block height for which this [`VoteExtension`] was made.
     pub block_height: BlockHeight,
+    /// The validator's address is temporarily being included
```
batconjurer, line 22: mark this as a TODO

batconjurer (changes requested):

shared/src/types/ethereum_events/vote_extensions.rs
```diff
@@ -252,16 +260,16 @@ mod tests {
...
-        let ext_1 = Signed::new(&sk_1, ext.clone());
-        let ext_2 = Signed::new(&sk_2, ext);
+        let ext_1 = Signed::new(&sk_1, ext(validator_1.clone()));
+        let ext_2 = Signed::new(&sk_2, ext(validator_2.clone()));
 
         let ext = vec![ext_1, ext_2];
 
         // we have the `Signed<VoteExtension>` instances we need,
         // let us now compress them into a single `VoteExtensionDigest`
         let signatures: Vec<(_, Address)> = vec![
```
sug0, line 270: We can probably remove the type signature (`Vec<(_, Address)>`). This was only here because the compiler was complaining about not being able to resolve the type of `.into()`.

james-chf: :+1:

sug0 (approved):

batconjurer (approved):

PR #210, opened 2022-07-21
Calculate `EthMsgUpdate`s based on vote extensions

james-chf:
Relates to https://github.com/anoma/namada/issues/144, depends on PR https://github.com/anoma/namada/pull/205

This PR:
- constructs a `TxEthBridgeData` which contains the information needed to update the `/eth_msgs` storage subspace and do mints. We include the last block's epoch and the voting power for each validator in that epoch who saw an event in the previous round's vote extensions. It is easier to look up these bits of information using `Shell` APIs than to do it in wasm when the transaction is being executed.
- constructs `EthMsgUpdate`s for each voted-for Ethereum event, that are included in the `TxEthBridgeData`. Each of these structs should contain the information needed to create/update an Ethereum event in storage.


james-chf:
Current dependencies on/for this PR:
* eth-bridge-integration
  * **PR #205** <a href="https://app.graphite.dev/github/pr/anoma/namada/205" target="_blank">Graphite</a> 
    * **PR #210** <a href="https://app.graphite.dev/github/pr/anoma/namada/210" target="_blank">Graphite</a>  👈
      * **PR #211** <a href="https://app.graphite.dev/github/pr/anoma/namada/211" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/anoma/namada/210?utm_source=stack-comment).

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msg_updates.rs
```diff
@@ -0,0 +1,162 @@
...
+    let body = multisigned.event;
+
+    let seen_by = multisigned
+        .signers
+        .into_iter()
+        .sorted_by(|a, b| a.cmp(b))
+        .collect();
+
+    EthMsgUpdate { body, seen_by }
+}
```
james-chf, line 25: I'm not sure now whether these fns need to be deterministic. When acting on Ethereum events we need to do so in a deterministic order, but for updating of `/eth_msgs` storage I'm not sure if this is strictly necessary. eg when adding voting powers to an EthMsgs, or updating the `seen_by` field. All nodes should arrive at the same result regardless, and the `seen_by` field should always be saved in some deterministic sorted order, but the intermediary calculations/additions can be in any order.

shared/src/types/ethereum_events.rs
```diff
@@ -219,3 +229,138 @@ pub struct TokenWhitelist {
...
+    BorshSerialize,
+    BorshDeserialize,
+)]
+pub struct EthMsgUpdate {
+    /// the event being seen
+    pub body: EthereumEvent,
+    /// addresses of the validators who have just seen this event
+    /// we use [`BTreeSet`] even though ordering is not important here, so that
+    /// we can derive [`Hash`] for [`EthMsgUpdate`]
+    pub seen_by: BTreeSet<Address>,
```
james-chf, line 251: Further up the stack the `/eth_msgs/<hash>/seen_by` is still a `Vec<Address>` but maybe a `BTreeSet` would be better?

james-chf:
Opening this for preliminary review now. I think all nodes should arrive at the same `/eth_msgs` state regardless of the order in which the apply `EthMsgUpdate`s (ie updating `seen_by` and `voting_power` fields of stored `/eth_msgs`). We will need to confirm what ordering requirements there are on executing Ethereum events when they become `seen = true`.

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -216,6 +252,24 @@ where
     }
 }
 
+/// This is a helper function for getting the set of active validators for a
+/// given epoch.
+pub fn get_active_validators<D, H>(
```
batconjurer, line 257: This function would be useful in the `queries.rs` module. In fact, it would probably simplify a lot of the existing functions in there. Just also make the epoch optional and if a `None` is passed in, use the current.

batconjurer (changes requested):
It looks like you are summing up the voting power of the vote extensions and putting that on every event in the digest. We need to sum up for each event, the power of the validators who saw it.

apps/src/lib/node/ledger/protocol/mod.rs
james-chf: :+1: I'll open a small PR to move this to `queries.rs` and some of the other voting power helper code so it can be shared between `Shell` and `apply_tx`

PR #219, opened 2022-07-26
Handle vote extensions in ProcessProposal

sug0:
This PR implements handling vote extensions in ProcessProposal.

---

In ProcessProposal, each validator must check:

* All the signatures are valid
* All signatures came from an validator in the active validator set at the previous block height,
* The block height is correct,
* The sum of the stake of the validators in the contained vector is at least 2/3 of the total amount staked at the previous block height.

### DIFF
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -195,7 +195,10 @@ mod prepare_block {
                     validator_voting_power,
                     total_voting_power,
                 )
-                .unwrap_or_default();
+                .expect(
+                    "The voting power we obtain from storage should always be \
+                     valid",
+                );
 
                 // register all ethereum events seen by `validator_addr`
                 for ev in vote_extension.data.ethereum_events {
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -355,21 +355,33 @@ mod prepare_block {
             // artificially change the block height
             shell.storage.last_height = LAST_HEIGHT;
 
-            let validator_addr = wallet::defaults::validator_address();
-
             let signed_vote_extension = {
-                // create a fake signature
-                let sig = common::Signature::Ed25519(ed25519::Signature(
-                    [0u8; 64].into(),
-                ));
+                let (protocol_key, _) = wallet::defaults::validator_keys();
+                let validator_addr = wallet::defaults::validator_address();
 
-                let data = VoteExtension {
+                // generate a valid signature
+                let mut ext = VoteExtension {
                     validator_addr,
                     block_height: LAST_HEIGHT,
                     ethereum_events: vec![],
+                }
+                .sign(&protocol_key);
+                assert!(ext.verify(&protocol_key.ref_to()).is_ok());
+
+                // modify this signature such that it becomes invalid
+                ext.sig = {
+                    let mut sig_bytes = match ext.sig {
+                        common::Signature::Ed25519(ed25519::Signature(
+                            ref sig,
+                        )) => sig.to_bytes(),
+                    };
+                    sig_bytes[0] = sig_bytes[0].wrapping_add(1);
+                    common::Signature::Ed25519(ed25519::Signature(
+                        sig_bytes.into(),
+                    ))
                 };
 
-                SignedExt { sig, data }
+                ext
             };
 
             check_vote_extension_filtering(&mut shell, signed_vote_extension);
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -283,7 +283,7 @@ mod prepare_block {
         use namada::types::address::xan;
         use namada::types::ethereum_events::vote_extensions::VoteExtension;
         use namada::types::ethereum_events::EthereumEvent;
-        use namada::types::key::{common, ed25519};
+        use namada::types::key::common;
         use namada::types::storage::{BlockHeight, Epoch};
         use namada::types::transaction::protocol::ProtocolTxType;
         use namada::types::transaction::{Fee, TxType};
@@ -369,18 +369,7 @@ mod prepare_block {
                 assert!(ext.verify(&protocol_key.ref_to()).is_ok());
 
                 // modify this signature such that it becomes invalid
-                ext.sig = {
-                    let mut sig_bytes = match ext.sig {
-                        common::Signature::Ed25519(ed25519::Signature(
-                            ref sig,
-                        )) => sig.to_bytes(),
-                    };
-                    sig_bytes[0] = sig_bytes[0].wrapping_add(1);
-                    common::Signature::Ed25519(ed25519::Signature(
-                        sig_bytes.into(),
-                    ))
-                };
-
+                ext.sig = test_utils::invalidate_signature(ext.sig);
                 ext
             };
 
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -4,18 +4,15 @@
 mod prepare_block {
     use std::collections::{BTreeMap, HashMap, HashSet};
 
-    use namada::ledger::pos::namada_proof_of_stake::types::VotingPower;
-    use namada::proto::Signed;
     use namada::types::ethereum_events::vote_extensions::{
-        FractionalVotingPower, MultiSignedEthEvent, VoteExtension,
-        VoteExtensionDigest,
+        FractionalVotingPower, MultiSignedEthEvent, VoteExtensionDigest,
     };
     use namada::types::transaction::protocol::ProtocolTxType;
     use tendermint_proto::abci::{
         ExtendedCommitInfo, ExtendedVoteInfo, TxRecord,
     };
 
-    use super::super::vote_extensions::{SignedExt, VoteExtensionError};
+    use super::super::vote_extensions::deserialize_vote_extensions;
     use super::super::*;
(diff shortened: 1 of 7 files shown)


apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -166,14 +166,23 @@ where
...
+                            //         .into(),
+                            // }
+                            // ```
+                            //
+                            // TODO: is unreachable really fine?
+                            unreachable!(
+                                "We should never get an insufficient backing \
+                                 stake on vote extensions published in \
+                                 proposals."
+                            )
```
sug0, line 194: I'm not sure if this `unreachable!()` condition is fine here. I guess the leader might censor some `VoteExtension` instances here, and if so, I need to get rid of this, and return the `TxResult` value again...

james-chf (commented):
I've taken a first look. I think separate from this PR there are some transaction types now that we would never expect an honest block proposer to include e.g. bytes which don't decode to `Tx`. So there should be some refactoring at some point to make sure the set of transactions that (honest) `PrepareProposal` might include in a block is the same as the set of transactions that `ProcessProposal` will accept.

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -103,18 +123,39 @@ mod extend_votes {
...
+            }
+            // verify if we have any duplicate Ethereum events,
+            // and if these are sorted in ascending order
+            let have_dupes_or_non_sorted = {
+                !ext.data
+                    .ethereum_events
+                    // TODO: move to `array_windows` when it reaches Rust stable
+                    .windows(2)
+                    .all(|evs| evs[0] < evs[1])
+            };
```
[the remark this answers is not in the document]
sug0: if two elements are duplicated, less then will evaluate to `false`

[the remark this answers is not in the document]
james-chf: ah cool!

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -238,30 +238,55 @@ mod prepare_block {
             Some(VoteExtensionDigest { events, signatures })
         }
 
+        /// Takes a list of signed vote extensions,
+        /// and filters out invalid instances, returning
+        /// all residual values (including invalid vote
+        /// extensions).
+        #[inline]
+        pub fn filter_invalid_vote_extensions_residuals(
```
batconjurer, line 246: Can this function go live in the `vote_extensions` module? It makes sense there and will tighten up this file a bit.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -238,30 +238,55 @@ mod prepare_block {
...
+                    self.storage.last_height,
+                )
+            })
+        }
+
         /// Takes a list of signed vote extensions,
         /// and filters out invalid instances.
-        fn filter_invalid_vote_extensions(
+        #[inline]
+        pub fn filter_invalid_vote_extensions(
```
batconjurer, line 266: Similarly, can this go live in `vote_extensions`?

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -238,30 +238,55 @@ mod prepare_block {
...
-                    &vote.vote_extension[..],
-                )
+            self.filter_invalid_vote_extensions_residuals(vote_extensions)
+                .filter_map(|ext| ext.ok())
+        }
+    }
+
+    /// Given a `Vec` of [`ExtendedVoteInfo`], return an iterator over the
+    /// deserialized [`SignedExt`] instances.
+    fn deserialize_vote_extensions(
```
batconjurer, line 277: Similarly, can this go live in `vote_extensions`?

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -34,16 +35,31 @@ where
...
+        // We should not have more than one `VoteExtensionDigest` in
+        // a proposal from some round's leader.
+        let too_many_vext_digests = vote_ext_digest_num > 1;
+
+        // Erroneous transactions were detected when processing
+        // the leader's proposal. We allow txs that do not
+        // deserialize properly, that have invalid signatures
+        // and that have invalid wasm code to reach FinalizeBlock.
+        // This is the reason behind that greater than three check.
+        let invalid_txs = tx_results.iter().any(|res| res.code > 3);
```
batconjurer, line 59: Maybe we make a constant `PROTOCOL_LEVEL_ERROR=3`. We allow user txs to go through that error, but if something is wrong at the protocol level, we reject. That might make this code a little clearer to understand

batconjurer: And if you think of a better name than `PROTOCOL_LEVEL_ERROR`, use it.

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -293,12 +383,217 @@ mod test_process_proposal {
...
+        };
+        let tx = ProtocolTxType::EthereumEvents(vote_extension_digest)
+            .sign(&protocol_key)
+            .to_bytes();
+        #[allow(clippy::redundant_clone)]
+        let request = ProcessProposal {
+            txs: vec![tx.clone(), tx],
+        };
+        let results = shell.process_proposal(request);
+        assert!(
```
batconjurer, line 430: you can use the `assert_matches!` macro

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -293,12 +383,217 @@ mod test_process_proposal {
...
+        let (protocol_key, _) = wallet::defaults::validator_keys();
+        let vote_extension_digest = {
+            let addr = wallet::defaults::validator_address();
+            let event = EthereumEvent::TransfersToNamada {
+                nonce: 1u64.into(),
+                transfers: vec![],
+            };
+            let ext = {
+                // create a fake signature
+                let sig = common::Signature::Ed25519(ed25519::Signature(
```
batconjurer, line 477: in `test_utils`, there is a function for generating random keypairs

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/process_proposal.rs
sug0: Boy I sure don't want my unit test to fail with a `1 / 2^256` chance

batconjurer (approved):

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -67,116 +87,183 @@ where
...
-                            }
-                        } else if verify_decrypted_correctly(&tx, privkey) {
+            TxType::Protocol(protocol_tx) => match protocol_tx.tx {
+                ProtocolTxType::EthereumEvents(digest) => {
+                    *vote_ext_digest_num += 1;
+
+                    let extensions =
+                        digest.decompress(self.storage.last_height);
+                    let filtered_extensions = self
+                        .filter_invalid_vote_extensions_residuals(extensions);
```
james-chf, line 148: similarly here we are not filtering anything, we are getting the validation result for each vote extension we have. (each result including some useful things like the signed extension and the voting power if the validation did pass).

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -67,116 +87,183 @@ where
...
+                    };
+
+                    if filtered_extensions.into_iter().all(|maybe_ext| {
+                        maybe_ext
+                            .map(|(power, _)| {
+                                voting_power += FractionalVotingPower::new(
+                                    u64::from(power),
+                                    total_power,
+                                )
+                                .unwrap_or_default();
```
james-chf, line 166: we should hard `.unwrap()`/`.expect()` here or otherwise match on the result of constructing the `FractionalVotingPower`. Although I'm not sure how we would handle the error (or if we should ever encounter it - the voting power values are coming from our storage so we should always be able to construct a valid `FractionalVotingPower`?)

james-chf (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -185,16 +179,23 @@ mod prepare_block {
...
                 let validator_addr = vote_extension.data.validator_addr;
 
                 // update voting power
-                voting_power += u64::from(validator_voting_power);
+                let validator_voting_power = u64::from(validator_voting_power);
+                voting_power += FractionalVotingPower::new(
+                    validator_voting_power,
+                    total_voting_power,
+                )
+                .unwrap_or_default();
```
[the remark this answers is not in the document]
sug0: Yeah, something has gone terribly wrong if our storage was corrupted in such a way that creating a new `FractionalVotingPower` would fail. I'll add an expect here.

sug0:
I have opened two issues for the remaining items which were left open.

https://github.com/anoma/namada/issues/238
https://github.com/anoma/namada/issues/239

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -88,13 +85,10 @@ mod prepare_block {
...
-                            "The genesis block should not contain vote \
-                             extensions"
-                        );
-                        // TODO: maybe slash validators who claim to have
-                        // seen vote extensions at H=0
-                        return vec![];
+                        unreachable!(
+                            "We already handle this scenario in \
+                             validate_vote_extension."
+                        )
```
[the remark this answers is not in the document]
james-chf: Sounds good, I will check this at some point as I'm not 100% sure either

james-chf (approved):

PR #225, opened 2022-07-27
Reduce logging and transactions in eth-bridge-integration

james-chf:
Some noise I've seen while testing eth bridge branches
- sleep between logging attempts to connect to `geth`
- don't log if no Ethereum events were received in vote extensions in the previous round (we also won't derive/apply a state update transaction either)
- make validators log only if they receive >=1 Ethereum events from their Ethereum node

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/mod.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/mod.rs
@@ -139,10 +139,7 @@ pub mod eth_fullnode {
                 if let Err(error) = client.eth_syncing().await {
                     // This is very noisy and usually not interesting.
                     // Still can be very useful
-                    tracing::debug!(
-                        ?error,
-                        "Couldn't connect to Geth, will retry"
-                    );
+                    tracing::debug!(?error, "Couldn't check Geth sync status");
                 }
                 tokio::time::sleep(SLEEP_DUR).await;
             }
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -132,11 +132,15 @@ pub mod eth_fullnode {
...
+                if let Err(error) = client.eth_syncing().await {
                     // This is very noisy and usually not interesting.
                     // Still can be very useful
-                    tracing::debug!("Error trying to connect to Geth: {:?}", e);
+                    tracing::debug!(
+                        ?error,
+                        "Couldn't connect to Geth, will retry"
+                    );
                 }
+                std::thread::sleep(std::time::Duration::from_secs(1));
```
sug0, line 143: Can we have a `const SLEEP_DUR: Duration = ...` constant value here? Also, we're in an async function, this sleep will block other tasks. Let's switch to `tokio::time::sleep(SLEEP_DUR).await`

sug0 (changes requested):

apps/src/lib/node/ledger/ethereum_node/mod.rs
sug0: The constant change is optional tbh. In my opinion it's better than leaving seemingly random params in a function/method call.

james-chf: Yeah a named constant is clearer, have changed to using `tokio::time::sleep` as well

batconjurer (approved):

sug0 (approved):

PR #230, opened 2022-07-28
Clarify inflation system calculation details

cwgoes:
This PR updates the inflation system section of the Namada specs to clarify PD-controller calculation details.

For simplicity, I think we can just use fully separate PD-controllers for each asset in the shielded pool for which Namada decides to pay rewards. This requires that we pick parameters and caps independently, but otherwise we have to deal with contention between controllers, which is complicated and likely confusing to users, so I think it's worth the additional parameters.

Also for simplicity, I'm assuming that we do all inflation calculations once-per-epoch and quantise the PD-controllers at that frequency.

### DIFF
--- a/documentation/specs/src/economics/inflation-system.md
+++ b/documentation/specs/src/economics/inflation-system.md
@@ -1,119 +1,80 @@
 ## Inflation system
 
-In general, inflation refers to the process of a currency losing its 
-purchasing power over time. While this is a classical economic phenomenon, the way cryptocurrencies are produced permits great control over money supply, and doing so cleverly can have positive effects such as increasing incentives. Here we use "inflation" as a synonym for "token printing". The protocol controls the Namada token NAM (the native staking token), which is programmatically minted to pay for algorithmically measurable public goods - proof-of-stake security and shielded pool usage - and out-of-band public goods.
-
-1. Proof-of-stake rewards, which are paid into the reward distribution mechanism in order to distribute them to validators and delegators.
-2. Shielded pool rewards, which are locked in a way such that they can be eventually paid to users who kept tokens in the shielded pool.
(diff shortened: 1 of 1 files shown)


adrianbrink (approved):
Looks good and massively simplifies things.

documentation/specs/src/economics/inflation-system.md
```diff
@@ -1,119 +1,80 @@
...
-If $I_{L}^{min}< I_{L}$ then $\frac{dI_{L}}{dt}=max(A(t),0)$.
+First, we start with the following fixed (governance-alterable) parameters:
 
-If $I_{L}< I_{L}^{max}$ then $\frac{dI_{L}}{dt}=min(A(t),0)$.
+- $Cap_{PoS}$ is the cap of proof-of-stake reward rate, in units of percent per annum
+- $Cap_{SP-A}$ is the cap of shielded pool reward rate for each asset $A$, in units of percent per annum
+- $R_{PGF}$ is the public goods funding reward rate, in units of percent per annum
+- $R_{PoS-Target}$ is the target staking ratio (genesis default 2/3)
+- $R_{SP-A-Target}$ is the target amount of asset $A$ locked in the shielded pool (separate value for each asset $A$)
+- $EpochsPerYear$ is the number of epochs per year (genesis default 365)
```
tzemanovic, line 28: we currently have these params https://github.com/anoma/namada/blob/958715216c93d76b29b59f4765c38a78c4d119df/shared/src/ledger/parameters/mod.rs#L141 
Should we use `min_duration` to estimate `EpochsPerYear`? We should probably make sure that `min_duration > min_num_of_blocks * consensus_timeout_commit` generously enough so that we're not waiting for blocks to switch epoch

cwgoes: Yes, we should use `min_duration` to estimate `EpochsPerYear` (but the latter should still be constant)

PR #234, opened 2022-07-28
CI run wasm step in container

Fraccaman:


juped:
still failing wasm checksums

tzemanovic:
> still failing wasm checksums

it looks like it just wasn't up-to-date on main

juped:
did you run make build-wasm-scripts-docker

tzemanovic:
sry, bad commit - accidentally added vp_masp

juped:
ok, so the main thing here is in fact changes to the dockerfile?

docker/namada-wasm/Dockerfile
```diff
@@ -1,18 +1,16 @@
 # This docker is used for deterministic wasm builds
 
 # The version should be matching the version set in wasm/rust-toolchain.toml
-FROM rust:1.61.0
+FROM rust:1.61
```
juped, line 4: should stay 1.61.0; right now 1.61 is an alias for it but it might have a patch release

Fraccaman: yeah, lets keep `rust:1.61.0`. I'll make a commit

PR #240, opened 2022-07-29
host env VP write check

tzemanovic:
to make sure that we're checking VPs written to storage in txs, the `tx_write` host function should disallow direct raw-bytes updates of VP storage keys. Instead, there's `tx_update_validity_predicate` that performs some validation on the VP code


shared/src/vm/host_env.rs
```diff
@@ -807,6 +812,10 @@ where
     tracing::debug!("tx_update {}, {:?}", key, value);
 
     let key = Key::parse(key).map_err(TxRuntimeError::StorageDataError)?;
+    // VPs must be updated via `tx_update_validity_predicate`
+    if key.is_validity_predicate().is_some() {
+        return Err(TxRuntimeError::CannotDirectlyUpdateVp);
+    }
```
tzemanovic, line 818: actually, maybe this would be a bit nicer

```
    // VP update must be validated
    if key.is_validity_predicate().is_some() {
        tx_add_gas(env, code.len() as u64 * WASM_VALIDATION_GAS_PER_BYTE)?;
        validate_untrusted_wasm(&code).map_err(TxRuntimeError::UpdateVpInvalid)?;
    }
```

with having common validation function with `update_validity_predicate`

PR #241, opened 2022-07-29
Refactor some helpers in queries.rs

james-chf:
Motivation is for some of this functionality to be available outside of the `Shell` (e.g. in `apply_tx` some of the current `queries.rs` functionality is duplicated in [this PR](https://github.com/anoma/namada/pull/210/files#diff-e02b60ae4f03bdbcdd62ba79596497e65f9f32ecad4d98a4f353cff428919f6e))

### DIFF
--- a/apps/src/lib/node/ledger/shell/queries.rs
+++ b/apps/src/lib/node/ledger/shell/queries.rs
@@ -282,12 +282,6 @@ pub(crate) trait QueriesExt {
         token: &Address,
         owner: &Address,
     ) -> std::result::Result<Amount, String>;
-    fn read_storage_value(
-        &self,
-        key: &Key,
-        height: BlockHeight,
-        is_proven: bool,
-    ) -> response::Query;
     fn get_evidence_params(
         &self,
         epoch_duration: &EpochDuration,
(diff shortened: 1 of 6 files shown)


batconjurer (changes requested):
It seems we can make all the query functions struct methods. Can we do that? `self.` is more readable than `fn(&self.storage,..` I think.

sug0:
right, these could be struct methods of `Storage`

james-chf:
sure it makes sense for most if not all of these methods to be on an `impl` for `Storage`

james-chf:
done a bigger refactor to put most of the `queries.rs` methods that were on `Shell` onto a `QueriesExt` trait that is then implemented for `Storage`. These methods are higher level than the ones directly implemented on `Storage`. 

I've left on `Shell` the methods that seemed to be related to the ABCI /query method (i.e. were returning a Tendermint `response::Query`). Had to refactor also the `get_balance()` method to not use an ABCI-ish method that was left on `Shell`.

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -281,22 +206,169 @@ where
...
+        &self,
+        address: &Address,
+        epoch: Option<Epoch>,
+    ) -> std::result::Result<(VotingPower, common::PublicKey), Error>;
+    fn get_validator_from_tm_address(
+        &self,
+        tm_address: &[u8],
+        epoch: Option<Epoch>,
+    ) -> std::result::Result<Address, Error>;
+}
```
sug0, line 305: Maybe move the documentation of each method you implemented in `Storage` here.

sug0: Otherwise lgtm!

sug0 (approved):

batconjurer (approved):

PR #245, opened 2022-08-01
Simplify fee mechanism

cwgoes:
Per discussion, l propose that we start with a simple fee system for Namada, which should be sufficient for the immediate future.


documentation/specs/src/economics/fee-system.md
```diff
@@ -1,20 +1,9 @@
...
 
-Namada follows a [tipless version](https://arxiv.org/pdf/2106.01340.pdf) of the EIP 1559 scheme. In contrast with the original EIP 1559, the transaction fee of this tipless version consists solely of a base fee, with no tip. The base fee increases whenever blocks are fuller than the desired capacity and decreases when the blocks haven't reached this capacity (i.e. a P-controller). Namada uses a target block fullness of 0.5 (adjustable by governance).
+Namada transaction fees are paid directly to the reward distribution scheme. Fees can be paid in any fungible token which is a member of a whitelist controlled by Namada governance. Governance also sets minimum fee rates (which can be periodically updated so that they are usually sufficient) which transactions must pay in order to be accepted (but they can always pay more to encourage the proposer to prioritise them). When using the shielded pool, transactions can also unshield tokens in order to pay the required fees.
```
tzemanovic, line 5: I might be missing some crucial piece, but what happens with the tokens from fees in the reward distribution scheme? Doesn't it only deal with inflation of the native token?

cwgoes: Ah yes, this is an excellent question. We should be able to implement the same reward distribution mechanism for other tokens, just without compounding inflation (since there is none), which can be done by adding periods instead of multiplying them. However, I think doing this is probably not worth it - let's just pay fees to the block producer (which is then incentive-compatible anyways).

cwgoes: Updated.

FatemeShirazi: Does this mean that we are giving the fees to future block producers and nothing to current block producers? Are any fees burned?

FatemeShirazi: I also have a question about using multiple whitelisted assets for paying fees, the asset that is chosen is revealed to everyone no? Wouldnt that hurt MASP privacy?

cwgoes: 1. for now, yes - current block producer only
2. yes, but users can choose which asset (so they can still send USDC and pay with NAM, for example)

FatemeShirazi (approved):

Issue #251, opened 2022-08-01
Restarting a sole validator node panics due to receiving an empty vote extension


james-chf:
The following panic can happen when running a chain with a single genesis validator, after having stopped and restarted the node.

```
2022-08-01T19:51:01.768890Z ERROR namada_apps::node::ledger::shell::vote_extensions::extend_votes: Failed to deserialize signed vote extension err=Custom { kind: InvalidInput, error: "Unexpected length of input" }
2022-08-01T19:51:01.768922Z ERROR namada_apps::node::ledger::shell::prepare_proposal::prepare_block: Tendermint has decided on a block including vote extensions reflecting <= 2/3 of the total stake
The application panicked (crashed).
Message:  internal error: entered unreachable code: Honest Namada validators will always sign a VoteExtension, even if no Ethereum events were observed at a given block height. In fact, a quorum of signed empty VoteExtension commits the fact no events were observed by a majority of validators. Likewise, a Tendermint quorum should never decide on a block including vote extensions reflecting less than or equal to 2/3 of the total stake. These scenarios are virtually impossible, so we will panic here.
Location: apps/src/lib/node/ledger/shell/prepare_proposal.rs:95

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: RecvError(())
Location: /Users/main/.cargo/git/checkouts/tower-abci-0d01b039e0b7a0c9/f646338/src/server.rs:163
```

## To reproduce

The issue definitely happens on Namada built from commit [5a2fd9d1e](https://github.com/anoma/namada/commit/5a2fd9d1eac5deb78f5d998dadb88b40f779708b) and Tendermint built from commit [29e5fbcc6](https://github.com/tendermint/tendermint/commit/29e5fbcc6).

- Download [anoma-dev.c21a27c05f23fae3d2f5bac3ad.tar.gz](https://github.com/anoma/namada/files/9299669/anoma-dev.c21a27c05f23fae3d2f5bac3ad.tar.gz) - this contains an `.anoma` directory preconfigured with a single genesis validator. Extract to the `namada` repo directory.
- Run the following to build and start running the ledger node:

```shell
# build namada with ABCI++ functionality enabled
cargo build \
    --no-default-features \
    --features 'namada/ABCI-plus-plus namada_apps/ABCI-plus-plus' \
    --package namada_apps

[...]

PR #252, opened 2022-08-02
Handle vote extensions when the chain has only a single validator

james-chf:
Fixes https://github.com/anoma/namada/issues/251

### DIFF
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -4,12 +4,13 @@
 mod prepare_block {
     use std::collections::{BTreeMap, HashMap, HashSet};
 
+    use eyre::eyre;
     use namada::types::ethereum_events::vote_extensions::{
         FractionalVotingPower, MultiSignedEthEvent, VoteExtensionDigest,
     };
     use namada::types::transaction::protocol::ProtocolTxType;
     use tendermint_proto::abci::{
-        ExtendedCommitInfo, ExtendedVoteInfo, TxRecord, Validator,
+        ExtendedCommitInfo, ExtendedVoteInfo, TxRecord,
     };
 
     use super::super::vote_extensions::deserialize_vote_extensions;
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -42,8 +42,18 @@ mod prepare_block {
...
-                let mut txs =
-                    self.build_vote_extensions_txs(req.local_last_commit);
+                let mut txs = match req.local_last_commit {
+                    Some(local_last_commit) => {
+                        self.build_vote_extensions_txs(local_last_commit)
+                    }
+                    None => {
+                        tracing::info!(
+                            "No local last commit, skipping including vote \
+                             extensions"
```
sug0, line 52: can you somehow fit "Ethereum events" into the log message? so that we can distinguish these vote extensions from the validator set update log messages.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -68,18 +78,22 @@ mod prepare_block {
...
         fn build_vote_extensions_txs(
             &mut self,
-            local_last_commit: Option<ExtendedCommitInfo>,
+            local_last_commit: ExtendedCommitInfo,
         ) -> Vec<TxRecord> {
+            if is_sole_validators(&local_last_commit.votes) {
+                tracing::info!(
+                    "Sole validator and block proposer, no vote extensions to \
+                     include this round"
+                );
```
sug0, line 87: same here

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -238,9 +238,17 @@ mod extend_votes {
...
-                        "Failed to deserialize signed vote extension",
+                        validator_tendermint_addr = ?vote.validator,
+                        vote_extension_len = vote.vote_extension.len(),
+                        "Failed to deserialize signed vote extension, will \
+                         ignore",
+                    );
+                    tracing::debug!(
+                        ?vote.vote_extension,
+                        validator_tendermint_addr = ?vote.validator,
+                        "Raw undeserializable vote extension for validator",
```
sug0, line 251: same here

sug0 (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
james-chf: In this case, we don't get a vote extension at all (so neither Ethereum events nor validator set updates). I'm not sure when or if this code path can happen - in the [protobuf](https://github.com/tendermint/tendermint/blob/a1104b98d2d2f04855d59af98036a2278aab23d2/proto/tendermint/abci/types.proto#L131) the `local_last_commit` field is marked as not nullable and the generated Go struct has it as a non-optional field. For the type we are using from `tendermint-rs`, this field is an `Option<ExtendedCommitInfo>`. Need to look into whether the types are out of sync somehow

james-chf: Re the error in https://github.com/anoma/namada/issues/251 , `local_last_commit` was `Some(_)` but then the single `ExtendedVoteInfo` inside it had an empty `vote_extension` field. So this code path wasn't being hit in the end

apps/src/lib/node/ledger/shell/prepare_proposal.rs
james-chf: We match here when we have a single empty vote extension from ourselves, similar to above we won't have either Ethereum events or validator set updates. I guess it raises the question of how we should handle a node restarting and launching into a round where where it needs to aggregate the validator set update vote extensions from the previous round.

apps/src/lib/node/ledger/shell/vote_extensions.rs
james-chf: :+1: mentioned Ethereum events in the logs here

james-chf:
Having a special match for when there is a sole validator doesn't seem like a robust way to handle this, still need to make some changes to this WIP PR. I'm not sure of the best way to handle this. I'm not even sure right now if the fact that Tendermint gives us a completely empty vote extension from ourselves as block proposer is an expected behaviour from Tendermint, or some kind of bug.

Some things we can change to be more robust:
- if we can't construct the `VoteExtensionDigest` for whatever reason, we should still propose the block without the protocol transaction - but log at error level that this block is likely to be rejected by other validators. This doesn't apply to the genesis block, which we expect not to have a vote extension protocol transaction. In the case of a single validator chain, we should be able to progress as `ProcessProposal` shouldn't be called. This isn't the solution to the issue though, as we will still be making a spurious error log.
 - right now in `compress_vote_extensions`, when checking that we got vote extensions that sum to >2/3 voting power, we ignore ones we can't deserialize to `SignedExt`. Instead we should count the voting power of any vote extensions we have (including undeserializable ones), and then we should always expect to reach the 2/3 threshold as an honest block proposer. This handles the case where we get an empty vote extension from ourselves - we can't construct the `VoteExtensionDigest` but won't log the "Tendermint has decided on a block including vote extensions reflecting <= 2/3 of the total stake" which isn't the case

james-chf:
There are multiple things to solve here which can be done in separate PRs, will make new PRs/open issue(s)

Issue #102, opened 2022-04-26
Have a Cargo workspace for wasm crates


james-chf:
We have the following crates
- [x] `wasm/wasm_source`
- [x] `wasm/tx_template`
- [x] `wasm/vp_template`
- [ ] `wasm_for_tests/wasm_source`

These crates share the same sort of directory structure, Makefile targets and Rust dependencies (`Cargo.lock`) so it could make sense to contain them in a Cargo workspace e.g. located at `wasm/Cargo.toml` (and have `wasm_for_tests/` move under the `wasm/` directory).

tzemanovic:
everything except for `wasm_for_tests` is now in wasm workspace (#258)

cwgoes:
@james-chf is this still relevant?

james-chf:
It's still relevant and should eventually be done - it would mean all our wasms are built under one workspace so they would be easier to work with and keep in sync e.g. only one `Cargo.toml` to patch and keep updated, one less `Cargo.lock` file in the repo, etc. 

One issue currently is that there are various duplicate references and hardcoded paths to our `wasm_for_tests` directory in the repo at the moment, that should be cleaned up first before we migrate the folder - I've linked https://github.com/anoma/namada/pull/893 which starts on doing that.

james-chf:
After https://github.com/anoma/namada/pull/893 lands this should be fairly straightforward to do

PR #258, opened 2022-08-02
Try out a Cargo workspace for wasm/ directory

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1096 without rebase

Relates to anoma/namada#102

If this works, we could later include `wasm_for_tests/` in it


tzemanovic:
merged to main

PR #267, opened 2022-08-02
Permit proposal of blocks without vote extension transactions

james-chf:
Workaround to https://github.com/anoma/namada/issues/251


james-chf:
We will not want to allow any blocks except the genesis block to not have vote extension transactions. We can workaround this issue for now by recreating the Anoma chain during development when we want to start up the node again.

PR #257, opened 2022-08-02
Murisi/masp history split

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1234

The MASP client and incentive integration work. Contained features:
* Multi asset shielded/shielding/unshielding transactions and balance checks
* Wallet support for spending keys, viewing keys, payment addresses, and pinned payment addresses
* Burn and mint for shielded assets allowing rewards to be received for shielded funds
* One time payment addresses enabling transaction scanning to be skipped
* Gas payments originating from shielded pool
* Historical transparent and shielded transaction viewing

PR #268, opened 2022-08-02
temp: masp history split debugging

tzemanovic:


tzemanovic:
pushed some fixes back to #257

PR #272, opened 2022-08-03
Ensure Ethereum oracle starts up

### LINKED ISSUE #269: Ethereum oracle may not be starting up properly
Adding a log line just before when the oracle starts running 

https://github.com/anoma/namada/blob/94735d3d553cb5bfa349fe9b818f28816a8a6621/apps/src/lib/node/ledger/mod.rs#L353-L356

e.g.

```rust
        let oracle = local.spawn_local(async move {
            tracing::info!("started oracle");
            ethereum_node::run_oracle(&ethereum_url, eth_sender, abort_sender)
                .await
        });
```

When starting and stopping the ledger, the "started oracle" log shows when the ledger is shutting down rather than starting up.

```
...
ERROR[08-03|10:17:48.305] Failed to journal state snapshot         err="snapshot [0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544] missing"
INFO [08-03|10:17:48.305] Writing clean trie cache to disk         path=/Users/main/Library/Ethereum/geth/triecache threads=10
2022-08-03T09:17:48.306262Z  INFO namada_apps::node::ledger: Tendermint node is no longer running.
INFO [08-03|10:17:48.325] Persisted the clean trie cache           path=/Users/main/Library/Ethereum/geth/triecache elapsed=19.998ms
INFO [08-03|10:17:48.325] Blockchain stopped
2022-08-03T09:17:52.916963Z ERROR namada_apps::node::ledger: Failed to receive a response from Ethereum: channel closed
2022-08-03T09:17:52.916964Z  INFO namada_apps::node::ledger: Ethereum fullnode is no longer running.
2022-08-03T09:17:52.917034Z  INFO namada_apps::node::ledger::broadcaster: Shutting down broadcaster...
2022-08-03T09:17:52.917047Z  INFO namada_apps::node::ledger: Broadcaster is no longer running.
2022-08-03T09:17:52.917065Z  INFO namada_apps::node::ledger: started oracle
2022-08-03T09:17:52.921400Z  INFO namada_apps::node::ledger: Anoma ledger node has shut down.
```

The oracle and the geth tasks should run in parallel.

james-chf:
Relates to #269

Tested manually (for now), both with the `namada_apps/eth-fullnode` feature enabled and disabled. A log is emitted indicating the oracle is starting right after we've checked `geth` is reachable

```
...
namada_apps::node::ledger::ethereum_node::eth_fullnode: Finished syncing
2022-08-03T16:11:16.241376Z  INFO namada_apps::node::ledger::ethereum_node::oracle::oracle_process: Ethereum event oracle is starting
...
```

A log is emitted at shutdown indicating it stopped.

```
...
2022-08-03T15:48:13.114802Z  INFO namada_apps::node::ledger: Broadcaster is no longer running.
2022-08-03T15:48:13.114785Z  INFO namada_apps::node::ledger: Anoma ledger node has shut down.
2022-08-03T15:48:13.115287Z  INFO namada_apps::node::ledger::ethereum_node::oracle::oracle_process: Ethereum event oracle is no longer running
```

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/oracle.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/oracle.rs
@@ -88,12 +88,29 @@ pub mod oracle_process {
     /// Set up an Oracle and run the process where the Oracle
     /// processes and forwards Ethereum events to the ledger
     pub async fn run_oracle(
-        url: &str,
+        url: impl AsRef<str>,
         sender: UnboundedSender<EthereumEvent>,
         abort_sender: Sender<()>,
     ) {
-        let oracle = Oracle::new(url, sender, abort_sender);
-        run_oracle_aux(oracle).await;
+        let url = url.as_ref().to_owned();
+        tokio::task::spawn_blocking(move || {
+            let rt = tokio::runtime::Handle::current();
+            rt.block_on(async move {
+                let local = tokio::task::LocalSet::new();
+                local
+                    .run_until(async move {
+                        tracing::info!("Ethereum event oracle is starting");
+
+                        let oracle = Oracle::new(&url, sender, abort_sender);
+                        run_oracle_aux(oracle).await;
+
+                        tracing::info!(
+                            "Ethereum event oracle is no longer running"
+                        );
+                    })
+                    .await
+            });
+        });
     }
 
     /// Given an oracle, watch for new Ethereum events, processing
--- a/apps/src/lib/node/ledger/ethereum_node/oracle.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/oracle.rs
@@ -87,11 +87,11 @@ pub mod oracle_process {
 
     /// Set up an Oracle and run the process where the Oracle
     /// processes and forwards Ethereum events to the ledger
-    pub async fn run_oracle(
+    pub fn run_oracle(
         url: impl AsRef<str>,
         sender: UnboundedSender<EthereumEvent>,
         abort_sender: Sender<()>,
-    ) {
+    ) -> tokio::task::JoinHandle<()> {
         let url = url.as_ref().to_owned();
         tokio::task::spawn_blocking(move || {
             let rt = tokio::runtime::Handle::current();
@@ -110,7 +110,7 @@ pub mod oracle_process {
                     })
                     .await
             });
-        });
+        })
     }
 
     /// Given an oracle, watch for new Ethereum events, processing
--- a/apps/src/lib/node/ledger/ethereum_node/oracle.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/oracle.rs
@@ -7,6 +7,7 @@ pub mod oracle_process {
     use num256::Uint256;
     use tokio::sync::mpsc::UnboundedSender;
     use tokio::sync::oneshot::Sender;
+    use tokio::task::LocalSet;
     #[cfg(not(test))]
     use web30::client::Web3;
 
@@ -93,11 +94,12 @@ pub mod oracle_process {
         abort_sender: Sender<()>,
     ) -> tokio::task::JoinHandle<()> {
         let url = url.as_ref().to_owned();
+        // we have to run the oracle in a [`LocalSet`] due to the web30
+        // crate
         tokio::task::spawn_blocking(move || {
             let rt = tokio::runtime::Handle::current();
             rt.block_on(async move {
-                let local = tokio::task::LocalSet::new();
-                local
+                LocalSet::new()
                     .run_until(async move {
                         tracing::info!("Ethereum event oracle is starting");
 
--- a/apps/src/lib/node/ledger/ethereum_node/test_tools.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/test_tools.rs
@@ -35,11 +35,12 @@ pub mod mock_oracle {
     pub fn run_oracle(
         _: impl AsRef<str>,
         _: UnboundedSender<EthereumEvent>,
-        mut abort: Sender<()>,
+        abort: Sender<()>,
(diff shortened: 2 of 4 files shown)


apps/src/lib/node/ledger/mod.rs
```diff
@@ -350,9 +345,25 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
...
+                            eth_sender,
+                            abort_sender,
+                        )
+                        .await;
+                        tracing::info!(
+                            "Ethereum event oracle is no longer running"
+                        );
+                    })
+                    .await
+            });
```
sug0, line 366: we can probably move this `LocalSet` logic to `ethereum_node::run_oracle`. better yet, move the whole `tokio::task::spawn_blocking` block there, so this ugly hack is not exposed here.

sug0 (changes requested):

apps/src/lib/node/ledger/mod.rs
sug0: we would return a `JoinHandle` from `ethereum_node::run_oracle`

sug0: yeah, it's necessary because of the `web30` crate.

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
```diff
@@ -30,17 +30,35 @@ pub mod mock_oracle {
...
+                LocalSet::new()
+                    .run_until(async move {
+                        tracing::info!(
+                            "Mock Ethereum event oracle is starting"
+                        );
+
+                        loop {
+                            if abort.is_closed() {
+                                break;
+                            }
```
james-chf, line 52: There is probably a better way to busy wait here (`tokio::time::sleep` inbetween `abort.is_closed()` checks?).

sug0: Use this
https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.poll_closed

sug0 (changes requested):

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
sug0: Another thing is that you probably don't need to use `LocalSet` in the mock `run_oracle`. Just do a regular `tokio::spawn`, and return the handle from that.
```rust
tokio::spawn(async move {
    tracing::info!(
        "Mock Ethereum event oracle is starting"
    );

    poll_fn(|cx| abort.poll_closed(cx)).await;

    tracing::info!(
        "Mock Ethereum event oracle is no longer running"
    );
})
```
This should probably work

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -87,13 +88,31 @@ pub mod oracle_process {
 
     /// Set up an Oracle and run the process where the Oracle
     /// processes and forwards Ethereum events to the ledger
-    pub async fn run_oracle(
-        url: &str,
+    pub fn run_oracle(
+        url: impl AsRef<str>,
```
batconjurer, line 92: What's the motivation for switching to `impl AsRef<str>`?

james-chf: Before we could use the `url: &str` reference directly, now we need an owned `url` string of some sort we can move into `tokio::task::spawn_blocking(...)`. We could also accept `url: String` or call `let url = url.to_owned()` and keep using `url: &str`. In actuality we were still passing `&str` at the sole call site, I updated it to pass the `ethereum_url` `String`.

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
james-chf: This works, had to change the `abort: Sender<()>` parameter to be `mut abort: Sender<()>` as `poll_closed` requires mutability on `abort`. Should we change the parameter on the real `run_oracle` as well to be consistent? I need to learn why we need to do this - since we were passing ownership of the `Sender<()>` struct already, I would have thought we should be able to mutate it if we wanted to without having to add an extra keyword.

james-chf: We can keep using `abort: Sender<()>` as a parameter and instead do

```rust
            let mut abort = abort;
            poll_fn(|cx| abort.poll_closed(cx)).await;
```

I guess the compiler just needs an indication we will mutate this variable. The caller gives us ownership so they don't need to worry about it.

sug0: if you indicate a variable will be mutated, at the function signature level, it will have no consequence on the type of the function. for instance:

This function

```rust
fn do_the_thing(mut x: i32) -> i32 { x = 0; x }
```

Is equivalent to this function, type wise.

```rust
fn do_the_thang(x: i32) -> i32 { x }
```

james-chf: Oh interesting, so the `mut` is just for the convenience of the code in the function so we don't have to redeclare. To the caller it is still the same. I thought it would change the type, I'll change this back quickly to use `mut` in the params

james-chf:
Force pushed a rebase onto `eth-bridge-integration`

apps/src/lib/node/ledger/ethereum_node/oracle.rs
batconjurer: I see. Ok :+1:

batconjurer (approved):

Issue #242, opened 2022-08-01
Implement validator set update vote extensions


sug0:
1. First, decouple vote extension logic from Ethereum events logic, by renaming types/functions/methods accordingly.
    1. For instance, instead of `VoteExtensionDigest`, we would have `EthereumEventsDigest`, or something equivalent.
    2. Merged in https://github.com/anoma/namada/pull/243
2. Introduce some baseline Rust types necessary to implement the rest of the changes.
    1. The types on the Rust side need to map to types on the Ethereum smart contract's side.
    2. Merged in https://github.com/anoma/namada/pull/273
3. Implement PrepareProposal changes.
    1. Merged in https://github.com/anoma/namada/pull/286
4. Implement ProcessProposal changes.
    1. Merged in https://github.com/anoma/namada/pull/372
5. Implement shimming of vote extensions, to accomodate to [our Tendermint 0.37.x version](https://github.com/tendermint/tendermint/tree/0b58342a46b1917a40d506e48e87448dcc8b3131), which ships without vote extensions enabled.
    1. Merged in https://github.com/anoma/namada/pull/437
6. Implement a way to call Ethereum bridge smart contract with the validator
   set update vote extension digest decision from Tendermint. In other words,
   we need a relayer from Namada to Ethereum.
    1. Merged in https://github.com/anoma/namada/pull/1157
7. Implement FinalizeBlock changes.
    1. Merged in https://github.com/anoma/namada/pull/683
8. Implement extend vote and verify vote extension calls.
    1. Depends on https://github.com/anoma/namada/issues/133.
    3. Partially implemented in https://github.com/anoma/namada/pull/286
    4. The rest of the required changes were merged in https://github.com/anoma/namada/pull/471
9. ??? Other misc. changes.
    1. We need two secp256k1 ethereum keys in `ShellMode`, one for governance (on Ethereum) and another for signing validator set update proofs. Signatures can't be from Namada validator keys. Related to https://github.com/anoma/namada/issues/133.
        i. Merged in https://github.com/anoma/namada/pull/278
        ii. Key storage was merged in https://github.com/anoma/namada/pull/471
    5. Store the version of `Bridge.sol` in storage. Related to https://github.com/anoma/namada/issues/249.
---
These changes can be implemented as separate PRs. Notes for these changes can be accessed [here](https://hackmd.io/aKXu06tkSTqsoRVCTn_XMA?view=).

cwgoes:
[...]

PR #273, opened 2022-08-03
Introduce baseline types required to implement validator set updates

sug0:
Implements the second item of issue #242


sug0:
I haven't added unit tests to these yet. I might leave that to the next PR (PrepareProposal), since I'll write the validation code then. idk, what do you think?

james-chf:
There is one more `Cargo.lock` -  `make -C wasm_for_tests/wasm_source check` - not currently compiling

shared/src/proto/types.rs
```diff
@@ -51,54 +52,47 @@ pub struct SignedTxData {
...
+/// to sign data in a [`Signed`] wrapper.
+#[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub enum SerializeWithBorsh {}
+
+/// A generic signed data wrapper for serialize-able types.
+///
+/// The default serialization method is [`BorshSerialize`].
 #[derive(
-    Clone, Debug, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+    Eq, Clone, Debug, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
```
james-chf, line 64: I'm guessing `Eq` should only be derived for the case where `T: PartialEq`? It might need to be manually implemented in that case

james-chf: *`T: PartialEq + Eq`

sug0: `Eq` doesn't add any methods, it just means we a have a full equivalence relation.
https://doc.rust-lang.org/stable/std/cmp/trait.Eq.html

sug0: if you have `PartialEq` implemented, `Eq` can be derived

james-chf: I mean we can (or should?) only derive `Eq` in the case where `T: Eq`. As you can sign `T`s which don't implement `PartialEq`/`Eq`. I'm surprised the compiler lets you annotate `Eq` on `Signed<T, S>` without `T` also having the `Eq` trait bound though so I could be misunderstanding something

james-chf: something like

sug0: the compiler already does that automatically. it'll only implement `Eq` if the type params are also `Eq`

sug0: actually not true, it'll implement it if the type params are `PartialEq`. ehhhh
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=095498c880c5653848415b9228642fb7

alright cool I'll manually implement `Eq`

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -0,0 +1,297 @@
...
+use crate::types::key::common::{self, Signature};
+use crate::types::key::{SigScheme, VerifySigError};
+
+// the namespace strings plugged into validator set hashes
+const BRIDGE_CONTRACT_NAMESPACE: &str = "bridge";
+const GOVERNANCE_CONTRACT_NAMESPACE: &str = "governance";
+
+/// Wrapper type for [`ethereum::Address`]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct EthAddr(pub ethereum::Address);
```
batconjurer, line 25: We already have the type in `src/types/ethereum_events` except that we call the `to_fixed_bytes` method on the `ethereum::Address` type to get the raw bytes.

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -0,0 +1,297 @@
...
+            ),
+            epoch_to_token(ext.epoch),
+        ])
+    }
+
+    /// Sign this [`Vext`] with an Ethereum key.
+    ///
+    /// For more information, check the Ethereum bridge smart contract code:
+    ///   - <https://github.com/anoma/ethereum-bridge/blob/main/contracts/contract/Bridge.sol#L186>
+    pub fn new_abi_encoded(keypair: &common::SecretKey, ext: Vext) -> Self {
```
batconjurer, line 127: Can't we just implement a `new` method in the `types/proto.rs` defns for `Signed<T,S>`? And add an ABI marker struct there. That seemed like the whole point of that change and we get type API consistency.

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -0,0 +1,297 @@
...
+        let to_sign = Self::serialize_vext(&ext);
+        let sig = common::SigScheme::sign(keypair, &to_sign);
+        Self::new_from(ext, sig)
+    }
+
+    /// Verify the signature of a [`Vext`], signed by some
+    /// Ethereum key.
+    pub fn verify_abi_encoded(
```
batconjurer, line 135: Similary, this should just be a `verify` method in `types/proto.rs`. In fact, it might be drying to change `Signed<T,S>` so that `S` implements a trait that gives access to a serialization function. Then all of this logic only needs to be written once and we only need to implement this trait on marker structs. So it is a much more extensible design.

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// TODO: the validator's address is temporarily being included
+    /// until we're able to map a Tendermint address to a validator
+    /// address (see https://github.com/anoma/namada/issues/200)
+    pub validator_addr: Address,
+    /// The new [`Epoch`].
+    ///
+    /// Since this is a monotonically growing sequence number,
+    /// it is signed together with the rest of the data to
+    /// prevent replay attacks on validator set updates.
+    pub epoch: Epoch,
```
batconjurer, line 166: It also allows us to look up with valset is needed to verify signatures of user txs coming to the eth side of the bridge.

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -0,0 +1,297 @@
...
+
+/// Provides a mapping between [`EthAddr`] and [`VotingPower`] instances.
+pub type VotingPowersMap = HashMap<EthAddr, VotingPower>;
+
+/// This trait contains additional methods for a [`HashMap`], related
+/// with validator set update vote extensions logic.
+pub trait VotingPowersMapExt {
+    /// Returns the keccak hash of this [`VotingPowersMap`]
+    /// to be signed by an Ethereum validator key.
+    fn get_bridge_hash(&self, epoch: Epoch) -> [u8; 32];
```
batconjurer, line 187: A super nit, but I like using the `Hash` type instead unless there is a really good reason not to. I think it just make intentions clearer in the code.

batconjurer (changes requested):

shared/src/proto/types.rs
james-chf: > the compiler already does that automatically. it'll only implement `Eq` if the type params are also `Eq`

Yeah I thought this is how it should work as well and still don't fully understand this. Like for `PartialEq` the compiler could only derive it if `T: PartialEq` it seems. Maybe will revisit this.

shared/src/proto/types.rs
```diff
@@ -51,54 +52,49 @@ pub struct SignedTxData {
     pub sig: common::Signature,
 }
 
-/// A generic signed data wrapper for Borsh encode-able data.
+/// Tag type that indicates we should use [`BorshSerialize`]
+/// to sign data in a [`Signed`] wrapper.
+#[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub enum SerializeWithBorsh {}
```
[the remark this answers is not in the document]
sug0: well it's a tag, we don't need to construct a value from it

batconjurer (approved):
LGTM!

shared/src/proto/types.rs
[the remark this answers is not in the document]
batconjurer: Good point.

shared/src/proto/types.rs
```diff
@@ -51,54 +52,69 @@ pub struct SignedTxData {
...
+}
+
+/// A generic signed data wrapper for serialize-able types.
+///
+/// The default serialization method is [`BorshSerialize`].
 #[derive(
     Clone, Debug, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
 )]
-pub struct Signed<T: BorshSerialize + BorshDeserialize> {
+pub struct Signed<T, S = SerializeWithBorsh> {
```
james-chf, line 86: It'd be good if we could bound `S` here (if it's possible?) so that it's impossible to construct a `Signed` with an invalid tag, something like

```rust
pub struct Signed<T, S: SignedSerialize<T> = SerializeWithBorsh>
```

james-chf (approved):

shared/src/proto/types.rs
sug0: we already tighten `Signed` at the impl block starting on line 148. you can only verify signatures of `Signed` instances with a tag which implements `SignedSerialize`. you can't use the struct itself to instantiate new values, since it has a private field, but you can use `Signed::new_from` instead. regardless, we hit the same wall, this instance can't do jack if `S` does not implement `SignedSerialize`.

tl;dr I don't think this is an issue

Issue #82, opened 2022-02-24
use zeroize for crypto secret keys


tzemanovic:
https://crates.io/crates/zeroize

tzemanovic:
done in #277

PR #277, opened 2022-08-04
Zeroize secret keys upon drop

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1166

based on #250

Closes issue anoma/namada#82. It seems the best way to get the memory of the `SecretKey` byte array to be reset to `0`s is to wrap the external secret key structs (`ed25519_consensus::SigningKey` for example) in a `Box` pointer when initializing our own `SecretKey` structs.

I found some inspiration from [this blog post](https://benma.github.io/2020/10/16/rust-zeroize-move.html), which described behavior I was observing while experimenting with this code.


tzemanovic:
let's rebase on #250 to check this through new CI

tzemanovic:
pls update wasm

tzemanovic:
pls update wasm

tzemanovic (commented):
I cannot approve here as I moved the PR, but I reviewed it and LGTM

PR #281, opened 2022-08-04
Update `eth-bridge-integration` with changes from `main`

james-chf:
For this PR, we should be able to get as much CI working as possible, with the exception of any ABCI tests. CI is only running for old ABCI right now so had to make some small changes so that things will still compile with the default feature flags (i.e. with the soon-to-be-removed ABCI) (see the commits made after the merge commit in this PR).


james-chf:
pls update wasm

juped:
main is not in a useful state to do this right now (soon, it should be in a better one)

james-chf:
Yeah we will have to merge again soon once CI is working further.

james-chf:
This PR is stale but I can cherry pick commits to make `eth-bridge-integration` compile with the default feature flags, then we can try merging latest `main` again

PR #283, opened 2022-08-05
PoS spec updates

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1075, partially addresses #488 

depends on #282 

- adds new rewards and commissions system
- updates specs for https://github.com/anoma/namada/issues/366


documentation/dev/src/explore/design/ledger/pos-integration.md
```diff
@@ -14,35 +14,31 @@ All [the data relevant to the PoS system](https://specs.namada.net/economics/pro
...
 - `slash/{validator_address}` (optional): a list of slashes, where each record contains epoch and slash rate
 - `bond/{bond_source}/{bond_validator} (optional)`
 - `unbond/{unbond_source}/{unbond_validator} (optional)`
-- `validator_set (required)`
-- `total_voting_power (required)`
+- `validator_set/consensus (required)`: the set of up to `max_validator_slots` (parameter) validators, ordered by their bonded stake
+- `validator_set/below_capacity (required)`: the set of validators below consensus capacity, but with bonded stake above the `min_validator_stake` parameter, also ordered by their bonded stake
+- `total_deltas (required)`
 
 - standard validator metadata (these are regular storage values, not epoched data):
```
brentstone, line 31: Do we particularly want these fields to be listed separately to the other validator-specific fields above? I would think to combine them

documentation/dev/src/explore/design/ledger/pos-integration.md
```diff
@@ -14,35 +14,31 @@ All [the data relevant to the PoS system](https://specs.namada.net/economics/pro
...
-  - `validator/{validator_address}/staking_reward_address` (required): an address that should receive staking rewards
   - `validator/{validator_address}/address_raw_hash` (required): raw hash of validator's address associated with the address is used for look-up of validator address from a raw hash
-  - TBA (e.g. alias, website, description, delegation commission rate, etc.)
+  - `validator/{validator_address}/max_commission_rate_change` (required): a validator-chosen maximum commission rate change per epoch, set when the validator account is created and cannot be changed
+  - TBA (e.g. alias, website, description, etc.)
 
-Only XAN tokens can be staked in bonds. The tokens being staked (bonds and unbonds amounts) are kept in the PoS account under `{xan_address}/balance/{pos_address}` until they are withdrawn.
+Only NAM tokens can be staked in bonds. The tokens being staked (amounts in the bonds and unbonds) are kept in the PoS account under `{nam_address}/balance/{pos_address}` until they are withdrawn.
```
brentstone, line 36: Don't want to forget that accumulated rewards are held in the PoS account and should maybe be included here. But how to treat these and whether or not to include them in the deltas may be a larger discussion

documentation/dev/src/explore/design/ledger/pos-integration.md
```diff
@@ -53,48 +49,49 @@ For slashing tokens, we implement a [PoS slash pool account](vp.md#pos-slash-poo
 
 The validator transactions are assumed to be applied with an account address `validator_address`.
 
-- `become_validator(consensus_key, staking_reward_address)`:
+- `become_validator(consensus_key, commission_rate, max_commission_rate_change)`:
```
brentstone, line 52: How precise do we care to be here with the input arguments? Technically, this function also requires the validator address and the current epoch as inputs at the moment.

documentation/dev/src/explore/design/ledger/pos-integration.md
```diff
@@ -53,48 +49,49 @@ For slashing tokens, we implement a [PoS slash pool account](vp.md#pos-slash-poo
...
   - creates a record in `validator/{validator_address}/consensus_key` in epoch `n + pipeline_length`
-  - creates a record in `validator/{validator_address}/staking_reward_address`
+  - creates a record in `validator/{validator_address}/commission_rate` in epoch `n + pipeline_length`
```
brentstone, line 54: Does `commission_rate` need to be epoched? Currently when the inflation is applied, the commission rate is read from storage as a single value, then it is applied to store the delegator rewards product, which is epoched and contains the commission rate information inherently.

documentation/dev/src/explore/design/ledger/pos-integration.md
tzemanovic: I think we can combine them yeah

documentation/dev/src/explore/design/ledger/pos-integration.md
tzemanovic: For current epoch, we can maybe extend the line 50 with e.g. `and assuming a transaction uses host environment function to find the current epoch`. I think we could also specify them in args to be more complete

documentation/dev/src/explore/design/ledger/pos-integration.md
tzemanovic: I see, good point! Let's just use a single value and document why

brentstone: I thought more about this... I now think it is right to be epoched to ensure that this can only change by at most `max_commission_rate_change` per epoch.

tzemanovic: ah, right! I guess it doesn't need to be fully "epoched", but for this it may be convenient

brentstone (approved):
LGTM!

Issue #140, opened 2022-06-23
Mint wrapped ERC20s when `EthereumEvent::TransfersToNamada` is seen


james-chf:
Depends on https://github.com/anoma/namada/issues/144

- should work similarly to `tx_mint_multitoken` in https://github.com/anoma/anoma-wasm-multitoken
- the transaction will be applied to storage natively by ledger code, rather than via the wasm environment
- state changes should be derived deterministically from `TransfersToNamada` events

Issue #127, opened 2022-06-01
Namada validators should process relevant events from Ethereum smart contracts


batconjurer:
- [x] Integrate a full node into the ledger
- [x] Add the ability to subscribe and deserialize ethereum events
- [x] Events with relevant number of confirmations only should be sent to the ledger
- [x] Events should be added to vote extensions
- [x] #144 
- [x] #197
- [x] Mocks of ethereum events should be created for testing.

cwgoes:
@batconjurer looks like this can be closed?

Issue #197, opened 2022-07-18
Process Ethereum event vote extensions correctly for chains with more than one validator


james-chf:
During `FinalizeBlock`, we may see Ethereum events in protocol transactions, where those events have already been stored in Namada as they were previously seen. In that case, we need to update the Ethereum event in storage with details of the new validators who have seen it, and change `seen` from `false` to `true` if appropriate. We should also act on the event if necessary. We should also store the event but not act upon it immediately, if it hasn't been seen by enough validators yet.

Related issue: https://github.com/anoma/namada/issues/515

Issue #166, opened 2022-07-08
common storage read trait


tzemanovic:
Depends on #110 

Add a common trait for client RPC storage read queries, ledger direct storage access (probably needs to include gas metering somehow), VPs posterior storage read (we shouldn't need to re-use code for reading from pre-state, that's typically only ever done by the VP that owns the given storage key) and tx storage read.

cwgoes:
@tzemanovic what's the status of this?

tzemanovic:
done in #324 (incorrectly showing closed, but it was merged)

PR #284, opened 2022-08-08
Update storage and mint new wrapped ERC20s for a single validator node

james-chf:
Closes https://github.com/anoma/namada/issues/144

Closes https://github.com/anoma/namada/issues/140

This PR makes it so that we apply a derived tx whenever we apply a `ProtocolTxType::EthereumEvents`. This derived tx:
- for each event, writes the event to storage if it is the first time is being seen
- for any `TransferToNamada` event which becomes `seen`, mints wrapped ERC20s

This PR is implementing the previous design of the Ethereum bridge where we would have had vote extensions.

For this PR, Ethereum updates are only written to storage and acted upon the first time they are seen. In the degenerate case of a chain with a single validator this works, as the validator has 100% of the voting power so Ethereum events are always acted upon as soon as they are seen. Making the state update transaction work across multiple rounds and with multiple validators (https://github.com/anoma/namada/issues/197) will come in a later PR.


apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -666,7 +666,7 @@ where
                     ShellParams {
                         block_gas_meter: &mut gas_meter,
                         write_log: &mut write_log,
-                        storage: &self.storage,
+                        storage: &mut self.storage,
                         wasm_dir: self.wasm_dir.as_path(),
```
james-chf, line 670: TODO: `wasm_dir` is no longer necessary. `storage` should still be able to be passed as immutable during a dry_run_tx or query.

james-chf: Being fixed up in https://github.com/anoma/namada/pull/310

james-chf:
pls update wasm

sug0:
Also closes https://github.com/anoma/namada/issues/127

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    BorshDeserialize,
+)]
+pub struct EthMsgUpdate {
+    /// the event being seen
+    pub body: EthereumEvent,
+    /// addresses of the validators who have just seen this event
+    /// we use [`BTreeSet`] even though ordering is not important here, so that
+    /// we can derive [`Hash`] for [`EthMsgUpdate`]. This also conveniently
+    /// orders addresses in the order in which they should be stored in
+    /// blockchain storage.
```
sug0, line 56: can you capitalize the first word of each of these docstrings?

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+#[derive(
+    Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, BorshSchema,
+)]
+pub struct EthMsg {
+    /// The event being stored
+    pub body: EthereumEvent,
+    /// The total voting power that's voted for this event across all epochs
+    pub voting_power: (u64, u64),
+    /// The addresses of validators that voted for this event, in sorted order.
+    pub seen_by: Vec<Address>,
```
sug0, line 32: you can pick a `HashSet` or `BTreeMap` here, even if we need deterministic data in storage. borsh should guarantee we get the same encoding every time.

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    tracing::info!(
+        ethereum_events = events.len(),
+        "Applying state updates derived from Ethereum events found in \
+         protocol transaction"
+    );
+
+    let last_epoch = store.get_last_epoch();
+    tracing::debug!(?last_epoch, "Got epoch of last block");
+
+    let active_validators = store.get_active_validators(Some(last_epoch));
```
sug0, line 87: you should be getting these based on some block height param, especially with the new shimming stuff

sug0: to be honest, I'm not entirely sure on this

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    voting_powers: &HashMap<Address, VotingPower>,
+) -> Result<(BTreeSet<storage::Key>, bool)> {
+    let eth_msg_keys = eth_msgs::Keys::from(&update.body);
+
+    // we arbitrarily look at whether the seen key is present to
+    // determine if the /eth_msg already exists in storage, but maybe there
+    // is a less arbitrary way to do this
+    let exists_in_storage = store.has_key(&eth_msg_keys.seen())?;
+
+    let (eth_msg, changed, newly_confirmed) = if !exists_in_storage {
```
sug0, line 190: is it possible to split each of these branches of the if expr into two different helper funs?

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    _voting_powers: &HashMap<Address, VotingPower>,
+) -> (EthMsg, BTreeSet<storage::Key>, bool) {
+    tracing::warn!(
+        "Updating Ethereum events is not yet implemented, so this Ethereum \
+         event won't change"
+    );
+    (eth_msg, BTreeSet::default(), false)
+}
+
+fn threshold() -> Ratio<u64> {
```
sug0, line 261: use `FractionalVotingPower::TWO_THIRDS`

apps/src/lib/node/ledger/protocol/transactions/store.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn read(&self, key: &storage::Key) -> Result<Option<Vec<u8>>>;
+    /// Check if the given key is present in storage.
+    fn has_key(&self, key: &storage::Key) -> Result<bool>;
+    /// Write a value to `key`
+    fn write(
+        &mut self,
+        key: &storage::Key,
+        value: impl AsRef<[u8]> + Clone,
+    ) -> Result<()>;
+}
```
sug0, line 30: tomas is adding a `StoreRead` and `StoreWrite` set of traits, soon, I think

sug0 (changes requested):

apps/src/lib/node/ledger/protocol/transactions/store.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    use eyre::Result;
+    use namada::types::storage;
+
+    use super::*;
+
+    /// Very simple fake storage for use in tests. In-memory map of
+    /// [`storage::Key`]s to raw bytes.
+    #[derive(Default)]
+    pub struct FakeStorage {
```
batconjurer, line 119: Why use this when we have the `Storage<MockDB, Sha256Hasher>` for testing?

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/events.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use eyre::Result;
+use namada::ledger::eth_bridge::storage::wrapped_erc20s;
+use namada::types::ethereum_events::{EthereumEvent, TransferToNamada};
+use namada::types::storage::Key;
+
+use super::update;
+use crate::node::ledger::protocol::transactions::store::Store;
+
+pub(super) fn act_on(
```
batconjurer, line 13: doc string please

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+        MultiSignedEthEvent { event, signers }: MultiSignedEthEvent,
+    ) -> Self {
+        Self {
+            body: event,
+            seen_by: signers.into_iter().collect(),
+        }
+    }
+}
+
+pub(crate) fn apply_derived_tx(
```
batconjurer, line 71: doc string please

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        let result =
+            get_voting_powers_for_selected(&active_validators, validators);
+
+        let voting_powers = match result {
+            Ok(voting_powers) => voting_powers,
+            Err(error) => panic!("error: {:?}", error),
+        };
+        assert_eq!(voting_powers.len(), 1);
+        assert!(
```
batconjurer, line 79: use the assert_matches! macro

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,179 @@
...
+            Ok(voting_powers) => voting_powers,
+            Err(error) => panic!("error: {:?}", error),
+        };
+        assert_eq!(voting_powers.len(), 2);
+        assert!(
+            matches!(voting_powers.get(&validator_1), Some(v) if *v == voting_power_1)
+        );
+        assert!(
+            matches!(voting_powers.get(&validator_2), Some(v) if *v == voting_power_2)
+        );
```
batconjurer, line 138: assert_matches!

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/events.rs
```diff
@@ -0,0 +1,73 @@
...
+    store: &mut impl Store,
+    event: &EthereumEvent,
+) -> Result<BTreeSet<Key>> {
+    match &event {
+        EthereumEvent::TransfersToNamada { transfers, .. } => {
+            act_on_transfers_to_namada(store, transfers)
+        }
```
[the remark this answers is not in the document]
james-chf: We shouldn't need to look at the nonce when processing `TransferToNamada` events, as they can be processed in any order

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/events.rs
```diff
@@ -0,0 +1,73 @@
...
+                "Existing value found",
+            );
+            supply.receive(amount);
+            tracing::debug!(
+                %supply_key,
+                ?supply,
+                "New value calculated",
+            );
+        })?;
+        _ = changed_keys.insert(supply_key);
```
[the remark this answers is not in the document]
james-chf: Yeah, I'm not sure whether it is better to use `let _` or `_` here

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
james-chf: I agree, we should use a `BTreeSet`. I would rather make this change in a new PR though (and update the specs at the same time), I can make an issue for it

james-chf: https://github.com/anoma/namada/issues/434

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/events.rs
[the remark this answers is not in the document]
james-chf: I wanted to make explicit that we can ignore the result of `.insert` i.e. whether the supply key is being written for the first time, or whether it is being updated from a previous value, is not important at this point in the code

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
james-chf: oh yeah good catch, this needs to change now. Leaving this comment thread open and will come back to it as need to think about it

apps/src/lib/node/ledger/protocol/transactions/store.rs
james-chf: These two issues?
https://github.com/anoma/namada/issues/166

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    Ok(changed_keys)
+}
+
+/// Apply an [`EthMsgUpdate`] to storage. Returns any keys changed and whether
+/// the event was newly seen.
+fn apply_single_update(
+    store: &mut impl Store,
+    update: EthMsgUpdate,
+    total_voting_power: VotingPower,
+    voting_powers: &HashMap<Address, VotingPower>,
```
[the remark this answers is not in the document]
james-chf: We work out the fractional voting power in the function itself (using `total_voting_power` as the denominator), though not using the `FractionalVotingPower` type (yet). I'll try to make this clearer the voting powers here are absolute and should be for the same epoch i.e.

`total_voting_power` - total voting power of all the active validators for epoch `n`
`voting_powers` - the absolute voting power for all active validators for epoch `n` who submitted an Ethereum event in their vote extension

james-chf:
I will try updating to use our `FractionalVotingPower` type whereever possible (i.e. where this PR was using `Ratio<u64>` or `(u64, u64)` before

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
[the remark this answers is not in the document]
batconjurer: That part is fine, but when I look at the unimplemented calculate_diff function, we don't have all the necessary info being passed in. This is because the previous voting powers may have come from a different epoch, but I only see the total_voting_power for the current epoch being passed int.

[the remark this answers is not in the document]
batconjurer: And to get that data, it think it is easier to change the signatures all the way up and use normalized voting powers (i.e. fractional voting powers)

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
james-chf: So in the previous design with vote extensions, using whoever the active validators were for the most recent committed block made sense, as the Ethereum events would have been voted for during the consensus round for that block. e.g. in the consensus round for proposing block `n`, validators for block `n` vote for Ethereum events. Then the Ethereum events protocol tx is made and applied in block `n+1`. The validator sets may be different between blocks `n` and `n+1` if we crossed into a new epoch, so we get the validators for block `n`. (I will try to make the code here clearer as well as all this is happening a bit implicitly)

I'm not so familiar yet with this part of the new design but that can be further issues/PRs to migrate this old code to using a specified block height (as provided in `ProtocolTxType::EthereumEvents`, I guess?). 

NB: I've now updated the PR description to make explicit this is for the old ABCI++ design.

james-chf:
pls update wasm

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
sug0: okay sounds good. maybe add a `TODO` here, though? so that we know we need to change this later when `abciplus` shims merge

apps/src/lib/node/ledger/protocol/transactions/store.rs
sug0: likewise, add a `TODO`, so we remember to change this later

sug0 (changes requested):

sug0:
I left comments in two open threads, if it's not clear what this review is about

james-chf:
Rebasing on top of the shim vext changes from https://github.com/anoma/namada/pull/437/files , will also add an integration test using the `Storage` component

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/events.rs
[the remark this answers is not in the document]
batconjurer: I didn't know this either because apparently my IDE marks it as a syntax error. :exploding_head:

batconjurer (changes requested):
So I've spent some time thinking about it and I really would prefer not to have the `StoreTrait` here. If it were needed for mocking, I would except it as a necessary evil, but the `FakeStorage` is almost identical to using the existing `MockDB`.

So if we have a mock we don't really need, that leaves us with a trait implemented by a single type, which I'm not in favor of as a general practice. I think it bloats the code and adds cognitive load by adding a layer of indirection over the code flow.  This would be fine if it was an api facing outward from a third party library, but since this is internal, I don't see it as a realistic goal to hide / abstract details away that we should actually be aware of.

This is a divisive issue in software development ( as it touches on the role of encapsulation and how much OOP should be allowed in a codebase ), so in order to resolve this philosophical dispute equitably, I have requested that Heliax define a set of code standards that everyone will be able to give their input on. This should happen some in the next weeks. 

As such, we should put merging this on hold until we have a clear set of guidelines on how to proceed. I will request changes for now to reflect this, and if need be I will change it to approve if it is decided this particular coding practice is acceptable.

james-chf:
> So I've spent some time thinking about it and I really would prefer not to have the `StoreTrait` here. If it were needed for mocking, I would except it as a necessary evil, but the `FakeStorage` is almost identical to using the existing `MockDB`.

> So if we have a mock we don't really need, that leaves us with a trait implemented by a single type, which I'm not in favor of as a general practice. I think it bloats the code and adds cognitive load by adding a layer of indirection over the code flow. This would be fine if it was an api facing outward from a third party library, but since this is internal, I don't see it as a realistic goal to hide / abstract details away that we should actually be aware of.

I think the layer of indirection and hiding away the details from the `ethereum_events` module is worth it (though it's more lines of code). There are some issues with injecting `Storage` directly, even though that would work as well. The main two things are:

- we don't need the majority of its API, only a few methods - read, write, has_key - so we are making this module depend on a large interface it doesn't need
- we couple our module to only being able to use that `Storage` struct, and not anything else

The read/write/has_key methods of `Storage` are not perfectly suited for the derived transaction logic, as they return a gas cost which isn't relevant in this module - so we would still want an adapter around the `Storage` imo to discard this (and provide the API we actually want). This is what the `LedgerStore` implementation does in this PR - discards irrelevant values like gas cost so we don't need to do so throughout the `ethereum_events` module.

Using the `Store` trait, we can provide any implementation of that trait which is really useful for tests (some to be written). So far this PR only has a simple `FakeStorage` but we could implement some other more involved stubs or fakes, and in theory we could autogenerate mocks even (like with [mockall](https://docs.rs/mockall/latest/mockall)). By injecting the `Storage` struct directly, we would have to make changes to the `Storage` struct if we wanted to do that (some of which might not be suitable for tests).

The fake `Storage<MockDB, Sha256Hasher>` would work in effect identically to the simple `FakeStorage` for the purposes of tests here (I think) but is a more complex component. That isn't a big deal though. I see it could be more practical to just use it rather than implementing `FakeStorage`. We should definitely have integration tests with a real `Storage` component anyway. 

The `Store` trait here as declared just for use by a single type seems a bit overkill, but it serves to explicitly document the  API that derived transactions want i.e. a simple read/write/has_key one without gas costs. Passing a `Storage` in directly, you need to look at the code in `ethereum_events` to see what parts of its API are actually being used rather than just looking at the trait.
 
> This is a divisive issue in software development ( as it touches on the role of encapsulation and how much OOP should be allowed in a codebase ), so in order to resolve this philosophical dispute equitably, I have requested that Heliax define a set of code standards that everyone will be able to give their input on. This should happen some in the next weeks.
> 
> As such, we should put merging this on hold until we have a clear set of guidelines on how to proceed. I will request changes for now to reflect this, and if need be I will change it to approve if it is decided this particular coding practice is acceptable.

It's not really to do with OOP, I think encapsulation is good though and it'd be good to have coding guidelines. The most important thing here (imo) is to inject only the interface that we need, for code clarity, and one that is suited to us. This should make the `ethereum_events` code simpler.

An equivalent struct with that API that the `Store` trait has (read/write/has_key) would at least do the job of interface segregation. Ideally the `Storage` struct could have some subcomponent which implemented a `Store`-like trait that we could pass in directly (and the `Storage` struct itself would use that internally also) - that would provide most of the benefits although we'd still be coupled to that one implementation.

I'll try making the changes requested - switch to using the `Storage` struct directly and the fake `Storage<MockDB, Sha256Hasher>` for testing, to compare, then we could merge that asap and come back to changing this if we want to later.

james-chf:
I've updated to use `Storage`, going to readapt now based on the logic that came in for ABCI 0.18, then will retest. Specifically we can no longer just get active validators/voting powers for the epoch of the last committed block, but must look at individual `MultiSignedEthEvent` to get the appropriate validators/voting powers.

james-chf:
pls update wasm

james-chf:
pls update wasm

james-chf:
This PR became a bit overgrown, there are a few small things that should be in separate PRs that aren't directly to do with what this PR is trying to achieve:
- the `impl Add`/`impl AddAssign` for `FractionalVotingPower`
- the extra debug logging when a validator votes for Ethereum events
- ensuring the `QueriesExt` trait is visible outside of the `node::ledger::shell` module (as we want to use it in the `node::ledger::protocol` module as well)

In terms of what this PR is concerned with, it's still too much, it could at least be split up into PRs for:
- adding the `EthMsg`/`EthMsgUpdate` types (`ethereum_events/eth_msgs.rs`)
- the logic for storing newly seen Ethereum events (most of `ethereum_events/mod.rs`)
- the logic for minting wrapped ERC20s (`ethereum_events/events.rs`)

james-chf:
pls update wasm

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/update.rs
```diff
@@ -0,0 +1,70 @@
...
+use eyre::Result;
+use namada::ledger::storage::{DBIter, Storage, StorageHasher, DB};
+use namada::types::storage;
+use namada::types::token::Amount;
+
+/// Reads the `Amount` from key, applies update then writes it back
+pub(super) fn amount<D, H>(
+    store: &mut Storage<D, H>,
+    key: &storage::Key,
+    update: impl Fn(&mut Amount),
```
sug0, line 12:
```suggestion
    update: impl FnOnce(&mut Amount),
```
I think you can make this a `FnOnce`

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/update.rs
```diff
@@ -0,0 +1,70 @@
...
+    store.write(key, amount.try_to_vec()?)?;
+    Ok(amount)
+}
+
+#[allow(dead_code)]
+/// Reads an arbitrary value, applies update then writes it back
+pub(super) fn value<D, H, T: BorshSerialize + BorshDeserialize>(
+    store: &mut Storage<D, H>,
+    key: &storage::Key,
+    update: impl Fn(&mut T),
```
sug0, line 29:
```suggestion
    update: impl FnOnce(&mut T),
```
same here

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,319 @@
...
+                            height
+                        )
+                    })?;
+                let individual_voting_power = active_validators
+                    .iter()
+                    .find(|&v| v.address == addr)
+                    .ok_or_else(|| {
+                        eyre!(
+                            "No active validator found with address {} for \
+                             height {}",
```
sug0, line 49: same thing here

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,319 @@
...
+                            height
+                        )
+                    })?
+                    .voting_power;
+                let total_voting_power = total_voting_powers
+                    .get(&height)
+                    .ok_or_else(|| {
+                        eyre!(
+                            "No total voting power provided for height {}",
+                            height
```
sug0, line 60: same thing here

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
```diff
@@ -0,0 +1,98 @@
...
+)]
+pub struct EthMsg {
+    /// The event being stored
+    pub body: EthereumEvent,
+    /// The total voting power that's voted for this event across all epochs
+    pub voting_power: FractionalVotingPower,
+    /// The addresses of validators that voted for this event, in sorted order.
+    pub seen_by: Vec<Address>,
+    /// Whether this event has been acted on or not
+    pub seen: bool,
```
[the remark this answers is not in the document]
james-chf: This struct represents an event as stored under `/eth_msgs`, which has an explicit `seen` key which maps to this field. So as long as the stored event has a `seen` key, we should have this field here. Though it does seem kind of redundant in a way as it can be derived whether an event is seen or not based on the `voting_power` being >2/3 like you say. Maybe we should simplify our design and not store an explicit `seen` field at all?

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/read.rs
```diff
@@ -0,0 +1,103 @@
...
+    H: 'static + StorageHasher + Sync,
+{
+    let (maybe_val, _) = storage.read(key)?;
+    let bytes = match maybe_val {
+        Some(bytes) => bytes,
+        None => return Ok(None),
+    };
+    let deserialized = T::try_from_slice(&bytes[..])?;
+    Ok(Some(deserialized))
+}
```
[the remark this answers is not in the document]
james-chf: Almost definitely, there is for certain functionality like this for VPs. This sort of wrapper logic is defined in a few places and we should consolidate

sug0 (approved):

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/read.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Returns the stored Amount, or 0 if not stored
+pub(super) fn amount_or_default<D, H>(
+    storage: &Storage<D, H>,
+    key: &storage::Key,
+) -> Result<Amount>
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+{
+    let amount = match maybe_value(storage, key)? {
```
batconjurer, line 17:
```suggestion
    Ok(maybe_value(storage, key)?.unwrap_or_default())
```

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        let result =
+            get_voting_powers_for_selected(&active_validators, validators);
+
+        let voting_powers = match result {
+            Ok(voting_powers) => voting_powers,
+            Err(error) => panic!("error: {:?}", error),
+        };
+        assert_eq!(voting_powers.len(), 1);
+        assert_matches!(voting_powers.get(&(sole_validator, BlockHeight(100))), Some(v) if *v == FractionalVotingPower::new(1, 1).unwrap());
```
batconjurer, line 130:
```suggestion
        assert_matches!(
            voting_powers.get(&(sole_validator, BlockHeight(100))),
            Some(v) if *v == FractionalVotingPower::new(1, 1).unwrap()
        );
```

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/utils.rs
```diff
@@ -0,0 +1,314 @@
...
+        let result =
+            get_voting_powers_for_selected(&active_validators, validators);
+
+        let voting_powers = match result {
+            Ok(voting_powers) => voting_powers,
+            Err(error) => panic!("error: {:?}", error),
+        };
+        assert_eq!(voting_powers.len(), 2);
+        assert_matches!(voting_powers.get(&(validator_1, BlockHeight(100))), Some(v) if *v == FractionalVotingPower::new(100, 300).unwrap());
+        assert_matches!(voting_powers.get(&(validator_2, BlockHeight(100))), Some(v) if *v == FractionalVotingPower::new(200, 300).unwrap());
```
batconjurer, line 211:
```suggestion
        assert_matches!(
            voting_powers.get(&(validator_1, BlockHeight(100))),
            Some(v) if *v == FractionalVotingPower::new(100, 300).unwrap()
        );
        assert_matches!(
            voting_powers.get(&(validator_2, BlockHeight(100))), 
            Some(v) if *v == FractionalVotingPower::new(200, 300).unwrap()
        );
```

batconjurer (approved):
LGTM! There are a couple of small suggestions, but you can take them or leave them.

PR #292, opened 2022-08-08
More logging during PrepareProposal, ProcessProposal, FinalizeBlock

james-chf:
This is a quick PR to add some more visibility around what is happening during block proposal and finalization. Especially:
- logging when we reject a block proposal as a validator
- logging when we receive an unsupported transaction in `apply_tx`

### DIFF
--- a/apps/src/lib/node/ledger/protocol/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/mod.rs
@@ -174,7 +174,11 @@ where
                 ..Default::default()
             })
         }
-        _ => {
+        tx_type @ _ => {
+            tracing::error!(
+                "Attempt made to apply an unsupported transaction! - {:#?}",
+                tx_type
+            );
             let gas_used = block_gas_meter
                 .finalize_transaction()
                 .map_err(Error::GasError)?;
--- a/apps/src/lib/node/ledger/protocol/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/mod.rs
@@ -164,8 +164,10 @@ where
                     ..
                 }),
             ..
-        }) if !events.is_empty() => {
-            tracing::debug!("Ethereum events received");
+        }) => {
+            if !events.is_empty() {
+                tracing::debug!("Ethereum events received");
+            }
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -174,7 +176,11 @@ where
                 ..Default::default()
             })
         }
-        _ => {
+        tx_type @ _ => {
+            tracing::error!(
+                "Attempt made to apply an unsupported transaction! - {:#?}",
+                tx_type
+            );
```
james-chf, line 183: `ProtocolTxType::EthereumEvents` with an empty vector of events was hitting this branch before, I've moved case that to be handled in the above branch. 

I'm not sure now if we should expect to hit this code path anymore. The only remaining tx types that would hit here are: `TxType::Wrapper`, non-Ethereum `TxType::Protocol` and `TxType::Decrypted(DecryptedTx::Undecryptable(_))`. We could replace this branch with returning an error like we do for `TxType::Raw(_)` i.e. `_ => Err(Error::TxTypeError)`.

sug0 (approved):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -36,6 +36,11 @@ mod prepare_block {
             &mut self,
             req: RequestPrepareProposal,
         ) -> response::PrepareProposal {
+            tracing::info!(
```
batconjurer, line 39: We already get a log for this from tendermint.

james-chf: :+1: I've removed this

apps/src/lib/node/ledger/protocol/mod.rs
james-chf: #313

james-chf: #313

batconjurer (approved):

Issue #227, opened 2022-07-28
Endpoint for submitting Ethereum events to a validator during testing


james-chf:
Every Namada validator should listen for events emitted by the Ethereum bridge smart contracts. In the ledger code, we provide the `Shell` with a channel of `EthereumEvent`s that is fed by an `Oracle` component listening to a local `geth` subprocess. 

For integration testing, it would be useful to have an option to run the `Shell` with a channel that is fed via something like an endpoint on the ledger node. We could push `EthereumEvent`s to this endpoint in e2e tests in CI, or when testing manually. It could also be used for load/fuzz testing.

PR #294, opened 2022-08-08
Add endpoint for direct submission of Ethereum events

james-chf:
Closes https://github.com/anoma/namada/issues/227

When the `namada_apps/eth-fullnode` feature is not enabled, there is a config option - `ledger.ethereum.oracle_event_endpoint` - which if set to true, makes it so that a server is started by the oracle exposing an endpoint at `http://127.0.0.1:3030` for submission of Borsh-serialized `EthereumEvent`s. Any events submitted there will be included in validator's vote extensions.

Enabling this functionality at runtime can be done like so:

```shell
ANOMA_LEDGER__ETHEREUM__ORACLE_EVENT_ENDPOINT=true target/debug/namadan ledger run
```

This functionality can be tested out using a devtool built from https://github.com/anoma/devtool/pull/2

```shell
devtool submit-fake-transfer-to-namada \
	--nonce 1 \
	--receiver atest1v4ehgw36x5myzdfcg5myg3feg4p5vvzyxfrrgdpnxccnvvp5gdrrqse3gdz5zv2yxsenqs3ntgy89n
```

### DIFF
--- a/apps/src/lib/config/ethereum.rs
+++ b/apps/src/lib/config/ethereum.rs
@@ -1,3 +1,4 @@
+//! Configuration settings to do with the Ethereum bridge.
 use serde::{Deserialize, Serialize};
 
 /// Default [Ethereum JSON-RPC](https://ethereum.org/en/developers/docs/apis/json-rpc/) endpoint used by the oracle
--- a/apps/src/lib/config/ethereum.rs
+++ b/apps/src/lib/config/ethereum.rs
@@ -1,4 +1,6 @@
 //! Configuration settings to do with the Ethereum bridge.
+#[allow(unused_imports)]
+use namada::types::ethereum_events::EthereumEvent;
 use serde::{Deserialize, Serialize};
 
 /// Default [Ethereum JSON-RPC](https://ethereum.org/en/developers/docs/apis/json-rpc/) endpoint used by the oracle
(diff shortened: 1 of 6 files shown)


apps/src/lib/config/ethereum.rs
```diff
@@ -8,12 +8,20 @@ pub struct Config {
     /// The Ethereum JSON-RPC endpoint that the Ethereum event oracle will use
     /// to listen for events from the Ethereum bridge smart contracts
     pub oracle_rpc_endpoint: String,
+    /// If this is set to `true`, then instead of the oracle listening for
+    /// events at a Ethereum JSON-RPC endpoint, an endpoint will be exposed by
+    /// the ledger for submission of Borsh-serialized
+    /// [`namada::types::ethereum_events::EthereumEvent`]s
```
sug0, line 14: nit: you can add a `use` statement with the imported path, accompanied by a `#[allow(unused_imports)]` tag. the generated docs reference the full path, with the import this would be shorter.

sug0 (approved):
lgtm

apps/src/lib/config/ethereum.rs
james-chf: It feels odd importing something just to use it in the docstring, but it makes sense cc https://github.com/rust-lang/rust/issues/79542

apps/src/lib/config/ethereum.rs
```diff
@@ -1,3 +1,6 @@
+//! Configuration settings to do with the Ethereum bridge.
+#[allow(unused_imports)]
+use namada::types::ethereum_events::EthereumEvent;
```
batconjurer, line 3: I'm confused, why is this import here?

batconjurer (approved):
Small question, but otherwise looks good

apps/src/lib/config/ethereum.rs
james-chf: https://github.com/anoma/namada/pull/294#discussion_r962311096

(I'm happy to use either the qualified or unqualified symbol in the docstring)

batconjurer: oh wow. Ok.

Issue #309, opened 2022-08-09
Add functionality to be able to apply derived transactions natively


james-chf:
We may want nodes to write deterministic changes to storage based on some protocol transaction that was included in the block. They should be able to do this natively rather than via wasm, during `apply_tx` for the given protocol transaction, in `FinalizeBlock`. Currently, we only get a read-only handle there via a `storage: &Storage<D, H>` parameter.

https://github.com/anoma/namada/blob/de53b7f9e9bf94d75957ea771b8d422dfebe555b/apps/src/lib/node/ledger/protocol/mod.rs#L69-L77

We should refactor so that `apply_tx` can take a `&'a mut Storage<D,H>`, but maintain immutability in the cases where that is needed (like if calling `apply_tx` from `dry_run_tx`).

james-chf:
This is in `eth-bridge-integration` branch

PR #310, opened 2022-08-09
Accommodate natively updating blockchain state based on protocol transactions

james-chf:
Relates to https://github.com/anoma/namada/issues/309 , functionality like in this PR will be needed for https://github.com/anoma/namada/pull/284

This PR isn't a pure refactor but changes how transactions are dispatched slightly during `FinalizeBlock`. `apply_tx` becomes `dispatch_tx` which may call either `apply_wasm_tx` or `apply_protocol_tx`. `apply_wasm_tx` meters gas and can update storage via the wasm environment, updates are checked by validity predicates. `apply_protocol_tx` lets us update storage directly without going via wasm. Gas doesn't apply and validity predicates aren't checked.

`dry_run_tx` and the application of governance proposal transactions now go directly via `apply_wasm_tx`. When iterating over transactions in `finalize_block.rs`, we do `dispatch_tx` and pass a `&'a mut Storage<D, H>` in case it is a protocol transaction being applied. Before we were passing `&Storage<D,H>` everywhere. We can also now get rid of `ShellParams`.

### DIFF
--- a/apps/src/lib/node/ledger/protocol/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/mod.rs
@@ -23,6 +23,8 @@ use namada::vm::{self, wasm, WasmCacheAccess};
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use thiserror::Error;
 
+use crate::node::ledger::shell::Shell;
+
 #[derive(Error, Debug)]
 pub enum Error {
     #[error("Storage error: {0}")]
@@ -59,6 +61,36 @@ pub enum Error {
     AccessForbidden(InternalAddress),
 }
 
+pub(crate) struct ShellParams<'a, D, H, CA>
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+    CA: 'static + WasmCacheAccess + Sync,
+{
+    pub block_gas_meter: &'a mut BlockGasMeter,
+    pub write_log: &'a mut WriteLog,
+    pub storage: &'a mut Storage<D, H>,
+    pub vp_wasm_cache: &'a mut VpCache<CA>,
+    pub tx_wasm_cache: &'a mut TxCache<CA>,
+}
+
+impl<'a, D, H> From<&'a mut Shell<D, H>>
+    for ShellParams<'a, D, H, namada::vm::WasmCacheRwAccess>
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+{
+    fn from(shell: &'a mut Shell<D, H>) -> Self {
+        Self {
+            block_gas_meter: &mut shell.gas_meter,
+            write_log: &mut shell.write_log,
+            storage: &mut shell.storage,
+            vp_wasm_cache: &mut shell.vp_wasm_cache,
+            tx_wasm_cache: &mut shell.tx_wasm_cache,
+        }
+    }
+}
+
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// Dispatch a given transaction to be applied based on its type. Some storage
@@ -71,11 +103,13 @@ pub type Result<T> = std::result::Result<T, Error>;
 pub(crate) fn dispatch_tx<'a, D, H, CA>(
     tx_type: TxType,
     tx_length: usize,
-    block_gas_meter: &'a mut BlockGasMeter,
-    write_log: &'a mut WriteLog,
-    storage: &'a mut Storage<D, H>,
-    vp_wasm_cache: &'a mut VpCache<CA>,
-    tx_wasm_cache: &'a mut TxCache<CA>,
+    ShellParams {
+        block_gas_meter,
+        write_log,
+        storage,
+        vp_wasm_cache,
+        tx_wasm_cache,
+    }: ShellParams<'a, D, H, CA>,
 ) -> Result<TxResult>
 where
     D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
@@ -87,11 +121,13 @@ where
         TxType::Decrypted(DecryptedTx::Decrypted(tx)) => apply_wasm_tx(
             tx,
             tx_length,
-            block_gas_meter,
-            write_log,
-            storage,
-            vp_wasm_cache,
-            tx_wasm_cache,
+            ShellParams {
+                block_gas_meter,
+                write_log,
+                storage,
+                vp_wasm_cache,
+                tx_wasm_cache,
+            },
         ),
         TxType::Protocol(ProtocolTx { tx, .. }) => {
             apply_protocol_tx(tx, storage)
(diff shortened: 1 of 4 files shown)


apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -95,22 +93,58 @@ where
...
+            vp_wasm_cache,
+            tx_wasm_cache,
+        ),
+        TxType::Protocol(ProtocolTx { tx, .. }) => {
+            apply_protocol_tx(tx, storage)
+        }
+        _ => {
+            // other transaction types we treat as a noop
+            Ok(TxResult::default())
+        }
```
james-chf, line 134: I think here we should return `Err(Error::TxTypeError)` rather than noop. The only transaction types we should ever be applying for Namada are decrypted transactions and protocol transactions.

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -121,68 +155,65 @@ where
...
+
+// Apply a derived transaction to storage based on some protocol transaction.
+// The logic here must be completely deterministic and will be executed by all
+// full nodes every time a protocol transaction is included in a block. Storage
+// is updated natively rather than via the wasm environment, so gas does not
+// need to be charged and validity predicates are bypassed. A [`TxResult`]
+// containing changed keys and the like should be returned in the normal way.
+pub(crate) fn apply_protocol_tx<'a, D, H>(
+    tx: ProtocolTxType,
+    _storage: &'a mut Storage<D, H>,
```
james-chf, line 200: This may eventually be tightened further to be an impl trait of some subset of `impl Storage`. So far for processing Ethereum events, only `storage.read`, `storage.write` and `storage.has_key` has been needed.

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -61,56 +59,60 @@ pub enum Error {
...
+            storage,
+            vp_wasm_cache,
+            tx_wasm_cache,
+        ),
+        TxType::Protocol(ProtocolTx { tx, .. }) => {
+            apply_protocol_tx(tx, storage)
+        }
+        _ => {
+            // other transaction types we treat as a noop
+            Ok(TxResult::default())
```
james-chf, line 101: I think here we should return `Err(Error::TxTypeError)` rather than noop. The only transaction types we should ever be applying for Namada are decrypted wasm transactions and native protocol transactions.

batconjurer: Sure

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -61,56 +59,60 @@ pub enum Error {
...
-pub type Result<T> = std::result::Result<T, Error>;
-
-/// Apply a given transaction
-///
-/// If the given tx is a successfully decrypted payload apply the necessary
-/// vps. Otherwise, we include the tx on chain with the gas charge added
-/// but no further validations.
-pub(crate) fn apply_tx<'a, D, H, CA>(
-    tx: TxType,
+// Apply a transaction going via the wasm environment. Gas will be metered and
```
batconjurer, line 106: This should be `///` comments, not `//`

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -121,68 +123,65 @@ where
...
-                ibc_event,
-            })
-        }
-        TxType::Protocol(ProtocolTx {
-            tx:
-                ProtocolTxType::EthereumEvents(ethereum_events::VextDigest {
-                    events,
-                    ..
-                }),
+// Apply a derived transaction to storage based on some protocol transaction.
```
batconjurer, line 160: This should be `///` comments

batconjurer (changes requested):
Two small nits, but otherwise looks good.

apps/src/lib/node/ledger/protocol/mod.rs
sug0: add a TODO there, maybe?

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -339,8 +340,16 @@ where
...
+            match protocol::dispatch_tx(
+                tx_type,
+                tx_length,
+                &mut self.gas_meter,
+                &mut self.write_log,
+                &mut self.storage,
+                &mut self.vp_wasm_cache,
+                &mut self.tx_wasm_cache,
+            )
+            .map_err(Error::TxApply)
```
sug0, line 352: without `ShellParams` this call becomes huge. well it's whatever really

sug0 (approved):

apps/src/lib/node/ledger/shell/finalize_block.rs
james-chf: yeah we added `ShellParams` once we needed the extra `wasm_dir` var (due to a clippy warning I think)? This code now is more similar to how it is in `main`, using this long list of parameters as well. I think we could later on further refactor `dispatch_tx` to be a method on `Shell` or something like that since it uses so many of its fields.

apps/src/lib/node/ledger/protocol/mod.rs
james-chf: I've opened an issue to look at doing this more in depth as it will break some tests - https://github.com/anoma/namada/issues/313

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -61,56 +59,60 @@ pub enum Error {
     AccessForbidden(InternalAddress),
 }
 
-pub(crate) struct ShellParams<'a, D, H, CA>
```
batconjurer, line 64: I liked the shell params. Can we leave it?

apps/src/lib/node/ledger/shell/finalize_block.rs
batconjurer: I like the shell params. It it tightens the code up a bit.

james-chf: I put back `ShellParams` though for `dispatch_tx` we still need a `&mut Storage` whereas `ShellParams` has an immutable `&Storage`. So here is still requiring to pass the params explicitly but at the other two call sites where we're changing to calling `apply_wasm_tx`, those can stay the same.

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -85,8 +84,7 @@ where
         Self {
             block_gas_meter: &mut shell.gas_meter,
             write_log: &mut shell.write_log,
-            storage: &mut shell.storage,
-            wasm_dir: shell.wasm_dir.as_path(),
+            storage: &shell.storage,
```
james-chf, line 87: We were passing here `&mut shell.storage` before though the `storage` field itself was `&'a Storage<D, H>` so I think the `mut` here wasn't doing anything?

batconjurer (approved):

PR #276, opened 2022-08-04
Fix output of `anoma wallet address find`

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1161

Fixes [#993](https://github.com/anoma/namada/issues/91). Based and depends on https://github.com/anoma/namada/pull/264

 The output is now:

image

PR #311, opened 2022-08-09
Fix output of `namada wallet address find`

### LINKED ISSUE #91: `anoma wallet address find` should provide a hint to the user of the missing `--alias` flag

brentstone:
Moved from anoma/anoma#1161. Closes #276.

Fixes #91.

Based on #250 and #264.

This PR both fixes the aforementioned CLI issue and adds some additional functionality. We use `clap::ArgGroup` here in order to supply only one of `--alias` or `--address` to `namadaw address find`. If the `alias` is supplied, then the command finds the corresponding `address`, and vice-versa.

Th output of running `namadaw address find` now looks like:
image


tzemanovic (approved):
LGTM, thanks!

PR #314, opened 2022-08-10
Events voted on by validators are not voted on again

### LINKED ISSUE #308: Validators should stop voting for Ethereum events once they are included on chain
Validators store an in-memory queue of Ethereum events that they have received from their oracle (`BTreeSet<EthereumEvent>`). Each Tendermint round, validators include these events in their vote extensions. Once an Ethereum event has been committed to storage with the validator's address in the `seen_by` field of the Ethereum event, they can stop voting for it. They should remove the recorded event from their in-memory queue rather than continuing to vote for it every round.

Separately from this issue, the state update transaction should remain robust regardless against validators voting for events they have already seen, as this may happen in any case due to bugs or malicious validators.

batconjurer:
Closes #308 

When a new ethereum event is seen, it is put in a queue to be included in a vote extension. Once that event is included on chain, it should be removed from the queue so that the validator that saw it stops including it in vote extensions.

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -326,7 +326,12 @@ where
                     continue;
                 }
                 TxType::Protocol(protocol_tx) => match protocol_tx.tx {
-                    ProtocolTxType::EthereumEvents(_) => {
+                    ProtocolTxType::EthereumEvents(ref digest) => {
+                        for event in
+                            digest.events.iter().map(|signed| &signed.event)
+                        {
+                            self.mode.deque_eth_event(event);
+                        }
                         Event::new_tx_event(&tx_type, height.0)
                     }
                     _ => {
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -512,8 +517,12 @@ where
 #[cfg(test)]
 mod test_finalize_block {
     use namada::types::address::xan;
+    use namada::types::ethereum_events::EthAddress;
     use namada::types::storage::Epoch;
     use namada::types::transaction::{EncryptionKey, Fee};
+    use namada::types::vote_extensions::ethereum_events::{
+        MultiSignedEthEvent, Vext, VextDigest,
+    };
```
sug0, line 525: can you use `Vext` with its associated mod, for clarity? e.g. `ethereum_events::Vext`. same thing for `VextDigest`.

sug0 (changes requested):

james-chf (approved):
I tested this patched on https://github.com/anoma/namada/pull/284 and it works

sug0 (approved):

PR #316, opened 2022-08-11
Add storage keys for the EthBridge internal account

james-chf:
(split out from PR https://github.com/anoma/namada/pull/284)

This PR is adds keys to do with:
- storing seen Ethereum events (`eth_msgs/...`)
- tracking the supply and balances of wrapped ERC20 tokens (i.e. this VP implements the multitoken interface and so should eventually work with `anomac transfer`)

There are also some small changes to the VP.

Currently, the EthBridge VP is rejecting any changes made to it's storage subspace. This is the protection we want for `/eth_msgs`, but a later PR based off of this one will allow certain changes to wrapped ERC20 balance keys so that users can do transfers (c.f. https://github.com/anoma/anoma-wasm-multitoken/blob/master/crates/vp_multitoken/src/vp.rs)


shared/src/ledger/eth_bridge/storage/eth_msgs.rs
```diff
@@ -0,0 +1,199 @@
...
+                transfers: vec![],
+            },
+            "06799912C0FD8785EE29E13DFB84FE2778AF6D9CA026BD5B054F86CE9FE8C017"
+                .to_owned(),
+        )
+    }
+
+    #[test]
+    fn test_prefix() {
+        assert!(matches!(&prefix().segments[..], [
```
batconjurer, line 116: the `assert_matches!` macro should be in scope here.

batconjurer (approved):

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+#[allow(missing_docs)]
+pub const BODY_KEY_SEGMENT: &str = "body";
+#[allow(missing_docs)]
+pub const SEEN_KEY_SEGMENT: &str = "seen";
+#[allow(missing_docs)]
+pub const SEEN_BY_KEY_SEGMENT: &str = "seen_by";
+#[allow(missing_docs)]
+pub const VOTING_POWER_KEY_SEGMENT: &str = "voting_power";
```
sug0, line 24: nit: can the visibility of these items be restricted to something like `pub(crate)`? this way cargo won't complain about missing docs, either, I don't think

sug0: maybe `pub` is not needed at all

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
```diff
@@ -0,0 +1,199 @@
...
+                .to_owned(),
+        )
+    }
+
+    #[test]
+    fn test_prefix() {
+        assert_matches!(&prefix().segments[..], [
+                DbKeySeg::AddressSeg(ADDRESS),
+                DbKeySeg::StringSeg(s),
+            ] if s == PREFIX_KEY_SEGMENT)
```
sug0, line 119: I think you can pattern match on `DbKeySeg::StringSeg(PREFIX_KEY_SEGMENT)`directly, kind of like you did for `DbKeySeg::AddressSeg(ADDRESS)`

sug0 (changes requested):

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
james-chf: These were intended to be public but I'm not sure what the use case would be tbh. I was thinking an external tool or test crate that tried to read/write storage under the EthBridge account might use the segments. But they would probably just use the `eth_msgs::Keys` struct instead

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
james-chf: I'm not sure how, this doesn't work?

```rust
        assert_matches!(
            &prefix().segments[..],
            [
                DbKeySeg::AddressSeg(ADDRESS),
                DbKeySeg::StringSeg(PREFIX_KEY_SEGMENT.to_owned()),
            ]
        );
```

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
james-chf: I've removed the `pub` from all these consts

shared/src/ledger/eth_bridge/storage/eth_msgs.rs
sug0: oh it's an owned string. nevermind then!

sug0 (approved):
the build is failing with wasm checksum errors. otherwise lgtm

james-chf:
pls update wasm

PR #317, opened 2022-08-11
Make ABCI compile again

james-chf:
Adapted from https://github.com/anoma/namada/pull/281

Although we are not using the default feature flag set in developing `eth-bridge-integration`, these are used in CI in `main` at the moment which we'll eventually merge in. This PR is the minimal changes to make it so that running "default" commands like `make check`, `make test-unit`, etc. will still compile and pass.


james-chf:
We are going to await https://github.com/anoma/namada/pull/250 to merge in instead

PR #266, opened 2022-08-02
re-add dev docs

tzemanovic:
based on #250

PR #322, opened 2022-08-11
Add stub OpenAPI spec

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1228

based on #266 

Relates to https://github.com/anoma/namada/issues/149

This is a stub OpenAPI spec that could be imported with Insomnia or Postman, with only one type of operation - making a call via [`abci_query`](https://docs.tendermint.com/master/rpc/#/ABCI/abci_query) of Tendermint RPC. The default server is the ledger running on your localhost (`http://127.0.0.1:26657`). Many more examples could be added, e.g. specific `value/<path>` queries for common VPs such as `vp_token` (or alternatively we could have VP-specific OpenAPI files?).

Unfortunately only the first example is imported into tools such as Insomnia or Postman. Maintaining a direct Postman collection of examples as well as/instead of this OpenAPI spec may be better. Unsure where the most appropriate place for this OpenAPI spec file is (maybe `specs/` or `docs/` somewhere?).


tzemanovic:
we should move the file under `documentation/dev` and link it from the book

tzemanovic:
> we should move the file under documentation/dev and link it from the book

done in [d758b1b](https://github.com/anoma/namada/pull/322/commits/d758b1ba12077f367a7877a9ef86173263d0ff5a)

PR #254, opened 2022-08-02
PoS tx tests

tzemanovic:
closes anoma/namada#22, fixes anoma/namada#164 ([5d65759](https://github.com/anoma/anoma/pull/1090/commits/5d65759db9170345a50ae95e30663852339319e4) added the same seed to confirm)

re-opened from https://github.com/anoma/anoma/pull/1090 and rebased on #250

PR #326, opened 2022-08-12
fix validator raw hash

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1246

based on #325 

closes https://github.com/anoma/namada/issues/200


tzemanovic:
the failing unit PoS VP test here is fixed in #254 (verified with seed in https://github.com/anoma/namada/pull/254/commits/a235b8043fe5ea18e9d48f6a7d6a8161d140c692)

tzemanovic:
there was one more issue with the changes - the validation was still using the raw hash derived from address instead of consensus key - fixed in [6be7b2e](https://github.com/anoma/namada/pull/326/commits/6be7b2e4a430fd39eeb956256358d536147f9d29)

PR #329, opened 2022-08-12
Permit transfers of wrapped ERC20 tokens between different addresses

james-chf:
Closes https://github.com/anoma/namada/issues/427

Currently, the EthBridge VP rejects any transactions which attempt to change keys in its space. This PR will change it to allow valid transfers of wrapped ERC20s between different users.

This PR does not include authorization of transfers by the user whose balance is decreasing - this will come in a later PR (https://github.com/anoma/namada/issues/432). End-to-end tests for CI will also come in a separate PR (it will be based on the shell example in https://github.com/anoma/namada/pull/394).


james-chf:
pls update wasm

james-chf:
pls update wasm

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+        &self,
+        tx_data: &[u8],
+        keys_changed: &BTreeSet<Key>,
+        verifiers: &BTreeSet<Address>,
+    ) -> Result<bool, Self::Error> {
+        tracing::debug!(
+            tx_data_len = tx_data.len(),
+            keys_changed_len = keys_changed.len(),
+            verifiers_len = verifiers.len(),
+            "Validity predicate triggered",
```
batconjurer, line 70:
```suggestion
            "Ethereum Bridge VP triggered",
```

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+                    "Rejecting transaction that is attempting to change two \
+                     supply keys"
+                );
+                return Ok(None);
+            }
+        };
+    let balance_a_pre = match reader.read_pre::<Amount>(&balance_a)? {
+        Some(value) => value,
+        None => Amount::from(0),
+    }
```
batconjurer, line 202:
```suggestion
    let balance_a_pre = match reader.read_pre::<Amount>(&balance_a)?.unwrap_or_default()
```

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            );
+            return Ok(None);
+        }
+    }
+    .change();
+    let balance_b_pre = match reader.read_pre::<Amount>(&balance_b)? {
+        Some(value) => value,
+        None => Amount::from(0),
+    }
+    .change();
```
batconjurer, line 219:
```suggestion
    let balance_b_pre = match reader.read_pre::<Amount>(&balance_b)? 
    .unwrap_or_default()
    .change();
```

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            "Rejecting transaction as keys are for different assets"
+        );
+        return Ok(None);
+    }
+    Ok(Some((key_a, key_b)))
+}
+
+/// Returns the `Address` which should be authorizing the balance change, or
+/// `None` if the balance change is invalid.
+fn extract_valid_sender(
```
batconjurer, line 156: The name of this method is not very accurate. This mainly checks that balances were adjust correctly,

shared/src/ledger/eth_bridge/vp/authorize.rs
```diff
@@ -0,0 +1,36 @@
+//! Functionality to do with checking whether a transaction is authorized by the
+//! "owner" of some key under this account
+use eyre::Result;
+
+use super::store;
+use crate::types::address::Address;
+
+pub(super) fn is_authorized(
```
batconjurer, line 8: I'm trying to follow the idea here. Is that idea that since the users vps won't be triggered otherwise, we need to manually trigger them to see if they approve a multitoken transfer? If so, shouldn't we pass in the verifiers set here and insert the necessary vps?

shared/src/ledger/eth_bridge/vp/store.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn deserialize_if_present<T: BorshDeserialize>(
+        maybe_bytes: Option<Vec<u8>>,
+    ) -> Result<Option<T>> {
+        let bytes = match maybe_bytes {
+            Some(bytes) => bytes,
+            None => return Ok(None),
+        };
+        let deserialized = T::try_from_slice(&bytes)
+            .wrap_err_with(|| "couldn't deserialize".to_string())?;
+        Ok(Some(deserialized))
```
batconjurer, line 34:
```suggestion
       maybe_bytes.map(| ref bytes | 
           T::try_from_slice(bytes).wrap_err_with(|| "couldn't deserialize".to_string())
       )
       .transpose()
```

shared/src/ledger/eth_bridge/vp/store.rs
```diff
@@ -0,0 +1,92 @@
...
+use eyre::{Context, Result};
+
+use crate::ledger::native_vp::Ctx;
+use crate::ledger::storage as ledger_storage;
+use crate::ledger::storage::StorageHasher;
+use crate::types::storage::Key;
+use crate::vm::WasmCacheAccess;
+
+/// Read pre/post storage
+pub(super) trait Reader {
```
batconjurer, line 13: Since I've added these `read_pre_value` and `read_post_value` to the `Ctx` type in the bridge pool PR, let's just use those instead (just steal them). They return a simpler type (`Option<T>` instead of `Result<Option<T>>`) so it is simpler to handle.

batconjurer (changes requested):

shared/src/ledger/eth_bridge/vp/mod.rs
james-chf: Yeah this was not well named, have updated the name + docstring

shared/src/ledger/eth_bridge/vp/store.rs
james-chf: We should share code functionality, though I'd rather we make an issue to update the `Ctx` API in `main` and then refactor everywhere at once. We shouldn't silently ignore errors though if they happen (i.e. we should use `Result<Option<T>>` rather than `Option<T>`).

james-chf:
pls update

james-chf:
pls update wasm

shared/src/ledger/eth_bridge/vp/store.rs
batconjurer: These methods are in `eth_bridge_integration` now, so lets use them. But I agree, we shouldn't siliently ignore errors. It should be super fast to refactor the ctx methods to return the `Result<Option<T>>`

shared/src/ledger/eth_bridge/vp/authorize.rs
james-chf: > Implicit VPs are required for Namada launch, so I think triggering the necessary sender and receiver vps manually in this vp is the way to go.

Implicit VP issue - https://github.com/anoma/namada/issues/8

I will scope out https://github.com/anoma/namada/issues/432 further. There are some issues I think with the implicit VP approach in general, not directly related to the eth bridge, I can write up. For now then we can at least get parity with what works for normal tokens (i.e. proper authorization for transfers sent from an established address). Authorization for transfers sent from implicit addresses does not currently work even for normal token transfers using `namadac transfer`, we will need to wait for https://github.com/anoma/namada/issues/8 to be implemented

shared/src/ledger/eth_bridge/vp/store.rs
james-chf: sure I can refactor there in our codebase quickly, I will make an issue before merging this PR to update the `Ctx` API in `main`. I need to look closer at why it returns `Result<Option<T>>` and not just `Result<T>` as at a glance it looks like these `read_pre`/`read_post` methods error if a key isn't present in storage, I would have thought in that case it would have defaulted to `None` and the errors would be reserved for things like e.g. rocksdb or OS-level errors.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -58,8 +58,18 @@ where
...
+            .unwrap_or_else(|error| {
+                tracing::warn!(?error, %account_key, "reading pre value");
+                None
+            })?;
+        let after: Amount = (&self.ctx)
+            .read_post_value(&account_key)
+            .unwrap_or_else(|error| {
+                tracing::warn!(?error, %account_key, "reading post value");
+                None
+            })?;
```
james-chf, line 72: This behaves the same way as before except it will log the error that happened in addition to returning `None`

batconjurer: Looks good!

batconjurer (approved):

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -117,11 +127,11 @@ where
         // but that will be a separate PR.
         let pending_key = get_pending_key();
         let pending_pre: HashSet<PendingTransfer> =
-            self.ctx.read_pre_value(&pending_key).ok_or(eyre!(
+            (&self.ctx).read_pre_value(&pending_key)?.ok_or(eyre!(
```
sug0, line 130: hm why do we need the explicit borrow here, on `ctx`? was the compiler (or clippy) complaining about something?

shared/src/ledger/eth_bridge/storage/mod.rs
```diff
@@ -10,3 +10,44 @@ use crate::types::storage::{Key, KeySeg};
...
+    }
+
+    #[test]
+    fn test_is_eth_bridge_key_returns_false_for_different_address_subkey() {
+        let key =
+            Key::from(address::testing::established_address_1().to_db_key())
+                .push(&"arbitrary key segment".to_owned())
+                .expect("Could not set up test");
+        assert!(!is_eth_bridge_key(&key));
+    }
```
sug0, line 52: good candidates for a proptest :D

sug0 (commented):
lgtm. just some random comments

sug0 (approved):

shared/src/ledger/eth_bridge/storage/mod.rs
james-chf: will have a go at this (though not in this PR)

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
james-chf: It's because the trait this method comes from is implemented only for `&Ctx` and not `Ctx`

https://github.com/anoma/namada/blob/635f68cbf39656c2390d784f28104a98e57ed991/shared/src/ledger/native_vp.rs#L337

If there is a nice way to implement for both `Ctx` and `&Ctx` without too much duplication, I'd like to add it (though not in this PR) for ergonomics. (This `StorageReader` thing actually needs to be contributed back to `main` or may even be able to be replaced by the new native storage traits coming in soon)

james-chf:
pls update wasm

PR #332, opened 2022-08-14
Lazy data structures for storage access

brentstone:
based #503 

Here we'll be finishing adding LazySet, LazyHashSet and LazyHashMap.


tzemanovic:
@brentstone I'm gonna close this for now, but let's re-open once we get to it - I'll link to it from the issue

PR #337, opened 2022-08-16
Specify --target-dir when building wasms

james-chf:
Handles the case where a custom $CARGO_TARGET_DIR is set. wasms need to be built into a `target` directory in the wasm folder when we are going to `cp` them


tzemanovic (dismissed):
LGTM, thanks!

Issue #307, opened 2022-08-09
Handle MASP Transactions Across Epoch Boundaries


murisi:
Right now in the client, the epoch can change during the construction of a MASP transaction. And when this happens, the ledger simply rejects the submitted transaction, and then the client prints out an error. The client ought to detect when an epoch boundary is near enough to invalidate a transaction it would create, sleep till after the epoch changed, and then attempt to construct and submit a transaction.

@tzemanovic had suggested the following:
> The client could wait for the new epoch and then continue, but we’ll need to add a new RPC path for this, probably something that exposes `storage.last_height`,  `next_epoch_min_start_height` and `next_epoch_min_start_time` from the shell.

If there are any unexpected problems with the above approach, the client could instead attempt resubmitting the transaction at most once. This trick would work under the assumption that two consecutively submitted transactions cannot both be on epoch boundaries  (for otherwise the epoch length for the network would be too short).

cwgoes:
@karbyshev can you take this on as part of the wallet work?

murisi:
@karbyshev Looking at `namada/main`, an attempt to solve this issue has been made. The approach taken is to resubmit (transparent -> shielded) transactions if the epoch changes during construction, while at the same time weakening the MASP VP to allow (shielded -> transparent) transactions for assets belonging to older epochs. But the implementation has some possible shortcomings:
1. The client only tests if the epoch changes during the construction of a shielded transaction (and repeats the construction if this is the case). However, theoretically it's possible for the epoch to change only after transaction creation, but before submission. In this (unlikely) case, the client would fail. Also, if a transaction is somehow rejected due to a epoch boundary crossing (caused for example by E2E epochs being too short), the cause is still obscure to the user.
2. The code for reconstructing a transaction in the case where the epoch did change is essentially duplicated from the code that initially attempts to create a transaction. This could end up being a maintenance burden if the two codes for constructing a transaction fall out of sync.
3. The `vp_masp` code for checking that the value balance is legal assumes that the MASP value balance has at least one component (i.e. it is non-zero) and that the asset type of the first component of the value balance is constructed from the token of the wrapper transparent `Transfer`. The first assumption not being met could crash the VP, the second assumption not being met could lead to validation bugs down the line.

Issue (1) could be solved by checking after a transaction has been rejected whether the cause was due to the epoch having changed. If the epoch did change, then an informative error message could be printed (or maybe a retry could be attempted).

[...]

PR #357, opened 2022-08-19
Handle MASP Transactions Across Epoch Boundaries

Scar26:
Solves: #307 
Depends on: #257 

Reduces the window for transaction invalidations caused by epoch change

- Store merkle root of conversion tree from previous epoch to check older conversions
- Change dummy VP to accept unshielding of old tokens while still rejecting shielding to older asset types.
- Check if epoch changes during a shielded transaction build, recompute output descriptions accordingly (everything else remains valid)


tzemanovic:
I tried to speed up the epochs in `e2e::ledger_tests::masp_incentives`, but the client seems to crash when the ledger crosses epochs during the shielded transaction construction:

```
...
Last committed epoch: 4
Last committed epoch: 5
Message:  called `Result::unwrap()` on an `Err` value: ChangeIsNegative(Amount({AssetType { identifier: [148, 243, 79, 253, 100, 239, 10, 195, 105, 8, 253, 223, 236, 5, 104, 88, 6, 41, 196, 86, 113, 15, 161, 134, 131, 18, 168, 127, 191, 10, 165, 9], nonce: Some(0) }: -10000, AssetType { identifier: [236, 12, 219, 18, 244, 149, 41, 63, 83, 160, 74, 65, 174, 240, 156, 191, 226, 128, 163, 249, 179, 171, 76, 131, 86, 103, 175, 163, 88, 201, 49, 132], nonce: Some(3) }: -20000000}))
Location: apps/src/lib/client/tx.rs:1568
```

Scar26:
> I tried to speed up the epochs in `e2e::ledger_tests::masp_incentives`, but the client seems to crash when the ledger crosses epochs during the shielded transaction construction:

Should work now

PR #297, opened 2022-08-08
multitoken transfer

tzemanovic:
moved from https://github.com/anoma/anoma/pull/1200

closes https://github.com/anoma/namada/issues/132

PR #359, opened 2022-08-20
multitoken transfer and query

### LINKED ISSUE #132: `anomac transfer` should support multitoken transfers
We will want accounts on Anoma which handle the ledgers for more than one token, for example this is how wrapped ERC20 assets will be kept track of. (https://github.com/anoma/namada/issues/114)

```
#<eth-bridge-bech32m-address>/ERC20/<erc20-ethereum-address>/balance/<owner-bech32m-address> : Amount
```

e.g.

```
#atest1v4ehgw36g4pyg3j9x3qnjd3cxgmyz3fk8qcrys3hxdp5xwfnx3zyxsj9xgunxsfjg5u5xvzyzrrqtn/ERC20/0x6B175474E89094C44Da98b954EedeAC495271d0F/balance/atest1d9khqw36x9zyxwfhgfpygv2pgc65gse4gy6rjs34gfzr2v69gy6y23zpggurjv2yx5m52sesu6r4y4
```

More generally we want to support transfers for any arbitrary "<token-id>/balance/<owner-bech32m-address>` key suffix that isn't necessarily top level. So in general we could have multitoken accounts like

```
#<some-top-level-bech32m-address>/.../<token-id>/balance/<owner-bech32m-address>
```

This issue is to update `anomac transfer` to work with such multitokens, specifically we need a tx wasm that does the transfer correctly - the validity predicate(s) for validating that such a transfer happened is out of scope for this issue and may differ slightly between different multitokens.

One way would do this issue would be to update the existing `tx_transfer` to work with any arbitrary storage key prefix (e.g. `#<eth-bridge-bech32m-address>/erc20/<erc20-ethereum-address>/...`) rather than a top-level Anoma address. Alternatively we could make a new separate `tx_multitoken_transfer` that the `anomac transfer` should delegate to using in the case of multitoken transfers.

yito88:
closes #132

split #297 which has lots IBC related changes

### DIFF
--- a/apps/src/lib/client/rpc.rs
+++ b/apps/src/lib/client/rpc.rs
@@ -27,7 +27,7 @@ use namada::types::governance::{
     OfflineProposal, OfflineVote, ProposalVote, TallyResult,
 };
 use namada::types::key::*;
-use namada::types::storage::{Epoch, PrefixValue};
+use namada::types::storage::{Epoch, Key, KeySeg, PrefixValue};
 use namada::types::token::{balance_key, Amount};
 use namada::types::{address, storage, token};
 use tendermint::abci::Code;
@@ -101,15 +101,29 @@ pub async fn query_balance(ctx: Context, args: args::QueryBalance) {
         (Some(token), Some(owner)) => {
             let token = ctx.get(&token);
             let owner = ctx.get(&owner);
-            let key = token::balance_key(&token, &owner);
+            let key = match &args.sub_prefix {
+                Some(sub_prefix) => {
+                    let sub_prefix = Key::parse(sub_prefix).unwrap();
+                    let prefix =
+                        token::multitoken_balance_prefix(&token, &sub_prefix);
+                    token::multitoken_balance_key(&prefix, &owner)
+                }
+                None => token::balance_key(&token, &owner),
+            };
             let currency_code = tokens
                 .get(&token)
                 .map(|c| Cow::Borrowed(*c))
                 .unwrap_or_else(|| Cow::Owned(token.to_string()));
             match query_storage_value::<token::Amount>(&client, &key).await {
-                Some(balance) => {
-                    println!("{}: {}", currency_code, balance);
-                }
+                Some(balance) => match &args.sub_prefix {
+                    Some(sub_prefix) => {
+                        println!(
+                            "{} with {}: {}",
+                            currency_code, sub_prefix, balance
+                        );
+                    }
+                    None => println!("{}: {}", currency_code, balance),
+                },
                 None => {
                     println!("No {} balance found for {}", currency_code, owner)
                 }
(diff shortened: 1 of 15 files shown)


yito88:
pls update wasm

apps/src/lib/client/rpc.rs
```diff
@@ -119,12 +133,44 @@ pub async fn query_balance(ctx: Context, args: args::QueryBalance) {
...
+                                            "{}: {}",
+                                            currency_code, balance
+                                        )
+                                        .unwrap();
+                                        found_any = true;
+                                    }
+                                }
+                            }
+                        }
+                    }
```
batconjurer, line 173: This code block exists three times identically.  Maybe use a helper func

shared/src/types/token.rs
```diff
@@ -297,6 +314,54 @@ pub fn is_non_owner_balance_key(key: &Key) -> Option<&Address> {
...
+}
+
+fn multitoken_balance_owner(key: &Key) -> Option<(Key, &Address)> {
+    let len = key.segments.len();
+    if len < 4 {
+        // the key of a multitoken should have 1 or more segments other than
+        // token, balance, owner
+        return None;
+    }
+    match key.get_at(len - 2) {
```
batconjurer, line 347: you can use `?` and I think it will simplify the code a bit.

shared/src/types/token.rs
```diff
@@ -297,6 +314,54 @@ pub fn is_non_owner_balance_key(key: &Key) -> Option<&Address> {
...
+        Some(DbKeySeg::StringSeg(balance))
+            if balance == BALANCE_STORAGE_KEY =>
+        {
+            match key.segments.last() {
```
batconjurer, line 351: same here, `?` should tighten this up

batconjurer (changes requested):

apps/src/lib/client/rpc.rs
yito88: I added `print_balances` func.

shared/src/types/token.rs
yito88: Thanks! I tried to use a delimiter to simplify it.

apps/src/lib/client/rpc.rs
```diff
@@ -1332,14 +1368,7 @@ where
...
                         match T::try_from_slice(&value[..]) {
-                            Err(err) => {
-                                eprintln!(
-                                    "Skipping a value for key {}. Error in \
-                                     decoding: {}",
-                                    key, err
-                                );
-                                None
-                            }
+                            Err(_) => None,
```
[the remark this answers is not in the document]
yito88: Good question!
`query_storage_prefix` is called for the balance read.
The error message frequently appears when the validity predicate is read even though we want to read just `Amount`.
I think it is more readable without the error message.

batconjurer (approved):

james-chf:
I am testing this out now with the Ethereum bridge multitoken functionality. I think there is some code/terminology we can standardize or share between IBC multitoken and Ethereum bridge multitoken, that can be a later PR/issue though. Specifically I'd suggest we use a similar sort of terminology to <https://eips.ethereum.org/EIPS/eip-1155#abstract>.

james-chf:
I tried out `namadac balance` and `namadac transfer` for ERC20 multitokens and it works well with it, lgtm!

For ERC20 tokens, checking balances will look like:
```shell
namadac balance \
 	--token atest1v9hx7w36g42ysgzzwf5kgem9ypqkgerjv4ehxgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpq8f99ew \
 	--sub-prefix ERC20/0x6b175474e89094c44da98b954eedeac495271d0f
```

Transfers will look like
```shell
namadac transfer \
 	--token atest1v9hx7w36g42ysgzzwf5kgem9ypqkgerjv4ehxgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpq8f99ew \
 	--sub-prefix ERC20/0x6b175474e89094c44da98b954eedeac495271d0f
 	--source "${VALIDATOR_ADDR}" \
 	--target "${RECEIVER_ADDR}" \
 	--amount 10
```

james-chf (approved):

tzemanovic:
pls update wasm

PR #363, opened 2022-08-21
Fix release pipeline

Fraccaman:


juped:
noting that if they ever address their issue https://github.com/actions/checkout/issues/882 in the future, we may be able to remove this

Fraccaman:
Tested this on my fork, seems like its working

PR #371, opened 2022-08-21
draft: Validator eth secp keys

tzemanovic:
depends on #254 and #318

This PR adds 2 required validator eth bridge secp256k1 keys - hot and cold. Hot key will be used for signing validator set updates and cross-chain transactions (done by the ledger) and cold key for bridge governance (done on-demand by validators).

to-do:
- [x] add fn to derive Eth address from secp256k1 address (or use some crate)
- [x] Add a PoS storage map from eth address to namada address and set the associations when we write eth cold and hot keys
- [ ] update PoS VP to check eth keys
- [ ] update Pos VP test


tzemanovic:
I think we can rebase this on https://github.com/anoma/namada/pull/379

sug0:
note:
```rs
fn read_validator_eth_hot_key(
    &self,
    key: &Self::Address
) -> Option<Self::PublicKey>
```
this method should return `Option<ValidatorEthKey<Self::PublicKey>>` instead

tzemanovic:
this is going to stay in draft for now - it's only partially finished, because some updates depend on other PoS changes that are not in eth-bridge-integration yet, but it unblocked the bridge dev itself

cwgoes:
@batconjurer @tzemanovic Is this PR still relevant? Should it be merged into the Ethereum bridge integration branch, perhaps?

sug0:
@cwgoes it's already in `eth-bridge-integration`, but not in `main`, due to missing some VPs, still

adrianbrink:
Closing this since it's stale. If we need commits from this then @batconjurer @sug0 can pick them into a new PR.

PR #382, opened 2022-08-25
Merge main into the ethereum bridge

batconjurer:
Title says it all


apps/src/lib/client/utils.rs
```diff
@@ -357,10 +353,21 @@ pub async fn join_network(
...
+            let _pk: ed25519::PublicKey = pk.try_to_pk().unwrap();
+            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
+            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
+        }
+        common::PublicKey::Secp256k1(_) => {
+            let _pk: secp256k1::PublicKey = pk.try_to_pk().unwrap();
+            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
+            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
+        }
+    }
```
sug0, line 370: this isn't really our concern for `eth-bridge-integration`, but for `main`, this function should return `Option<TendermintNodeId>`, since AFAIK tendermint keys can never be of type Secp256k1

sug0 (approved):
LGTM. left a semi-unrelated comment

james-chf (approved):

apps/src/lib/client/utils.rs
batconjurer: I guess make an issue for it? I think Tomas looked into it but I forget what the outcome was.

PR #384, opened 2022-08-25
fix pred epoch height

tzemanovic:
based on #327 

The value written to pred_epoch.height was off by one - we're switching on the last block of an epoch, so the start height that should be recorded for the new epoch is `current + 1`


sug0 (approved):

sug0:
integrated these changes here https://github.com/anoma/namada/pull/286

PR #387, opened 2022-08-25
Fix up CI on `eth-bridge-integration`

james-chf:
Quick PR to fix up CI on `eth-bridge-integration`
- run `make fmt`
- update wasms


james-chf:
pls update wasm

sug0:
@james-chf seems like e2e tests are also broken here. I was looking into fixing them yesterday, but got nowhere 😬

james-chf:
> should we try to fix them in this PR as well?

I'd say not in this PR, we can merge these easy fixes but the e2e test fixing will be more involved (should probably open issues for them)

james-chf:
If the changes from this PR are in https://github.com/anoma/namada/pull/286 , then when that merges this PR can be closed

PR #393, opened 2022-08-30
Update specs with batching and lack of vote extensions

batconjurer:
Added the specs for batching proofs Namada side for transfer to Ethereum.
Added details in places where vote extensions were used to specify how the bridge will work without them.


documentation/specs/src/interoperability/ethereum-bridge.md
```diff
@@ -13,13 +13,24 @@ The Namada Ethereum bridge system consists of:
   [ICS20](https://docs.cosmos.network/v0.42/modules/ibc/) fungible token 
   transfers.
 * A set of Ethereum smart contracts.
-* A relayer for submitting transactions to Ethereum.
+* An automated process to send validator set updates to the Ethereum smart 
+  contracts.
```
sug0, line 17: should we mention that without vote extensions this process may fallback to a governance vote?

sug0: as in, we reach the end of an epoch without getting >2/3 of voting power on a validator set update certificate (comprised of signatures from a quorum of validators)

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -54,104 +54,93 @@ keys involved are:
...
-1. The storage value `/eth_msgs/\$msg_hash/seen_by` does not include their
-   address.
-2. It's correctly formatted.
-3. It's reached the required number of confirmations on the Ethereum chain
+For every Namada block proposal, block proposer should include the votes for
+events from other validators into their proposal. If the underlying Tendermint
+version supports vote extensions, consensus invariants guarantee that a
+quorum of votes from the previous block height can be included. Otherwise,
+validators can only submit votes by broadcasting protocol transactions,
+which comes with less guarantees.
```
sug0, line 80: maybe add:
```
.. which comes with less guarantees (i.e. no consensus finality)
```

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -54,104 +54,93 @@ keys involved are:
...
-Each event that a validator is voting to include must be individually signed by 
+The vote of a validator should include the events of the Ethereum blocks they
+have seen via their full node such that:
+1. It's correctly formatted.
+2. It's reached the required number of confirmations on the Ethereum chain
+
+Each event that a validator is voting to include must be individually signed by
 them. If the validator is not voting to include any events, they must still
-provide a signed voted extension indicating this.
+provide a signed empty vector of events to indicate this.
```
sug0, line 89: mention that we sign block heights as a way to prevent replay attacks? or do we already mention this elsewhere?

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -54,104 +54,93 @@ keys involved are:
...
 
-It is also checked that each vote extension came from the previous round, 
-requiring validators to sign over the Namada block height with their vote
-extension. Furthermore, the vote extensions included by the block proposer
-should have at least 2 / 3 of the total voting power of the previous round 
-backing it. Otherwise the block proposer would not have passed the 
-`FinalizeBlock` phase of the last round. These checks are to prevent censorship 
-of events from validators by the block proposer.
+If vote extensions are supported, it is also checked that each vote extension
+came from the previous round, requiring validators to sign over the Namada block
```
sug0, line 119: came from the previous block height

sug0: ah and we did already mention we signed over the block height with the vext, albeit we didn't give a reason for that.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -54,104 +54,93 @@ keys involved are:
...
+height with their vote extension.
+
+Furthermore, the vote extensions included by
+the block proposer should have at least 2 / 3 of the total voting power of the
+previous round backing it. Otherwise the block proposer would not have passed the
```
sug0, line 124: of the previous block height

sug0: or more accurately, of the epoch of the previous block height

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -54,104 +54,93 @@ keys involved are:
...
+`FinalizeBlock` phase of the last round.
```
sug0, line 125: `phase of the last round of the last block.`

sorry I know I'm being a pain in the ass lol

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,88 @@
+# Proofs
+
+A proof for the bridge is a quorum of signatures by a valid validator set
+attached to a message understandable to the Ethereum smart contracts. For
```
sug0, line 4: question: do proofs only include >2/3 of the voting power worth of signatures, or do they require a signature from every validator in the set?

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,88 @@
...
+However, any user may choose to submit this transaction anyway.
+
+This necessitates a Namada node whose job it is to submit these transactions on
+Ethereum at the conclusion of each Namada epoch. This node is called the
+__bridge relayer__. In theory, since this message is publicly available
+on the blockchain, anyone can submit this transaction, but only the
+bridge relayer will be directly compensated by Namada.
+
+All Namada validators will have an option to serve as bridge relayer and
+the Namada ledger will include a process that does the relaying. Since all
```
sug0, line 77: we can probably do the relaying with a tokio task. there's already a smart contract call stub in `eth-bridge-integration` I think

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,88 @@
...
+Namada validators are running Ethereum full nodes, they can monitor
+that the message was relayed correctly by the bridge relayer.
+
+During the `FinalizeBlock` call in the ledger, if the transfer of power
+message is placed on chain, a flag should be set alerting the next block  
+proposer that they are the bridge relayer for this epoch.
+
+If the Ethereum event spawned by relaying their  message gets accepted by the
```
sug0, line 85: extra space between `their` and `message`

documentation/specs/src/interoperability/ethereum-bridge.md
batconjurer: Oh, I do remember. Just had to start typing up the details to realize it.. Still coffeeing up. :sweat_smile:

batconjurer: Actually, I think we should not use governance for this.

james-chf (commented):
I've only really looked at `proofs.md` so far, I have a lot of thoughts on the bridge relayer part of the spec so will write up a longer comment separately

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
...
+Validator changes must be turned into a message that can be communicated to
+smart contracts on Ethereum. These smart contracts need this information
+to verify proofs of actions taken on Namada.
+
+Since this is protocol level information, it is not user prompted and thus
+should not be the responsibility of any user to submit such a transaction.
+However, any user may choose to submit this transaction anyway.
+
+This necessitates a Namada node whose job it is to submit these transactions on
+Ethereum at the conclusion of each Namada epoch. This node is called the
```
james-chf, line 61:
```suggestion
Ethereum by the conclusion of each Namada epoch. This node is called the
```

james-chf:
I had some thoughts on the relaying of transfers of power (i.e. validator set updates).

# With explicit transfer of power message on chain

This is my understanding of the current design in `proofs.md`. An explicit transfer of power message needs to be submitted on chain once there are enough signatures on chain. Then, the next block proposer after that transfer of power message lands on Namada, is identified as the __block relayer__ and they should submit it to Ethereum.

## Submission of the transfer of power message onto Namada chain

The way I think this could work - the transfer of power message should be a protocol transaction injectable by a block proposer only. So if you're block proposer, and you notice that there are enough validator set update signatures on chain for this epoch (i.e. >2/3 by voting power), you can inject a transfer of power message during `PrepareProposal`. Other validators should check it in `ProcessProposal`. Then, after `FinalizeBlock`, the block proposer (or anyone) should submit the transfer of power to Ethereum. The block relayer gets rewarded once a `ValidatorSetUpdate` event is received by Namada.

So a block proposer may choose to be a relayer, or not to, without need of a flag. They should naturally be incentivized to do so though if it's possible, as then they will get the reward. We could also enforce as part of the protocol that validators must not reject valid transfer of power messages included in a block by a block proposer.

## Example flow

- epoch `n` starts
- every (honest) validator broadcasts a protocol transaction containing their signature over the validator set for epoch `n+1`, asap
- after block `h` has been commited, there are >2/3 worth of signatures by voting power on chain (in practice, `h` should be very soon after epoch `n` has started)
- someone broadcasts a transfer of power message (or a block proposer injects it) that is included in block `i`
- the block proposer for block `i` will get the reward, so they're  incentivized to submit the transfer to Ethereum, and do so
- a `ValidatorSetUpdate` event is emitted by the Ethereum bridge smart contracts and the block relayer (i.e. block proposer for block `i` which included the transfer of power message) receives the reward

# Alternative: no explicit transfer of power message on chain

We could not have an explicit transfer of power message on chain, rather anyone could craft such a message offchain from any arbitrary set of validator signatures over the next epoch's validator set, as long as the signatures altogether had >2/3 of voting power. The transfer of power message should include the Namada address to be rewarded, and the emitted `ValidatorSetUpdate` event should contain that Namada address to be rewarded as well.

In this case, anyone can relay the transfer of power, and the only aspect of it recorded on the Namada chain is when the `ValidatorSetUpdate` event is received. The Namada chain should halt if such an event isn't received by the end of the epoch.

## Example flow

From the point at which there are >2/3 worth of signatures by voting power on chain.

- at this point, some actor is incentivized to craft a transfer of power message offchain, using any arbitrary set of signatures that has >2/3 voting power. They submit it to Ethereum and provide a Namada address in the transfer of power message that will receive the reward
- the Ethereum bridge smart contract has at most `n` signatures to check (where `n` is the number of active validators in the current epoch)
- a `ValidatorSetUpdate` event is emitted by the Ethereum bridge smart contracts which includes the Namada address which should be rewarded (this can be selected by the relayer)

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
...
+Such a message is not prompted by any user transaction and thus will have
+to be carried out by a _bridge relayer_. Once the transfer of power
+message is on chain, any time afterwards a Namada bridge process may take
+it to craft the appropriate message to the Ethereum smart contracts.
+
+The details on bridge relayers are below in the corresponding section.
+
+Signing incorrect headers is considered a slashable offense. Anyone witnessing
+an incorrect header that is signed may submit a complaint (a type of transaction)
+to initiate slashing of the validator who made the signature.
```
james-chf, line 48:
```suggestion
Signing incorrect validator sets is considered a slashable offense. Anyone witnessing
an incorrect validator set that is signed may submit a complaint (a type of transaction)
to initiate slashing of the validator who made the signature.
```

documentation/specs/src/interoperability/ethereum-bridge.md
```diff
@@ -1,9 +1,9 @@
 # Ethereum bridge
 
-The Namada - Ethereum bridge exists to mint wrapped ERC20 tokens on Namada 
+The Namada - Ethereum bridge exists to mint ERC20 tokens on Namada 
 which naturally can be redeemed on Ethereum at a later time. Furthermore, it 
-allows the minting of wrapped Namada-native NAM (wNAM) on Ethereum backed 
-by escrowed NAM on Namada.
+allows the minting of wrapped tokens on Ethereum backed by escrowed assets on 
+Namada.
```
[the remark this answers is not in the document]
batconjurer: Probably true. I'll add a clarification.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
+# Proofs
+
+A proof for the bridge is a quorum of signatures by a valid validator set
+attached to a message understandable to the Ethereum smart contracts. For
+transferring value to Ethereum, a proof is a signed Merkle tree root and
+inclusion proofs of assert transfer messages understandable to the Ethereum
+smart contractions, as described in the section on
```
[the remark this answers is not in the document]
batconjurer: A quorum is by definition >2/3 by stake. I got tired of writing >2/3 stake all the time, so I switched to quorum

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
...
+
+This means the at the end of every Namada epoch, a special transaction must be
+sent to the Ethereum smart contracts detailing the new public keys and stake
+of the new validator set. This message must also be signed by at least 2 / 3
+of the current validators as a "transfer of power".
+
+If vote extensions are available, this signed data can be constructed
+using them. Otherwise, validators must send protocol txs to be included on
+the ledger. Once a quorum exist on chain, they can be aggregated into a
+single message that can be relayed to Ethereum. Signing an
```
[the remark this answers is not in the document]
batconjurer: Absolutely. If the block relayer doesn't do their job, there is no reason to halt the chain. So anyone should be able to do it.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
...
+invalid  validator transition set will be considered a slashable offense.
+
+Due to asynchronicity concerns, this message should be submitted well in
+advance of the actual epoch change, perhaps even at the beginning of each
+new epoch. Bridge headers to ethereum should include the current Namada epoch
```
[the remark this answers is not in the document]
batconjurer: Agreed

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
...
+so that the smart contract knows how to verify the headers. In short, there
```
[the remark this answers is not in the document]
batconjurer: Will clarify.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
batconjurer: It's not just validator set updates. It is also just normal transfers.

batconjurer:
> 

This is out of scope for this PR. We should take this discussion elsewhere. Also, the alternative is almost exactly the current design. The main difference is the compensation scheme.

batconjurer:
> I had some thoughts on the relaying of transfers of power (i.e. validator set updates).
> # With explicit transfer of power message on chain
> 
> This is my understanding of the current design in `proofs.md`. An explicit transfer of power message needs to be submitted on chain once there are enough signatures on chain. Then, the next block proposer after that transfer of power message lands on Namada, is identified as the **block relayer** and they should submit it to Ethereum.
> ## Submission of the transfer of power message onto Namada chain
> 
> The way I think this could work - the transfer of power message should be a protocol transaction injectable by a block proposer only. So if you're block proposer, and you notice that there are enough validator set update signatures on chain for this epoch (i.e. >2/3 by voting power), you can inject a transfer of power message during `PrepareProposal`. Other validators should check it in `ProcessProposal`. Then, after `FinalizeBlock`, the block proposer (or anyone) should submit the transfer of power to Ethereum. The block relayer gets rewarded once a `ValidatorSetUpdate` event is received by Namada.
> 
> So a block proposer may choose to be a relayer, or not to, without need of a flag. They should naturally be incentivized to do so though if it's possible, as then they will get the reward. We could also enforce as part of the protocol that validators must not reject valid transfer of power messages included in a block by a block proposer.
> ## Example flow
> 
>     * epoch `n` starts
> 
>     * every (honest) validator broadcasts a protocol transaction containing their signature over the validator set for epoch `n+1`, asap
> 
>     * after block `h` has been commited, there are >2/3 worth of signatures by voting power on chain (in practice, `h` should be very soon after epoch `n` has started)
> 
>     * someone broadcasts a transfer of power message (or a block proposer injects it) that is included in block `i`
> 
>     * the block proposer for block `i` will get the reward, so they're  incentivized to submit the transfer to Ethereum, and do so
> 
>     * a `ValidatorSetUpdate` event is emitted by the Ethereum bridge smart contracts and the block relayer (i.e. block proposer for block `i` which included the transfer of power message) receives the reward
> 
> 
> # Alternative: no explicit transfer of power message on chain
> 
> We could not have an explicit transfer of power message on chain, rather anyone could craft such a message offchain from any arbitrary set of validator signatures over the next epoch's validator set, as long as the signatures altogether had >2/3 of voting power. The transfer of power message should include the Namada address to be rewarded, and the emitted `ValidatorSetUpdate` event should contain that Namada address to be rewarded as well.
> 
> In this case, anyone can relay the transfer of power, and the only aspect of it recorded on the Namada chain is when the `ValidatorSetUpdate` event is received. The Namada chain should halt if such an event isn't received by the end of the epoch.
> ## Example flow
> 
> From the point at which there are >2/3 worth of signatures by voting power on chain.
> 
>     * at this point, some actor is incentivized to craft a transfer of power message offchain, using any arbitrary set of signatures that has >2/3 voting power. They submit it to Ethereum and provide a Namada address in the transfer of power message that will receive the reward
> 
>     * the Ethereum bridge smart contract has at most `n` signatures to check (where `n` is the number of active validators in the current epoch)
> 
>     * a `ValidatorSetUpdate` event is emitted by the Ethereum bridge smart contracts which includes the Namada address which should be rewarded (this can be selected by the relayer)

Although, that didn't seem obvious when I reread the relevant section. Clarified it a bit.

james-chf:
> > 
> 
> This is out of scope for this PR. We should take this discussion elsewhere. Also, the alternative is almost exactly the current design. The main difference is the compensation scheme.

Sure, I wasn't sure about how/when the explicit transfer of power message would be crafted/submitted, will look again at the relaying part of the spec and comment there

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,78 @@
+# Proofs
+
+A proof for the bridge is a quorum of signatures by a valid validator set
+attached to a message understandable to the Ethereum smart contracts. For
+transferring value to Ethereum, a proof is a signed Merkle tree root and
+inclusion proofs of assert transfer messages understandable to the Ethereum
```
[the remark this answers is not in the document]
batconjurer: Yep, should say asset, not assert.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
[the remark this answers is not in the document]
james-chf: We should still halt the chain if a `ValidatorSetUpdate` event is not received by the end of the epoch i.e. it's not confirmed on Namada that the transfer of power was completed successfully. Or rather we should not transition to the next epoch until that happens. Unsure still a bit about this

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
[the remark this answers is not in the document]
james-chf: How come we use the term "headers"? At least in my mind I associate it with the old design and block headers which we are not using. What about just "message"?

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
james-chf: Sorry should have been clearer here, it's the term "header" I thought we should change (thought it was a holdover usage from an older version of the spec)

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -0,0 +1,96 @@
...
+
+This necessitates a Namada node whose job it is to submit these transactions on
+Ethereum by the conclusion of each Namada epoch. This node is called the
+__bridge relayer__. In theory, since this message is publicly available
+on the blockchain, anyone can submit this transaction, but only the
+bridge relayer will be directly compensated by Namada.
+
+The bridge relayer will be chosen to be the proposer of the first block of the 
+new epoch. Anyone else may relay this message, but must pay for the fees out of
+their own pocket.
```
james-chf, line 69: How about letting a block proposer choose whether they want to be the bridge relayer by injecting a transfer of power message into their block? (https://github.com/anoma/namada/pull/393#issuecomment-1232809428 - "Submission of the transfer of power message onto Namada chain").

We should add more details of how/when the transfer of power message is submitted on chain to Namada and who actually submits it. e.g. maybe anyone can craft this transfer of power message and submit it on chain as a normal transaction, or maybe it needs to be a validator or the block proposer. It should happen asap after there are enough signatures on chain of the next epoch's validator set.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
```diff
@@ -0,0 +1,58 @@
...
+
+/// An event transferring some kind of value from Ethereum to Anoma
+pub struct TransferToNamada {
+    /// Quantity of ether in the transfer
+    pub amount: Amount,
+    /// Address on Ethereum of the asset
+    pub asset: EthereumAsset,
+    /// The Namada address receiving wrapped assets on Anoma
+    pub receiver: Address,
+}
```
james-chf, line 37:
```suggestion
/// An event transferring some kind of value from Anoma to Ethereum
pub struct TransferToEthereum {
    /// Quantity of wrapped Asset in the transfer
    pub amount: Amount,
    /// Address of the smart contract issuing the token
    pub asset: EthAddress,
    /// The address receiving assets on Ethereum
    pub receiver: EthAddress,
}
```
to update: `asset` will always be an ERC20 address (at least for this initial version of the bridge). In the case of transferring NAM to Ethereum, `asset` is the ERC20 address of wNAM which is set when we deploy the Ethereum bridge smart contracts.

https://github.com/anoma/ethereum-bridge/blob/a1fcd58f5f51f161d9d70a142aa0e9bc0ae5a52e/scripts/deploy.js#L139-L140

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
```diff
@@ -0,0 +1,58 @@
...
+
+##### Example
+
+For 10 DAI i.e. ERC20([0x6b175474e89094c44da98b954eedeac495271d0f](https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f)) to `atest1v4ehgw36xue5xvf5xvuyzvpjx5un2v3k8qeyvd3cxdqns32p89rrxd6xx9zngvpegccnzs699rdnnt`
+```
+#EthBridge
+    /erc20
+        /0x6b175474e89094c44da98b954eedeac495271d0f
+            /balances
+                /atest1v4ehgw36xue5xvf5xvuyzvpjx5un2v3k8qeyvd3cxdqns32p89rrxd6xx9zngvpegccnzs699rdnnt
```
james-chf, line 48:
```suggestion
#EthBridge
    /ERC20
        /0x6b175474e89094c44da98b954eedeac495271d0f
            /balance
                /atest1v4ehgw36xue5xvf5xvuyzvpjx5un2v3k8qeyvd3cxdqns32p89rrxd6xx9zngvpegccnzs699rdnnt 
```

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
```diff
@@ -0,0 +1,58 @@
...
+  ["multitoken"](https://github.com/anoma/anoma/issues/1102) hierarchy
+- `#EthBridgeEscrow` which will hold in escrow wrapped Namada tokens which have
+  been sent to Ethereum.
+
+#### Wrapped ERC20
+
+If an ERC20 token is transferred to Namada, once the associated 
+`TransferToNamada` Ethereum event is included into Namada, validators mint 
+the  appropriate amount to the corresponding  multitoken balance key for 
+the receiver.
```
james-chf, line 18:
```suggestion
`TransferToNamada` Ethereum event is included into Namada, validators mint 
the appropriate amount to the corresponding  multitoken balance key for 
the receiver, or release the escrowed native Namada token.
```

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
```diff
@@ -0,0 +1,58 @@
...
+                += 10
+```
+
+#### Namada tokens
+
+Any wrapped Namada tokens being redeemed from Ethereum must have an 
+equivalent amount of the native token held in escrow by `#EthBridgeEscrow`.
+Once the associated`TransferToNamada` Ethereum event is included into 
+Namada, validators should simply make a transfer from `#EthBridgeEscrow` to 
+the `receiver` for the appropriate amount and asset.
```
james-chf, line 58: I can spec this out further but we could actually do the release of escrowed NAM as a free derived transaction. In which case the sender only has to make the transaction on the Ethereum side for redeeming wNAM and everything else will happen automatically (as is the case with sending other ERC20s).

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -0,0 +1,135 @@
...
+They will then pay out the fees for each submitted transaction to the user who
+relayed these transactions (still in NAM). These will be paid out from the
+Bridge Pool Escrow.
+
+The idea is that users will only relay transactions from the Bridge Pool
+that make economic sense. This prevents DoS attacks by underpaying fees as
+well as obviating the need for Ethereum gas price oracles. It also means
+that transfers to Ethereum are not ordered, preventing other attack vectors.
+
+The Bridge Pool will be organized as a Merkle tree. Every time it is updated,
```
james-chf, line 71: to clarify, it's a Merkle tree of height 1 including all `TransferToEthereum`s that have ever been added to `/pending_transfers` (in the order in which they were added)?

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -0,0 +1,135 @@
...
+the root of tree must be signed by a quorum of validators. When a user
+wishes to construct a batch of transactions to relay to Ethereum, they
+include the signed tree root and inclusion proofs for the subset of the pool
+they are relaying. This can be easily verified by the Ethereum smart contracts.
+
+If vote extensions are available, these are used to collect the signatures
+over the Merkle tree root. If they are not, these must be submitted as protocol
+transactions, introducing latency to the pool. A user wishing to relay will
+need to wait until a Merkle tree root is signed for a tree that
+includes all the transactions they wish to relay.
```
james-chf, line 81: How often should validators sign a new Merkle root? In the absence of vote extensions, I think we should have some regular schedule like every 500 blocks (~1 hour or so) validators should submit a signature over whatever the Merkle root is in that block (assuming the Merkle root has changed since the last time).

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -0,0 +1,135 @@
...
+    token: NamadaAddress,
+    /// The number of wrapped Namada tokens to mint on Ethereum
+    amount: Amount,
+}
+```
+
+If a user wishes to mint a wrapped Namada token on Ethereum, they must
+submit a transaction on Namada that:
+- stores `MintWrappedNam` on chain somewhere - TBD
+- sends the correct amount of Namada token to `#EthBridgeEscrow`
```
james-chf, line 41:
```suggestion
To redeem wrapped Ethereum assets, a user should make a transaction to burn
their wrapped tokens, which the `#EthBridge` validity predicate will accept.
For sending NAM over the bridge, a user should send their NAM to 
`#EthBridgeEscrow`. In both cases, it's important that the user also adds a
`PendingTransfer` to the [Bridge Pool](#bridge-pool-validity-predicate).
```

I can spec this out further also but in short this will be simpler now that we will only handle NAM, and `PendingTransfer` replaces what was `MintWrappedNam`.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Batching
+
+Ethereum gas fees make it prohibitively expensive in many cases to submit
+the proof for a single transaction over the bridge. Instead, it is typically
+more economical to submit proofs of many transactions in bulk. This batching
+is described in this section.
+
+A pool of transaction from Namada to Ethereum will be kept by Namada. Every
```
james-chf, line 50:
```suggestion
A pool of transfers from Namada to Ethereum will be kept by Namada. Every
```

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -0,0 +1,135 @@
...
+
+The idea is that users will only relay transactions from the Bridge Pool
+that make economic sense. This prevents DoS attacks by underpaying fees as
+well as obviating the need for Ethereum gas price oracles. It also means
+that transfers to Ethereum are not ordered, preventing other attack vectors.
+
+The Bridge Pool will be organized as a Merkle tree. Every time it is updated,
+the root of tree must be signed by a quorum of validators. When a user
+wishes to construct a batch of transactions to relay to Ethereum, they
+include the signed tree root and inclusion proofs for the subset of the pool
```
james-chf, line 74: When a user submits a batch of transactions, should they only need the inclusion proofs and not the signed tree root? As the signed tree root should be being relayed/updated separately.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
batconjurer: The problem is that whoever does it gets paid. So the protocol should specify who gets paid, otherwise we incentivize other validators to reject valid proposals to get funds for themselves.

batconjurer: The latter part is easy, there is just a scattered bunch of protocol txs on chain with a `Signed<ValidatSet>` struct inside. Someone just collects all those together and submits to ethereum. I can add some struct defs. The when is as soon as possible, but it is impossible to enforce a timeline.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
batconjurer: This should be a transfer to Namada, not Ethereum.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
batconjurer: It is a free derived transaction. I'll make that wording clearer.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
batconjurer: This is a hard question. Not signing can halt the bridge since no one can generated proofs of their relayed events. But we do have rate limiting, I should look into this.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
batconjurer: I haven't decided on the underlying tree type yet. I'm actually leaning toward binary so I can reuse existing third party code. It will also likely store the transfers in order of their hashes (so a random order). We remove transfers once we are notified by ethereum that they have been relayed or they time out.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
batconjurer: We don't relay the root separately.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
james-chf: Validators shouldn't reject valid blocks (including ones that have a valid transfer of power message injected by the block proposer), if they do so they should be slashed. As a more general issue, I think validators should have to provide a signed rejection of a block proposal that could later be submitted as evidence, but I'm guessing we don't have anything like this implemented yet. 

If we go with the bridge relayer being designated as the first block proposer of an epoch, I guess they should be naturally incentivized to submit the transfer of power message on chain as soon as it is possible. We should add details to the spec of what this transfer of power will look like

batconjurer: Adding slashing would fix the problem, it's just a lot more engineering effort. So I'm more inclined to have the protocol specify who gets paid.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
batconjurer: Anyway, this is an implementation detail that doesn't affect the final functionality, so it doesn't need to be included in the specs.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
james-chf: > The latter part is easy, there is just a scattered bunch of protocol txs on chain with a `Signed<ValidatSet>` struct inside. Someone just collects all those together and submits to ethereum. I can add some struct defs. The when is as soon as possible, but it is impossible to enforce a timeline.

(submitted my last comment before seeing this)

I might be misunderstanding the spec, can the transfer of power be crafted offchain arbitrarily and submitted, without first being submitted onto Namada?

batconjurer: Yes, I will clarify this.

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
james-chf: sorry the struct should be
```rust
pub struct TransferToNamada {
    /// Quantity of the ERC20 token in the transfer
    pub amount: Amount,
    /// Address of the smart contract issuing the token
    pub asset: EthAddress,
    /// The address receiving wrapped assets on Anoma
    pub receiver: Address,
}
```

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
james-chf: makes sense, so when a batch transfer is submitted to the Ethereum bridge smart contracts, the submitter also provides a signed tree root that is at least recent enough to cover the transfer they are doing (ideally it would be whatever the most recent signed tree root is)

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
james-chf: ok, that was a big misunderstanding on my part, I've been reading this part of the spec as if there was this step where we submit a transfer of power message onto Namada first

james-chf (approved):
lgtm, this improves/updates our current specs a lot, I think more things can be clarified in later PRs. Specifically it would be good to add some example flows around happy paths, edge cases, etc. and it should also make clearer how things work.

I think we need to clearly separate out parts of our design that use vote extensions (since for v1 of the Ethereum bridge we won't have them). We could use mdbook-admonish or something to factor our where they are mentioned - this could be done in some other PR.

PR #395, opened 2022-08-30
dev encoding spec update

tzemanovic:


tzemanovic:
note that the rustdoc links in encoding_spec (`fn with_rust_doc_link`) are currently invalid, because we don't have these hosted anywhere yet

Fraccaman (approved):

Issue #386, opened 2022-08-25
change token units to nano


tzemanovic:

tzemanovic:
we decided to keep micro for now

PR #379, opened 2022-08-24
PoS integration

tzemanovic:
based on #351

Merged PRs that touch PoS:
- #254 
- #256 
- #318 
- #324
- #325 
- #326 
- #327 
- #331
- #335 
- #334
- #380 
- #384 
- #465
- #458
- #442 (already merged, fixes CI)
- #503

PR #400, opened 2022-08-30
Update token parameters and fundamental units

### LINKED ISSUE #391: change votes per token PoS param to be per smallest denomination
depends on #386 

In genesis, we want to configure the `votes_per_token` to be 1/100 vote per namnam. Currently, it's calculated from whole tokens, so we want to change to be able express votes per nano tokens (in parts-per-billion instead of basis points).

Note that our total staked whole tokens will be capped at around `i64::MAX / 8 / 10^7 = 115_292_150_460` (tendermint total voting power has upper limit of `i64::MAX / 8`) (related issue https://github.com/anoma/namada/issues/159)

brentstone:
Closes #391 and #386.

Based on #379.

Previously, fundamental token units were such that 10^6 fundamental units (namnam) comprised a whole NAM token. This is changed to 10^9. The proof-of-stake parameter `votes_per_token` is also changed such that now it is 1/100 votes per fundamental unit, rather than 1/1000 votes per whole token.


tzemanovic:
both issues are closed as design decision has changed

Issue #402, opened 2022-08-31
Genesis validator panics soon after chain start


james-chf:
Commit: 70adf7987ac1aae10f2447adc363b786723cdf8a (`eth-bridge-integration` branch)

`namadan ledger run` crashes while proposing block 2 of a chain. 

To reproduce, using an example `.anoma` directory with a fresh chain - 
[anoma-v0.7.1.tar.gz](https://github.com/anoma/namada/files/9462308/anoma-v0.7.1.tar.gz), e.g.

```shell
tar -xzvf anoma-v0.7.1.tar.gz
cargo build --no-default-features --features namada_apps/eth-fullnode --package namada_apps
target/debug/namadan ledger run
```

Log snippet
```
2022-08-31T14:20:00.415288Z DEBUG namada_apps::node::ledger: Request Commit
2022-08-31T14:20:00.416748Z  INFO namada_apps::node::ledger::shell: Committed block hash: 66e147812197a4d249bece765af8f5c848772eed1c0514de54621927c360f4ce, height: 1
2022-08-31T15:20:00+01:00 INFO committed state app_hash=66E147812197A4D249BECE765AF8F5C848772EED1C0514DE54621927C360F4CE height=1 module=state num_txs=0
2022-08-31T15:20:01+01:00 INFO Timed out dur=921.257 height=2 module=consensus round=0 step=1
2022-08-31T14:20:01.357332Z DEBUG namada_apps::node::ledger: Request PrepareProposal
2022-08-31T14:20:01.357416Z DEBUG namada::ledger::storage: storage read key #atest1v9hx7w362pex7mmxyphkvgznw3skkefqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqq8ylv7/validator_set
2022-08-31T14:20:01.357950Z DEBUG namada::ledger::storage: storage read key #atest1v9hx7w362pex7mmxyphkvgznw3skkefqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqq8ylv7/validator_set
2022-08-31T14:20:01.358217Z DEBUG namada::ledger::storage: storage read key #atest1v4ehgw36x5myzdfcg5myg3feg4p5vvzyxfrrgdpnxccnvvp5gdrrqse3gdz5zv2yxsenqs3ntgy89n/protocol_public_key
[...]

PR #407, opened 2022-08-31
Use current epoch's voting powers when compressing validator set updates

james-chf:
Closes https://github.com/anoma/namada/issues/402

### DIFF
--- a/apps/src/lib/node/ledger/shell/vote_extensions/validator_set_update.rs
+++ b/apps/src/lib/node/ledger/shell/vote_extensions/validator_set_update.rs
@@ -155,13 +155,13 @@ where
         &self,
         vote_extensions: Vec<validator_set_update::SignedVext>,
     ) -> Option<validator_set_update::VextDigest> {
-        let events_epoch =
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/shell/vote_extensions/validator_set_update.rs
```diff
@@ -151,25 +151,17 @@ where
...
-                return None;
-            }
-            let prev_valset_epoch = current_valset_epoch - 1;
-            u64::from(
-                self.storage.get_total_voting_power(Some(prev_valset_epoch)),
-            )
-        };
+        let events_epoch =
+            self.storage.get_epoch(self.storage.last_height).expect(
+                "The epoch of the last block height should always be known",
```
james-chf, line 160: alternatively we could use `self.storage.get_current_epoch().0;` (here and also in `compress_ethereum_events`)

sug0 (approved):
lgtm. it's what we discussed yesterday

batconjurer (approved):

PR #436, opened 2022-09-05
Replace `f64` with `rust_decimal` in token conversions

### LINKED ISSUE #435: Use `rust_decimal` instead of `f64` to perform token conversions and calculations
The `rust_decimal` crate is safer for avoiding round-off errors that can result from some imprecision with floats. This crate is also starting to be used elsewhere in proof-of-stake. We should phase out the use of floats in the code altogether.

brentstone:
Closes #435.

Based on #379 and #400.


tzemanovic:
pls update wasm

tzemanovic (approved):

tzemanovic (dismissed):

brentstone:
pls update wasms

Issue #178, opened 2021-11-29
Wrapper tx replay protection 


tzemanovic:
As noted in https://github.com/anoma/anoma/blob/625bab46c32da6aeb35418869974bfb1ac98cd3e/docs/src/specs/ledger.md: "DKG transactions will include replay protection (this is because we can simply check a counter against the source (i.e. gas payer) of the transaction before the transactions order is committed to by the DKG protocol, which could affect the expected counter order for sources with multiple queued transactions)".

We should add reply protection to the wrapper tx. This can simply be a counter in implicit accounts that is checked and incremented by the wrapper transactions.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763318619271) by [Unito](https://www.unito.io)

grarco:
Closed in favor of #1009

PR #440, opened 2022-09-06
Replay protection specs

grarco:
Specs for replay protection (#178)

### DIFF
--- a/documentation/specs/src/base-ledger/replay-protection.md
+++ b/documentation/specs/src/base-ledger/replay-protection.md
@@ -85,7 +85,7 @@ Replay protection will require interaction with the storage from both the protoc
 
 ## Batching
 
-The implementation proposed in this document doesn't support batching of multiple transactions from a same address in a single block. This is because the order in which transactions will be included in the block by the proposer is not guarateed. An out of order execution of multiple transaction would lead to the failure of some of them (in the worst case, the failure of all of them but the first one executed, in the best case, the failure of only the last transaction). This problem will be amplified by the introduction of Ferveo for DKG which will be able to reorder transactions.
+The implementation proposed in this document doesn't support batching of multiple transactions from a same address in a single block. More specifically, the transactions will all succeed only if they are executed in the intended order, but the order in which transactions will be included in the block by the proposer is not guaranteed. An out of order execution of multiple transactions would lead to the failure of some of them (in the worst case, the failure of all of them but the first one executed, in the best case, the failure of only the last transaction). This problem will be amplified by the introduction of Ferveo for DKG which will be able to reorder transactions.
 
 The Wasm implementation of replay protection can't cope with this problem because every wasm run (of either a transaction or a validity predicate) is only aware of its context, meaning the wasm bytecode and the serialized transaction data. The lack of awareness of the other transactions makes it impossible to develop a replay protection mechanism supporting batching in wasm.
 
@@ -94,7 +94,7 @@ To address this issue there could be two ways:
 - Keep the proposed replay protection in Wasm and implement a batching mechanism in both the client and the ledger to embed more than one transaction in a single `Tx` struct
 - Implement replay protection in protocol for the inner transaction (as discolsed in section [InnerTx](#InnerTx))
 
-Following the second option, the ledger would be able to analyze the validity of the counters, of all the transcations relating to a single address, against the value in storage at the beginning of the block.
+Following the second option, the ledger would be able to analyze the validity of the counters, of all the transactions relating to a single address, against the value in storage at the beginning of the block.
 Finally, it could increment the counter in storage a single time by the correct amount (given by the amount of transactions that were executed with success).
 
 The first option, though, seems to have more benefits. In addition to allowing batching, it's more flexible and it also enables the in-order execution of the transactions included in the batch, which may come in handy in certain cases.
--- a/documentation/specs/src/base-ledger/replay-protection.md
+++ b/documentation/specs/src/base-ledger/replay-protection.md
@@ -79,7 +79,7 @@ The actual Wasm code and data for the transaction are encapsulated inside a stru
 
 `WrapperTx` is the only type of transaction currently accepted by the ledger. It must be protected from replay attacks because, if it wasn't, a malicious user could replay the transaction as is. Even if the inner transaction implemented replay protection or, for any reason, wasn't accepted, the signer of the wrapper would still pay for gas and fees, effectively suffering economic damage.
 
-To prevent the replay of both these transactions we will rely on a set of already processed transactions' digests that will be kept in storage. These digests will be computed on the **signed** transactions. To support this, we'll need a subspace in storage headed by a `ReplayProtection` internal address:
+To prevent the replay of both these transactions we will rely on a set of already processed transactions' digests that will be kept in storage. These digests will be computed on the **unsigned** transactions, to support replay protection even for [multisigned](multisignature.md) transactions: in this case, if hashes were taken from the signed transactions, a different set of signatures on the same tx would produce a different hash, effectively allowing for a replay. To support this, we'll need a subspace in storage headed by a `ReplayProtection` internal address:
 
 ```
 /$ReplayProtectionAddress/$tx0_hash: None
--- a/documentation/specs/src/base-ledger/replay-protection.md
+++ b/documentation/specs/src/base-ledger/replay-protection.md
@@ -115,7 +115,7 @@ This new field will be signed just like the other ones and is therefore subject 
 
 In general, a transaction is valid at the moment of submission, but after that, a series of external factors (ledger state, etc.) might change the mind of the submitter who's now not interested in the execution of the transaction anymore. 
 
-We have to introduce the concept of a lifetime (or timeout) for the transactions: basically, the `Tx` struct will hold an extra field called `expiration` stating the maximum block height up until which the submitter is willing to see the transaction executed. After the specified block height, the transaction will be considered invalid and discarded regardless of all the other checks.
+We have to introduce the concept of a lifetime (or timeout) for the transactions: basically, the `Tx` struct will hold an extra field called `expiration` stating the maximum `DateTimeUtc` up until which the submitter is willing to see the transaction executed. After the specified time, the transaction will be considered invalid and discarded regardless of all the other checks.
 
 By introducing this new field we are setting a new constraint in the transaction's contract, where the ledger will make sure to prevent the execution of the transaction after the deadline and, on the other side, the submitter commits himself to the result of the execution at least until its expiration. If the expiration is reached and the transaction has not been executed the submitter can decide to submit a new transaction if he's still interested in the changes carried by it.
 
@@ -123,11 +123,11 @@ In our design, the `expiration` will hold until the transaction is executed: onc
 
 - Transaction is invalid regardless of the specific state
 - Transaction is executed (either with success or not) and the transaction hash is saved in the storage
-- Expiration block height has passed
+- Expiration time has passed
 
 The first condition satisfied will invalidate further executions of the same tx.
 
-In anticipation of DKG implementation, the current struct `WrapperTx` holds a field `epoch` stating the epoch in which the tx should be executed. This is because Ferveo will produce a new public key each epoch, effectively limiting the lifetime of the transaction (see section 2.2.2 of the [documentation](https://eprint.iacr.org/2022/898.pdf)). Unfortunately, for replay protection, a resolution of 1 epoch (~ 1 day) is too low for the possible needs of the submitters, therefore we need the `expiration` field to hold a maximum `BlockHeight` to increase resolution down to a single block (~ 10 seconds).
+In anticipation of DKG implementation, the current struct `WrapperTx` holds a field `epoch` stating the epoch in which the tx should be executed. This is because Ferveo will produce a new public key each epoch, effectively limiting the lifetime of the transaction (see section 2.2.2 of the [documentation](https://eprint.iacr.org/2022/898.pdf)). Unfortunately, for replay protection, a resolution of 1 epoch (~ 1 day) is too low for the possible needs of the submitters, therefore we need the `expiration` field to hold a maximum `DateTimeUtc` to increase resolution down to a single block (~ 10 seconds).
 
 ```rust
 pub struct Tx {
@@ -136,7 +136,7 @@ pub struct Tx {
     pub timestamp: DateTimeUtc,
     pub chain_id: ChainId,
     /// Lifetime of the transaction, also determines which decryption key will be used
-    pub expiration: BlockHeight,
+    pub expiration: DateTimeUtc,
 }
 
 pub struct WrapperTx {
@@ -156,7 +156,7 @@ pub struct WrapperTx {
 
 Since we now have more detailed information about the desired lifetime of the transaction, we can remove the `epoch` field and rely solely on `expiration`. Now, the producer of the inner transaction should make sure to set a sensible value for this field, in the sense that it should not span more than one epoch. If this happens, then the transaction will be correctly decrypted only in a subset of the desired lifetime (the one expecting the actual key used for the encryption), while, in the following epochs, the transaction will fail decryption and won't be executed. In essence, the `expiration` parameter can only restrict the implicit lifetime within the current epoch, it can not surpass it as that would make the transaction fail in the decryption phase.
 
-The subject encrypting the inner transaction will also be responsible for using the appropriate public key for encryption relative to the targeted block height.
+The subject encrypting the inner transaction will also be responsible for using the appropriate public key for encryption relative to the targeted time.
 
 The wrapper transaction will match the `expiration` of the inner for correct execution. Note that we need this field also for the wrapper to anticipate the check at mempool/proposal evaluation time, but also to prevent someone from inserting a wrapper transaction after the corresponding inner has expired forcing the wrapper signer to pay for the fees.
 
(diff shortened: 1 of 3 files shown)


documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,100 @@
...
+    pub fee: Fee,
+    /// Used to determine an implicit account of the fee payer
+    pub pk: common::PublicKey,
+    /// The epoch in which the tx is to be submitted. This determines
+    /// which decryption key will be used
+    pub epoch: Epoch,
+    /// Max amount of gas that can be used when executing the inner tx
+    pub gas_limit: GasLimit,
+    /// Transaction counter for replay-protection
+    pub tx_counter: u32,
```
tzemanovic, line 27: I think we can use u64, majority of targets running node will be 64 bit so might as well

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,100 @@
...
+}
+``` 
+
+In addition, we need a counter in the storage subspace of every implict address:
+
+```
+/$Address/tx_counter: u32
+```
+
+In `process_proposal` we check that `tx_counter` field in `WrapperTx` is greater or equal to the value currently in storage; if this doesn't hold, the transaction is considered invalid.
```
tzemanovic, line 42: I think the check here has to be for strict equality, because we only increment it by one - if we allow greater than the current value, that could still be replayed, because even after its updated, it would still be at least equal to it. We'd need to increment it to the highest counter value from all txs in a block, but that's still affected by the other issues you describe in batching and it's tricky with the wrapper/inner tx counter interplay, so I think there will be better options for tx batching.

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,100 @@
...
+
+Note that the address whose counter is involved in this process is the one signing the transaction (the same as the `pk` field of the struct).
+
+## InnerTx
+
+The `EncryptedTx` incapsulated inside `WrapperTx` should be protected too. That's because, if it wasn't, an attacker could extract the inner tx, rewrap it, and replay it.\
+To simplify this check we will perform it entirely in Wasm: the check of the counter will be carryed out by the validity predicates while the actual writing of the counter in storage will be done by the transactions themselves. 
+
+To do so, the `SignedTxData` attached to the transaction will hold the current value of the counter in storage. The transaction will simply read the value from storage and increase its value by one. The VP will then check the validity of the signature and, if it's deemed valid, will proceed to checking if the pre value of the counter in storage was less or equal than the one contained in the transaction data and if the post value of the key in storage has been incremented by one.
```
tzemanovic, line 53: same as above, I think this should be "checking if the pre value of the counter in storage was equal to the one contained in the transaction data"

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,100 @@
...
+```rust
+pub struct Tx {
+    pub code: Vec<u8>,
+    pub data: Option<Vec<u8>>,
+    pub timestamp: DateTimeUtc,
+    pub tx_counter: Option<(Address, u32)>
+}
+```
+
+The check would run in `process_proposal` and the update in `finalize_block`, just like for the wrapper transaction. The drawback of this implementation is that it implies the need for an hard fork in case of a modification of the replay protection mechanism.
```
tzemanovic, line 68: I think another issue is that from protocol's perspective, it's not clear who is the source of an inner transaction as that depends on what happens in the code and the validity predicates that get triggered from it should decide when the counter is required and when it's not. However, it's not decidable when `tx_counter` should be `Some` (although we could tie it the specific whitelisted txs set we'll have in Namada, but it wouldn't work with programmable txs in future), or who the address should be (this is however a blocker)

tzemanovic (approved):
LGTM, thanks!

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
+# Replay Protection
+
+Namada supports a replay protection mechanism to prevent the execution of already processed transactions.
+
+## Tendermint
+
+[Tendermint]((https://docs.tendermint.com/v0.33/app-dev/app-development.html#replay-protection)) provides a first layer of protection against replay attacks in its mempool. The mechanism is based on a cache of previously seen transactions. This check though, like all the checks performed in `CheckTx`, is weak, since a malicious validator could always propose a block containing invalid transactions. There's therefore the need for a more robust replay protection mechanism implemented directly in the application.
```
cwgoes, line 7: presumably we aren't using v0.33? (link)

cwgoes: the preface should describe what replay protection _is_ and why we need it in the first place

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+}
+``` 
+
+In addition, we need a counter in the storage subspace of every implict address:
+
+```
+/$Address/tx_counter: u64
+```
+
+In `process_proposal` we check that `tx_counter` field in `WrapperTx` is equal to the value currently in storage; if this doesn't hold, the transaction is considered invalid.
```
cwgoes, line 42: and what happens in this case? (the transaction should be completely rejected, with no storage changes)

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+At last, in `finalize_block`, the protocol updates the counter key in storage, increasing its value by one. Now, if a malicious user tried to replay this transaction, the `tx_counter` in the struct would no longer be equal to the one in storage and the transaction would be deemed invalid.
```
cwgoes, line 44: for transactions for which a fee is paid, I presume? can we clarify this?

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+Note that the address whose counter is involved in this process is the one signing the transaction (the same as the `pk` field of the struct).
+
+## InnerTx
+
+The `EncryptedTx` incapsulated inside `WrapperTx` should be protected too. That's because, if it wasn't, an attacker could extract the inner tx, rewrap it, and replay it.\
+To simplify this check we will perform it entirely in Wasm: the check of the counter will be carryed out by the validity predicates while the actual writing of the counter in storage will be done by the transactions themselves.
```
cwgoes, line 51:
```suggestion
To simplify this check we will perform it entirely in Wasm: the check of the counter will be carried out by the validity predicates while the actual writing of the counter in storage will be done by the transactions themselves. 
```

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+To do so, the `SignedTxData` attached to the transaction will hold the current value of the counter in storage. The transaction will simply read the value from storage and increase its value by one. The VP will then check the validity of the signature and, if it's deemed valid, will proceed to checking if the pre value of the counter in storage was equal to the one contained in the transaction data and if the post value of the key in storage has been incremented by one.
+
+In the specific case of a transfer, since MASP already comes with replay protection as part of the Zcash design (see the [MASP specs](https://specs.namada.net/masp.html) and [Zcash protocol specs](https://zips.z.cash/protocol/protocol.pdf)), the counter in `SignedTxData` should be optional.
+
+An alternative implementation could place the protection for the inner tx in protocol, just like the wrapper one. In this case we would need to extend the `Tx` struct to hold an additional optional field (again, because of MASP) for the transaction counter and the address involved in replay protection, like so:
```
cwgoes, line 57: it's possible, but I think it's preferable to put the replay protection for the inner tx in WASM, it's a bit more appropriate to how the abstractions of the system are intended to work.

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Storage Interface
+
+Replay protection will require interaction with the storage from both the protocol and Wasm. To do so we can take advantage of the `StorageRead` and `StorageWrite` traits to work with a single interface.
+
+## Batching
+
+The implementation proposed in this document doesn't support batching of multiple transactions from a same address in a single block. More specifically, the transactions will all succeed only if they are executed in the intended order, but the order in which transactions will be included in the block by the proposer is not guaranteed. An out of order execution of multiple transactions would lead to the failure of some of them (in the worst case, the failure of all of them but one, in the best case, the failure of only one). This problem will be amplified by the introduction of Ferveo for DKG which will be able to reorder transactions.
```
cwgoes, line 91: well, the outer tx nonce can be incremented in the submitted transactions, right? this at least gives the proposer an incentive to order them correctly since they receive more fees, right?

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+The Wasm implementation of replay protection can't cope with this problem because every wasm run (of either a transaction or a validity predicate) is only aware of its context, meaning the wasm bytecode and the serialized transaction data. The lack of awareness of the other transactions makes it impossible to develop a replay protection mechanism supporting batching in wasm.
```
cwgoes, line 93: I don't follow - why can't the user just increment the counter in the transactions which they submit?

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+To address this issue there could be two ways:
+
+- Keep the proposed replay protection in Wasm and implement a batching mechanism in both the client and the ledger to embed more than one transaction in a single `Tx` struct
```
cwgoes, line 97: we shouldn't need to do this, I think, see above comments

cwgoes (changes requested):
Thanks for the spec, very nice! See comments

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,103 @@
...
+
+In `process_proposal` we check that `tx_counter` field in `WrapperTx` is equal to the value currently in storage; if this doesn't hold, the transaction is considered invalid.
+
+At last, in `finalize_block`, the protocol updates the counter key in storage, increasing its value by one. Now, if a malicious user tried to replay this transaction, the `tx_counter` in the struct would no longer be equal to the one in storage and the transaction would be deemed invalid.
+
+Note that the address whose counter is involved in this process is the one signing the transaction (the same as the `pk` field of the struct).
+
+## InnerTx
+
+The `EncryptedTx` incapsulated inside `WrapperTx` should be protected too. That's because, if it wasn't, an attacker could extract the inner tx, rewrap it, and replay it.\
```
leontiad, line 50: it is not very clear that, even if  EncryptedTx is extracted isn't the purpose of `tx_counter` to protect the replay thereof?What are the steps when unwrapping a transaction to check the counter? Also it would clarify a lot the document to describe the end-to-end communication between involved entities. What is assumed end to end authenticated and encrypted? how wrapperTx is being put on the network? How the integrity of the counter is protected? Are we assuming symmetric encryption?How keys are being exhanged between t he involved parties?

documentation/specs/src/base-ledger/replay-protection.md
grarco: Sorry I don't get it, could you explain better?

documentation/specs/src/base-ledger/replay-protection.md
grarco: Agree, would you like me to delete this part or move it to a "Discarded solutions" section of this spec to keep track of it?

documentation/specs/src/base-ledger/replay-protection.md
grarco: Yes, the outer counter needs to be incremented every time. If the order of transactions in the mempool of the proposer is the intended one, all the transactions will be executed correctly. The issue is when the order is off: at the moment, the protocol doesn't have a way to reorder the txs based on this counter (we could implement it). This of course works assuming that we are going to charge the fee (or at least part of it) for the inclusion in the block only, regardless of the transaction being correctly executed, otherwise, the proposer would just have an incentive to add them to the block without worrying about the order.
But if we will randomize the tx order for DKG then we are stuck with this problem.  
Another scenario is if, for any reason, a transaction doesn't get to the proposer's mempool: in this case, all the transactions with a tx_counter greater than the missing one will fail, even if placed in the correct (ascending) order.

documentation/specs/src/base-ledger/replay-protection.md
cwgoes: under what conditions exactly does the protocol update the counter key?

documentation/specs/src/base-ledger/replay-protection.md
cwgoes: moving it to an "alternatives considered" section sounds good

documentation/specs/src/base-ledger/replay-protection.md
cwgoes: the protocol doesn't need to reorder the transactions as long as it is incentive compatible for the proposer to order them correctly

we don't need to worry about the DKG for now (perhaps such a counter would have to be outside the encrypted inner transaction if the user wants a particular execution order to be enforced)

cwgoes: for the latter case, we should probably prevent transactions with counter different by more than 1 from executing / paying any fee at all

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,285 @@
...
+
+The wasm transaction will simply read the value from storage and increase its value by one. The target key in storage will be the following:
+
+```
+/$Address/inner_tx_counter: u64
+```
+
+The VP of the _source_ address will then check the validity of the signature and, if it's deemed valid, will proceed to check if the pre-value of the counter in storage was equal to the one contained in the `SignedTxData` struct and if the post-value of the key in storage has been incremented by one: if any of these conditions doesn't hold the VP will discard the transactions and prevent the changes from being applied to the storage.
+
+In the specific case of a transfer, since MASP already comes with replay protection as part of the Zcash design (see the [MASP specs](https://specs.namada.net/masp.html) and [Zcash protocol specs](https://zips.z.cash/protocol/protocol.pdf)), the counter in `SignedTxData` is not required and therefore should be optional.
```
cwgoes, line 108:
```suggestion
In the specific case of a shielded transfer, since MASP already comes with replay protection as part of the Zcash design (see the [MASP specs](https://specs.namada.net/masp.html) and [Zcash protocol specs](https://zips.z.cash/protocol/protocol.pdf)), the counter in `SignedTxData` is not required and therefore should be optional.
```

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,285 @@
...
+
+Unfortunately, at the moment, Wasmer doesn't allow [yielding](https://github.com/wasmerio/wasmer/issues/1127) from the execution. For now, the responsibility will be up to the user to provide a valid inner transaction, and, in case of an invalid one, to take actions to prevent a possible replay attack: in essence, the user will be required to submit a new valid transaction to invalidate the counter of the previous one.
```
cwgoes, line 181: this is not safe because there is a race condition - instead, users will need to craft WASM code to wrap logic which might possibly fail in a try/catch block or something like that - is this possible? what happens if the transaction runs out of gas? if we can't make this safe we'll have to find another solution (unfortunately)...

cwgoes (changes requested):
Unfortunately I don't think the proposed InnerTx method is safe, due to a race condition - a proposer (or other user) who sees a failed inner transaction can replay it if the original submitter doesn't have an opportunity to invalidate the inner counter in time. See comment.

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,503 @@
...
+}
+```
+
+This new field will be signed just like the other ones and is therefore subject to the same guarantees explained in the [initial](#encryption-authentication) section. The validity of this identifier will be checked in `process_proposal` for both the outer and inner tx: if a transaction carries an unexpected chain id, it won't be applied, meaning that no modifications will be applied to storage.
+
+### Transaction lifetime
+
+In general, a transaction is valid at the moment of submission, but after that, a series of external factors (ledger state, etc.) might change the mind of the submitter who's now not interested in the execution of the transaction anymore. 
+
+We have to introduce the concept of a lifetime (or timeout) for the transactions: basically, the `Tx` struct will hold an extra field called `expiration` stating the maximum block height up until which the submitter is willing to see the transaction executed. After the specified block height, the transaction will be considered invalid and discarded regardless of all the other checks.
```
tzemanovic, line 118: I think it might be bit more user friendly to use expiry time (we can check it against last committed block time) as blocks duration can vary slightly

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,503 @@
...
+```rust
+pub struct Tx {
+    pub code: Vec<u8>,
+    pub data: Option<Vec<u8>>,
+    pub timestamp: DateTimeUtc,
+    pub chain_id: ChainId
+}
+```
+
+This new field will be signed just like the other ones and is therefore subject to the same guarantees explained in the [initial](#encryption-authentication) section. The validity of this identifier will be checked in `process_proposal` for both the outer and inner tx: if a transaction carries an unexpected chain id, it won't be applied, meaning that no modifications will be applied to storage.
```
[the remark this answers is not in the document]
tzemanovic: ah, I see this is further down under `### Wrapper checks`

tzemanovic (dismissed):

documentation/specs/src/base-ledger/replay-protection.md
```diff
@@ -0,0 +1,503 @@
...
+
+Both in `mempool_validation` and `process_proposal` we will perform a check (together with others, see the [relative](#wrapper-checks) section) on both the digests against the storage to check that neither of the transactions has already been executed: if this doesn't hold, the `WrapperTx` will not be included into the mempool/block respectively. If both checks pass then the transaction is included in the block and executed. In the `finalize_block` function we will add the transaction's hash to storage to prevent re-executions. We will first add the hash of the wrapper transaction. After that, in the following block, we deserialize the inner transaction, check the correct order of the transactions in the block and execute the tx: if it runs out of gas then we'll avoid storing its hash to allow rewrapping and executing the transaction, otherwise we'll add the hash in storage (both in case of success or failure of the tx).
```
cwgoes, line 95: Hmm I don't think this is right - if the transaction is executed and pays some fee, but runs out of gas, we should still prevent replay (the user can issue a new transaction)

otherwise the user may pay more fees than intended

cwgoes (changes requested):

documentation/specs/src/base-ledger/replay-protection.md
grarco: Actually, by the time we get to the execution of the inner transaction, the wrapper hash has already been saved to storage (in the previous block), so you can't replay the wrapper. You can only take the encrypted inner transaction and include it in a new wrapper

cwgoes: I see. I think this is probably fine but we should put a TTL in inner transactions then, it's weird to have some signed data around for a long time which could potentially be executed at any moment

grarco: We do that with this last spec revision. The struct `Tx` will carry an `expiration` field that will define the lifetime of the transaction. The user will be able to set this value and we'll perform a set of checks on it. Since it's implemented in struct `Tx`, this will apply to both wrappers and inners.

cwgoes: OK sounds good

cwgoes (dismissed):

PR #443, opened 2022-09-06
make: use build-doc to only check rustdoc build

tzemanovic:
this is follow-up to https://github.com/anoma/namada/pull/405


Fraccaman (approved):

tzemanovic:
the failing rustdoc build is fixed in https://github.com/anoma/namada/pull/442

Issue #444, opened 2022-09-06
Remove `eth-fullnode` flag and make starting managed `geth` a runtime option


james-chf:
Managed `geth` should still start up by default, but there should be the option to not do. Similarly the oracle should be configurable at runtime to either be on or off (i.e. the mock oracle).

Issue #365, opened 2022-08-21
Make a governance parameter for the Ethereum bridge smart contract address


james-chf:
Currently we hardcode dummy Ethereum addresses for a couple of the bridge smart contracts

https://github.com/anoma/namada/blob/67d9d43b66b0714132cab4e0d18d26b6d353b4f1/apps/src/lib/node/ledger/ethereum_node/oracle.rs#L21-L23

We should instead have a governance parameter for the Ethereum address of the "hub" smart contract and use that for getting the "bridge" and "governance" smart contract addresses. This should be initialized to the zero Ethereum address at genesis as the Ethereum bridge should not launch straight away at chain launch. The Ethereum oracle component should switch away from using the hardcoded dummy addresses.

james-chf:
Will start being used in https://github.com/anoma/namada/issues/686

PR #445, opened 2022-09-06
Attempt to remove `eth-fullnode` feature flag

james-chf:
Closes https://github.com/anoma/namada/issues/444

Whether `geth` will be started up or not becomes a runtime option


apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -196,7 +231,21 @@ pub mod eth_fullnode {
...
+            test_tools::mock_eth_fullnode::EthereumNode::new().await?;
+        Ok((Subprocess::Mock(node), sender))
+    }
+}
+
+/// Represents a subprocess running an Ethereum full node
+pub enum Subprocess {
+    Mock(test_tools::mock_eth_fullnode::EthereumNode),
+    Geth(eth_fullnode::EthereumNode),
+}
```
james-chf, line 251: Having this enum is unfortunate. I tried to make both `EthereumNode` types implement the same (async) trait but ran into issues with futures not being sendable across threads. I think the best refactor to do later on would be to change `run_aux` so that if `ledger.ethereum.geth_startup = false`, we don't start any sort of mock node at all and then we could remove it altogether - but at the moment `run_aux` is sort of hardcoded in our branch to expect five handles to have to manage.

apps/src/lib/node/ledger/mod.rs
```diff
@@ -353,16 +358,10 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
                     abort_sender,
                 )
             }
-            #[cfg(feature = "eth-fullnode")]
-            ethereum_node::oracle::run_oracle(
-                ethereum_url,
-                eth_sender,
-                abort_sender,
-            )
```
sug0, line 361: can this all be collpased into a single match case (more readable IMO), by changing the config struct to take an enum with the different options?

apps/src/lib/config/ethereum.rs
```diff
@@ -8,22 +8,23 @@ pub const DEFAULT_ORACLE_RPC_ENDPOINT: &str = "http://127.0.0.1:8545";
...
     /// The Ethereum JSON-RPC endpoint that the Ethereum event oracle will use
     /// to listen for events from the Ethereum bridge smart contracts
     pub oracle_rpc_endpoint: String,
     /// If this is set to `true`, then instead of the oracle listening for
     /// events at a Ethereum JSON-RPC endpoint, an endpoint will be exposed by
     /// the ledger for submission of Borsh-serialized
     /// [`EthereumEvent`]s
-    #[cfg(not(feature = "eth-fullnode"))]
     pub oracle_event_endpoint: bool,
 }
```
sug0, line 21: see my comment below in `apps/src/lib/node/ledger/mod.rs` regarding oracle startup

sug0 (changes requested):

apps/src/lib/node/ledger/mod.rs
sug0, line 361: can this all be collpased into a single match case (more readable IMO), by changing the config struct to take an enum with the different options?

sug0 (changes requested):

apps/src/lib/node/ledger/mod.rs
sug0: sorry if it isn't clear what this is referring to, it's the entire `let oracle = ...` expr basically.

apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -28,15 +28,51 @@ pub enum Error {
...
+    ethereum_node: Subprocess,
+    abort_recv: Receiver<Sender<()>>,
+) -> Result<()> {
+    match ethereum_node {
+        Subprocess::Mock(node) => monitor_mock(node, abort_recv).await,
+        Subprocess::Geth(node) => monitor_real(node, abort_recv).await,
+    }
+}
+
+async fn monitor_real(
```
batconjurer, line 43: Could we implement a common trait for the nodes the has the `kill` and `wait` methods and then only have one `monitor` method instead of two?

batconjurer (changes requested):

apps/src/lib/node/ledger/ethereum_node/mod.rs
james-chf: :+1: will try implementing an `[async_trait]` again, I tried initially and was defeated because of something to do with sending futures across threads in the real `EthereumNode::new`, but with just the `kill` and `wait` methods it should hopefully just work

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -1,534 +1,520 @@
...
+use super::events::{signatures, PendingEvent};
+#[cfg(test)]
+use super::test_tools::mock_web3_client::Web3;
+
+/// Minimum number of confirmations needed to trust an Ethereum branch
+pub(crate) const MIN_CONFIRMATIONS: u64 = 50;
+
+/// Dummy addresses for smart contracts
+const MINT_CONTRACT: EthAddress = EthAddress([0; 20]);
+const GOVERNANCE_CONTRACT: EthAddress = EthAddress([1; 20]);
```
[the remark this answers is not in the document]
james-chf: Sorry I should've mentioned in the PR description, I only unindented `oracle.rs` (all the code in there was wrapped in a `mod oracle_process` which was no longer necessary) so there's no real code changes there, I can update these though anyway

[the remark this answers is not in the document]
james-chf: Actually these two specific consts will have to be removed soon anyway, we will instead need to get the address of the proxy contract from a governance parameter - https://github.com/anoma/namada/issues/365

apps/src/lib/node/ledger/mod.rs
james-chf: yeah it would be better not to have mutually exclusive config options as well, will try this

james-chf: using an enum is much nicer!

apps/src/lib/node/ledger/ethereum_node/mod.rs
james-chf: added a `Monitorable` trait which gets rid of some code duplication, though this code can still be simplified a lot further. When merging this PR, I can open an issue to make it so that `run_aux` does not need dummy node/oracle handles, since the dummy node/oracle are only needed afaict because `run_aux` needs a hardcoded five tasks to manage at the moment (shell, abci server, broadcaster, eth-fullnode, oracle)

james-chf:
changed the config to use an enum as suggested, I tested out the modes as in the PR description

apps/src/lib/config/ethereum.rs
```diff
@@ -6,25 +6,37 @@ use serde::{Deserialize, Serialize};
...
+    /// oracle is configured to listen for events from the Ethereum bridge
+    /// smart contracts using this endpoint.
+    Managed,
+    /// Do not start a managed `geth` subprocess. Instead of the oracle
+    /// listening for events using a Ethereum JSON-RPC endpoint, an endpoint
+    /// will be exposed by the ledger itself for submission of Borsh-
+    /// serialized [`EthereumEvent`]s. Mostly useful for testing purposes.
+    EventsEndpoint,
+    /// Do not run any components of the Ethereum bridge
+    Off,
```
james-chf, line 23: NB: will add and test a mode for running the oracle but not the managed `geth` subprocess in a separate PR - https://github.com/anoma/namada/pull/253

apps/src/lib/node/ledger/mod.rs
```diff
@@ -340,29 +342,28 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
...
                     eth_sender,
                 )
-            } else {
+            }
+            ethereum::Mode::Off => {
                 ethereum_node::test_tools::mock_oracle::run_oracle(
                     ethereum_url,
                     eth_sender,
                     abort_sender,
                 )
```
sug0, line 361: this should only be on if `testing` or `dev` are enabled, no?

james-chf: Re. discussing before, I was initially thinking the Ethereum event endpoint mode could be guarded behind our current `dev` feature as in production, you should never want to use this. So it could be safer to distribute binaries without that functionality to prevent accidental use. That said, I don't think it's worth it. At least currently, the `dev` flag changes multiple things all at once to make local development easier (e.g. where we look for wasms) but there is no fine grained control so putting this mode behind that flag right now means having to turn on that other functionality. Also, the fake Ethereum event endpoint will be most useful to have in CI e2e tests, which use production binaries.

Separately from the eth bridge/this PR, I think the functionality behind the `dev` flag should be split out into separate runtime option(s) rather than hidden behind one compilation flag. It's a bigger discussion though. For now I'd prefer to keep just the one set of options in `eth-bridge-integration` without anything hidden behind a feature, but we should revisit how to prevent accidental misuse in production.

sug0 (approved):

PR #437, opened 2022-09-05
Shim vote extensions for Tendermint 0.37.x

sug0:
Implements the fifth item of issue https://github.com/anoma/namada/issues/242

With the `abcipp` feature flag enabled, test this PR with this Tendermint version:
https://github.com/heliaxdev/tendermint/releases/tag/v0.1.1-abcipp

Otherwise, use this version:
https://github.com/tendermint/tendermint/tree/0b58342a46b1917a40d506e48e87448dcc8b3131

---

**Note:** Either websockets rpc code, or the pubsub layer in Tendermint are broken, so client transfers performed with `namadac` may hang indefinitely, and e2e tests are definitely broken because of them.

Issue #418, opened 2022-09-01
Fix end-to-end tests in CI for `eth-bridge-integration`


james-chf:
Currently, many e2e tests don't pass for `eth-bridge-integration` in CI (although some that don't pass in CI might pass locally). This can be fixed now we are using a (relatively) stable v0.37 Tendermint version.

We run CI for every merge to `eth-bridge-integration` in https://github.com/anoma/namada/pull/394.

This issue can be closed once all the `#[ignore]`s introduced by https://github.com/anoma/namada/pull/514 have been removed, and those tests are passing.

sug0:
Closing this issue, as https://github.com/anoma/namada/pull/646 seems to have fixed it

PR #447, opened 2022-09-06
Temporarily ignore e2e tests that are failing due to Tendermint restart issue

james-chf:
Relates to https://github.com/anoma/namada/issues/418

The current e2e tests failing for the `eth-bridge-integration` branch are all to do with stopping and then restarting the ledger. These fail due to the version of Tendermint we're currently using not being able to parse vote extensions (https://github.com/anoma/namada/issues/251). We can try reenabling these tests after we have switched to Tendermint v0.37.x (https://github.com/anoma/namada/pull/437) - they should work again without any other changes.


james-chf:
At some point, all the end-to-end tests started failing in CI for this branch (as opposed to just ones to do with stopping and restarting the ledger)

Issue #448, opened 2022-09-06
Spec for the new Ethereum bridge smart contracts


james-chf:
Some main changes from the previous design
- addition of a fourth contract (vault contract) for holding escrowed ERC20s
- flow rate limiting for Namada -> Ethereum transfers
- bridge contract must keep the signed root of a Merkle tree of all Namada -> Ethereum transfers

cwgoes:
@james-chf @Fraccaman what's the status of this?

PR #450, opened 2022-09-07
Updated specs for Ethereum bridge smart contracts

james-chf:
Relates to https://github.com/anoma/namada/issues/448

This PR is a preliminary updated spec for how the Ethereum bridge smart contracts will work, covering:

- at a high level, the five contracts that will make up the bridge
- how we store details of controlling validator sets
- transfers of power 

c.f. code/docs in https://github.com/anoma/ethereum-bridge

### DIFF
--- a/documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
@@ -3,8 +3,12 @@
 ## Overview
 
 The Ethereum bridge is not enabled at the launch of a Namada chain. Instead,
-there is a governance parameter, `eth_bridge_proxy_address`, which is
-initialized to the zero Ethereum address
+there are two governance parameters:
+
+- `eth_bridge_proxy_address`
+- `eth_bridge_wnam_address`
+
+Both are initialized to the zero Ethereum address
 (`"0x0000000000000000000000000000000000000000"`). An overview of the steps to
 enable the Ethereum bridge for a given Namada chain are:
 
@@ -58,8 +62,7 @@ do the Ethereum bridge smart contract deployment.
 If for some reason the validity of the smart contract deployment cannot be
 agreed upon by the validators who will responsible for restarting Namada, it
 must remain possible to restart the chain with the Ethereum bridge still not
-enabled i.e. with `eth_bridge_proxy_address =
-"0x0000000000000000000000000000000000000000"`.
+enabled.
 
 ## Example
 
@@ -102,7 +105,8 @@ validator set does not change at any point.
 
 - Validators coordinate to craft a new genesis file for the chain restart at
   `3400`, with the governance parameter `eth_bridge_proxy_address` set to
-  `0x00000000000000000000000000000000DeaDBeef`
+  `0x00000000000000000000000000000000DeaDBeef` and `eth_bridge_wnam_address` at
+  `0x000000000000000000000000000000000000Cafe`
 
 - The chain restarts at `3400` (the first block of epoch `34`)
 
--- a/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
@@ -133,9 +133,9 @@ Given this data, anyone should be able to make the relevant Ethereum smart
 contract method call, if they are willing to pay the Ethereum gas. A call is
 then authorized to happen if:
 
-- the active validator set specified in the call hashes to *either* of the
+- The active validator set specified in the call hashes to *either* of the
   validator set hashes stored in the smart contract
-- a quorum (i.e. more than 2/3 by voting power) of the signatures over the
+- A quorum (i.e. more than 2/3 by voting power) of the signatures over the
   message are valid
 
 ### Validator set updates
--- a/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
@@ -66,8 +66,8 @@ secp256k1 keys:
   withdrawal from the bridge
 
 These keys are used to control the bridge smart contracts, via signing of
-messages. Validators may be challenged periodically to prove they still retain
-knowledge of these keys.
+messages. Validators should be challenged periodically to prove they still retain
+knowledge of their governance key, which is not regularly used.
 
 ## Deployment
 
--- a/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
@@ -32,9 +32,9 @@ by the _Governance_ contract.
 ### Governance
 
 The _Governance_ contract may "upgrade" the bridge by updating the _Proxy_
-contract to point to a new _Bridge_ contract. It may also withdraw all funds
-from the _Vault_ to any specified Ethereum address, if a quorum of validators
-choose to do so.
+contract to point to a new _Bridge_ contract and/or a new _Governance_ contract.
+It may also withdraw all funds from the _Vault_ to any specified Ethereum
+address, if a quorum of validators choose to do so.
 
 ### wNAM
 
--- a/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
@@ -61,8 +61,8 @@ The _Vault_ contract is fixed forever once the bridge has been deployed.
 When an account on Namada becomes a validator, they must provide two Ethereum
 secp256k1 keys:
 
-- a "hot" key for normal operations
-- a "cold" key for exceptional operations, like emergency withdrawal from the
+- the bridge key - a hot key for normal operations
+- the governance key - a cold key for exceptional operations, like emergency withdrawal from the
   bridge
 
 These keys are used to control the bridge smart contracts, via signing of
--- a/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
+++ b/documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
@@ -1,27 +1,180 @@
 # Ethereum Smart Contracts
 
-The set of Ethereum contracts should perform the following functions:
+## Contracts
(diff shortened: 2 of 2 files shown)


documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
```diff
@@ -3,8 +3,12 @@
 ## Overview
 
 The Ethereum bridge is not enabled at the launch of a Namada chain. Instead,
-there is a governance parameter, `eth_bridge_proxy_address`, which is
-initialized to the zero Ethereum address
+there are two governance parameters:
+
+- `eth_bridge_proxy_address`
+- `eth_bridge_wnam_address`
```
james-chf, line 9: I think any contracts that will not be upgradeable we could store as governance parameters. So perhaps the _Vault_ contract as well. Or we could store only the _Proxy_ address and have validators look up via their oracle the addresses of the other contracts as required. I'm not sure which is the better approach.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+  specifically, for each validator:
+  - their hot Ethereum address
+  - their cold Ethereum address
+  - their voting power on Namada for the epoch when the bridge will launch
+- the total supply of the wNAM ERC20 token, which will represent Namada-native
+  NAM on the EVM chain
+- an initial whitelist of ERC20 tokens that may cross the bridge from Ethereum
+  to Namada - specifically, for each whitelisted ERC20:
+    - the Ethereum address of the ERC20 contract
+    - a cap on the total amount that may cross the bridge, in units of ERC20
```
james-chf, line 88: Not explicitly mentioned here (yet), but the flow rate cap should be the same for every ERC20 (roughly ~5% per day max per ERC20 from Namada to Ethereum)

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+### From Namada to Ethereum
+
+At any time, the _Governance_ and _Bridge_ contracts must store:
+
+- a hash of the current Namada epoch's active validator set
+- a hash of another epoch's active validator set. When the bridge is first
+  deployed, this will also be the current Namada epoch's active validator set,
+  but after the first valset update is submitted to the _Governance_ smart
+  contract, this hash will always be an adjacent Namada epoch's active validator
+  set i.e. either the previous epoch's, or the next epoch's
```
james-chf, line 117: This is probably not the clearest way to explain it, and there is a reference here to `valset update` which isn't defined until later in the spec. The idea is at any time we have two hashes, and (at least) one of them is the current validator set's hash. The other hash should be the next validator set's hash most of the time, but after an epoch transition, we'll be in the situation where we have the current + previous validator sets' hashes, usually only briefly until a valset update has been submitted.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+Given this data, anyone should be able to make the relevant Ethereum smart
+contract method call, if they are willing to pay the Ethereum gas. A call is
+then authorized to happen if:
+
+- the active validator set specified in the call hashes to *either* of the
+  validator set hashes stored in the smart contract
+- a quorum (i.e. more than 2/3 by voting power) of the signatures over the
+  message are valid
+
+### Valset updates
```
sug0, line 141:
```suggestion
### Validator set updates
```

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+- signatures over the message by validators from the this active validator set
+
+Given this data, anyone should be able to make the relevant Ethereum smart
+contract method call, if they are willing to pay the Ethereum gas. A call is
+then authorized to happen if:
+
+- the active validator set specified in the call hashes to *either* of the
+  validator set hashes stored in the smart contract
+- a quorum (i.e. more than 2/3 by voting power) of the signatures over the
+  message are valid
```
sug0, line 139:
```suggestion
- The active validator set specified in the call hashes to *either* of the
  validator set hashes stored in the smart contract.
- A quorum (i.e. more than 2/3 by voting power) of the signatures over the
  message are valid.
```

sug0 (changes requested):

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
james-chf: :+1: have changed all usages of valset to validator set

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+
+### Proxy
+
+The _Proxy_ contract serves as a dumb storage for holding the addresses of other
+contracts, specifically the _Governance_ contract, the _Vault_ contract and the
+current _Bridge_ contract. Once deployed, it is modifiable only by the
+_Governance_ contract, to update the address for which contract is the current
+_Bridge_ contract.
+
+The _Proxy_ contract is fixed forever once the bridge has been deployed.
```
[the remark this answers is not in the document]
james-chf: IIUC as it is implemented right now, the proxy contract stores a map of `bytes32` to `address`es which could be of arbitrary size, but we only store values for keys `"bridge"` and `"governance"`, so the current implementation has this sort of functionality. We should go in the other direction though I think, bounding the contract to exactly what is needed i.e. only 2 slots, to keep things simpler.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+from the _Vault_ for transfers made from Namada to Ethereum. It holds a
+whitelist of ERC20s that may cross the bridge, and this whitelist may be updated
+by the _Governance_ contract.
+
+### Governance
+
+The _Governance_ contract may "upgrade" the bridge by updating the _Proxy_
+contract to point to a new _Bridge_ contract. It may also withdraw all funds
+from the _Vault_ to any specified Ethereum address, if a quorum of validators
+choose to do so.
```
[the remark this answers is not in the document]
james-chf: oh yeah, have added that it can upgrade itself

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
batconjurer: I think we will allow Vault to be upgradeable, but hopefully that won't ever be necessary

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,182 @@
...
+
+### Proxy
+
+The _Proxy_ contract serves as a dumb storage for holding the addresses of other
+contracts, specifically the _Governance_ contract, the _Vault_ contract and the
+current _Bridge_ contract. Once deployed, it is modifiable only by the
+_Governance_ contract, to update the address for which contract is the current
+_Bridge_ contract.
+
+The _Proxy_ contract is fixed forever once the bridge has been deployed.
```
batconjurer, line 21: Actually, it could possibly be replaced with a hard fork. Hopefully never necessary

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,182 @@
...
+
+When an account on Namada becomes a validator, they must provide two Ethereum
+secp256k1 keys:
+
+- the bridge key - a hot key for normal operations
+- the governance key - a cold key for exceptional operations, like emergency
+  withdrawal from the bridge
+
+These keys are used to control the bridge smart contracts, via signing of
+messages. Validators may be challenged periodically to prove they still retain
```
batconjurer, line 69: The hot keys don't need a challenge I guess since once an epoch they are used to sign over valset updates. Of course that only tests a quorum of validators. For cold keys, this is an interesting point.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
```diff
@@ -1,27 +1,180 @@
...
+The contracts should be deployable by anyone to any EVM chain using an automated
+script. The following configuration should be agreed up front by Namada
+governance before deployment:
+
+- details of the initial active validator set that will control the bridge -
+  specifically, for each validator:
+  - their hot Ethereum address
+  - their cold Ethereum address
+  - their voting power on Namada for the epoch when the bridge will launch
+- the total supply of the wNAM ERC20 token, which will represent Namada-native
```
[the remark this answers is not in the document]
batconjurer: That would be my guess as well

batconjurer (approved):

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
sug0: if we use that strategy we discussed some 2 weeks ago or so, to prove ownership of all keys, then you also solve the problem of proving ownership of eth cold keys.

https://excalidraw.com/#json=GzskiES12LxvLl-7Y_wGk,qPn9jmoqX2VUnMbJSm86hQ

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
james-chf: :+1: for this current spec I'll leave it as not specified to be upgradeable, we will need to consider an upgrade process for it if we want it

documentation/specs/src/interoperability/ethereum-bridge/ethereum_smart_contracts.md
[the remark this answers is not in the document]
james-chf: leaving this as a configuration variable for now, as NAM supply will be variable in any case, cc @Fraccaman

james-chf:
There are still some things to update that weren't included in this PR, so leaving the GH issue open for the time being

PR #452, opened 2022-09-07
Update rocksdb version

Fraccaman:


apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -323,10 +323,14 @@ impl DB for RocksDB {
...
         let mut pred_epochs = None;
         let mut address_gen = None;
-        for (key, bytes) in self.0.iterator_opt(
+        for value in self.0.iterator_opt(
             IteratorMode::From(prefix.as_bytes(), Direction::Forward),
             read_opts,
         ) {
+            let (key, bytes) = match value {
+                Ok(data) => data,
+                Err(_) => continue,
```
tzemanovic, line 332: this hides an error (skips to iter's next, if any), so could be confusing if the iterator fails for whatever reason - I think better to make the whole function error out if that happens

tzemanovic:
rebased on #442 and added a changelog

tzemanovic (approved):

PR #335, opened 2022-08-15
refactor storage prefix iter

tzemanovic:
depends on #167

PR #458, opened 2022-09-08
make prefix iterator ordered by storage keys and add reverse order prefix iterator

### LINKED ISSUE #409: storage prefix iter - deterministic order with reverse
for the storage prefix iterator, we need to make sure that we specify a deterministic order of the iteration as this can be used in tx and VPs. Additionally, we should allow to iterate in reverse, which will be needed for certain operations like finding min/max in a prefix without iterating all the sub-keys

tzemanovic:
based on #335 

fixes #409


tzemanovic:
pls update wasm

PR #486, opened 2022-09-20
wip: fee spec

juped:


documentation/specs/src/economics/fee-system.md
```diff
@@ -4,4 +4,36 @@ In order to be accepted by the Namada ledger, transactions must pay fees. Transa
...
+for tokens is defined above. Whitelisted fees may be updated by
+governance using a standard proposal.
+
+When a transaction is submitted with $n$ units of data, its
+responsibility is to pay the predictable flat fee $n * f_{tx} *
+GP_{min}$, hereafter $F$, in some whitelisted token $T$. Generically,
+this may be done by transferring the token (which may be a simple
+transparent NAM transfer, a bridged multitoken transfer, or the
+transparent output of a shielded transaction) to a native VP address
+designated for fee burning. It shall be lawful for a block proposer to
```
cwgoes, line 26: why is this necessary? we don't need to burn any fees - this address can just be the block proposer address directly

cwgoes: yeah - for the MASP we have to use a special protocol address I guess - for the other txs we could just fill in the address at the time of transaction execution, but a protocol address is fine too

cwgoes: but the block proposer shouldn't need to make an extra transaction, the fees should just be distributed to them in EndBlock

bengtlofgren: I agree with Ray that a native account/VP for handling fee distribution to block producers makes the most sense and is the most user-friendly. To my understanding, coinbase is a transaction that is included by the block producer, so in my head it is as simple as extending this coinbase transaction (or adding an additional one depending on ease) to also include withdrawing funds from the address of the native transaction-fee VP?

cwgoes:
@grarco Can you merge this into your PR (if there is any difference worth merging)?

grarco:
> @grarco Can you merge this into your PR (if there is any difference worth merging)?

I think we can drop this one. The important things from it are already in my pr, no need to merge

PR #503, opened 2022-09-22
lazy vec and map

tzemanovic:


tzemanovic:
pls rebuild wasm

tzemanovic:
pls update wasm

PR #505, opened 2022-09-23
Avoid starving other Tokio tasks while waiting for geth to exit

sug0:
Waiting for `geth` to finish had no `.await` points in an infinite loop. This meant waiting for it to finish starved other tasks from making progress on one of the runtime's threads. This PR fixes that.

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/mod.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/mod.rs
@@ -99,7 +99,6 @@ pub mod eth_fullnode {
 
     use async_trait::async_trait;
     use tokio::process::{Child, Command};
-    use tokio::sync::oneshot::error::TryRecvError;
     use tokio::sync::oneshot::{channel, Receiver, Sender};
     use tokio::task::LocalSet;
     use web30::client::Web3;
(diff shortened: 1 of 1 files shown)


apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -209,22 +208,19 @@ pub mod eth_fullnode {
...
-                        } else {
-                            Err(Error::Runtime(status.to_string()))
-                        };
-                    }
-                    Ok(None) => {}
-                    Err(err) => return Err(Error::Runtime(err.to_string())),
+            use futures::future::{self, Either};
+
+            let child_proc = self.process.wait();
+            futures::pin_mut!(child_proc);
```
james-chf, line 214: How come we have to pin the future?

apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -209,22 +208,19 @@ pub mod eth_fullnode {
...
+
+            match future::select(&mut self.abort_recv, child_proc).await {
+                Either::Left((_abort_received, _)) => Err(Error::Oracle),
```
james-chf, line 217:
```suggestion
                Either::Left(_) => Err(Error::Oracle),
```

james-chf (approved):

apps/src/lib/node/ledger/ethereum_node/mod.rs
james-chf: on second thoughts idk about this suggestion, maybe having `_abort_received` is more explicit about what is happening

apps/src/lib/node/ledger/ethereum_node/mod.rs
sug0: because the future we return from `let child_proc = self.process.wait();` is not [`Unpin`](https://doc.rust-lang.org/std/marker/trait.Unpin.html). it fails to compile if we don't pin it to the stack.

sug0:
```
error[E0277]: `from_generator::GenFuture<[static generator@tokio::process::Child::wait::{closure#0}]>` cannot be unpinned
    --> apps/src/lib/node/ledger/ethereum_node/mod.rs:216:56
     |
216  |             match future::select(&mut self.abort_recv, child_proc).await {
     |                   --------------                       ^^^^^^^^^^ within `impl futures::Future<Output = std::result::Result<ExitStatus, std::io::Error>>`, the trait `Unpin` is not implemented for `from_generator::GenFuture<[static generator@tokio::process::Child::wait::{closure#0}]>`
     |                   |
     |                   required by a bound introduced by this call
     |
    ::: /home/sugo/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.0/src/process/mod.rs:1116:37
```

sug0: @james-chf basically this just means that if you `poll()` a future and return `Poll::Pending`, you can't guarantee you'll access the same value on a later call to `poll()` unless it is `Pin`. simultaneously, you can move values out of a `Pin<T>` if `T: Unpin`.

batconjurer (approved):

PR #509, opened 2022-09-26
Fix ABCI+ vote extensions logic

sug0:
Much of our `abciplus` related code still had `abcipp` logic remnants. This PR addresses that.

If there's anything missing, and you notice it, leave it in the comments so we can still include it under this PR.


sug0:
we can probably remove the duped vote checking. this has to be done at the FinalizeBlock stage anyway...

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -1047,4 +1047,62 @@ mod test_prepare_proposal {
             assert_eq!(received, expected_txs);
         }
     }
+
+    /// Test if we are filtering out duped votes for the
+    /// same Ethereum event at different block heights.
+    #[cfg(not(feature = "abcipp"))]
+    #[test]
+    fn test_prepare_proposal_filter_duped_eth_event_votes() {
```
james-chf, line 1055: This is allowed though (for the Tendermint v0.37 design)? We should be testing the opposite I think. Deduping (or rather ignoring of duplicate votes) will happen during `FinalizeBlock`.

apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
```diff
@@ -127,6 +156,9 @@ where
     /// and returns another iterator. The latter yields
     /// valid Ethereum events vote extensions, or the reason why these
     /// are invalid, in the form of a [`VoteExtensionError`].
+    ///
+    /// If `abciplus` is enabled, we check if a validator has not voted more
+    /// than once for some Ethereum event, at different block heights.
```
james-chf, line 161: We will allow this now

james-chf (changes requested):
We need the fix for the last height being used instead of the height specified in vote extensions, but the deduping is not (any longer) necessary I think

james-chf:
just noticed this is a draft PR, the `last_height` fix is worth getting into `eth-bridge-integration` as it's own PR before https://github.com/anoma/namada/pull/517

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -26,7 +26,6 @@ const VALIDATOR_EXPECT_MSG: &str = "Only validators receive this method call.";
 pub enum VoteExtensionError {
     #[error("The vote extension was issued at block height 0.")]
     IssuedAtGenesis,
-    #[cfg(feature = "abcipp")]
     #[error(
         "The vote extension has an unexpected sequence number (e.g. block \
          height)."
```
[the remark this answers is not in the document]
sug0: I think I chose this name because I was trying to be generic over invalid epochs or block numbers. might just rename it to `UnexpectedBlockNumber`, since we abandoned epochs as nonces in validator set updates

apps/src/lib/node/ledger/shell/prepare_proposal.rs
sug0: yeah this test was checking if the deduping logic was correct. I'm gonna purge that from this branch, so this test can also go along with it. it should still be adapted to finalize block though, I think.

apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
sug0: yup. gotta remove this

batconjurer (approved):

james-chf:
When running the ledger, it outputs these sorts of error logs every block?

```
2022-10-03T14:44:37.891091Z ERROR namada_apps::node::ledger::shell::vote_extensions::eth_events: Ethereum events vote extension issued for a block height higher than the chain's last height. ext_height=BlockHeight(2) last_height=BlockHeight(1)
2022-10-03T14:44:37.891142Z ERROR namada_apps::node::ledger::shell::vote_extensions::val_set_update: Validator set update vote extension issued for a block height higher than the chain's last height. ext_height=BlockHeight(2) last_height=BlockHeight(1)
```

james-chf (changes requested):
It is logging those errors each block, when I run the ledger with 
[anoma-ethbridge-2022-09-30.tar.gz](https://github.com/anoma/namada/files/9699419/anoma-ethbridge-2022-09-30.tar.gz)

sug0:
@james-chf nice catch! the underlying cause seems to be the fact we are now extending votes during the `Commit` stage, after we have called `FinalizeBlock`, which updates the chain's height. so, instead of signing over the current decision height, we now should be signing the `last_height` param, during `craft_extension`.

james-chf (approved):

PR #522, opened 2022-09-27
Check Tendermint version is compatible with the ledger on start up

james-chf:


apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -74,8 +114,29 @@ pub async fn run(
...
+            %version_reqs,
+            "Running with supported Tendermint version",
+        );
+    } else {
+        tracing::warn!(
+            %tendermint_path,
+            %version,
+            %version_reqs,
+            "Running with a Tendermint version which is not supported"
+        );
```
james-chf, line 136: Here I actually think we should terminate the ledger by default if the `tendermint version` isn't compatible, to avoid accidentally clobbering our `tendermint` directory, but there needs to also be some easy way to override this for development and using custom Tendermint builds. This functionality could be a separate issue/PR to this one.

batconjurer (approved):

apps/src/lib/node/ledger/tendermint_node.rs
sug0: you can override version requirements when the `TENDERMINT` env var is set, and the `dev` feature flag is on or something. or completely lift version requirements altogether when `dev` is on

sug0: I'm in the middle of testing custom tendermint builds, and this is a bit inconvenient

sug0 (approved):

apps/src/lib/node/ledger/tendermint_node.rs
james-chf: This PR won't cause the ledger to terminate if an unsupported version of Tendermint is being used, just log a warning. Have opened https://github.com/anoma/namada/issues/525 though to change this behaviour after this PR lands, but it is inconvenient if it's not easily overrideable so won't make that change without having an easy way to override it as well.

james-chf: @sug0 I've updated this log message, after #525 only users who explicitly override the defaults and use a custom/unsupported Tendermint build will be able to see this warning

Issue #527, opened 2022-09-29
Make it easier to patch tendermint-rs in our codebase


james-chf:
We depend on packages from a fork of the [tendermint-rs](https://github.com/heliaxdev/tendermint-rs), and some of our other forks that we depend on (e.g. [ibc-rs](https://github.com/heliaxdev/tendermint-rs), [tower-abci](https://github.com/heliaxdev/tower-abci)) also use this tendermint-rs fork. Currently, if we want to update any tendermint-rs packages in the namada workspace, we have to update patching in any of our other forks that are using them as well. This is cumbersome and it should be possible to patch tendermint-rs in the namada codebase without having to patch it in our other forks as well, so long as the version of tendermint-rs we are patching in remains code-compatible with our forks.

For more context on why this is painful, see https://github.com/rust-lang/cargo/issues/7497 and associated issues. 

Fix this in the following branches
- [x] `eth-bridge-integration`
- [x] `main` (this should be fixed when tendermint changes are pushed from `eth-bridge-integration` branch to `main`)

PR #528, opened 2022-09-29
Patch some crates that we have forked at the workspace level

james-chf:
Resolves https://github.com/anoma/namada/issues/527

This PR changes `tendermint-rs`, `ibc-rs` and `tower-abci` forks that we depend on to be patched in the workspace `Cargo.toml`, rather than pulled in directly via git in individual crate `Cargo.toml`s. Feature flags we are using stay specified in individual crate `Cargo.toml`s. We are already following this sort of pattern for our `borsh` fork.

Also in tandem with this PR, our `ibc-rs` and `tower-abci` forks should be updated to do something similar. 

ibc-rs - https://github.com/heliaxdev/ibc-rs/pull/3
tower-abci - https://github.com/heliaxdev/tower-abci/pull/2

Specifically, they shouldn't depend on our `tendermint-rs` fork directly via git but should instead use a `[patch.crates-io]` entry. This means when we patch `tendermint-rs` in `namada`, e.g. changing the `rev` to something different, the patch will propagate to `ibc-rs` and `tower-abci` without us having to go and update the hardcoded `rev` in those forks as well.

TODO
- patch wasm crates as well, to avoid upstream dependencies and wasm-bindgen being pulled in


james-chf:
pls update wasm

james-chf:
This PR just served as a proof of concept and shouldn't actually land since in `main`, since we are using an old Tendermint v0.36-based fork there and are upgrading to Tendermint v0.37 in the `eth-bridge-integration` branch. Instead we should land https://github.com/anoma/namada/pull/531 and then update `main` to use Tendermint v0.37 all at once by merging in the needed changes from `eth-bridge-integration`.

PR #530, opened 2022-09-30
Fix chrono in `eth-bridge-integration` to avoid pulling in wasm-bindgen

james-chf:
Do this by updating to chrono >=0.4.22 and turning off default features, like suggested in Slack.
To test, do
```
make -C wasm/wasm_source tx_transfer
wasm2wat wasm/tx_transfer.wasm > tx_transfer.wat
cat tx_transfer.wat | grep wbindgen
```

After this PR, `wbindgen` doesn't appear anymore

PR #531, opened 2022-09-30
Start using `eth-bridge-integration` for tendermint-rs/ibc-rs/tower-abci dependencies

james-chf:
Depends on https://github.com/anoma/namada/pull/530 (as that fixes some wasm-bindgen issue which blocks this).

tendermint-rs `eth-bridge-integration` branch is the same as the current `bat/abciplus` branch, except it was cleanly rebased onto `v0.23.6` (`bat/abciplus` was based on `v0.23.5`)


sug0 (approved):
tested this PR. seems to work fine!

wasm/tx_template/Cargo.toml
```diff
@@ -25,6 +25,16 @@ borsh-derive = {git = "https://github.com/heliaxdev/borsh-rs.git", rev = "cd5223
...
+# patched to a commit on the `eth-bridge-integration` branch of our fork
+tendermint = {git = "https://github.com/heliaxdev/tendermint-rs.git", rev = "bc0b6ac47b3bfc1ee8b944341b654b867b3b5d0b"}
+tendermint-proto = {git = "https://github.com/heliaxdev/tendermint-rs.git", rev = "bc0b6ac47b3bfc1ee8b944341b654b867b3b5d0b"}
+tendermint-testgen = {git = "https://github.com/heliaxdev/tendermint-rs.git", rev = "bc0b6ac47b3bfc1ee8b944341b654b867b3b5d0b"}
+tendermint-light-client-verifier = {git = "https://github.com/heliaxdev/tendermint-rs.git", rev = "bc0b6ac47b3bfc1ee8b944341b654b867b3b5d0b"}
+
+# patched to a commit on the `eth-bridge-integration` branch of our fork
+ibc = {git = "https://github.com/heliaxdev/ibc-rs.git", rev = "451ff75591d9de48b7147a651858c3daeb5e2436"}
+ibc-proto = {git = "https://github.com/heliaxdev/ibc-rs.git", rev = "451ff75591d9de48b7147a651858c3daeb5e2436"}
+
```
james-chf, line 37: We didn't have to patch these things in our wasm crates before, as the namada crates were pulling in the forks for us indirectly. Now, because the namada crates are patched at the workspace level, we have to also patch anywhere we use the individual namada crates.

In some sense, it would make sense for individual namada crates to pull in the forks directly rather than via workspace-level `Cargo.toml` patching, but it's not possible to patch individual crates in a workspace - also, we are then unable to patch tendermint-rs and the like whenever we are using those crates outside of the workspace due to https://github.com/rust-lang/cargo/issues/5478 :(

batconjurer (approved):

PR #533, opened 2022-10-03
Fixed the confirmations calculation for ethereum events

batconjurer:
In the audit on the Ethereum bridge, an error in the formula checking that the minimum number of confirmations was reached was discovered. This PR fixes that.

Old:
`self.confirmations >= height.clone() - self.block_height.clone()`
New:
`self.confirmations <= height.clone() - self.block_height.clone()`

This also highlighted a bug in testing the oracle. This PR fixes that as well.

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/events.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/events.rs
@@ -202,7 +202,7 @@ pub mod eth_events {
         /// Check if the minimum number of confirmations has been
         /// reached at the input block height.
         pub fn is_confirmed(&self, height: &Uint256) -> bool {
-            self.confirmations >= height.clone() - self.block_height.clone()
+            self.confirmations <= height.clone() - self.block_height.clone()
         }
     }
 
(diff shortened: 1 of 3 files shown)


sug0 (approved):

james-chf (approved):

apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -202,7 +202,7 @@ pub mod eth_events {
         /// Check if the minimum number of confirmations has been
         /// reached at the input block height.
         pub fn is_confirmed(&self, height: &Uint256) -> bool {
```
james-chf, line 204: We should add a test to cover the issue that this PR fixes i.e. asserting that if the block height difference is greater than `self.confirmations`, then `is_confirmed` does return `true`

batconjurer: I agree.

batconjurer: Actually, we do have a test that catches this bug. The `test_oracle::test_finality_gadget`. It caught a few more bugs even. Fixing now.

batconjurer:
So a bug appeared in the oracle tests when I fixed confirmation formula that I didn't catch at first somehow. I fixed it, but best to review it again.

sug0 (approved):

PR #534, opened 2022-10-03
Remove Websockets from Client code

sug0:
Based on https://github.com/anoma/namada/pull/531.

This PR replaces Websockets client code with a `tendermint-rs` built-in HTTP client. These changes are a precursor to fixing e2e tests in the ledger, most of which are presently failing.

---

Associated dependencies and their respective PRs:

1. `tendermint-rs`: https://github.com/heliaxdev/tendermint-rs/pull/7
2. `ibc-rs`: https://github.com/heliaxdev/ibc-rs/pull/5
3. `tower-abci`: this dep doesn't need a PR. we do have a new `eth-bridge-integration` branch for it though
4. `tendermint`:  https://github.com/heliaxdev/tendermint/pull/162


james-chf:
No change/PR should be necessary for `tower-abci` I think

apps/src/lib/client/tendermint_rpc_types.rs
```diff
@@ -40,77 +40,59 @@ impl TxResponse {
...
+        let code = events[&format!("{evt_key}.code")][index].clone();
+        let gas_used = events[&format!("{evt_key}.gas_used")][index].clone();
+        let initialized_accounts = events
+            [&format!("{evt_key}.initialized_accounts")]
+            .get(index)
+            .as_ref()
+            .map(|initialized_accounts| {
+                serde_json::from_str(initialized_accounts).unwrap()
+            })
+            .unwrap_or_else(|| {
```
sug0, line 82: this code might be incorrect. needs further testing

apps/src/lib/client/tendermint_rpc_types.rs
```diff
@@ -40,77 +40,59 @@ impl TxResponse {
...
                 eprintln!(
-                    "Couldn't find tx with hash {} in the event string {}",
-                    tx_hash, json
+                    "Tendermint omitted one of the expected indices in events"
```
batconjurer, line 84: This is not an error. Most events should not initialize any accounts

apps/src/lib/client/tx.rs
```diff
@@ -1140,28 +1142,12 @@ pub async fn broadcast_tx(
...
-        WebSocketAddress::try_from(address.clone())?,
-        Some(websocket_timeout),
-    )?;
+    let rpc_cli = HttpClient::new(address)?;
 
-    let response = wrapper_tx_subscription
-        .broadcast_tx_sync(tx.to_bytes().into())
-        .await
-        .map_err(|err| WsError::Response(format!("{:?}", err)))?;
+    // TODO: timeout?
```
batconjurer, line 1147: yes, one that can be set with an env var. Needed for the ci

apps/src/lib/client/tx.rs
```diff
@@ -1140,28 +1142,12 @@ pub async fn broadcast_tx(
...
+    let response = rpc_cli.broadcast_tx_sync(tx.to_bytes().into()).await?;
 
-    wrapper_tx_subscription.close();
+    drop(rpc_cli);
```
batconjurer, line 1150: why is this needed? It will drop in 10 lines anyway

batconjurer: If you really need it to drop, use an anonymous scope and return the `response` var out of it

batconjurer (changes requested):

apps/src/lib/client/tendermint_rpc_types.rs
sug0:
```
    "events": {
      "applied.code": [
        "0",
        "0",
        "1"
      ],
      "applied.initialized_accounts": [
        "[]",
        "[]"
      ],
      "tm.event": [
        "NewBlock"
      ],
      "applied.info": [
        "Transaction is valid. Gas used: 0; VPs result: ",
        "Transaction is valid. Gas used: 0; VPs result: ",
        "Tx rejected: Unsupported protocol transaction type"
      ],
      "applied.hash": [
        "3CEDD1EB1A23D3235B10F114F997EB05C3748F59DC69A62984DE8B0019166CEA",
        "B1D3BF49B333C4B80680301E7F5F4A7718B9E44BE74CD95A1E292FD2633E3EFB",
        "7793D575C0903458DD94BE8BAF0D79D19551A86FD38E9AA0152A8781E5E1DF47"
      ],
      "applied.log": [
        "",
        "",
        ""
      ],
      "applied.height": [
        "2",
        "2",
        "2"
      ],
      "applied.gas_used": [
        "0",
        "0",
        "0"
      ]
    }
```
if a tx is rejected (in this example, a valset upd protocol tx), tendermint omits one of the initialized accounts arrays from the RPC response. it's an error because we're relying on fixed indexes to join related data together, which is why we do that index lookup earlier in the file.

sug0: a later pr should honestly just change the way we lookup event data. there's an alternative, more reliable way to do this, through the `EndBlock` data.

sug0: https://p.sicp.me/k64uK example json data

apps/src/lib/client/tx.rs
sug0: no reason, I was experimenting with this to fix websockets errors. this is a remnant of that code.

apps/src/lib/client/tx.rs
sug0: we need to hack away at tendermint-rs once more then. their `HttpClient` uses a hardcoded 30 second timeout iirc.

apps/src/lib/client/tendermint_rpc_types.rs
batconjurer: Oh, ok

apps/src/lib/client/tx.rs
batconjurer: Ah, as long as there is a timeout, we should be fine. We just can't let it run forever.

sug0:
pls update wasm

apps/src/lib/client/tx.rs
```diff
@@ -1198,52 +1184,28 @@ pub async fn submit_tx(
         _ => panic!("Cannot broadcast a dry-run transaction"),
     };
 
-    let websocket_timeout =
-        if let Ok(val) = env::var(ENV_VAR_ANOMA_TENDERMINT_WEBSOCKET_TIMEOUT) {
+    let rpc_timeout =
```
james-chf, line 1187: nit: could we call this max_wait_time or smth since that's the parameter it's used as? A server side expected timeout rather than a client side enforced timeout. At a glance I interpreted this as the latter.

apps/src/lib/client/tx.rs
```diff
@@ -57,8 +54,13 @@ const TX_UNBOND_WASM: &str = "tx_unbond.wasm";
...
 
-const ENV_VAR_ANOMA_TENDERMINT_WEBSOCKET_TIMEOUT: &str =
-    "ANOMA_TENDERMINT_WEBSOCKET_TIMEOUT";
+/// Timeout for jsonrpc requests to the `/events` endpoint in Tendermint.
+const ENV_VAR_ANOMA_TENDERMINT_RPC_TIMEOUT: &str =
+    "ANOMA_TENDERMINT_RPC_TIMEOUT";
+
+/// Default timeout in seconds for jsonrpc requests to the `/events` endpoint in
+/// Tendermint.
+const DEFAULT_ANOMA_TENDERMINT_RPC_TIMEOUT: u64 = 30;
```
james-chf, line 63: We could include events in the name since this is just for the events endpoint e.g. `ENV_VAR_ANOMA_TENDERMINT_EVENTS_RPC_TIMEOUT` or `ENV_VAR_ANOMA_TENDERMINT_EVENTS_MAX_WAIT_TIME`

james-chf (approved):
lgtm!

apps/src/lib/client/tx.rs
```diff
@@ -1140,28 +1142,12 @@ pub async fn broadcast_tx(
...
-    let response = wrapper_tx_subscription
-        .broadcast_tx_sync(tx.to_bytes().into())
-        .await
-        .map_err(|err| WsError::Response(format!("{:?}", err)))?;
-
-    wrapper_tx_subscription.close();
+    // TODO: configure an explicit timeout value? we need to hack away at
+    // `tendermint-rs` for this, which is currently using a hard-coded 30s
+    // timeout.
+    let response = rpc_cli.broadcast_tx_sync(tx.to_bytes().into()).await?;
```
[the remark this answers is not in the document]
james-chf: could be a separate PR/issue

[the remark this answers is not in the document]
sug0: that could work I guess! but yeah, it could be a separate PR

PR #536, opened 2022-10-03
Ethereum oracle should use a bounded channel

batconjurer:
In the audit, they identified an attack vector whereby paying for many invalid ethereum events could crash namada validators with an out of memory error. This is because the channel from the ethereum oracle to the shell was unbounded. They recommended using the bounded channels instead. 

This PR uses the bounded channels in the oracle instead. If the channel fills up, the oracle will halt until there is space again. This does allow an attacker to halt the bridge but:
 * It is an active attack. As soon as the stop sending events, the bridge resumes
 * It is an expensive attack. In order to keep sending events, they must pay ethereum gas fees.


apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
```diff
@@ -309,16 +309,16 @@ mod test_vote_extensions {
             address: EthAddress([0; 20]),
         };
 
-        oracle.send(event_1.clone()).expect("Test failed");
-        oracle.send(event_3.clone()).expect("Test failed");
+        tokio_test::block_on(oracle.send(event_1.clone())).expect("Test failed");
+        tokio_test::block_on(oracle.send(event_3.clone())).expect("Test failed");
```
sug0, line 313: cool, I didn't know `tokio_test` was a thing

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
```diff
@@ -87,7 +87,7 @@ pub mod event_endpoint {
                         }
                     };
                     tracing::debug!("Serialized event - {:#?}", event);
-                    match sender.send(event) {
+                    match sender.try_send(event) {
```
sug0, line 90:
```suggestion
                    match tokio::block_on(sender.send(event)) {
```
not sure if this type checks, but something along these lines, to avoid losing events arbitrarily

sug0 (changes requested):

apps/src/lib/node/ledger/mod.rs
```diff
@@ -44,6 +44,10 @@ const ENV_VAR_TOKIO_THREADS: &str = "ANOMA_TOKIO_THREADS";
 /// Env. var to set a number of Rayon global worker threads
 const ENV_VAR_RAYON_THREADS: &str = "ANOMA_RAYON_THREADS";
 
+/// The maximum number of Ethereum events the channel between
+/// the oracle and the shell can hold.
+const ORACLE_CHANNEL_BUFFER_SIZE: usize = 1000;
```
james-chf, line 49: This const should live elsewhere or even be a `ledger.ethereum` config option (in `apps/src/lib/config/ethereum.rs`), but I think it's fine to leave it as is in this PR and make that change separately

apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -8,7 +8,7 @@ use futures::future::FutureExt;
 use namada::types::ethereum_events::EthereumEvent;
 #[cfg(not(feature = "abcipp"))]
 use namada::types::hash::Hash;
-use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+use tokio::sync::mpsc::{Receiver as BoundedReceiver, UnboundedSender};
```
james-chf, line 11: nit: we can just import `Receiver` as normal, we should get used to interpreting `Sender`/`Receiver` as bounded. We are importing it as `Receiver` in `shell/mod.rs`

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -3,7 +3,7 @@ use std::ops::Deref;
 use clarity::Address;
 use namada::types::ethereum_events::{EthAddress, EthereumEvent};
 use num256::Uint256;
-use tokio::sync::mpsc::UnboundedSender;
+use tokio::sync::mpsc::{Sender as BoundedSender};
```
james-chf, line 6: can we import it as `Sender`? This is what we do elsewhere in the codebase

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -69,12 +69,16 @@ impl Oracle {
...
-            .all(|res| res.is_ok())
-            && !self.sender.is_closed()
+    ///
+    /// N.B. this will block if the internal channel buffer
+    /// is full.
+    async fn send(&self, events: Vec<EthereumEvent>) -> bool {
+        for event in events.into_iter() {
+            if self.sender.send(event).await.is_err() {
+                return false;
+            }
```
james-chf, line 79:
```suggestion
            match self.sender.send(event).await {
                Ok(()) => continue,
                Err(error) => {
                    tracing::warn!(
                        ?error,
                        "Failed to send Ethereum event to ledger",
                    );
                    return false;
                }
            }
```

If it fails, let's log the error. It also looks like we will actually shut down the ledger if this fn returns false?

https://github.com/anoma/namada/blob/64f73c7186d78d73f80f51a5851ff24468d9ad9b/apps/src/lib/node/ledger/ethereum_node/oracle.rs#L199-L205

but we can fix this in another PR

james-chf (changes requested):

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: We probably want to make an issue to refactor how this works a bit separately. We could return a `Result` or smth indicating how many of the `Vec<EthereumEvent>` we were able to send, and then the caller should wait and try again in a bit (when the channel might have some free space)

apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
batconjurer: It is pretty awesome

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
batconjurer: there is no such thing as `tokio:;block_on`. Do you mean `tokio_test::block_on`?

apps/src/lib/node/ledger/mod.rs
batconjurer: Sounds good, will make an issue.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
batconjurer: Yeah, this one has to stay. See the line directly below.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
batconjurer: Since we have to shut down here, I don't think the error message is helpful since a better error message will be immediately issued at line 199. I do think I can remove the `!self.sender.is_closed()` as that is now redundant.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: o i c, we could import `oneshot::Sender` as `oneshot::Sender`, that could be clearer in the code anyway but think it's also fine as is then

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: ahh right, so `.send()` will hang until there's space (compared to `try_send()` which errors if there isn't space), that makes sense. The error returned by `.send` could still be useful as it indicates the `EthereumEvent` that failed to send (so we should be able to work out how many of `Vec<EthereumEvent>`s we managed to actually send). Really this fn should return a `Result` though instead of `bool` and then the error could be logged by the caller with extra context, but we could refactor this later

james-chf (approved):

apps/src/lib/node/ledger/ethereum_node/test_tools.rs
sug0: https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.current

fetch the current handle to the runtime, then call `block_on`

sug0: so instead of calling `map`, we would call `then`. this way we can still call async stuff from this ctx

sug0: another nice option, idk if you prefer, is to use [`flume`](https://docs.rs/flume) channels, which have both a [sync](https://docs.rs/flume/latest/flume/struct.Sender.html#method.send) and [async](https://docs.rs/flume/latest/flume/struct.Sender.html#method.send_async) sendind method

batconjurer: Took a little bit of finagling, but I got the `then` function to work.

sug0 (approved):

Issue #507, opened 2022-09-23
Ledger hangs when shutting down if the Ethereum events endpoint is enabled


james-chf:
As of 09ac240d6a831f56d6f067323fe9c4c2337e552c in the `eth-bridge-integration` branch, starting the ledger with the configuration value `ledger.ethereum.mode = 'EventsEndpoint'`, the ledger will hang and not shut down properly if you send it a SIGINT (e.g. by doing <kbd>Ctrl+C</kbd> in the terminal).

```shell
ANOMA_LEDGER__ETHEREUM__MODE="EventsEndpoint" target/debug/namadan ledger run
```

This is likely to do with the warp server that starts up to serve the events endpoint.

https://github.com/anoma/namada/blob/09ac240d6a831f56d6f067323fe9c4c2337e552c/apps/src/lib/node/ledger/ethereum_node/test_tools.rs#L105

PR #542, opened 2022-10-04
Ensure Ethereum events endpoint shuts down correctly

james-chf:
This is a quick PR to improve the fake Ethereum events endpoint
- refactor and add some docstrings
- make it so that it shuts down correctly (fixes https://github.com/anoma/namada/issues/507)
- listen on `0.0.0.0:3030` rather than `127.0.0.1:3030` so that the endpoint can be more easily exposed from within Docker containers (as will be used in [anoma/ethereum-bridge-e2e-tests](https://github.com/anoma/ethereum-bridge-e2e-tests))

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/test_tools/events_endpoint.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/test_tools/events_endpoint.rs
@@ -2,7 +2,7 @@ use borsh::BorshDeserialize;
 use namada::types::ethereum_events::EthereumEvent;
 use tokio::sync::mpsc::UnboundedSender;
 
-const DEFAULT_ENDPOINT: ([u8; 4], u16) = ([127, 0, 0, 1], 3030);
+const DEFAULT_ENDPOINT: ([u8; 4], u16) = ([0, 0, 0, 0], 3030);
 
 /// The path to which Borsh-serialized Ethereum events should be submitted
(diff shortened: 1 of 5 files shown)


apps/src/lib/node/ledger/ethereum_node/test_tools/events_endpoint.rs
```diff
@@ -0,0 +1,66 @@
...
+        });
+
+    let (_, server) = warp::serve(eth_events).bind_with_graceful_shutdown(
+        DEFAULT_ENDPOINT,
+        async move {
+            tracing::info!(
+                ?DEFAULT_ENDPOINT,
+                "Starting to listen for Borsh-serialized Ethereum events"
+            );
+            poll_fn(|cx| abort_sender.poll_closed(cx)).await;
```
james-chf, line 57: We now shut down in a similar way to how the mock oracle does. That is, we await the `abort_sender` channel that was passed to us to be closed.

sug0 (changes requested):

batconjurer (approved):
Looks good. Will need some changes to handled the bounded channels.

james-chf:
I've made sure the bounded sender changes aren't lost due to the move of that module to new file `events_endpoint.rs`, have restacked and force pushed again since the unit tests CI job seemed stuck, but maybe GitHub actions is having issues?

PR #544, opened 2022-10-05
Remove unused deps

Fraccaman:
According to #236

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4121,7 +4121,6 @@ dependencies = [
  "num_cpus",
  "once_cell",
  "orion",
- "pathdiff",
  "proptest",
  "prost 0.9.0",
  "prost-types 0.9.0",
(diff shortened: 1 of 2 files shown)


tzemanovic:
I think we need to run it with `--all-targets` as winapi is used in windows build (however the win build is broken atm because of another dep that's not win compatible - file-lock)

tzemanovic:
> I think we need to run it with `--all-targets` as winapi is used in windows build (however the win build is broken atm because of another dep that's not win compatible - file-lock)

looks like it's still reported even with `--all-targets` because it is indeed unused in targets other than windows. We need to specify it in `apps/Cargo.toml` as platform specific:

```toml
[target.'cfg(windows)'.dependencies]
winapi = "0.3.9"
```

tzemanovic:
this is probably outdated

PR #548, opened 2022-10-06
Shutdown oracle when the receiver of it's channel is dropped

batconjurer:
This is a bug fix. When sending an message over a bounded channel, if the receiver is dropped, the send method will still block indefinitely. This is despite the  documentation stating that it should error. So we need to add a check that the channel is still open.

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/events.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/events.rs
@@ -202,7 +202,7 @@ pub mod eth_events {
         /// Check if the minimum number of confirmations has been
         /// reached at the input block height.
         pub fn is_confirmed(&self, height: &Uint256) -> bool {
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -202,7 +202,7 @@ pub mod eth_events {
         /// Check if the minimum number of confirmations has been
         /// reached at the input block height.
         pub fn is_confirmed(&self, height: &Uint256) -> bool {
-            self.confirmations >= height.clone() - self.block_height.clone()
+            self.confirmations <= height.clone() - self.block_height.clone()
```
sug0, line 205: wasn't this changed in some other pr?

batconjurer: yep. Got put back somehow

james-chf (approved):

PR #549, opened 2022-10-06
Upgrade to ics23 v0.7.0. 

batconjurer:
This version of ics23 is necessary to work with arse merkle trees. It means that some ibc component have been updated as well.

### DIFF
--- a/shared/src/ledger/ibc/handler.rs
+++ b/shared/src/ledger/ibc/handler.rs
@@ -1109,17 +1109,17 @@ pub fn packet_from_message(
 
 /// Returns a commitment from the given packet
 pub fn commitment(packet: &Packet) -> PacketCommitment {
-    let input = packet
-        .timeout_timestamp
-        .nanoseconds()
-        .to_be_bytes()
-        .to_vec();
+    let timeout = packet.timeout_timestamp.nanoseconds().to_be_bytes();
     let revision_number = packet.timeout_height.revision_number.to_be_bytes();
-    let input = [input.as_slice(), revision_number.as_slice()].concat();
     let revision_height = packet.timeout_height.revision_height.to_be_bytes();
-    let input = [input.as_slice(), revision_height.as_slice()].concat();
     let data = sha2::Sha256::digest(&packet.data);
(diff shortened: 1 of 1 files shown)


yito88 (dismissed):

shared/src/ledger/ibc/handler.rs
```diff
@@ -1111,13 +1111,19 @@ pub fn packet_from_message(
...
+    let mut input = packet
+        .timeout_timestamp
+        .nanoseconds()
+        .to_be_bytes()
+        .to_vec();
+    let revision_number = packet.timeout_height.revision_number.to_be_bytes();
+    input.append(&mut revision_number.to_vec());
+    let revision_height = packet.timeout_height.revision_height.to_be_bytes();
+    input.append(&mut revision_height.to_vec());
+    let data = sha2::Sha256::digest(&packet.data);
```
tzemanovic, line 1124: Just checking here, is the `packet.data` meant to be hashed on its own first?

Unrelated nit, `extend_from_slice` is cheaper than `append`, but slice concat function is better still as it allocates only once for the total size

shared/src/ledger/storage/write_log.rs
```diff
@@ -334,6 +334,7 @@ impl WriteLog {
             HashMap::with_capacity(100),
         );
         self.block_write_log.extend(tx_write_log);
+        self.take_ibc_event();
```
tzemanovic, line 337: ahhh, good catch!

batconjurer: I had to pull Yuji in a call to hunt that down. Luckily missing it fails a test.

shared/src/ledger/ibc/handler.rs
yito88: Yes, the `packet.data` is hashed first, then appended.
Ref. https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/core/04-channel/types/packet.go#L19

yito88 (approved):

tzemanovic (dismissed):

yito88 (dismissed):

Issue #563, opened 2022-10-07
Add new ABCI query endpoints in the ledger to reply with tx events


sug0:
As an alternative to [patching Tendermint 0.37.x with a new `/events` endpoint](https://github.com/heliaxdev/tendermint/pull/164), to fetch ABCI events, we could add new endpoints to the ledger, in the form of ABCI query paths.

Tasks:
- [x] Remove websockets from client code.
- [x] Implement an event log to persist tx events.
- [x] Add new endpoints to the ABCI queries.
- [x] Change client code to call into the ABCI queries events endpoint; clients should poll for events.

Issue #521, opened 2022-09-27
Refactor the `Shell` and associated structs so that the fullnode and validator APIs are split up


james-chf:
When users run a ledger node, whether as a fullnode or a validator, they are ultimately using a `Shell` component, which has an API which includes methods only relevant to validators e.g. `Shell::prepare_proposal` and other ABCI-related methods. On a `Shell`'s storage field, we also import an implementation of a trait called `QueriesExt` (introduced by the `eth-bridge-integration` branch), which has methods only relevant to validators such as `can_send_validator_set_update`. We should refactor so that fullnodes aren't exposed to the validator-only APIs, and the validator-only APIs don't need to have checks that they are being called by a validator and not a full node.

This is maybe worth doing post-Namada launch

cwgoes:
No longer applicable.

PR #564, opened 2022-10-07
Add event log sender to the Shell

sug0:
Based on https://github.com/anoma/namada/pull/534

Completes the first task of issue https://github.com/anoma/namada/issues/563

### DIFF
--- a/apps/src/lib/node/ledger/mod.rs
+++ b/apps/src/lib/node/ledger/mod.rs
@@ -34,7 +34,9 @@ use crate::facade::tendermint_proto::abci::CheckTxType;
 use crate::facade::tower_abci::{response, split, Server};
 use crate::node::ledger::broadcaster::Broadcaster;
 use crate::node::ledger::shell::{Error, MempoolTxType, Shell};
-use crate::node::ledger::shims::abcipp_shim::AbcippShim;
+use crate::node::ledger::shims::abcipp_shim::{
+    AbcippShim, NewAbcippShimParams,
+};
 use crate::node::ledger::shims::abcipp_shim_types::shim::{Request, Response};
 use crate::{config, wasm_loader};
 
@@ -406,7 +408,16 @@ fn start_abci_broadcaster_shell(
 
     // Channels for validators to send protocol txs to be broadcast to the
     // broadcaster service
-    let (broadcaster_sender, broadcaster_receiver) = mpsc::unbounded_channel();
+    let (broadcast_sender, broadcast_receiver) = mpsc::unbounded_channel();
+
+    // Channels for validators to send events to an event log, which will be
+    // queried from the node's RPC `/events` endpoint.
+    //
+    // TODO: make the sender a param of `start_abci_broadcaster_shell`;
+    // the sender will be returned by another task which starts the
+    // event log; we will also need yet another task for the rpc
+    // server itself
+    let (event_log_sender, _event_log_receiver) = mpsc::unbounded_channel();
 
     // Start broadcaster
     let broadcaster = if matches!(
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/mod.rs
```diff
@@ -442,15 +453,17 @@ fn start_abci_broadcaster_shell(
     // Construct our ABCI application.
     let tendermint_mode = config.tendermint.tendermint_mode.clone();
     let ledger_address = config.shell.ledger_address;
-    let (shell, abci_service) = AbcippShim::new(
+    let event_log_sender = Some(event_log_sender);
```
batconjurer, line 456: Shouldn't we pass in a `None` value if we are not a validator?

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -854,35 +894,45 @@ mod test_utils {
...
+        /// Returns a new test shell, and a number of channels to mock
+        /// shell sub-processes.
+        ///
+        /// The returned channels include:
+        ///
+        ///    - A channel which will receive any protocol txs sent by the shell
+        ///      via the [`Shell::broadcast`] method.
+        ///    - A channel that can send Ethereum events into the ledger,
+        ///      mocking the Ethereum fullnode process.
         pub fn new_at_height<H: Into<BlockHeight>>(
```
batconjurer, line 906: I think we need to pass out the events receiver as well. 1. that way we can test it and 2. if it is dropped, finalize will fail when pushing events to it.

sug0: ah yeah for sure, this is still a TODO item for another task (implementing the event log). I'll update the docstring and return type after that's implemented, this is pretty much just a PR in the spirit of "let's add this thing to the ledger, but we won't actually introduce any new functionality yet"

apps/src/lib/node/ledger/mod.rs
sug0: when I implement the event log, I'll update this. this value is pretty much just a placeholder, for the ledger to compile. there's a TODO item above the binding of `event_log_sender` explaining this

apps/src/lib/node/ledger/mod.rs
```diff
@@ -406,7 +408,17 @@ fn start_abci_broadcaster_shell(
...
+    let (broadcast_sender, broadcast_receiver) = mpsc::unbounded_channel();
+
+    // Channels for validators to send events to an event log, which will be
+    // queried from the node's RPC `/events` endpoint.
+    //
+    // TODO: make the sender a param of `start_abci_broadcaster_shell`;
+    // the sender will be returned by another task which starts the
+    // event log; we will also need yet another task for the rpc
+    // server itself. if we are not a validator, `event_log_sender`
+    // should be `None`
```
james-chf, line 420: Shouldn't regular full nodes be providing an `/events` endpoint as well? In which case `event_log_sender` shouldn't be an `Option`

james-chf:
Full nodes should provide this `/events` endpoint - if the events are deterministic based on committed blockchain state, they should be able to produce them as well

sug0:
> Full nodes should provide this `/events` endpoint - if the events are deterministic based on committed blockchain state, they should be able to produce them as well

there are also seed nodes, which don't need the RPC endpoint. but you're right, we do need some changes to deal with full nodes.

sug0:
these APIs with optionals will become a lot better once we refactor the shell (https://github.com/anoma/namada/issues/521)

james-chf (approved):

batconjurer (approved):

PR #566, opened 2022-10-07
Use an enum for the types of events emitted by Ethereum bridge smart contracts

james-chf:


apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -67,6 +42,106 @@ pub mod eth_events {
...
+            vec![
+                EventType::Bridge(BridgeEventType::TransferToNamada),
+                EventType::Bridge(BridgeEventType::TransferToEthereum),
+                EventType::Governance(GovernanceEventType::ValidatorSetUpdate),
+                EventType::Governance(GovernanceEventType::NewContract),
+                EventType::Governance(GovernanceEventType::UpgradedContract),
+                EventType::Governance(
+                    GovernanceEventType::UpdateBridgeWhiteList,
+                ),
+            ]
```
james-chf, line 80: I was trying to avoid having something like this - there is a nice macro for deriving an iterator over all variants of an in commit https://github.com/anoma/namada/pull/566/commits/7417daa8ad15fe9b2294b4dd1fad956a5346c72e , but it requires Rust >=1.63 , so we'd need to wait to update to use it

Issue #561, opened 2022-10-07
Update event signatures for the new Ethereum bridge smart contracts


james-chf:
We have some event signatures hard coded in the ledger:

https://github.com/anoma/namada/blob/85d554217bc5951f1ac77af42a963733ca88086a/apps/src/lib/node/ledger/ethereum_node/events.rs#L2-L11

These signatures need to checked to ensure they are up to date with changes made in https://github.com/anoma/ethereum-bridge as of commit https://github.com/anoma/ethereum-bridge/blob/87056f4c1fa5472e49adaf0793af1f2ecda9b5da

e.g. `TransferToNamada(uint256,address[],string[],uint256[],uint32)` should become `TransferToNamada(uint256,(address,uint256,string)[],uint256)`. 

Also the ABI decoding logic needs to be updated for raw event types, e.g. `RawTransfersToNamada`.

- [x] TransferToNamada (https://github.com/anoma/namada/pull/567)
- [ ] InvalidTransferToNamada (new event type)
- [ ] TransferToErc
- [ ] ValidatorSetUpdate
- [ ] NewContract
- [ ] UpgradedContract
- [ ] UpdateBridgeWhiteList

cwgoes:
@james-chf could we do this as part of the autogenerated types, perhaps?

james-chf:
Yep, the best way to solve this issue would be as part of https://github.com/anoma/namada/issues/986

PR #567, opened 2022-10-07
Update TransferToNamada event signature

james-chf:
Implements part of https://github.com/anoma/namada/issues/561

Update the `TransferToNamada` event signature to match the latest smart contracts - https://github.com/anoma/ethereum-bridge/blob/87056f4c1fa5472e49adaf0793af1f2ecda9b5da/contracts/interface/IBridge.sol#L7

Also update the decoding logic

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/events.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/events.rs
@@ -707,10 +707,10 @@ pub mod eth_events {
         }
 
         fn parse_transfer_to_namada(self) -> Result<TransferToNamada> {
-            if let Token::Tuple(tuple) = self {
-                let asset = tuple[0].clone().parse_eth_address()?;
-                let amount = tuple[1].clone().parse_amount()?;
-                let receiver = tuple[2].clone().parse_address()?;
+            if let Token::Tuple(mut items) = self {
+                let asset = items.remove(0).parse_eth_address()?;
+                let amount = items.remove(0).parse_amount()?;
+                let receiver = items.remove(0).parse_address()?;
                 Ok(TransferToNamada {
                     asset,
                     amount,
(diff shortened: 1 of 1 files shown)


apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -750,6 +783,45 @@ pub mod eth_events {
     mod test_events {
         use super::*;
 
+        #[test]
+        fn test_transfer_to_namada_decode() {
+            let data: Vec<u8> = vec![
```
james-chf, line 788: This event data was copied straight from ledger logs while I was testing out decoding an event from the real smart contracts. This might be better as an example rather than a test.

sug0 (approved):

apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -711,6 +687,43 @@ pub mod eth_events {
...
+            for token in array.into_iter() {
+                let transfer = token.parse_transfer_to_namada()?;
+                transfers.push(transfer);
+            }
+            Ok(transfers)
+        }
+
+        fn parse_transfer_to_namada(self) -> Result<TransferToNamada> {
+            if let Token::Tuple(tuple) = self {
+                let asset = tuple[0].clone().parse_eth_address()?;
```
batconjurer, line 711: I'm surprised we need a `clone` here. The function has owner of tuple and it's not used again elsewhere

batconjurer (approved):

apps/src/lib/node/ledger/ethereum_node/events.rs
james-chf: Calling `parse_` directly on `tuple`'s elements won't work as `tuple` still has ownership of the individual elements (though the `parse_` methods want ownership as well). Have updated the code to rename `tuple` to `items` (since it's actually a `Vec`), and avoid `clone` by moving items out of the vec first before calling the `.parse_` methods, so that they can take ownership.

PR #569, opened 2022-10-07
RPC shell sub-router

tzemanovic:
based on #553 

This PR moves the current RPC paths into a "shell" sub-router. This is a breaking change for all the current RPC paths which are all prefixed with `"/shell"` now :
- `"/shell/dry_run_tx"`
- `"/shell/epoch"`
- `"/shell/value"`
- `"/shell/prefix"`
- `"/shell/has_key"`

This is to take advantage of tree-based pattern matching, organize code with the handlers and to add other sub-routers. Concretely, I'll be adding `"/vp"` sub-router with `"/pos"` inside it for PoS specific queries next.

relevant commits:
- [3835cd5](https://github.com/anoma/namada/pull/569/commits/3835cd5d80a3194e4cac7d7ad851029f687ffde3) - moves current patterns under "shell" sub-router
- [276845c](https://github.com/anoma/namada/pull/569/commits/276845c55189e9ea0aed2b71db4e6c28f5a5a12e) - simplified handlers to be able to specify `(with_options $handler)` for handlers that want to use the request/response, otherwise it's not expected in the handler's signature

### DIFF
--- a/.changelog/unreleased/improvements/569-rpc-sub-shell.md
+++ b/.changelog/unreleased/improvements/569-rpc-sub-shell.md
@@ -1,0 +1,2 @@
+- Move all shell RPC endpoints under the /shell path. This is a breaking change
+  to RPC consumers. ([#569](https://github.com/anoma/namada/pull/569))


tzemanovic:
pls update wasm

Issue #560, opened 2022-10-07
Ethereum oracle should ensure all relevant Ethereum blocks are checked for events


james-chf:
https://github.com/anoma/namada/blob/85d554217bc5951f1ac77af42a963733ca88086a/apps/src/lib/node/ledger/ethereum_node/oracle.rs#L135-L139

Currently, the Ethereum event oracle loops and checks for events in the latest block according to the Ethereum RPC endpoint. If an iteration of the loop takes too long, this means the oracle can miss checking blocks e.g. `latest_block = 100`, then at the next iteration of the loop, `latest_block = 110`. The oracle also only starts checking from the latest block, so if the ledger goes down and starts back up again later, it will not check blocks that it missed inbetween.

Validators could keep track of blocks they checked somehow, and this should persist between ledger restarts. Validators should probably be checking all Ethereum blocks from when they become an active validator, up to the point at which they are no longer an active validator.

Tasks
- [x] oracle should check all blocks in sequence from some appropriate start point (temporarily we can start from the genesis Ethereum block) https://github.com/anoma/namada/pull/571

PR #571, opened 2022-10-10
Update the Ethereum bridge oracle to process blocks in sequence from zero

james-chf:
Relates to https://github.com/anoma/namada/issues/560

This PR makes some changes to the Ethereum oracle
- add more logging
- when the oracle starts up, process all Ethereum blocks in sequence starting from block #0, rather than the latest confirmed block
- make the oracle sleep time configurable (it is currently hardcoded to be 1s between blocks in production, but in tests we want it much shorter)
- make the oracle emit the block heights it processes for observability (in tests)

Processing blocks from 0 every time the ledger starts up is not the final behaviour - validators should process Ethereum blocks only from when they become an active validator, and should stop processing Ethereum blocks once they are no longer an active validator.

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/oracle.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/oracle.rs
@@ -729,7 +729,7 @@ mod test_oracle {
 
     /// Test that if the Ethereum RPC endpoint returns a latest block that is
     /// more than one block later than the previous latest block we received, we
-    /// still check all the blocks inbetween
+    /// still check all the blocks in between
     #[tokio::test]
     async fn test_all_blocks_checked() {
         let TestPackage {
--- a/apps/src/lib/node/ledger/ethereum_node/oracle.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/oracle.rs
@@ -90,12 +90,6 @@ impl Oracle {
         }
         true
     }
-
-    /// Check if the receiver in the ledger has hung up.
-    /// Used to help determine when to stop the oracle
-    fn connected(&self) -> bool {
-        !self.sender.is_closed()
-    }
 }
 
 /// Set up an Oracle and run the process where the Oracle
@@ -128,6 +122,8 @@ pub fn run_oracle(
     })
 }
 
+const SLEEP_DUR: std::time::Duration = std::time::Duration::from_secs(1);
+
 /// Given an oracle, watch for new Ethereum events, processing
 /// them into Anoma native types.
 ///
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -32,6 +37,12 @@ pub struct Oracle {
     /// A channel to signal that the ledger should shut down
     /// because the Oracle has stopped
     abort: Option<Sender<()>>,
+    /// How long the oracle should wait between checking blocks
+    backoff: Duration,
```
james-chf, line 41: This is crude linear backoff for now, mainly for the purpose of stopping the ledger from spamming the Ethereum RPC endpoint once it has caught up syncing, but there should be some more involved logic around this backoff later

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -32,6 +37,12 @@ pub struct Oracle {
...
+    /// If provided, the oracle will attempt to send block heights that it has
+    /// processed to this channel, but won't pause if this channel is full
+    /// or disconnected.
+    blocks_processed: Option<BoundedSender<Uint256>>,
```
james-chf, line 45: We don't initialize this in production right now, only for tests so we can check what blocks the oracle has actually processed. Alternatively we could have something like:

```rust
latest_block_processed: Arc<Mutex<Uint256>>
```

i.e. something which lets us look at what is the latest block height that the oracle has processed, from any thread

sug0: https://docs.rs/tokio/latest/tokio/sync/watch/index.html
use a single producer, multiple consumer buffer?

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -46,8 +57,17 @@ impl Drop for Oracle {
...
+            Ok(()) => tracing::info!("Oracle sent abort signal"),
+            Err(()) => {
+                // this isn't necessarily an issue as the ledger may have shut
+                // down first
+                tracing::debug!(
+                    "Oracle was unable to send an abort signal as the abort \
+                     channel was already closed"
+                )
+            }
+        };
```
sug0, line 70: I think this is impossible in practice. we wait for every sub-task in the ledger to exit gracefully, with the `try_join!()` macro.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -127,100 +153,181 @@ pub fn run_oracle(
...
+            ?next_block_to_process,
+            "Checking Ethereum block for bridge events"
+        );
+        tokio::select! {
+            result = process(&oracle, &mut pending, next_block_to_process.clone()) => {
+                match result {
+                    Ok(()) => {
+                        if let Some(blocks_processed) = &oracle.blocks_processed {
+                            if let Err(error) = blocks_processed.try_send(next_block_to_process.clone()) {
+                                tracing::warn!(?error, block = ?next_block_to_process, "Failed to send processed block height to `blocks_processed` channel");
```
sug0, line 175: can you manually format this line to be shorter? `rustfmt` gets confused with `tracing` macro calls, and does not format them properly. basically just split the arguments into different lines. idem for the string, perhaps.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -127,100 +153,181 @@ pub fn run_oracle(
...
+                            };
+                        }
+                        next_block_to_process += 1u8.into()
+                    },
+                    Err(error) => tracing::warn!(?error, block = ?next_block_to_process, "Error while trying to process Ethereum block"),
```
sug0, line 180: same thing here

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -127,100 +153,181 @@ pub fn run_oracle(
...
+) -> Result<()> {
+    // update the latest block height
+    let latest_block = loop {
+        let latest_block = match oracle.eth_block_number().await {
+            Ok(height) => height,
+            Err(error) => {
+                return Err(eyre!(
+                    "Couldn't get the latest synced Ethereum block height \
+                     from the RPC endpoint: {:?}",
+                    error
```
sug0, line 210:
```suggestion
                     from the RPC endpoint: {error:?}",
```

sug0 (changes requested):

apps/src/lib/node/ledger/ethereum_node/oracle.rs
batconjurer: Is there an obvious use case where a ledger would need this?

batconjurer: If it's just for testing, we can use a oneshot from the mock web3 client like we do for checking the events have been processed.

james-chf: No obvious use case for the ledger - I think we can dispatch the oracle to do it's thing without the ledger having to know which individual block(s) it has processed. Publishing Ethereum blocks processed metrics could be useful but not a priority right now. We need some way to observe which block(s) have been processed for the tests, the `tokio::sync::watch` looks like it would be suitable here although I'm finding it fiddlier (currently trying it out), I'm not sure how to do it with a `oneshot`?

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: Do you mean you think it's impossible for the oracle to send an abort signal? When running the ledger and pressing <kbd>Ctrl+C</kbd> I do only ever see the second message (unable to send abort signal). I think the flow of control for the oracle will have to change soon anyway in that it should be controlled from the ledger rather than as a "top level" task, as it will need access to parameters from blockchain storage like `MIN_CONFIRMATIONS` and will probably need to be epoched in some way (in that it should perhaps only run for certain epochs when a node is an active validator, and in each epoch it might have different parameters and also should check a different range of Ethereum blocks)

sug0 (approved):

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -32,6 +38,10 @@ pub struct Oracle {
     /// A channel to signal that the ledger should shut down
     /// because the Oracle has stopped
     abort: Option<Sender<()>>,
+    /// How long the oracle should wait between checking blocks
+    backoff: Duration,
+    /// If provided, the oracle will put here the latest block it has processed
+    latest_block_processed: Option<watch::Sender<Option<Uint256>>>,
```
batconjurer, line 44: What if we made this `watch::Sender<Uint256>`? Then we don't need the other counter as this always retains the last height.

james-chf: will have a go at this and removing the other counter

james-chf: This is turning out to be pretty fiddly - while trying this approach, I noticed something that makes me think we should go back to having a channel where we publish every block height that is processed. So the other counter in `fn process()` is `next_block_to_process` which can start from `0` when the oracle is created - this current `latest_block_processed` fied starts from `None` then go `Some(0)`, `Some(1)`, `Some(2)`, etc. Replacing the field in this struct with something like:

```rust
next_block_to_process: watch::Sender<Uint256>
```

works and gets rid of the `Option`s, and we can just use this one field to keep track, but then in `test_all_blocks_checked` when we set the height straightaway to be some height more than `MIN_CONFIRMATIONS`, the oracle can start processing blocks immediately and the `next_block_to_process` field can be quickly updated to some block more than 0. This is also an issue with the current `Option<watch::Sender<Option<Uint256>>>` approach as well - if I add a sleep in that test, the oracle races ahead.

james-chf: This is what I tried - https://github.com/anoma/namada/commit/302a0f2322dd46e4f1de1314c101d892adacd972

batconjurer: I'm not sure I understand the issue as you've described it

batconjurer: Ah, I see! The latest block processessed shouldn't start at 0. It should query geth to get the latest block height and set `next_block_to_process` to that. And that should be the last time it queries the block height from geth. So then it just increments the block height one by one.

james-chf: Yeah something like this needs to be done (have opened https://github.com/anoma/namada/issues/589). The issue for the tests is that we want to see every block height that was checked and in order, not just the `latest_block_processed` or `next_block_to_process`, so that we can assert the oracle is checking all blocks in sequence. I'm not sure of the best way to do this but a channel which the oracle sends block heights it has checked works, or we could provide a callback to the oracle for it to notify us when it has processed a block (that we can then use in our tests).

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -127,100 +152,189 @@ pub fn run_oracle(
...
-            match oracle.eth_block_number().await {
-                Ok(height) => break height,
-                Err(error) => {
-                    tracing::warn!(
+        tracing::info!(
+            ?next_block_to_process,
+            "Checking Ethereum block for bridge events"
+        );
+        tokio::select! {
+            result = process(&oracle, &mut pending, next_block_to_process.clone()) => {
```
sug0, line 169: this `tokio::select!` might be unsafe, https://carllerche.com/2021/06/17/six-ways-to-make-async-rust-easier/

sug0 (commented):
check the `select!` comment

sug0:
(sorry about that ref, pasted the wrong link to an issue)

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: I reverted to the original use of a `BoundedSender` on which we send all block heights that we process, so that we can check every one for `test_all_blocks_checked`

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -127,100 +153,183 @@ pub fn run_oracle(
...
+        let confirmed = process_queue(&latest_block, pending);
+        if !confirmed.is_empty() {
+            tracing::info!(
+                ?block_to_process,
+                ?addr,
+                ?sig,
+                pending = pending.len(),
+                confirmed = confirmed.len(),
+                ?MIN_CONFIRMATIONS,
+                "Some events that have reached the minimum number of \
```
batconjurer, line 321:
```suggestion
                "Some events have reached the minimum number of \
```

james-chf:
`test_node_connectivity` has started consistently failing on this PR, need to investigate

james-chf:
nvm the last comment, looks like CI is passing! (but it failed a couple of times I think?)

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -57,11 +77,14 @@ impl Oracle {
...
         sender: BoundedSender<EthereumEvent>,
         abort: Sender<()>,
+        backoff: Duration,
     ) -> Self {
         Self {
             client: Web3::new(url, std::time::Duration::from_secs(30)),
             sender,
             abort: Some(abort),
+            backoff,
+            blocks_processed: None,
```
batconjurer, line 87: Essentially, this is what bothers me. We have a test artifact in our production code. If you don't like the idea of PR #666, then we better make an Oracle trait so that we can make a mock oracle for tests. The only method that would differ is `run_oracle_aux`.

batconjurer (changes requested):

apps/src/lib/node/ledger/ethereum_node/oracle.rs
james-chf: I'd like to do this as well - have opened semi-related https://github.com/anoma/namada/issues/672 to fix up our `web30` dependency so that it should be easier to switch the `Oracle` to using a trait in the future

PR #572, opened 2022-10-10
Implement an event log to persist FinalizeBlock events

sug0:
Based on https://github.com/anoma/namada/pull/564

Completes the second task of issue https://github.com/anoma/namada/issues/563

### DIFF
--- a/apps/src/lib/node/ledger/events.rs
+++ b/apps/src/lib/node/ledger/events.rs
@@ -21,8 +21,8 @@ pub enum EventLevel {
     Tx,
 }
 
-/// Custom events that can be queried from Tendermint
-/// using a websocket client
+/// Custom events that can be queried from Namada
+/// validators or fullnodes.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Event {
     pub event_type: EventType,
--- a/apps/src/lib/node/ledger/events.rs
+++ b/apps/src/lib/node/ledger/events.rs
@@ -13,8 +13,8 @@ use thiserror::Error;
 
 use crate::facade::tendermint_proto::abci::EventAttribute;
 
-/// Indicates if an event is emitted do to
-/// an individual Tx or the nature of a finalized block
+/// Indicates if an event is emitted from an individual tx or from
+/// a `FinalizeBlock` call in the ledger.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum EventLevel {
     Block,
@@ -30,7 +30,7 @@ pub struct Event {
     pub attributes: HashMap<String, String>,
 }
 
-/// The two types of custom events we currently use
+/// The types of custom events we currently use.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum EventType {
     // The transaction was accepted to be included in a block
--- a/apps/src/lib/node/ledger/events/log.rs
+++ b/apps/src/lib/node/ledger/events/log.rs
@@ -99,7 +99,6 @@ pub struct EventLog {
 #[allow(dead_code)]
 struct EventLogSnapshot {
     oldest_height: BlockHeight,
-    num_entries: usize,
     num_events: usize,
     head: Arc<LogNode>,
 }
@@ -116,8 +115,6 @@ struct EventLogInner {
 /// Data which needs lock protection, in the [`EventLog`].
 #[derive(Debug)]
 struct EventLogInnerMux {
-    /// The total number of entries in the log.
-    num_entries: usize,
     /// The total number of events stored in the log.
     ///
     /// This value is the sum of every batch of events in
@@ -240,7 +237,6 @@ impl EventLog {
                 notifier: event_listener::Event::new(),
                 lock: RwLock::new(EventLogInnerMux {
                     num_events: 0,
-                    num_entries: 0,
                     oldest_height: 0.into(),
                     head: None,
                 }),
@@ -249,15 +245,10 @@ impl EventLog {
     }
 
     /// Prune the event log, ejecting old [`Event`] instances.
-    fn prune(
-        &self,
-        head: Option<Arc<LogNode>>,
-        num_entries: usize,
-        num_events: usize,
-    ) {
+    fn prune(&self, head: Option<Arc<LogNode>>, num_events: usize) {
         let _ = MAX_LOG_EVENTS;
         let _ = LOG_BLOCK_HEIGHT_DIFF;
-        let _ = (head, num_entries, num_events);
+        let _ = (head, num_events);
         // TODO
     }
 
@@ -269,16 +260,15 @@ impl EventLog {
         }
 
         // update the log head
-        let (head, entries, events) = {
+        let (head, events) = {
             let mut log = self.inner.lock.write().unwrap();
-            log.num_entries += 1;
             log.num_events += entry.events.len();
             log.head = Some(Arc::new(LogNode {
                 entry,
                 next: log.head.take(),
             }));
             let new_head = log.head.clone();
-            (new_head, log.num_entries, log.num_events)
+            (new_head, log.num_events)
         };
 
         // notify all event listeners
@@ -286,7 +276,7 @@ impl EventLog {
 
         // we don't need to hold a lock to check
         // if the log needs to be pruned
-        self.prune(head, entries, events);
+        self.prune(head, events);
     }
 
     /// Snapshot the current state of the event log, and return it.
@@ -295,7 +285,6 @@ impl EventLog {
         log.head.clone().map(|head| EventLogSnapshot {
             head,
             num_events: log.num_events,
-            num_entries: log.num_entries,
             oldest_height: log.oldest_height,
         })
     }
@@ -307,7 +296,6 @@ impl EventLogInnerMux {
     #[allow(dead_code)]
     fn install_snapshot(&mut self, snapshot: EventLogSnapshot) {
         self.oldest_height = snapshot.oldest_height;
-        self.num_entries = snapshot.num_entries;
         self.num_events = snapshot.num_events;
         self.head = Some(snapshot.head);
     }
--- a/apps/src/lib/node/ledger/events/log.rs
+++ b/apps/src/lib/node/ledger/events/log.rs
@@ -327,7 +327,7 @@ impl EventLog {
         }
     }
 
-    /// Prune events from the log, keeping as many as `max_events`
+    /// Prune events from the log, keeping at most `max_events + excess`
     /// of the most recent events.
     fn prune_too_many_events(
         &self,
@@ -378,7 +378,7 @@ impl EventLog {
         let mut oldest_height = 0.into();
 
         let head = LogNode::iter(head.as_ref())
-            // filter out excess events in the log
+            // filter out events in the log
             .take_while(|n| {
                 total_events += n.entry.events.len();
                 match predicate(n, total_events) {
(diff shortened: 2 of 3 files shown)


sug0:
diff this against `tiago/ethbridge/impl-events-endpoint-part-1` for review

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -0,0 +1,893 @@
...
+pub struct LogEntry {
+    /// The block height at which we emitted the events.
+    pub block_height: BlockHeight,
+    /// The events emitted by a `FinalizeBlock` call.
+    pub events: Vec<Event>,
+}
+
+/// Represents a node in the linked list of log entries.
+#[derive(Debug)]
+struct LogNode {
```
batconjurer, line 93: Linked lists are terrible data structures :sweat_smile: I think we can just use a vec here and then all the iter definitions simplifiy. The only question is the snapshot, which I haven't figured the purpose of yet.

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -0,0 +1,893 @@
...
+            let snapshot = if nodes_to_keep > 0 {
+                self.prune_old_events(head, nodes_to_keep)
+            } else {
+                None
+            };
+            self.inner.lock.write().unwrap().install_snapshot(snapshot);
+        }
+    }
+
+    /// Prune events from the log, keeping at most `max_events + excess`
```
batconjurer, line 340: what is `excess`?

apps/src/lib/node/ledger/events/log.rs
batconjurer: Or better yet some sort of automatically sorted data structure, like a btreeset.

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -0,0 +1,893 @@
...
+}
+
+/// Represents a log of [`Event`] instances emitted by
+/// `FinalizeBlock` calls, in the ledger.
+///
+/// __INVARIANT:__ All logged events should be ordered
+/// in reverse by the shell's block height at the time
+/// of calling `FinalizeBlock`.
+#[derive(Debug, Clone)]
+pub struct EventLog {
```
batconjurer, line 131: It might tighten up the code to implement `Deref` and `DerefMut` for this type

batconjurer: It essentially is just a pointer.

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -0,0 +1,893 @@
...
+
+impl EventLogInnerMux {
+    /// Modifies the state of the [`EventLogInnerMux`] with the provided
+    /// [`EventLogSnapshot`].
+    fn install_snapshot(&mut self, snapshot: Option<EventLogSnapshot>) {
+        if let Some(snapshot) = snapshot {
+            self.oldest_height = snapshot.oldest_height;
+            self.num_events = snapshot.num_events;
+            self.head = Some(snapshot.head);
+        } else {
```
batconjurer, line 480: If you derive `Default` on this type, the `else` can shorten to `*self = Default::default;`

batconjurer (changes requested):

apps/src/lib/node/ledger/events.rs
```diff
@@ -13,23 +15,23 @@ use crate::facade::tendermint_proto::abci::EventAttribute;
...
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum EventLevel {
     Block,
     Tx,
 }
 
 /// Custom events that can be queried from Tendermint
 /// using a websocket client
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
james-chf, line 26: we can update the docstring (`from Tendermint using a websocket client` no more 😄 )

apps/src/lib/node/ledger/events.rs
```diff
@@ -13,23 +15,23 @@ use crate::facade::tendermint_proto::abci::EventAttribute;
...
 pub struct Event {
     pub event_type: EventType,
     pub level: EventLevel,
     pub attributes: HashMap<String, String>,
 }
 
 /// The two types of custom events we currently use
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
james-chf, line 34: Sorry this isn't due to your PR but we can update the docstring here as well `/// The types of custom events we currently use`

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -286,19 +287,26 @@ impl ShellMode {
...
     /// These events can then be queried from this node's `/events` RPC
     /// endpoint, until they are ejected fron the event log some time later.
+    ///
     /// Ejecting events from a node's log happens after they have expired,
     /// which is configured on a per-node basis based on a specific number
-    /// of block heights parameter.
-    pub fn log_events(&self, events: Vec<Event>) {
+    /// of block heights parameter. This is the reason for including a
+    /// `block_height` parameter in this call.
+    pub fn log_events(&self, block_height: BlockHeight, events: Vec<Event>) {
```
james-chf, line 295: What type of places will this function be called from? I'm guessing end of `FinalizeBlock`? It would be cleaner to return a `Result`

james-chf: (I just saw the docstring for `EventLog` mentions `FinalizeBlock`)

apps/src/lib/node/ledger/events/log.rs
sug0: ah nevermind this. forgot to clean this docstring, after I changed a part of the design.

apps/src/lib/node/ledger/events/log.rs
sug0: closing this after our discussion on a call

apps/src/lib/node/ledger/events/log.rs
sug0: unfortunately I can't implement `Deref`, because `EventLogInner` is a private type

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -0,0 +1,156 @@
...
+impl<'q> QueryMatcher<'q> {
+    /// Checks if this [`QueryMatcher`] validates the
+    /// given [`Event`].
+    pub fn matches(&self, event: &Event) -> bool {
+        event.event_type == self.event_type
+            && event
+                .attributes
+                .get(&self.attr)
+                .as_ref()
+                .map(|value| value == &self.value)
```
james-chf, line 41:
```suggestion
                .get(&self.attr)
                .map(|value| value == self.value)
```

apps/src/lib/node/ledger/shell/mod.rs
sug0: yeah from `FinalizeBlock`. idk, `Shell::broadcast` does the same. besides, we wouldn't be able to recover from the `Result` anyway...

james-chf (commented):
Will try out https://github.com/anoma/namada/pull/587

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Parses a Tendermint-like events query.
+    pub fn parse(query: &'q str) -> Option<Self> {
+        let captures = QUERY_PARSING_REGEX.captures(query)?;
+
+        let event_type = match captures.get(1)?.as_str() {
+            "accepted" => EventType::Accepted,
+            "applied" => EventType::Applied,
+            // NOTE: the regex only matches `accepted`
+            // and `applied`
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
batconjurer: The `?` on the regex does that.

apps/src/lib/node/ledger/shell/mod.rs
batconjurer: Yes, if this fails, the node should shut down because of catastophic failure.

apps/src/lib/node/ledger/events/log/dumb_queries.rs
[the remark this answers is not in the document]
batconjurer: Oh, I guess it's an option, not a result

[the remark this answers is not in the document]
batconjurer: But, the unreachable!() is still necessary I think.

batconjurer (approved):

apps/src/lib/node/ledger/events/log/dumb_queries.rs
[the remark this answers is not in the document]
james-chf: Oh yeah you're right, the capture can only get `"accepted"` or `"applied"` otherwise we won't reach this point, so the `unreachable!()` makes sense. We should return a `Result` still indicating this is what happened (where we are currently returning `None`)

apps/src/lib/node/ledger/shell/mod.rs
james-chf: If the issue here to not return the `Result` is that we should terminate the program instead, we should do this in a more ordered way than panicking. We should propagate an error to the caller which may then be responsible for shutting down, or passing the error along further. 

This fn shouldn't be responsible for either forwarding events or potentially shutting down the ledger - we should centralize this responsibility somewhere. Right now though it's spread out across multiple places and it would be clearer if we start returning errors where they can occur, even ones that should cause the process to terminate, if this makes sense.

sug0: for that to be possible, all `Shell` ABCI methods would need to return a `Result`. or at least the ones we knew for sure would have failing ops, like `FinalizeBlock`, to send events to the event log.

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -0,0 +1,893 @@
...
+        {
+            ::tokio::task::block_in_place(|| $expr)
+        }
+    }};
+}
+
+/// Parameters to configure the pruning of the event log.
+#[derive(Debug, Copy, Clone)]
+pub struct Params {
+    /// Soft lock on the maximum number of events the event log can hold.
```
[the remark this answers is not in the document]
sug0: yeah

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -0,0 +1,156 @@
...
+            && event
+                .attributes
+                .get(&self.attr)
+                .as_ref()
+                .map(|value| value == &self.value)
+                .unwrap_or_default()
+    }
+
+    /// Parses a Tendermint-like events query.
+    pub fn parse(query: &'q str) -> Option<Self> {
```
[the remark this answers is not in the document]
sug0: this only gets called from the event log, which has an error type reflecting this failure.

apps/src/lib/node/ledger/events/log/dumb_queries.rs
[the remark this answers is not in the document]
sug0: we're now logging invalid queries

sug0:
Closing this PR, as the design of the RPC event log will change substantially. This is due to us no longer implementing an HTTP server in Namada to reply to RPC requests, but rather relying on Tendermint's ABCI queries method to reply with tx events.

This implementation still lives [here](https://github.com/heliaxdev/eventlog/tree/namada), for future reference.

PR #573, opened 2022-10-10
Bridge pool merkle tree

### LINKED ISSUE #420: Add the Ethereum bridge pool
The Ethereum bridge pool is a mempool-like store on Namada which holds transfers bound for Ethereum. User make batches of transactions from this pool along with merkle proofs and relay this over to our Ethereum smart contracts. 

 - [x] #421 
 - [x] #422
 - [x] #423
 - [x] #424
 - [x] #425
 - [x] #426

batconjurer:
This PR partially closes #420. It adds a merkle tree that will be used for the bridge pool. It also refactors the merkle tree storage to better accomodate multiple types of merkle trees in a uniform way (based off of PR #547 ).

This PR will depend on merging in #547 and #549 once they are approved for `main`.

This PR does not actually change the vp or wasm code, so the bridge pool is still help under the accounts sub-storage for now. This will be addressed in a subsequent PR.


batconjurer:
pls update wasm

batconjurer:
pls update wasm

batconjurer:
pls update wasm

shared/src/types/hash.rs
```diff
@@ -86,6 +88,26 @@ impl TryFrom<&[u8]> for Hash {
     }
 }
 
+impl TryFrom<String> for Hash {
+    type Error = self::Error;
+
+    fn try_from(string: String) -> HashResult<Self> {
+        let bytes: Vec<u8> =
+            Vec::from_hex(string).map_err(Error::FromStringError)?;
+        Self::try_from(bytes.as_slice())
```
sug0, line 97:
```suggestion
        string.as_str().try_into()
```

shared/src/types/hash.rs
```diff
@@ -86,6 +88,26 @@ impl TryFrom<&[u8]> for Hash {
...
+    }
+}
+
+impl TryFrom<&str> for Hash {
+    type Error = self::Error;
+
+    fn try_from(string: &str) -> HashResult<Self> {
+        let bytes: Vec<u8> =
+            Vec::from_hex(string).map_err(Error::FromStringError)?;
+        Self::try_from(bytes.as_slice())
```
sug0, line 107: you could add the length check at this level, to avoid allocating a buffer with a potentially huge capacity. depending on where this is being used, it can be a DoS vector, crashing the ledger's process with the OOM killer or smth

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+
+    use itertools::Itertools;
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::types::eth_bridge_pool::{GasFee, TransferToEthereum};
+    use crate::types::ethereum_events::EthAddress;
+
+    /// An established user address for testing & development
+    fn bertha_address() -> Address {
```
sug0, line 368: haha it's always bertha

shared/src/types/hash.rs
sug0: albeit to be honest, the source string is going to be twice as big

batconjurer: I need to change this to `<[u8; 32]>::from_hex(..)`

batconjurer: That's what I mentioned early that you reminded me of.

james-chf (commented):
Have just started looking at this

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
batconjurer: Damn straight!

sug0 (approved):

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+    /// Checks if the hash is in the tree.
+    pub fn contains_key(&self, key: &Key) -> Result<bool, Error> {
+        Ok(self.store.contains(&Self::parse_key(key)?))
+    }
+
+    /// Update the tree with a new value.
+    ///
+    /// Returns the new root if successful. Will
+    /// return an error if the key is malformed.
+    pub fn update_key(&mut self, key: &Key) -> Result<Hash, Error> {
```
james-chf, line 91: nit: would call this `insert_key` rather than `update_key`

shared/src/types/eth_bridge_pool.rs
```diff
@@ -49,6 +55,27 @@ pub struct PendingTransfer {
     pub gas_fee: GasFee,
 }
 
+impl keccak::encode::Encode for PendingTransfer {
+    fn tokenize(&self) -> Vec<Token> {
+        let from = Token::String(self.gas_fee.payer.to_string());
+        let fee = Token::Uint(u64::from(self.gas_fee.amount).into());
+        let to = Token::Address(self.transfer.recipient.0.into());
+        let amount = Token::Uint(u64::from(self.transfer.amount).into());
+        let nonce = Token::Uint(self.transfer.nonce.clone().into());
```
james-chf, line 64: How come we don't include `self.transfer.asset` as part of the encoding (which eventually feeds into the hash)?

james-chf: The gas fee payer and amount feeds into the hash of `PendingTransfer`, so it looks like the same `TransferFromEthereum` could be added to the pool multiple times but with different gas fees? This is the case with the current `HashSet` implementation as well I think (which uses the default `Hash` derivation).

james-chf (commented):
Have started looking at `bridge_pool.rs` mainly

shared/src/types/storage.rs
```diff
@@ -233,14 +237,33 @@ impl FromStr for Key {
...
+///
+/// This is a multi-store organized as
+/// several Merkle trees, each of which is
+/// responsible for understanding how to parse
+/// this value.
+pub enum MerkleValue {
+    /// raw bytes
+    Bytes(Vec<u8>),
+    /// A transfer to be put in the Ethereum bridge pool.
+    Transfer(PendingTransfer),
```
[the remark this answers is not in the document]
batconjurer: :+1:

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+    fn parse_key(key: &Key) -> Result<KeccakHash, Error> {
+        if key.segments.len() == 1 {
+            match &key.segments[0] {
+                DbKeySeg::StringSeg(str) => {
+                    str.as_str().try_into().map_err(|_| {
+                        eyre!("Could not parse key segment as a hash").into()
+                    })
+                }
+                _ => Err(eyre!(
+                    "Bridge pool keys should be strings, not addresses"
```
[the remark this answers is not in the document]
batconjurer: :+1:

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+                )
+                .into()),
+            }
+        } else {
+            Err(eyre!(
+                "Key for the bridge pool should not have more than one segment"
```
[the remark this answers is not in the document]
batconjurer: These hashes are namespaced, but to one of our merkle trees in the multistore. Those prefixes are stripped when interacting with them throughout the code base. So keys are of the form `bridge_pool/hash`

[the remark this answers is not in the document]
batconjurer: https://github.com/anoma/namada/blob/038360dac981920ef804c1bca635bb5b011bf0de/shared/src/ledger/storage/merkle_tree.rs#L168

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+
+            for left in left_leaves {
+                let right = right_leaves.next().cloned().unwrap_or_default();
+                next_hashes.push(hash_pair(left.clone(), right));
+            }
+            hashes = next_hashes;
+        }
+
+        if hashes.is_empty() {
+            Default::default()
```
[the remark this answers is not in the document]
batconjurer: We have to do this. The root of an empty tree is by definition `[0u8; 32]`. We can't have an option, otherwise we can't compute the root of our total merkle store in a way that is committing to this subtree and we lose security properties.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+        Ok(self.root())
+    }
+
+    /// Delete a key from storage and update the root
+    pub fn delete_key(&mut self, key: &Key) -> Result<(), Error> {
+        let hash = Self::parse_key(key)?;
+        _ = self.store.remove(&hash);
+        self.root = self.compute_root();
+        Ok(())
+    }
```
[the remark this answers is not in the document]
batconjurer: We could do this logging here, or in the VP, or when we apply the txs in the ledger. What do you think is the best place?

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
 }
+
+/// A simple Merkle tree for the Ethereum bridge pool
+#[derive(
+    Debug, Default, Clone, BorshSerialize, BorshDeserialize, BorshSchema,
+)]
+pub struct BridgePoolTree {
+    /// Root of the tree
+    root: KeccakHash,
+    /// The underlying storage
```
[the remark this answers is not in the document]
batconjurer: We actually don't put the dummy leaves into this set, it really does only contain hashes of `PendingTransfer`s. We use an `unwrap_or_default` to get dummy leaves.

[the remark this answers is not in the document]
batconjurer: But I will update the doc string to say they are hashes of `PendingTransfer`s

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
batconjurer: We can overwrite an existing key in theory, but that would imply a hash collision so that would never happen in practice. I can change it though.

shared/src/types/eth_bridge_pool.rs
batconjurer: Good question! I should probably add that.
And yes, the same transfer can be submitted multiple times with different gas fees. I'm not sure exactly why someone would do that, but there is no reason to stop them from doing it.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,847 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+    }
+
+    /// Compute the root of the merkle tree
+    pub fn compute_root(&self) -> KeccakHash {
+        let mut hashes: Vec<KeccakHash> = self.store.iter().cloned().collect();
+        while hashes.len() > 1 {
+            let mut next_hashes = vec![];
+            let left_leaves = hashes.iter().step_by(2);
+            let mut right_leaves = hashes.iter();
+            _ = right_leaves.next();
```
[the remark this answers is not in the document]
batconjurer: I actually intended to refactor this to use chunks. Hopefully that will make this clearer. The root node is not stored in this set. Only the leaves are. Since its a binary tree, to compute the root we start hashing from the leafs upwards in pairs.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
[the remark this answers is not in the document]
james-chf: If we want to log whether the key was newly inserted/deleted, that information is only in this function so we'd have to log here, or alternatively we could change these fns to return `Result<bool>` where the returned bool is whether the key was newly inserted/deleted (though then the callers of `update_key` would have to fetch the root hash themselves after calling `update_key`). 

Re. where to log, logging it in the native VP would be more practical I think as logging inside wasms is less functional it seems. I guess the main thing is we should be able to see when running a node what is happening with the bridge pool, to help us with development and debugging. I'd be tempted even to log at INFO so that real validators/nodes should always have this logged as well, in case we need to debug something later on in a testnet, they could give us the logs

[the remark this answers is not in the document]
batconjurer: If we log at this point, it could be confusing. Because this code is called when changing the write ahead log, which might then gets rejected by the VP and so this update/delete doesn't actually happen. I think I'll log at the VP level.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
james-chf: Yeah if it should never happen, logging here at `tracing::error` level would also document nicely that that should never be the case (and if it does happen somehow, we can see what happened)

batconjurer:
pls update wasm

shared/src/types/eth_bridge_pool.rs
james-chf: It feels like the `PendingTransfer`s that go through the pool should be unique by nonce, since on the Ethereum smart contract side, we shouldn't expect to have multiple transfers redeemed with the same nonce? I need to familiarize myself still with the nonce/replay protection so not sure if this is correct. Maybe we could allow multiple `PendingTransfer`s with the same nonce (in the `TransferFromEthereum` struct), but only the first one to be redeemed on the smart contract works, later ones fail. In which case we may as well prevent transfers with the same nonce going through the pool somehow, but I'm not sure

re specs: https://github.com/anoma/namada/blob/e88eeac2f7eae59466fce12b22f02d50415fd4f6/documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md#L126-L133

batconjurer:
pls update wasm

shared/src/types/eth_bridge_pool.rs
batconjurer: The first one that uses a nonce in an Ethereum smart contract will pass and the rest fail. It is the smart contracts job to enforce this. One possible scenario is that someone under pays gas fees, so submits with higher fees to get it relayed. Then the cheaper one will time out.

batconjurer: But in theory, if you don't pay enough gas, you can submit the same transfer with more gas to get it relayed. Then the older one will time out. And this way we don't have to do complicated replacement / remove from pool user logic.

batconjurer:
pls update wasm

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,837 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+        let mut flags = vec![];
+        let mut hashes: Vec<_> = self
+            .store
+            .iter()
+            .cloned()
+            .map(|hash| {
+                if leaves.contains(&hash) {
+                    Node::OnPath(hash)
+                } else {
+                    Node::Sibling(hash)
```
james-chf, line 174: Re. the `Node::Sibling` docstring, we could just call this `Node::NotOnPath` or something. Here some of the leaves we're wrapping could be not siblings of any of the `leaves` we want to make a proof for?

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,837 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+}
+
+/// A multi-leaf membership proof
+pub struct BridgePoolProof {
+    /// The hashes other than the provided leaves
+    pub proof: Vec<KeccakHash>,
+    /// The leaves; must be sorted
+    pub leaves: Vec<PendingTransfer>,
+    /// flags to indicate how to combine hashes
+    pub flags: Vec<bool>,
```
james-chf, line 283: Could we document what the elements of `flags` are meant to mean, if they are `true` or `false`? Could even use an enum something like

```rust
enum Flag {
	BothNodesOnPath,
	OneNodeOnPath,
}
```

I'm guessing from looking at how we set `flags` in `get_membership_proof`

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,837 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+                        proof_hashes.push(sib.clone());
+                        next_hashes
+                            .push(Node::OnPath(hash_pair(hash.clone(), sib)));
+                    }
+                    (Node::Sibling(sib), Node::OnPath(hash)) => {
+                        flags.push(false);
+                        proof_hashes.push(sib.clone());
+                        next_hashes
+                            .push(Node::OnPath(hash_pair(hash, sib.clone())));
+                    }
```
james-chf, line 202: We could combine these two match arms, something like:

```rust
                    (Node::OnPath(hash), Node::Sibling(sib))
                    | (Node::Sibling(sib), Node::OnPath(hash)) => {
                        flags.push(false);
                        proof_hashes.push(sib.clone());
                        next_hashes.push(Node::OnPath(hash_pair(
                            hash.to_owned(),
                            sib.to_owned(),
                        )));
                    }
```

shared/src/ledger/storage/traits.rs
```diff
@@ -159,6 +165,51 @@ impl<'a, H: StorageHasher + Default> SubTreeWrite for &'a mut Amt<H> {
...
+        key: &Key,
+        value: MerkleValue,
+    ) -> Result<Hash, Error> {
+        if let MerkleValue::BridgePoolTransfer(_) = value {
+            self.insert_key(key)
+                .map_err(|err| Error::MerkleTree(err.to_string()))
+        } else {
+            Err(Error::InvalidValue)
+        }
+    }
```
james-chf, line 204: We should revisit these traits in some other PR, `key` is deriveable from `value` so we shouldn't need to pass in both, at least for `BridgePoolTree`. I'm guessing for other impls this isn't the case and the key and value aren't related.

james-chf (changes requested):

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,837 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+
+    /// Get a reference to the backing store
+    pub fn store(&self) -> &BTreeSet<KeccakHash> {
+        &self.store
+    }
+
+    /// Create a batched membership proof for the provided keys
+    pub fn get_membership_proof(
+        &self,
+        keys: &[Key],
```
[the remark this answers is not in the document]
batconjurer: Yeah, actually we don't need it.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
batconjurer: Sure.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
batconjurer: These have to end up being bool for solidity, and I wanted to make sure that the test code in rust was as close a translation to the solidity code as possible. But I can add more doc strings about it.

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
batconjurer: I tried that at first, but rust wouldn' have it. :weary:

shared/src/ledger/storage/traits.rs
batconjurer: I've changed this method to ignore the keys.

batconjurer:
pls update wasm

shared/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -44,3 +58,799 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
...
+    }
+}
+
+/// A multi-leaf membership proof
+pub struct BridgePoolProof {
+    /// The hashes other than the provided leaves
+    pub proof: Vec<KeccakHash>,
+    /// The leaves; must be sorted
+    pub leaves: Vec<PendingTransfer>,
+    /// Flags to indicate how to combine hashes.
```
james-chf, line 271:
```suggestion
```
duplicate docstring

james-chf (approved):
lgtm!

Issue #562, opened 2022-10-07
Make a governance parameter for the Ethereum events minimum confirmations


james-chf:
Currently this is a hardcoded constant (100)

https://github.com/anoma/namada/blob/85d554217bc5951f1ac77af42a963733ca88086a/apps/src/lib/node/ledger/ethereum_node/oracle.rs#L16-L17

james-chf:
Will start being used in https://github.com/anoma/namada/issues/686

PR #575, opened 2022-10-10
Add chain parameters for the Ethereum bridge

james-chf:
Relates to https://github.com/anoma/namada/issues/365 and https://github.com/anoma/namada/issues/562

This PR adds *preliminary* chain parameters for the Ethereum bridge, and also renames ledger runtime configuration to be under `[ledger.ethereum_bridge]` rather than just `[ledger.ethereum]`. We can change/add extra validation for these configuration parameters in other PRs.

In a network config file, Ethereum bridge parameters will be able to be specified like so:

```toml
[ethereum_bridge_params]
min_confirmations = 100

[ethereum_bridge_params.contracts]
native_erc20 = "0x1721b337BBdd2b11f9Eef736d9192a8E9Cba5872"

[ethereum_bridge_params.contracts.bridge]
address = "0x237d915037A1ba79365E84e2b8574301B6D25Ea0"
version = 1

[ethereum_bridge_params.contracts.governance]
address = "0x308728EEa73538d0edEfd95EF148Eb678F71c71D"
version = 1
```

These parameters as a whole are optional but if any are set, then all must be set. i.e. a Namada chain can either start without any Ethereum bridge parameters set, or with all Ethereum bridge parameters set.

This PR only adds the parameter fields, but they aren't used yet by the oracle. When validator nodes start up, they will need to check blockchain storage for the appropriate parameter values for whatever epoch they have synced to, and configure their events oracle accordingly. If no Ethereum bridge parameters have been set for a chain yet, then validator nodes don't need to start any Ethereum bridge componentry (i.e. the events oracle or `geth`).

Specs need to be updated for no more proxy contract (have added an item to https://github.com/anoma/namada/issues/448)

I've tested this manually and have been able to `namadac utils init-network` etc. with this network config file which has Ethereum bridge parameters: [network-config.toml.txt](https://github.com/anoma/namada/files/9832307/network-config.toml.txt)

### DIFF
--- a/apps/src/lib/config/ethereum_bridge/ledger.rs
+++ b/apps/src/lib/config/ethereum_bridge/ledger.rs
@@ -1,4 +1,4 @@
-//! Configuration settings to do with the Ethereum bridge.
+//! Runtime configuration for a validator node
 #[allow(unused_imports)]
 use namada::types::ethereum_events::EthereumEvent;
 use serde::{Deserialize, Serialize};
@@ -28,7 +28,8 @@ pub enum Mode {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Config {
-    /// The mode in which to run the Ethereum bridge
+    /// The mode in which to run the Ethereum node and oracle setup of this
+    /// validator
     pub mode: Mode,
     /// The Ethereum JSON-RPC endpoint that the Ethereum event oracle will use
     /// to listen for events from the Ethereum bridge smart contracts
--- a/apps/src/lib/config/ethereum_bridge/ledger.rs
+++ b/apps/src/lib/config/ethereum_bridge/ledger.rs
@@ -29,7 +29,7 @@ pub enum Mode {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Config {
     /// The mode in which to run the Ethereum node and oracle setup of this
-    /// validator
+    /// validator.
     pub mode: Mode,
     /// The Ethereum JSON-RPC endpoint that the Ethereum event oracle will use
     /// to listen for events from the Ethereum bridge smart contracts
--- a/apps/src/lib/config/ethereum_bridge/ledger.rs
+++ b/apps/src/lib/config/ethereum_bridge/ledger.rs
@@ -1,4 +1,4 @@
-//! Runtime configuration for a validator node
+//! Runtime configuration for a validator node.
 #[allow(unused_imports)]
 use namada::types::ethereum_events::EthereumEvent;
 use serde::{Deserialize, Serialize};
--- a/apps/src/lib/config/ethereum_bridge/params.rs
+++ b/apps/src/lib/config/ethereum_bridge/params.rs
@@ -1,6 +1,14 @@
 //! Blockchain-level parameters for the configuration of the Ethereum bridge.
 use serde::{Deserialize, Serialize};
 
+/// Represents a configuration value for an Ethereum address e.g.
+/// 0x6B175474E89094C44Da98b954EedeAC495271d0F
+#[derive(Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct Address(String);
+
+/// Represents a configuration value for the minimum number of
+/// confirmations an Ethereum event must reach before it can be acted on.
 #[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
 #[repr(transparent)]
 pub struct MinimumConfirmations(u64);
@@ -11,29 +19,32 @@ impl Default for MinimumConfirmations {
     }
 }
 
+/// Represents chain parameters for the Ethereum bridge.
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
 pub struct Config {
     /// Minimum number of confirmations needed to trust an Ethereum branch.
     /// This must be at least one.
     pub min_confirmations: MinimumConfirmations,
     /// The addresses of the Ethereum contracts that need to be directly known
-    /// by validators
-    pub contract_addresses: Addresses,
+    /// by validators.
+    pub contracts: Contracts,
 }
 
+/// Represents all the Ethereum contracts that need to be directly know about by
+/// validators.
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
-pub struct Addresses {
+pub struct Contracts {
     /// The Ethereum address of the ERC20 contract that represents this chain's
-    /// native token e.g. 0x6B175474E89094C44Da98b954EedeAC495271d0F
-    pub native_erc20: String,
-    /// The Ethereum address of the bridge contract e.g.
-    /// 0x6B175474E89094C44Da98b954EedeAC495271d0F
-    pub bridge: EthereumContract,
-    /// The Ethereum address of the governance contract e.g.
-    /// 0x6B175474E89094C44Da98b954EedeAC495271d0F
-    pub governance: EthereumContract,
+    /// native token.
+    pub native_erc20: Address,
+    /// The Ethereum address of the bridge contract.
+    pub bridge: UpgradeableContract,
+    /// The Ethereum address of the governance contract.
+    pub governance: UpgradeableContract,
 }
 
+/// Represents a configuration value for the version of a contract that can be
+/// upgraded. Starts from 1.
 #[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
 #[repr(transparent)]
 pub struct ContractVersion(u64);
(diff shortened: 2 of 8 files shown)


apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -0,0 +1,106 @@
...
+/// 0x6B175474E89094C44Da98b954EedeAC495271d0F
+#[derive(Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct Address(String);
+
+/// Represents a configuration value for the minimum number of
+/// confirmations an Ethereum event must reach before it can be acted on.
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct MinimumConfirmations(u64);
```
sug0, line 14:
```suggestion
pub struct MinimumConfirmations(NonZeroU64);
```
maybe overkill, but this prevents you from configuring a value of 0 confirmations

apps/src/lib/config/ethereum_bridge/ledger.rs
```diff
@@ -1,4 +1,4 @@
-//! Configuration settings to do with the Ethereum bridge.
+//! Runtime configuration for a validator node
```
sug0, line 1:
```suggestion
//! Runtime configuration for a validator node.
```

apps/src/lib/config/ethereum_bridge/ledger.rs
```diff
@@ -28,7 +28,8 @@ pub enum Mode {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Config {
-    /// The mode in which to run the Ethereum bridge
+    /// The mode in which to run the Ethereum node and oracle setup of this
+    /// validator
```
sug0, line 32:
```suggestion
    /// validator.
```

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -0,0 +1,106 @@
+//! Blockchain-level parameters for the configuration of the Ethereum bridge.
+use serde::{Deserialize, Serialize};
+
+/// Represents a configuration value for an Ethereum address e.g.
+/// 0x6B175474E89094C44Da98b954EedeAC495271d0F
```
sug0, line 5:
```suggestion
/// Represents a configuration value for an Ethereum address.
///
/// For instance:
/// `0x6B175474E89094C44Da98b954EedeAC495271d0F`
```

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub bridge: UpgradeableContract,
+    /// The Ethereum address of the governance contract.
+    pub governance: UpgradeableContract,
+}
+
+/// Represents a configuration value for the version of a contract that can be
+/// upgraded. Starts from 1.
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct ContractVersion(u64);
```
sug0, line 50:
```suggestion
pub struct ContractVersion(NonZeroU64);
```
same thing here, since you want the version to start counting from 1

tests/src/e2e/setup.rs
```diff
@@ -667,7 +667,7 @@ where
         .env("ANOMA_LOG", "info")
         .env("ANOMA_TM_STDOUT", "true")
         .env("TM_LOG_LEVEL", "info")
-        .env("ANOMA_LEDGER__ETHEREUM__MODE", "Off")
+        .env("ANOMA_LEDGER__ETHEREUM_BRIDGE__MODE", "Off")
```
sug0, line 670: eventually this should be removed, it's a bit hacky to include it here

sug0 (changes requested):

apps/src/lib/config/ethereum_bridge/params.rs
james-chf: Definitely not overkill!! I had no idea there was this type in the std library, this works perfectly here

tests/src/e2e/setup.rs
james-chf: agreed, have opened https://github.com/anoma/namada/issues/665

We will need to revisit this soon probably as for some e2e tests, we will want to enable some components of the Ethereum bridge e..g `ANOMA_LEDGER__ETHEREUM_BRIDGE__MODE='EventsEndpoint'`

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -47,11 +49,11 @@ pub struct Contracts {
 /// upgraded. Starts from 1.
 #[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
 #[repr(transparent)]
-pub struct ContractVersion(u64);
+pub struct ContractVersion(NonZeroU64);
 
 impl Default for ContractVersion {
     fn default() -> Self {
-        Self(1)
+        Self(unsafe { NonZeroU64::new_unchecked(1) })
```
sug0, line 58: maybe add a `SAFETY` comment here. it might be somewhat obvious why it's needed, but I think it's good practice to accompany `unsafe{}` blocks with `SAFETY` comments. for instance:
```rs
        // SAFETY: The only way the API contract of `NonZeroU64`can be violated is if we construct values of this type using 0 as argument.
        Self(unsafe { NonZeroU64::new_unchecked(1) })
```

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -11,11 +13,11 @@ pub struct Address(String);
 /// confirmations an Ethereum event must reach before it can be acted on.
 #[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
 #[repr(transparent)]
-pub struct MinimumConfirmations(u64);
+pub struct MinimumConfirmations(NonZeroU64);
 
 impl Default for MinimumConfirmations {
     fn default() -> Self {
-        Self(100)
+        Self(unsafe { NonZeroU64::new_unchecked(100) })
```
sug0, line 22: same here

tests/src/e2e/setup.rs
sug0: a pr I have open allows this to be configured on demand https://github.com/anoma/namada/pull/646

apps/src/lib/config/ethereum_bridge/params.rs
james-chf: agreed, have added the comment

sug0 (approved):

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn default() -> Self {
+        // SAFETY: The only way the API contract of `NonZeroU64` can be violated
+        // is if we construct values of this type using 0 as argument.
+        Self(unsafe { NonZeroU64::new_unchecked(100) })
+    }
+}
+
+/// Represents chain parameters for the Ethereum bridge.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct Config {
```
batconjurer, line 30:
```suggestion
pub struct GenesisConfig {
```

batconjurer: Confused me at first because I though this was something that could be configured when starting a node. Reading the later code, I realized that wasn't true.

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -0,0 +1,115 @@
...
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct ContractVersion(NonZeroU64);
+
+impl Default for ContractVersion {
+    fn default() -> Self {
+        // SAFETY: The only way the API contract of `NonZeroU64` can be
+        // violated is if we construct values of this type using 0 as
+        // argument.
+        Self(unsafe { NonZeroU64::new_unchecked(1) })
```
batconjurer, line 63: Amazes me that this type does not implement Default.

apps/src/lib/config/mod.rs
```diff
@@ -83,7 +83,7 @@ pub struct Ledger {
     pub chain_id: ChainId,
     pub shell: Shell,
     pub tendermint: Tendermint,
-    pub ethereum: ethereum::Config,
+    pub ethereum_bridge: ethereum_bridge::ledger::Config,
```
batconjurer, line 86: Oh wait, you do have this as a config. We can't configure minimum confirmations or addresses in a normal config. They have to be set at genesis and changed only with governance votes.

batconjurer (changes requested):

apps/src/lib/config/mod.rs
sug0: ah okay that makes sense, since you can run into consensus failures if two validators have different ideas of what the min number of blocks should be for an event to be considered as valid.

james-chf: With the Tendermint v0.37 design, we shouldn't get consensus failures if validators have different ideas of min confirmations. For the addresses of the contracts, those should change only via hard fork as far as I understand it, if we do a contract upgrade, though technically it would be possible for a a governance proposal to change them (right now as the code currently is).

apps/src/lib/config/ethereum_bridge/params.rs
james-chf: Have renamed to `GenesisConfig`

apps/src/lib/config/mod.rs
batconjurer: Even if it didn't lead to consensus failures, it is supposed to be that all validators agree on this value. Otherwise, events may get skipped, or someone trying to slash some one else, or generally undesired behavior. Anything that is protocol specified shouldn't be in a config file for running the node software, it should only be in a genesis config.

batconjurer (approved):

Issue #434, opened 2022-09-05
Use a `BTreeSet` to store the set of validators who have seen an Ethereum event


james-chf:
After https://github.com/anoma/namada/pull/284 has landed

For any Ethereum event we include into Namada, we store a vector of validator addresses who have seen an event: 

```
seen_by: Vec<Address>
````

We should instead use a `BTreeSet<Address>` as this field should always be in order, and each `Address` may appear only once.

The specs must also be updated as well for this issue to be closed.

PR #585, opened 2022-10-12
Use `BTreeSet` instead of `Vec`/`HashSet` for storing validators who have seen an Ethereum event

james-chf:
Closes https://github.com/anoma/namada/issues/434

Updates the spec and code to:
- update specs to specify a `BTreeSet<Address>` instead of `Vec<Address>` for the `eth_msgs/$msg_hash/seen_by` field
- update `EthMsg::seen_by` to be a `BTreeSet`
- update `MultiSignedEthEvent::signers` to be a `BTreeSet`

This is a better data structure for doing https://github.com/anoma/namada/issues/197 (updating Ethereum events in storage) as when new validators see an event, we can read the `seen_by` from storage and insert new validators without having to worry about reordering a `Vec` before we store it again. We also get a uniqueness constraint on validators (the elements of the `BTreeSet`) which `Vec` doesn't give us.


shared/src/types/vote_extensions/ethereum_events.rs
```diff
@@ -79,11 +79,11 @@ pub struct MultiSignedEthEvent {
...
     /// List of addresses of validators who signed this event
     #[cfg(feature = "abcipp")]
-    pub signers: HashSet<Address>,
+    pub signers: BTreeSet<Address>,
     /// List of addresses of validators who signed this event
     /// and block height at which they signed it
     #[cfg(not(feature = "abcipp"))]
-    pub signers: HashSet<(Address, BlockHeight)>,
+    pub signers: BTreeSet<(Address, BlockHeight)>,
 }
```
james-chf, line 87: `MultiSignedEthEvent` and `EthMsgUpdate` are now the same for ABCI++ and may become the same for ABCI+ as well soon. These structs could probably be combined in the future - a `MultiSignedEthEvent` translates straightfowardly into an `EthMsgUpdate`.

james-chf:
pls update wasm

sug0 (approved):

james-chf:
pls update wasm

james-chf:
pls update wasm

PR #590, opened 2022-10-13
fix tx bond test condition

tzemanovic:
fixes issue spotted here: https://github.com/anoma/namada/actions/runs/3241666911/jobs/5313917717

The problem was that the delegation detection didn't check for equality on the source and validator addresses, which is what is produced from this seed (1ec12c4cefbc88748f77cf8f1a8c1484ad09fb62):

```
input: ... Bond { 
    validator: Established: atest1v4ehgw36xgunxvj9xqmny3jyxycnzdzxxqeng33ngvunqsfsx5mnwdfjgvenvwfk89prwdpjd0cjrk,
  source: Some(Established: atest1v4ehgw36xgunxvj9xqmny3jyxycnzdzxxqeng33ngvunqsfsx5mnwdfjgvenvwfk89prwdpjd0cjrk), 
  amount: Amount { micro: 0 } 
})
```

This is a false negative from a test so I think we can skip devnet for this change


wasm/wasm_source/src/tx_bond.rs
```diff
@@ -154,71 +156,63 @@ mod tests {
...
+            // It was a self-bond
+            for epoch in 0..pos_params.pipeline_len {
+                let expected_bond =
+                    HashMap::from_iter([(genesis_epoch, initial_stake)]);
+                let bond: Bond<token::Amount> = bonds_post
+                    .get(epoch)
+                    .expect("Genesis validator should already have self-bond");
+                assert_eq!(
+                    bond.pos_deltas, expected_bond,
+                    "Delegation before pipeline offset should be equal to the \
```
brentstone, line 196: Change assert failure message away from "Delegation" (perhaps "Self-bond")?

wasm/wasm_source/src/tx_bond.rs
```diff
@@ -154,71 +156,63 @@ mod tests {
...
                     );
-                }
+                let expected_bond = HashMap::from_iter([
+                    (genesis_epoch, initial_stake),
+                    (start_epoch, bond.amount),
+                ]);
+                let bond: Bond<token::Amount> = bonds_post.get(epoch).unwrap();
+                assert_eq!(
+                    bond.pos_deltas, expected_bond,
+                    "Delegation at and after pipeline offset should contain \
```
brentstone, line 212: Same as above regarding "Delegation"

wasm/wasm_source/src/tx_bond.rs
tzemanovic, line 196:
```suggestion
                    "Self-bond before pipeline offset should be equal to the \
```

wasm/wasm_source/src/tx_bond.rs
tzemanovic, line 212:
```suggestion
                    "Self-bond at and after pipeline offset should contain \
```

brentstone (approved):
LGTM

PR #607, opened 2022-10-14
Implement event log in the ledger

sug0:
Based on https://github.com/anoma/namada/pull/587

Implements the second task of issue https://github.com/anoma/namada/issues/563

PR #610, opened 2022-10-14
Add two new ABCI queries for ledger txs

sug0:
Based on https://github.com/anoma/namada/pull/607

Implements the third task of issue https://github.com/anoma/namada/issues/563


james-chf (changes requested):
lgtm, some comments around making it more explicit we're working with (hex encoded) transaction hashes

shared/src/types/hash.rs
```diff
@@ -145,3 +147,81 @@ impl Value for Hash {
...
+}
+
+impl Deref for HashString {
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        // SAFETY: We can only construct a `HashString`
+        // from valid hex encoded data.
+        unsafe { std::str::from_utf8_unchecked(&self.inner) }
+    }
```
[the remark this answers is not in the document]
sug0: the API is basically a contract that guarantees that as long we are using safe Rust, we can't construct a `HashString` with non utf8 data. therefore, I don't see a reason to call `expect(...)`, if the API guarantees we always have valid utf8.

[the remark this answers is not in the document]
james-chf: Makes sense!

apps/src/lib/node/ledger/rpc.rs
```diff
@@ -4,24 +4,29 @@ use std::fmt::Display;
...
 
 use namada::types::address::Address;
+use namada::types::hash::{self, HashString};
 use namada::types::storage;
 use thiserror::Error;
 
 use crate::facade::tendermint::abci::Path as AbciPath;
 
-/// RPC query path
+/// RPC query path.
```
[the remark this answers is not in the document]
sug0: yeah, consistency would be nice, in the comments

shared/src/types/hash.rs
```diff
@@ -145,3 +154,80 @@ impl Value for Hash {
         Hash([0u8; 32])
     }
 }
+
+/// A hex encoded hash.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct HexEncodedHash {
```
batconjurer, line 160: It looks like your branch doesn't have the `From<&str>` methods for hash yet, but they are coming in my PRs. Since `Display` on `Hash` is hex, the from string method will do all of this. So I'm not sure you need a special type here. In fact, I don't see why you couldn't just use a `&str`

sug0: the main motivation to add `HexEncodedHash` was to parse hex encoded hashes `$tx_hash` we receive in `/<accepted|applied>/<$tx_hash>` into a stack allocated value. we know these hashes can only of 64 bytes, so we don't need to heap-alloc a `String` for them.

sug0: I also personally prefer small domain specific types, which uphold some invariants

batconjurer: https://github.com/anoma/namada/blob/93000b0d28c9ed54b3a597b9cfddc7a99adda6bf/shared/src/types/hash.rs#L101

batconjurer: That code should work. It does pass through a heap allocation, so it could be adjusted to use a `try_into` to go directlu into a 32 byte array.

batconjurer: Anyway, the motivation is clear, I'm just trying to prevent double writing code in two separate PRs. So I'm suggesting you steal some upcoming changes.

batconjurer: > I also personally prefer small domain specific types, which uphold some invariants

But our `Hash` type should be hex encoded as an invariant. Is there some other invariant I am unaware of?

sug0: not really, I don't think. the bytes in the hash itself are completely arbitrary.

batconjurer (approved):

james-chf (changes requested):

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -6,118 +6,71 @@
...
     pub fn matches(&self, event: &Event) -> bool {
         event.event_type == self.event_type
             && event
                 .attributes
                 .get(&self.attr)
-                .map(|value| value == self.value)
+                .and_then(|value| {
+                    value
+                        .as_str()
+                        .try_into()
```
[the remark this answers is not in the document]
sug0: at the rpc level we already go from a hex encoded hash to a regular hash, and perform sanity checks at that level. this just does a roundtrip encoding, to compare with the hash in the query.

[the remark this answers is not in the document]
james-chf: From the perspective of this fn this isn't infallible though, it could be passed an `Event` that could not have a hex hash at the `attr`. I guess `Event::attributes` should be something more typed than `HashMap<String, String>`.

More generally than this PR I want to look at how we are handling errors in our code, I strongly think we should handle all `Err`s even if they should never occur, in case one of our underlying assumptions is wrong (or becomes invalid later on), or the function comes to be used somewhere else where our assumption why some error will never occur doesn't hold. So imo we should log or refactor such there is no `Err` to deal with.

james-chf (approved):

PR #619, opened 2022-10-17
Make client code call into the new ABCI queries tx endpoints

sug0:
Based on https://github.com/anoma/namada/pull/610

Implements the fourth task of issue https://github.com/anoma/namada/issues/563


apps/src/lib/client/rpc.rs
```diff
@@ -40,8 +41,66 @@ use crate::facade::tendermint_rpc::query::Query;
...
+                    );
+                    break Err(());
+                }
+            };
+            if let Some(e) = events.pop() {
+                // we should only have one event matching the query
+                break Ok(e);
+            }
+            // simple linear backoff - if an event is not available,
+            // increase the backoff duration by one second
```
james-chf, line 91: Could indicate we are doing a the retry to the user somewhere in this loop e.g. something like

```rust
println!("No events received, may retry in $backoff seconds")
```

apps/src/lib/client/tendermint_rpc_types.rs
```diff
@@ -34,65 +35,69 @@ pub struct TxResponse {
...
+            .map(String::as_str)
+            // TODO: fix finalize block, to return initialized accounts,
+            // even when we reject a tx?
+            .or(Some("[]"))
+            // NOTE: at this point we only have `Some(vec)`, not `None`
+            .ok_or_else(|| unreachable!())
+            .and_then(|initialized_accounts| {
+                serde_json::from_str(initialized_accounts)
+                    .map_err(|err| format!("JSON decode error: {err}"))
+            })?;
```
james-chf, line 81: What do we expect `event.get("initialized_accounts")` to look like now? It could be useful to factor out and have a test specifically for this parsing workaround we're doing. Is this an issue with Tendermint?

sug0: no it's an issue with our finalize block implementation. at first I thought it was tendermint's fault, but it's namada code. when we reject a tx, `initialized_accounts` is not set in the `Event` attributes, so we need to do this crap, which is a bit unfortunate.

apps/src/lib/client/tendermint_rpc_types.rs
```diff
@@ -34,65 +35,69 @@ pub struct TxResponse {
...
+
+impl TxResponse {
+    /// Convert an [`Event`] to a [`TxResponse`], or error out.
+    pub fn from_event(event: Event) -> Self {
+        event.try_into().unwrap_or_else(|err| {
+            eprintln!("Error fetching TxResponse: {err}");
+            safe_exit(1);
+        })
     }
 }
```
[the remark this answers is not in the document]
james-chf: To do later - https://github.com/anoma/anoma/issues/1049 - these `safe_exit`s need to be removed

apps/src/lib/client/rpc.rs
batconjurer: I think we normally take longer than a second to get a response. So we might want to change the backoff or else we will always print messages that make people worry :sweat_smile:

apps/src/lib/client/tendermint_rpc_types.rs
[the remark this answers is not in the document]
batconjurer: This sounds like an issue with the test runner. It is supposed to watch the exit status and return an error if it gets a non-zero value.

apps/src/lib/client/tx.rs
```diff
@@ -1241,24 +1237,26 @@ pub async fn submit_tx(
...
             .and_then(|val| val.parse().ok())
-            .unwrap_or(DEFAULT_ANOMA_TENDERMINT_EVENTS_MAX_WAIT_TIME),
+            .unwrap_or(DEFAULT_NAMADA_EVENTS_MAX_WAIT_TIME),
     );
-    tracing::debug!("Tenderming address: {:?}", address);
-    let rpc_cli = HttpClient::new(address.clone())?;
-
-    // Broadcast the supplied transaction
-    broadcast_tx(address, &to_broadcast).await?;
+    let deadline = Instant::now() + max_wait_time;
```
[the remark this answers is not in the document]
batconjurer: `broadcast_tx` logs that transaction is being submitted. I don't know if we need to log env vars the user sets.

batconjurer (approved):
These look fine to me, but see my comments on the PR it is built off of

sug0:
@james-chf  left the `safe_exit` discussion a bit open ended. for this PR, I'm just going to follow the general pattern of the rest of the code already in the client.

apps/src/lib/client/tendermint_rpc_types.rs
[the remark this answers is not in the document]
james-chf: It would be good if the test runner could catch that, but it's still bad that we have these functions maybe taking a decision to terminate the process

apps/src/lib/client/tendermint_rpc_types.rs
james-chf: Ah right, I made an issue - https://github.com/anoma/namada/issues/643 , we can eventually fix it and simplify this

james-chf (approved):
tested locally and works well, the `safe_exit` calls should be cleared up at some point

PR #620, opened 2022-10-17
Decouple vote tracking keys/logic from Ethereum events

james-chf:
Relates to https://github.com/anoma/namada/issues/621

This PR is a refactor only and there should be no change in functionality (although a fair bit of code has changed/ moved around).

- In the `protocol::transactions` module, unbundle the three fields which are used for tracking votes from the `ethereum_events::EthMsg` struct and put them in a self-contained `votes::VoteTracking` struct
- change the EthBridge VP `storage::eth_msgs::Keys` struct to a more generic `storage::vote_tracked::Keys<T>` (we now use `Keys<EthereumEvent>` for `/eth_msgs/$msg_hash/...` keys)
- move functionality out of the `protocol::transactions::ethereum_events` module and into the `protocol::transactions` module which should be useful for tracking validator set update votes i.e. vote tracking and reading/updating associated keys

Some further logic could be moved out of the `ethereum_events` module but that could be done after https://github.com/anoma/namada/pull/618 is done

We could track voting power behind validator set updates similarly to Ethereum events, e.g. in storage like:

```
/valset_upds/$msg_hash/body
/valset_upds/$msg_hash/seen
/valset_upds/$msg_hash/seen_by
/valset_upds/$msg_hash/voting_power
```

The big differences between valset updates and Ethereum events being:

- there should only be one (valid) validator set update per epoch (as opposed to many potentially valid Ethereum events per epoch), that validators are submitting onchain as protocol transactions
- for most nodes, there is nothing to do once valset update has enough voting power behind it i.e. `seen = true`, as opposed to Ethereum events where every node must update their storage to account for minting of wrapped ERC20s or release of native NAM from escrow


apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
```diff
@@ -48,12 +49,8 @@ impl From<MultiSignedEthEvent> for EthMsgUpdate {
 pub struct EthMsg {
     /// The event being stored
     pub body: EthereumEvent,
-    /// The total voting power that's voted for this event across all epochs
-    pub voting_power: FractionalVotingPower,
-    /// The addresses of validators that voted for this event
-    pub seen_by: BTreeSet<Address>,
-    /// Whether this event has been acted on or not
-    pub seen: bool,
```
james-chf, line 56: A motivation for factoring these fields out now while at the same time implementing https://github.com/anoma/namada/issues/197 is that these three fields need to be read/updated but the body doesn't, and it should never change after an event is first recorded to storage. Most of the logic for updating vote tracking in storage can work with this `VoteTracking` struct now rather than an `EthMsg`.

apps/src/lib/node/ledger/protocol/transactions/votes.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// set type as validators should only be able to vote at most once,
+    /// and [`BTreeSet`] specifically as we want this field to be
+    /// deterministically ordered for storage.
+    pub seen_by: BTreeSet<Address>,
+    /// Whether this event has been acted on or not - this should only ever
+    /// transition from `false` to `true`, once there is enough voting power
+    pub seen: bool,
+}
+
+pub fn calculate_new(
```
james-chf, line 34: Formerly `calculate_new_eth_msg`

apps/src/lib/node/ledger/protocol/transactions/votes.rs
```diff
@@ -0,0 +1,115 @@
...
+        voting_power: seen_by_voting_power,
+        seen_by: seen_by
+            .iter()
+            .map(|(validator, _)| validator.to_owned())
+            .collect(),
+        seen: newly_confirmed,
+    })
+}
+
+pub fn calculate_updated<D, H, T>(
```
james-chf, line 65: Formerly `calculate_updated_eth_msg` - this is being implemented in https://github.com/anoma/namada/pull/618 which will be rebased on this PR. The `calculate_diff` placeholder fn has been removed altogether.

james-chf:
pls update wasm

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
```diff
@@ -48,12 +49,8 @@ impl From<MultiSignedEthEvent> for EthMsgUpdate {
...
+    /// Tracking of votes for this event
+    pub vote_tracking: VoteTracking,
```
james-chf, line 53: Suggestions for better naming around `VoteTracking`, `votes`, etc. appreciated!

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
```diff
@@ -97,29 +102,33 @@ mod test {
...
     }
 
     #[test]
     fn test_prefix() {
-        assert_matches!(&prefix().segments[..], [
+        assert_matches!(&eth_msgs_prefix().segments[..], [
                 DbKeySeg::AddressSeg(ADDRESS),
                 DbKeySeg::StringSeg(s),
             ] if s == PREFIX_KEY_SEGMENT)
     }
```
sug0, line 126: renaming these tests to something like `test_eth_...`

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
```diff
@@ -6,27 +6,21 @@ use crate::types::storage::Key;
...
 const VOTING_POWER_KEY_SEGMENT: &str = "voting_power";
 
-/// Generator for the keys under which details of an [`EthereumEvent`] is stored
-pub struct Keys {
-    /// The prefix under which the details are an [`EthereumEvent`] is stored
+/// Generator for the keys under which details of a vote tracked event is stored
+pub struct Keys<T> {
+    /// The prefix under which the details of a vote tracked event is stored
     pub prefix: Key,
+    _phantom: std::marker::PhantomData<T>,
```
sug0, line 18:
```suggestion
    _phantom: std::marker::PhantomData<&'static T>,
```
we don't own values of `T`, no need to trigger the drop checker

sug0: why do we need to include a type parameter here though? is it being used anywhere? it would be nice if you could use this `T` to derive a new sub-key or smth, but it's not being used that way.

james-chf: Oh nice, I didn't know we could use `'static` here. Re. using a type parameter, it was so that `vote_tracked::Keys` to do with different things (i.e. Ethereum events vs valset updates) would have different types. So could do things like

```rust
impl From<&Hash> for Keys<EthereumEvent>
```

where the returned `keys` struct has a prefix which uses `eth_msgs_prefix()`, a `Keys<ValsetUpd>` or something might still be buildable from a hash but use a `valset_upd_prefix()` for example.

then we can also do stuff like

```rust
let keys: Keys<EthereumEvent> = (&hash).into();
```

Maybe there is a better way to handle this though, can change this.

james-chf: Maybe `T` should be bounded here by `BorshSerialize + BorshDeserialize` as well. I guess there are only specific `T` that should ever be allowed here (right now, just `EthereumEvent`).

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
```diff
@@ -4,29 +4,23 @@ use crate::types::hash::Hash;
...
 const BODY_KEY_SEGMENT: &str = "body";
 const SEEN_KEY_SEGMENT: &str = "seen";
 const SEEN_BY_KEY_SEGMENT: &str = "seen_by";
 const VOTING_POWER_KEY_SEGMENT: &str = "voting_power";
 
-/// Generator for the keys under which details of an [`EthereumEvent`] is stored
-pub struct Keys {
-    /// The prefix under which the details are an [`EthereumEvent`] is stored
+/// Generator for the keys under which details of a vote tracked event is stored
+pub struct Keys<T: 'static> {
```
sug0, line 15:
```suggestion
pub struct Keys<T> {
```
you don't need to add trait bounds at the struct def level, generally

sug0 (changes requested):

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
james-chf: It does't compile if the `: 'static` is removed

```
error[E0310]: the parameter type `T` may not live long enough
  --> shared/src/ledger/eth_bridge/storage/vote_tracked.rs:18:15
   |
15 | pub struct Keys<T> {
   |                 - help: consider adding an explicit lifetime bound...: `T: 'static`
...
18 |     _phantom: std::marker::PhantomData<&'static T>,
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...so that the reference type `&'static T` does not outlive the data it points at

For more information about this error, try `rustc --explain E0310`.
error: could not compile `namada` due to previous error
```

sug0: ah. well that's kinda odd. alright!

sug0 (approved):

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
batconjurer: If you remove the `'static` on both is works fine.

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
batconjurer: Hope the find / replace wouldn't be too much of a hassle, but I like `VoteTallies` and variations thereof

apps/src/lib/node/ledger/protocol/transactions/votes.rs
batconjurer: Can we get a doc string?

apps/src/lib/node/ledger/protocol/transactions/votes.rs
batconjurer: Can we get a doc string?

shared/src/ledger/eth_bridge/storage/vote_tracked.rs
batconjurer, line 15: Is there a plan to use this also for val set updates? I'm curious about the generic paramter.

sug0: that would be the idea yeah

james-chf: Yeah I've put an idea in the PR description but it needs to be properly specced out still. Now thinking about it further, I don't think we absolutely need vote tallying or even including validator set update transactions in digests for the transfers of power. As some off chain relayer could just watch for protocol txs that contain the validator set signatures, and then do a relay once it had gathered enough. Without any vote extension digest ever needing to be constructed.

If we did go with tallying votes for the validator set updates, then once a validator set update had a quorum of voting power behind it, one of the nodes (probably whoever gets the reward) could "act on" the validator set update by doing the relay, and other nodes don't need to do anything else extra. 

We would need to track of at which block heights the validator set update signatures (i.e. the fake "vote extensions") were included, so the relaying node could know in which blocks to look. This could be done by changing the `seen_by` field to be a `BTreeSet<(Address, BlockHeight)>` rather than just a `BTreeSet<Address>`. i.e. we could record the heights at which validators submitted their "vote" for a validator set update, rather than just the addresses of validators that had voted for a validator set update.

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
james-chf: Have updated to `votes::Tally` etc

batconjurer (approved):

batconjurer (approved):

PR #622, opened 2022-10-17
Update secp256k keys with recovery id

batconjurer:
This ports the changes from #617 to our branch and updates the key files to be the same as main.


shared/src/types/key/mod.rs
```diff
@@ -356,6 +357,26 @@ impl<PK: PublicKey> From<&PK> for PublicKeyHash {
...
+            pkh.0
+        }
+        common::PublicKey::Secp256k1(pk) => {
+            let pkh = PublicKeyHash::from(pk);
+            pkh.0
+        }
+    }
+}
+
+/// Convert Tendermint validator's raw hash bytes to Anoma raw hash string
```
sug0, line 375:
```suggestion
/// Convert Tendermint validator's raw hash bytes to Namada raw hash string
```

shared/src/types/key/secp256k1.rs
```diff
@@ -316,7 +323,7 @@ impl<'de> Deserialize<'de> for Signature {
         struct ByteArrayVisitor;
 
         impl<'de> Visitor<'de> for ByteArrayVisitor {
-            type Value = [u8; libsecp256k1::util::SIGNATURE_SIZE];
+            type Value = [u8; libsecp256k1::util::SIGNATURE_SIZE + 1];
```
sug0, line 326: this has probably already gone into main, but we could add a const here with a docstr explaining why this `+ 1` is needed

sug0 (changes requested):

batconjurer:
pls update wasm

shared/src/types/key/mod.rs
```diff
@@ -356,6 +357,26 @@ impl<PK: PublicKey> From<&PK> for PublicKeyHash {
...
+            let pkh = PublicKeyHash::from(pk);
+            pkh.0
+        }
+    }
+}
+
+/// Convert Tendermint validator's raw hash bytes to Namada raw hash string
+pub fn tm_raw_hash_to_string(raw_hash: impl AsRef<[u8]>) -> String {
+    HEXUPPER.encode(raw_hash.as_ref())
+}
```
james-chf, line 378: I'm not sure about these helpers here, could they live with the Tendermint related code that will use them? (they aren't used yet in this PR). These might belong in a separate PR

shared/src/types/key/mod.rs
```diff
@@ -356,6 +357,26 @@ impl<PK: PublicKey> From<&PK> for PublicKeyHash {
...
+            pkh.0
+        }
+        common::PublicKey::Secp256k1(pk) => {
+            let pkh = PublicKeyHash::from(pk);
+            pkh.0
+        }
+    }
+}
+
+/// Convert Tendermint validator's raw hash bytes to Namada raw hash string
```
james-chf, line 375: `validator's raw hash bytes` - not sure what this is, test/example would be useful here

shared/src/types/key/secp256k1.rs
```diff
@@ -316,7 +323,10 @@ impl<'de> Deserialize<'de> for Signature {
         struct ByteArrayVisitor;
 
         impl<'de> Visitor<'de> for ByteArrayVisitor {
-            type Value = [u8; libsecp256k1::util::SIGNATURE_SIZE];
+            // [`libsecp265k1::util::SIGNATURE_SIZE`] is for a traditional
+            // signature on this curve. For Ethereum, an extra byte is included
+            // that prevents malleability attacks.
+            type Value = [u8; libsecp256k1::util::SIGNATURE_SIZE + 1];
```
james-chf, line 329: The `libsecp256k1::util::SIGNATURE_SIZE + 1` and comment should be pulled out into a const and docstring e.g. `SIG_BYTES_LEN`, then whereever we are using `[u8; 65]`/`libsecp256k1::util::SIGNATURE_SIZE + 1` or similar we should use the const instead e.g. `[u8; SIG_BYTES_LEN]`.

shared/src/types/key/secp256k1.rs
```diff
@@ -304,6 +310,7 @@ impl Serialize for Signature {
         for elem in &arr[..] {
```
james-chf, line 310: On L309
```
        let mut seq = serializer.serialize_tuple(arr.len())?;
```

should this be `arr.len() + 1` now that we have the extra `seq.serialize_element()` call? We should test our Serialize/Deserialize impls match up (we should already have had tests probably, I could have a go at this even as a separate PR)

james-chf (commented):
have started taking a look, will review https://github.com/anoma/namada/pull/579 and other PRs before this for context

shared/src/types/key/mod.rs
batconjurer: Those got grabbed when I synced this file with main. There was a bug when looking up tendermint addresses (@sug0  should remember. Something about not being able to slash if someones address wasn't valid utf-8). I can remove these as they aren't directly relevant to this PR, but they will reappear if we ever merge in main.

shared/src/types/key/mod.rs
batconjurer: Same thing, this got pulled in from main. Can remove if necessary

shared/src/types/key/secp256k1.rs
batconjurer: good catch! Will change and cover with tests.

shared/src/types/key/mod.rs
james-chf: Oh right, I thought these were being added new for some further PR, if they are already merged in `main` then I guess it's immaterial. I'm guessing the bug was re. https://github.com/anoma/namada/blob/a669ca573f1d3f61b87cdcb8eba7cb9a6f872a6a/shared/src/types/vote_extensions/ethereum_events.rs#L24-L27

james-chf: I've opened https://github.com/anoma/namada/issues/635. I don't fully recall the issue either but I think it was that a validator should be able to look at it's own storage to get a validator's address, rather than trust an address provided in vote extensions.

sug0: we were trying to parse utf8 data from raw hashes, which are of course completely arbitrary, and there was a piece of code that effectively skipped slashing if you failed to parse a hash as utf8 data.

sug0:
ci is throwing a fit. could we fix these issues before merging?

batconjurer:
pls update wasm

batconjurer:
pls update wasm

shared/src/types/key/secp256k1.rs
```diff
@@ -405,12 +435,19 @@ impl PartialOrd for Signature {
...
-impl TryFrom<&[u8; 64]> for Signature {
+impl TryFrom<&[u8; 65]> for Signature {
     type Error = ParseSignatureError;
 
-    fn try_from(sig: &[u8; 64]) -> Result<Self, Self::Error> {
-        libsecp256k1::Signature::parse_standard(sig)
-            .map(Self)
+    fn try_from(sig: &[u8; 65]) -> Result<Self, Self::Error> {
+        let sig_bytes = sig[..64].try_into().unwrap();
+        let recovery_id = RecoveryId::parse(sig[64]).map_err(|err| {
```
james-chf, line 443: We could use `SIGNATURE_LENGTH`/`SIGNATURE_LENGTH - 1` here as well instead of `65`/`64`

james-chf (approved):
lgtm

PR #581, opened 2022-10-11
Update bridge pool vp to use merkle tree

batconjurer:
This PR closes #422 

This PR depends on #579 which added a merkle tree to our storage for the bridge pool. This PR actually switches to using this tree and the necessary updates to the VP and wasm have been made.

PR #624, opened 2022-10-17
Add query endpoints for bridge merkle proofs

### LINKED ISSUE #425: Make the Ethereum bridge pool queryable
 
batconjurer:
This PR close #425 

This PR depends on #581. 
* This PR adds types than can be ABI encoded and relayed to Ethereum. 
* The bridge pool transfers can be read using an ABCI query endpoint.
* Merkle proofs for transfers can be requested for batches via an ABCI query endpoint.

This PR does not add CLI commands for querying these endpoints.


shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -301,6 +301,38 @@ fn compute_hash(
...
+                .collect(),
+        );
+        let powers = Token::Array(
+            self.powers
+                .iter()
+                .map(|power| Token::Uint(power.clone().into()))
+                .collect(),
+        );
+        let nonce = Token::Uint(self.nonce.clone().into());
+        [Token::FixedArray(vec![addrs, powers, nonce])]
```
sug0, line 332:
```suggestion
        [Token::Tuple(vec![addrs, powers, nonce])]
```
I think it should be a tuple. a `FixedArray` would be wrong, because it would be receiving heterogeneous type args.

shared/src/types/key/secp256k1.rs
```diff
@@ -392,6 +421,16 @@ impl BorshSchema for Signature {
     }
 }
 
+impl Encode<1> for Signature {
+    fn tokenize(&self) -> [Token; 1] {
+        let sig_serialized = libsecp256k1::Signature::serialize(&self.0);
+        let r = Token::FixedBytes(sig_serialized[..32].to_vec());
+        let s = Token::FixedBytes(sig_serialized[32..].to_vec());
+        let v = Token::FixedBytes(vec![self.1.serialize()]);
+        [Token::FixedArray(vec![r, s, v])]
```
sug0, line 430:
```suggestion
        [Token::Tuple(vec![r, s, v])]
```
same thing

shared/src/types/vote_extensions/validator_set_update.rs
sug0: I'm also not sure if this serializes to the same type anyway, since it's not a solidity primitive...

batconjurer: I have no idea. I asked @Fraccaman  and this is what he said. Probably won't know for sure until we do integrations.

batconjurer:
pls update wasm

batconjurer:
pls update wasm

shared/src/types/eth_bridge_pool.rs
```diff
@@ -96,3 +104,59 @@ pub struct GasFee {
...
+    pub sigs: Vec<crate::types::key::secp256k1::Signature>,
+    /// The Merkle root being signed
+    pub root: KeccakHash,
+    /// The block height at which this root was valid
+    pub height: BlockHeight,
+}
+
+impl Encode<2> for MultiSignedMerkleRoot {
+    fn tokenize(&self) -> [Token; 2] {
+        let MultiSignedMerkleRoot { sigs, root, .. } = self;
```
james-chf, line 123: how come we don't include the height in the tokenization?

shared/src/types/eth_bridge_pool.rs
```diff
@@ -96,3 +104,59 @@ pub struct GasFee {
...
+    }
+}
+
+/// All the information to relay to Ethereum
+/// that a set of transfers exist in the Ethereum
+/// bridge pool.
+pub struct RelayProof {
+    /// Information about the signing validators
+    pub validator_args: ValidatorSetArgs,
+    /// A merkle root signed by ta quorum of validators
```
james-chf, line 139:
```suggestion
    /// A merkle root signed by a quorum of validators
```

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -301,6 +301,38 @@ fn compute_hash(
...
 
+/// Struct for serializing validator set
+/// arguments with ABI for Ethereum smart
+/// contracts.
+#[derive(Debug, Clone, Default)]
+pub struct ValidatorSetArgs {
+    /// Ethereum address of validators
+    pub validators: Vec<EthAddress>,
+    /// Voting powers of validators
+    pub powers: Vec<Uint>,
```
james-chf, line 312: Will the format of these voting powers be "normalized" uints? If so, we should document that, it could even have its own type in the namada codebase so we know where we are using it. 

I'm not sure if there is a proper term for that or what the proper term is. We normalize voting powers to be evenly distributed over some range (1 - 2**32), if I remember correctly

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -337,6 +356,133 @@ where
...
+            <Vec<PendingTransfer>>::try_from_slice(request_bytes.as_slice())
+        {
+            // get the latest signed merkle root of the Ethereum bridge pool
+            let signed_root: MultiSignedMerkleRoot =
+                match self.storage.read(&get_signed_root_key()) {
+                    Ok((Some(bytes), _)) => {
+                        BorshDeserialize::try_from_slice(bytes.as_slice())
+                            .unwrap()
+                    }
+                    _ => {
```
james-chf, line 419: can match on `Err(error)` here and log if it happens

shared/src/types/eth_bridge_pool.rs
batconjurer: It might be later that we use the height for the nonce. But that's a later PR where I figure out all the nonce stuff. Otherwise, it's not used in the smart contracts.

shared/src/types/vote_extensions/validator_set_update.rs
batconjurer: No, these shouldn't normalized AFAIK

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -337,6 +356,133 @@ where
...
+                                .into(),
+                            info: "Could not deserialize the signed Ethereum \
+                                   bridge pool merkle root"
+                                .into(),
+                            ..Default::default()
+                        };
+                    }
+                };
+            // get the merkle tree corresponding to the above root.
+            let tree = if let Ok(Some(stores)) =
```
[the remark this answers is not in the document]
batconjurer: I should actually put an expect here. This should never fail.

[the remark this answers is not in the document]
batconjurer: tbh, I should put an expect here. This should never fail.

shared/src/types/vote_extensions/validator_set_update.rs
sug0: ah actually, these should probably be normalized! I already have some code that does this here: https://github.com/anoma/namada/blob/b977845e4fdda64f571000c0508b72a1fca88ce0/shared/src/types/vote_extensions/validator_set_update.rs#L215-L224

sug0: it could become a separate type, since it's being used in other places

batconjurer: good to know! Will add a new type for this

james-chf (approved):
lgtm, we should document/add a type for the normalized voting powers (in this PR or another one)

batconjurer:
> lgtm, we should document/add a type for the normalized voting powers (in this PR or another one)

I added a normalized voting power type

sug0 (approved):
lgtm, but we should wait for #704 to merge

PR #634, opened 2022-10-18
ledger: remove nft tx/vp

Fraccaman:


Fraccaman:
pls update wasm

tzemanovic (approved):

bengtlofgren:
Do we want to remove rather than comment out? Will we never want this in the future?

PR #636, opened 2022-10-18
Replace HEXENCODEDHASH with hash

batconjurer:
Hoping to convince you this simplifies the code :sweat_smile:

### DIFF
--- a/apps/src/lib/node/ledger/events/log.rs
+++ b/apps/src/lib/node/ledger/events/log.rs
@@ -67,13 +67,12 @@ impl EventLog {
 
     /// Returns a new iterator over this [`EventLog`].
     #[inline]
-    pub fn iter_with_matcher<'query, 'log>(
-        &'log self,
-        matcher: dumb_queries::QueryMatcher<'query>,
-    ) -> impl Iterator<Item = &'log Event> + 'query
+    pub fn iter_with_matcher<'a, T>(
+        &'a self,
+        matcher: dumb_queries::QueryMatcher<'a, T>,
+    ) -> impl Iterator<Item = &Event> + 'a
     where
-        // the log should outlive the query
-        'log: 'query,
+        T: 'a + PartialEq + TryFrom<&'a str>
     {
         self.queue
             .iter()
@@ -83,7 +82,7 @@ impl EventLog {
 
 #[cfg(test)]
 mod tests {
-    use namada::types::hash::HexEncodedHash;
+    use namada::types::hash::Hash;
 
     use super::*;
     use crate::node::ledger::events::{EventLevel, EventType};
@@ -95,7 +94,7 @@ mod tests {
     macro_rules! accepted {
         ($hash:expr) => {
             dumb_queries::QueryMatcher::accepted(
-                &HexEncodedHash::try_from($hash).unwrap(),
+                Hash::try_from($hash).unwrap(),
             )
         };
     }
--- a/apps/src/lib/node/ledger/events/log/dumb_queries.rs
+++ b/apps/src/lib/node/ledger/events/log/dumb_queries.rs
@@ -6,46 +6,59 @@
 //! tm.event='NewBlock' AND <accepted|applied>.<$attr>='<$value>'
 //! ```
 
-use namada::types::hash::HexEncodedHash;
+use std::marker::PhantomData;
+use namada::types::hash::Hash;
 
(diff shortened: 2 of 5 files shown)


sug0:
hash clash. LOL

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -6,46 +6,59 @@
...
+    pub fn matches(&self, event: &'a Event) -> bool {
         event.event_type == self.event_type
             && event
-                .attributes
-                .get(&self.attr)
-                .map(|value| value == self.value)
-                .unwrap_or_default()
+            .attributes
+            .get(&self.attr)
+            .map(|value| T::try_from(value.as_ref()).map(|v| v == self.value).unwrap_or(false))
```
sug0, line 38: see this is the part that bothers me: to do these kind of checks, you need to decode the hex encoded string into a `Hash`, for every iteration over the event log

apps/src/lib/node/ledger/events/log.rs
```diff
@@ -67,13 +67,12 @@ impl EventLog {
...
-        &'log self,
-        matcher: dumb_queries::QueryMatcher<'query>,
-    ) -> impl Iterator<Item = &'log Event> + 'query
+    pub fn iter_with_matcher<'a, T>(
+        &'a self,
+        matcher: dumb_queries::QueryMatcher<'a, T>,
+    ) -> impl Iterator<Item = &Event> + 'a
     where
-        // the log should outlive the query
-        'log: 'query,
```
sug0, line 76: we do get rid of this bit of lifetime gymnastics though, which is nice to the reader

shared/src/types/hash.rs
```diff
@@ -95,6 +92,26 @@ impl TryFrom<&[u8]> for Hash {
     }
 }
 
+impl TryFrom<String> for Hash {
+    type Error = self::Error;
+
+    fn try_from(string: String) -> HashResult<Self> {
+        Ok(
+            Self(<[u8; HASH_LENGTH]>::from_hex(string).map_err(Error::FromStringError)?)
+        )
```
sug0, line 101: here you could have used `try_into()` to avoid copy and pasting the same def

sug0: `string.as_str().try_into()`

apps/src/lib/node/ledger/events/log/dumb_queries.rs
```diff
@@ -6,46 +6,59 @@
...
 use crate::node::ledger::events::{Event, EventType};
 
 /// A [`QueryMatcher`] verifies if a Namada event matches a
 /// given Tendermint query.
 #[derive(Debug, Clone)]
-pub struct QueryMatcher<'q> {
+pub struct QueryMatcher<'a, T>
+where
+    T: 'a +  PartialEq + TryFrom<&'a str>
+{
```
sug0, line 20: I think it's cleaner to leave trait bounds to impl blocks

sug0: i.e. remove the trait bounds here

apps/src/lib/node/ledger/events/log/dumb_queries.rs
batconjurer: Yeah, but had to do a utf8 conversion before instead.

sug0: https://doc.rust-lang.org/stable/src/core/str/converts.rs.html#170-174
it just does a pointer cast, basically

shared/src/types/hash.rs
batconjurer: true. Copy pasta game weeeeaaak

apps/src/lib/node/ledger/events/log/dumb_queries.rs
batconjurer: Fair enough. Can literally just remove that and the code will work.

apps/src/lib/node/ledger/events/log/dumb_queries.rs
batconjurer: that's true because you were using the unchecked version.

sug0:
Merged here 9e7858732f8e02a77b8ac7924ff0f67add7ff542

PR #638, opened 2022-10-18
v0.9.0 draft

juped:


Fraccaman:
do we need to change tendermint version in CI?

Issue #644, opened 2022-10-19
E2E tests might be failing because of a race cond with the Tendermint RPC server starting up


sug0:
# Context

In e2e tests, before we make any client transfer, we grep the ledger's log for `"Anoma ledger node started"`. After this, we proceed to test different ledger functionality.

# Problem

When ran locally, most e2e tests seem to be failing arbitrarily. Upon closer inspection of log timestamps, we find that client requests are made, correctly, after `"Anoma ledger node started"` is logged, but not after `"module=rpc-server msg="Starting RPC HTTP server on..."`, which happens after the RPC server service in Tendermint starts. All client transfers make RPC requests, therefore they fail with a TCP connection reset error.

![](https://pomf2.lain.la/f/r0qa185.png)
![](https://pomf2.lain.la/f/2byk6z2l.png)

# Potential fix

Grep the Tendermint log for "`Starting RPC HTTP server"`.

sug0:
Fixed in https://github.com/anoma/namada/pull/646

PR #646, opened 2022-10-19
Fix e2e tests

sug0:
Very unassuming title for this issue.

Closes the following issues, hopefully:
- https://github.com/anoma/namada/issues/644
- https://github.com/anoma/namada/issues/418


tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -19,6 +19,8 @@ fn storage_key(path: &str) -> String {
 }
 
 #[test]
+#[ignore]
+// this test is outdated, so it is ignored
```
james-chf, line 23: Yep (also badly named...). There was this PR to clean it up (https://github.com/anoma/anoma/pull/1229) but that's a bit stale now probably. This test is still testing something we should continue testing (that a random actor cannot write arbitarily to `EthBridge` VP storage) but this test needs cleaning up.

sug0:
We might need to port some changes from this PR in `main`: https://github.com/anoma/namada/pull/661

sug0:
I've ran CI a couple of times, after the last fixes I pushed, and it passed every time. We could merge it as is, I guess? what do you think @james-chf @batconjurer 

image

batconjurer (approved):

james-chf:
If the CI job ends up flaking a lot it could become annoying to have to keep restarting it on our PRs. If we notice any super flaky tests though we can temporarily ignore them again and open an issue, so think this PR is good to merge

james-chf (approved):

PR #651, opened 2022-10-20
Fix wasms build

sug0:
The wasm `tx_bond` is not being built because of a small mistake in the `Makefile`. This PR addresses that.


james-chf (approved):

sug0:
pls update wasm

batconjurer (approved):

PR #655, opened 2022-10-20
Governance specs update

grarco:


documentation/specs/src/base-ledger/governance.md
```diff
@@ -44,6 +48,15 @@ Each proposal will be stored in a sub-key under the internal proposal address. T
...
 
+The `ProposalType` imply different combinations of:
+
+- the optional payload (memo) attached to the vote
+- which actors should be allowed to vote (delegators and validators or validators only)
+- the threshold to be used in the tally process
+- the optional wasm code attached to the proposal
+
+The correct logic to handle these different types will be hardcoded in protocol. We'll also rely on type checking to strictly enforce the correctness of a proposal given its type. These two approaches combined will prevent a user from deviating from the intended logic for a certain proposal type (e.g. providing a wasm code when it's not needed or allowing only validators to vote when also delegators should, etc...)
+
```
Fraccaman, line 59: I would list the 3 type of proposal and their constrains (default, eth, pgf)

documentation/specs/src/base-ledger/governance.md
```diff
@@ -108,81 +122,97 @@ This means that corresponding VPs need to handle these cases.
...
-    author: Address
-    votingStartEpoch: Epoch
-    votingEndEpoch: Epoch
-    graceEpoch: Epoch
-    proposalCode: Option<Vec<u8>>
+struct Proposal {
+    id: u64,
+    content: Vec<u8>,
+    author: Address,
+    proposal_type: ProposalType,
```
Fraccaman, line 132: toCamelCase()

documentation/specs/src/base-ledger/governance.md
```diff
@@ -108,81 +122,97 @@ This means that corresponding VPs need to handle these cases.
...
+
+Tallying, when no `memo` is required, is computed with the following rules:
+
+1. Sum all the voting power of validators that voted `yay`
+2. For any validator that voted `yay`, subtract the voting power of any delegation that voted `nay`
+3. Add voting power for any delegation that voted `yay` (whose corresponding validator didn't vote `yay`)
+4. If the aforementioned sum divided by the total voting power is greater or equal to the threshold set by `ProposalType`, the proposal outcome is positive otherwise negative.
 
-All the computation above must be made at the epoch specified in the `start_epoch` field of the proposal.
+If votes carry a `memo`, instead, the `yay` votes must be evaluated net of it. The protocol will implement the correct logic to make sense of these memos and compute the tally correctly:
```
Fraccaman, line 184: Is there a case where we have to evaluate the memo to compute the tally? we should be able to tally just with `yay` or `nay`, without looking the the memo

documentation/specs/src/base-ledger/governance.md
grarco: To use `type` as the name of the field we actually need to use the raw identifier `r#type` since `type` is a reserved keyword. Is it ok?

documentation/specs/src/base-ledger/governance.md
grarco: For example, for PGF, don't we need to look at the council in the memo and then compute the tally based on the different memos we received?

Fraccaman: ah, right!

Fraccaman (dismissed):

documentation/specs/src/base-ledger/governance.md
```diff
@@ -73,33 +86,68 @@ The governance machinery also relies on a subkey stored under the `NAM` token ad
...
+- Allows both validators and delegators to vote
+- Requires 2/3 of the total voting power to succeed
+- Doesn't expect any memo attached to the votes
+
+`PGFCouncil` is a specific proposal to elect the council for _Public Goods Funding_:
+
+- Doesn't carry any wasm code
+- Allows both validators and delegators to vote
+- Requires 1/3 of the total voting power to vote for the same council
+- Expect every vote to carry a memo in the form of a tuple `(Vec<Address>, BudgetCap)`
```
cwgoes, line 114: yes but people can vote for multiple councils & budget caps (approval voting) so it should be a set of tuples, right?

cwgoes (changes requested):

Fraccaman (approved):

cwgoes (approved):

Issue #459, opened 2022-09-08
`run_aux` should be able to manage any arbitrary number of tasks


james-chf:
Currently `run_aux` in the `eth-bridge-integration` branch needs a hardcoded five `JoinHandle`s to manage (tendermint node, ABCI server, broadcaster, Ethereum fullnode, oracle). 

https://github.com/anoma/namada/blob/65fb781db1f5194de11a983a39b2166f40b7fb8f/apps/src/lib/node/ledger/mod.rs#L539-L545

This means if we don't want to run the Ethereum fullnode and/or oracle, we still need to provide dummy `JoinHandle`s that don't do anything. We could simplify our code by making `run_aux` work with only handles for mandatory components (i.e. tendermint node + ABCI server + broadcaster) so that we don't need to provide dummy handles for the Ethereum fullnode and/or oracle components if they aren't needed.

james-chf:
Extra context: https://github.com/anoma/namada/pull/663#discussion_r1003023292

sug0:
nope, we still spawn a dummy task that exits immediately for the eth oracle

PR #663, opened 2022-10-21
Split oracle and geth initialization

james-chf:
Also remove the old mock node/oracles and associated code, since we can now just use `spawn_dummy_task()` since the `run_aux` refactor.

This is in preparation for moving of the oracle to be initialized by the `Shell` with configuration values from block storage that are being added in https://github.com/anoma/namada/pull/575, rather than using hardcoded constants


apps/src/lib/node/ledger/mod.rs
```diff
@@ -609,27 +614,63 @@ fn start_tendermint(
...
+
+    (Some(eth_receiver), oracle)
+}
+
+/// Launches a new task managing a `geth` process into the asynchronous
+/// runtime, and returns its [`task::JoinHandle`].
+///
+/// An oracle is also returned, along with its associated channel,
+/// for receiving Ethereum events from `geth`.
+async fn maybe_start_geth(
```
sug0, line 654: tbh, I'm not a big fan of returning those dummy tasks. a new refactor could be in order to return `Option<JoinHandle<T>>` values from some of these futures. then, we could use `futures::try_join_all`, on a `Vec` we collected with the flattening of all futures we want to wait on, from `Option<F>` for some `F: Future`. we'd probably need `F` to be `Pin<Box<dyn Future<Output = ShellResult<()>>>>` I guess

sug0 (approved):

apps/src/lib/node/ledger/mod.rs
james-chf: Yeah we need to refactor this further - dummy tasks should eventually no longer required. Linking to https://github.com/anoma/namada/issues/459. I briefly tried something like the `Vec` approach you mention but we need to update all our tasks to return the same type (e.g. `JoinHandle<Result<()>>`) which they don't currently do, also we run the shell in a regular thread rather than a tokio task for some reason.

sug0: the shell could potentially execute on a blocking tokio task I think, which can also have a `JoinHandle` of its own.
https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

batconjurer (approved):

PR #625, opened 2022-10-17
ICS20 transfer

yito88:
closes #117

depends on #261 

Ref: https://github.com/anoma/namada/blob/yuji/design_ibc/documentation/specs/src/interoperability/ibc.md#fungible-token-transfer

PR #664, opened 2022-10-21
IBC E2# test

yito88:
depends on #625 

NOTE: need v0.8.1 for tendermint `block_results()`


apps/src/lib/client/rpc.rs
```diff
@@ -1293,20 +1295,37 @@ pub async fn query_storage_value<T>(
...
+    let (value, _proof) =
+        query_storage_value_bytes(client, key, None, false).await;
+    match value {
+        Some(v) => match T::try_from_slice(&v[..]) {
+            Ok(value) => return Some(value),
+            Err(err) => eprintln!("Error decoding the value: {}", err),
+        },
+        None => return None,
+    }
+    cli::safe_exit(1)
```
batconjurer, line 1307:
```suggestion
    if let Some(Err(err)) =  value {
        eprintln!("Error deconding the value: {}", err);
        cli::safe_exit(1)
    } 
    value.map(|inner| inner.unwrap())
```

batconjurer (dismissed):

tests/src/e2e/ibc_tests.rs
```diff
@@ -0,0 +1,1404 @@
...
+) -> Result<()> {
+    let xan = find_address(test_a, XAN)?;
+    let sender = find_address(test_a, ALBERT)?;
+    let receiver = find_address(test_b, BERTHA)?;
+
+    let token = Some(Coin {
+        denom: xan.to_string(),
+        amount: "100000".to_string(),
+    });
+    let msg = MsgTransfer {
```
tzemanovic, line 700: can we use the cmd from #626 for the transfers?

yito88: Yes, we can use the `ibc-transfer` cmd for transaction requests with MsgTransfer.

apps/src/lib/client/rpc.rs
yito88: I renamed these variables. `value` is used twice.

tzemanovic (dismissed):

tests/src/e2e/setup.rs
```diff
@@ -89,8 +92,28 @@ pub fn single_node_net() -> Result<Test> {
     network(|genesis| genesis, None)
 }
 
+/// Setup two networks with a single genesis validator node.
+pub fn two_single_node_nets() -> Result<(Test, Test)> {
```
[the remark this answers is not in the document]
yito88: Thank you for your suggestion!
It looks better. I've changed `add_validators` to `set_validators` for changing the port of the first validator.

batconjurer (approved):

Issue #27, opened 2021-11-05
Add a metrics endpoint


SimSaladin:
A Prometheus-compatible metrics endpoint (`/metrics` on some configurable port) would be great for monitoring nodes. A Prometheus instance somewhere does all the heavy lifting of scraping the data and storing it as time series. The metrics endpoint just needs to respond with the current values of interesting application-specific data points.

Tendermint also has a metrics endpoint that can be enabled in its config. Currently it's a bit awkward to enable though, because you first need to generate the tendermint config by launching anoma once, edit the tendermint config and then restart anoma.

So maybe there should be just one place in anoma configuration to configure metrics, and it affects tendermint too. We could even aggregate the metrics from tendermint in order to avoid having two metrics endpoints (i.e. two ports for metrics) per node.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763045226887) by [Unito](https://www.unito.io)

tzemanovic:
@Fraccaman is there anything left for this (other than updates with ABCI++)?

Fraccaman:
We are currently just using the tendermint prometheus endpoint, we could integrate some more specific anoma/namada metrics.

diceride:
The modern approach would be to add OpenTelemetry to the p2p layer and allow the user to configure a custom [collector](https://opentelemetry.io/docs/collector/) endpoint to receive, process and export telemetry data. Instead of being locked-in by a single vendor, OpenTelemetry provides an open standard for generating, collecting, and exporting application telemetry across any number of backend systems, including Prometheus.

You can find an overview of configurable exporters in the [opentelemetry-collector-contrib/exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter) directory.

For Namada, OpenTelemetry can be configured with the [tracing](https://github.com/tokio-rs/tracing) framework. OpenTelemetry only needs to be configured once, for example when initializing a new Namada node.

https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/

Example

![example](https://user-images.githubusercontent.com/72995223/197396388-1ea2669c-85ec-491e-a469-f7bd6d9fc591.png)

diceride:
[...]

PR #666, opened 2022-10-21
Remove block processing channel

batconjurer:
The tests pass at least. And there isn't that unbounded channel.


sug0:
We should pin this in the namada hall of fame

sug0 (approved):

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -318,8 +306,8 @@ async fn process(
                 pending = pending.len(),
                 confirmed = confirmed.len(),
                 ?MIN_CONFIRMATIONS,
-                "Some events have reached the minimum number of confirmations \
-                 and will be sent onwards"
+                "Some events that have reached the minimum number of \
+                 confirmations and will be sent onwards"
```
james-chf, line 310: The typo here was fixed already in the other PR

apps/src/lib/node/ledger/ethereum_node/test_tools/mod.rs
```diff
@@ -206,6 +215,13 @@ pub mod mock_web3_client {
                         client.events.push((event_ty, data, height, seen));
                     }
                 }
+                if client.last_block_processed < Some(block_to_check.clone()) {
```
sug0, line 218:
```suggestion
                if client.last_block_processed.as_ref() < Some(&block_to_check) {
```
I think we might even be able to compare by ref here

james-chf:
This PR looks like it would change it so that we're no longer directly testing the oracle fully processing block heights. Instead we test indirectly that it is calling the fake `Web3Client::check_for_events` at least once for each expected block height. Why not have the oracle directly make observable what it is doing, and test on that?

batconjurer:
> This PR looks like it would change it so that we're no longer directly testing the oracle fully processing block heights. Instead we test indirectly that it is calling the fake `Web3Client::check_for_events` at least once for each expected block height. Why not have the oracle directly make observable what it is doing, and test on that?

I would love it if we could test directly, but that seems to involve invasive code changes and test artifacts in production code flows, which I really don't like. I was just trying to come up with an alternate solution using an existing mock. Otherwise, the best option is probably to make an oracle with generics in order to make a mock version for tests.

james-chf (approved):
Sure, I think adding code just for test paths is not great, that said more generally I think it would be good to make our components observable for production, by exposing things that could be published as a metric (https://github.com/anoma/namada/issues/27) - then that same instrumentation for observability can be used in tests. I think we should definitely revisit this before launch.

Feel free to take over the other PR to merge this and then that, though this isn't the approach I would have taken.

Re. making an oracle with generics for tests (I'm guessing making the oracle generic over it's `Web3Client`?), I think that would be useful, I had a go at this a while back but wasn't able to do it because of the `web30` thread issue (https://github.com/althea-net/web30/issues/4). Specifically trying to implement an `[async_trait]` `Web3Client` was difficult.

apps/src/lib/node/ledger/ethereum_node/oracle.rs
```diff
@@ -439,6 +424,7 @@ mod test_oracle {
             oracle,
             mut eth_recv,
             admin_channel,
+            blocks_processed_recv: _processed,
```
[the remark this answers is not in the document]
batconjurer: If we leave it out, Rust eager drops it and causes a panic on the channel send method in the mock oracle.

apps/src/lib/node/ledger/ethereum_node/test_tools/mod.rs
sug0: yeah, jacob was also surprised about this lol. it's definitely a bit of an obscure-ish feature of `Option` types, I think.

PR #667, opened 2022-10-21
switch to tendermint fork release v0.1.4-abciplus

tzemanovic:
the versioning is not semver compliant - it's just for temp fork releases

### DIFF
--- a/.changelog/unreleased/miscellaneous/667-tendermint-v0.1.4-abciplus.md
+++ b/.changelog/unreleased/miscellaneous/667-tendermint-v0.1.4-abciplus.md
@@ -1,0 +1,2 @@
+- Update tendermint to v0.1.4-abciplus.
+  ([#667](https://github.com/anoma/namada/pull/667))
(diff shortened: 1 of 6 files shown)


Fraccaman:
this should also update the docker image in `docker/namada/Dockerfile` line `28`

tzemanovic:
> this should also update the docker image in `docker/namada/Dockerfile` line `28`

done, thanks!

tzemanovic:
> in the interests of standardizing everything, this branch should update the CI to whatever version it specifies as well (it may only differ in the version number constant, but we should stick to it)

@juped I missed this before, it's a good call, updated

tzemanovic:
> this is held up for now on another tendermint update with prometheus fix

tested with v0.1.4-abciplus and ready

tzemanovic:
we may be able to switch back to official releases instead - see https://github.com/anoma/namada/pull/838

james-chf:
Yes it looks like the only difference now between our fork and upstream `v0.37.0-rc1` is the events endpoint, which we aren't using. https://github.com/tendermint/tendermint/compare/v0.37.0-rc1...heliaxdev:tendermint:ledger-main

So we could switch to using upstream v0.37.0-rc1

PR #671, opened 2022-10-24
add consensus commit timeout

tzemanovic:


tzemanovic:
pls update wasm

tzemanovic:
> this doesn't implicate the tendermint go version, right?

it works with v0.1.3-abciplus - this setting is back in the tendermint's config (they moved it to genesis, but that was reverted - we still have this in "abcipp"), we were just missing it in our tendermint-rs. The consensus timeout settings is used in `ledger_many_txs_in_a_block`, I checked that it's being applied in there

Fraccaman (approved):

PR #676, opened 2022-10-24
Record the block heights at which validators vote for Ethereum events

james-chf:
This PR:
- changes the `seen_by` field that we store for every Ethereum event from `BTreeSet<Address>` to `BTreeMap<Address, BlockHeight>` (`Votes` type)
- propagates this change through the rest of the Ethereum event code that exists in `eth-bridge-integration`

This change to start recording the block heights will be necessary for tallying votes against validator set updates - when a validator set update has enough votes behind it, the relayer should look in the specified `BlockHeight`s for the protocol txs containing the signatures.

### DIFF
--- a/apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
@@ -4,7 +4,7 @@
 mod eth_msgs;
 mod events;
 
-use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use eth_msgs::{EthMsg, EthMsgUpdate};
 use eyre::Result;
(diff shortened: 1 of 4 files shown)


apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -168,9 +174,24 @@ where
...
+            if let Some(already_present_fvp) =
+                votes.insert(address.to_owned(), fvp.to_owned())
+            {
+                tracing::warn!(
+                    ?address,
+                    ?already_present_fvp,
+                    new_fvp = ?fvp,
+                    "Validator voted more than once, arbitrarily using later value",
+                )
+            }
```
sug0, line 191: change `fvp` to `fractional_voting_power`, or even just `voting_power`. the former is somewhat confusing

sug0 (approved):
lgtm

batconjurer (approved):

PR #680, opened 2022-10-25
Multisignature specs

grarco:
Specs addressing #81


tzemanovic (dismissed):

documentation/specs/src/base-ledger/multisignature.md
```diff
@@ -1,3 +1,114 @@
...
+
+Namada does not provide a layer to support this process, so the involved parties will need to rely on an external communication mechanism.
+
+## Replay protection
+
+The [replay protection](./replay-protection.md) mechanism of Namada will prevent third-party malicious users from replaying a transaction having a multisig account as the source. This mechanism, though, is not enough to completely protect multisigned transactions: in this case, in fact, the threat could come from the members of the account themselves.
+
+With the current hash-based replay protection mechanism, once the transaction has been executed, its hash gets stored to prevent a replay. The issue, with a multisig transaction, is that the same transaction with a different set or amount of signatures would have a different hash, meaning that it can be replayed.
```
tzemanovic, line 112: I wonder if it might be worth adding a counter to multisig for this reason

documentation/specs/src/base-ledger/multisignature.md
```diff
@@ -1,3 +1,114 @@
-## k-of-n multisignature
+# k-of-n multisignature
 
-The k-of-n multisignature validity predicate authorises transactions on the basis of k out of n parties approving them.
\ No newline at end of file
+The k-of-n multisignature validity predicate authorizes transactions on the basis of k out of n parties approving them. This document targets the encrypted wasm transactions: at the moment there's no support for multisignature on wrapper or protocol transactions.
```
cwgoes, line 3: We do not need multisignatures for wrapper or protocol transactions, as far as I am aware

cwgoes (changes requested):
Thanks! see comments

documentation/specs/src/base-ledger/multisignature.md
```diff
@@ -1,3 +1,114 @@
...
+- The list of public keys of the signers
+
+`Internal` addresses may want a multi-signature scheme on top of their validation process as well. Among the internal ones, `PGF` will require multisignature for its council (see the [relative](../economics/public-goods-funding.md) spec). The storage data necessary for the correct working of the multisig for an internal address are written in the genesis file: these keys can be later modified through governance.
+
+`Implicit` addresses are not generated by a transaction and, therefore, are not suitable for a multisignature scheme since there would be no way to properly construct them. More specifically, an implicit address doesn't allow for:
+
+- A custom, modifiable VP
+- An initial transaction to be used as an initializer for the relevant data
+
+## Multisig account init validation
```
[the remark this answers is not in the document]
grarco: Yeah in the end we won't apply any kind of validation

documentation/specs/src/base-ledger/multisignature.md
tzemanovic: that's a better idea! We can use the hash of an unsigned tx, which we can reconstruct from a signed tx (we already do this for sig verification)

documentation/specs/src/base-ledger/multisignature.md
```diff
@@ -1,3 +1,114 @@
...
+In the transaction initiating the established address, the submitter will be able to provide a custom VP if the provided one doesn't impose the desired constraints.
+
+## Addresses
+
+The multisig vp introduced in the previous section is available for `established` addresses. To generate a multisig account we provide a new `tx_init_multisig_account` wasm transaction to be used instead of the already available `tx_init_account`. A multisig account can be created by anyone and the creator is responsible for providing the correct data, represented by the following struct:
+
+```rust
+pub struct InitMultiSigAccount {
+    /// The VP code
+    pub vp_code: Vec<u8>,
```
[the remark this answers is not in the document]
tzemanovic: we're doing this with `init-account` that uses vp_user by default too - the default is set in the client, but VPs on-chain enforce whitelist

documentation/specs/src/base-ledger/multisignature.md
grarco: I think we can do both: if we go the counter way we have to make it optional since other transactions don't need it. We can also decide to use the hash of the transaction without the signature, it shouldn't be a problem

documentation/specs/src/base-ledger/multisignature.md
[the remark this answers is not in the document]
cwgoes: Sounds good to me.

cwgoes (approved):
lgtm - thanks!

Fraccaman (approved):

PR #682, opened 2022-10-25
Make the Ethereum events oracle configurable via a channel

james-chf:
Relates to https://github.com/anoma/namada/issues/686

This PR makes it so that:
- the `Oracle` requires the receive half of a command channel to be passed in when it is created
- the `Oracle` doesn't do anything until it has received an `SendConfig { config }` command over this channel

The next PR will pass the send half of the command channel to the `Shell`, which should send a `SendConfig { config }` command based on values in storage


apps/src/lib/node/ledger/ethereum_node/oracle/config.rs
```diff
@@ -0,0 +1,22 @@
...
+//  default config - initialization should always be from storage
+impl std::default::Default for Config {
+    fn default() -> Self {
+        Self {
+            min_confirmations: 10,
+            mint_contract: EthAddress([0; 20]),
+            governance_contract: EthAddress([1; 20]),
+        }
+    }
+}
```
james-chf, line 22: This default config should be moved to a `[cfg(test)]` section in the next PR, as this is only really appropriate for tests.

apps/src/lib/node/ledger/ethereum_node/oracle/mod.rs
```diff
@@ -147,7 +165,22 @@ pub fn run_oracle(
 ///
 /// It also checks that once the specified number of confirmations
 /// is reached, an event is forwarded to the ledger process
-async fn run_oracle_aux(oracle: Oracle) {
+async fn run_oracle_aux(mut oracle: Oracle) {
+    tracing::info!("Oracle is awaiting initial configuration");
+    let config = match await_initial_configuration(&mut oracle.control).await {
```
james-chf, line 170: `config` is not a field on the `Oracle` because as we don't have an initial configuration when constructing it, such a field ends up being an `Option<Config>` (or equivalently some sort of dummy config that doesn't completely make sense in production, like we had before). Then in `process()`, we end up having to do things like `oracle.config.min_confirmations.unwrap()` even though we know by the point that we reach `process` that the oracle must have been configured.

Configuring the oracle after it has been constructed via a channel feels a bit overwrought, I think ideally the `Shell` would construct the `Oracle` with an initial `Config` as from storage (or not construct it at all if an oracle wasn't necessary) - and then we could just have a `config` field on the oracle that wasn't an `Option`.

james-chf:
pls update wasm

apps/src/lib/node/ledger/ethereum_node/oracle/config.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Config {
+    /// The minimum number of block confirmations an Ethereum block must have
+    /// before it will be checked for bridge events.
+    pub min_confirmations: u64,
+    /// The Ethereum address of the current bridge contract.
+    pub mint_contract: EthAddress,
+    /// The Ethereum address of the current governance contract.
+    pub governance_contract: EthAddress,
+}
```
sug0, line 14: I still think these parameters should be configured at genesis, and not through ledger config files

apps/src/lib/node/ledger/ethereum_node/oracle/config.rs
```diff
@@ -0,0 +1,26 @@
+//! Configuration for an oracle.
+use namada::types::ethereum_events::EthAddress;
+
+/// Configuration for an oracle.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Config {
+    /// The minimum number of block confirmations an Ethereum block must have
+    /// before it will be checked for bridge events.
+    pub min_confirmations: u64,
```
sug0, line 9:
```suggestion
    pub min_confirmations: NonZeroU64,
```
otherwise, we accept forged events that aren't even recorded on an ethereum block by setting `min_confirmations = 0`

apps/src/lib/node/ledger/ethereum_node/oracle/mod.rs
sug0: if we switch from the `futures::try_join!()` macro to `futures::future::try_join_all()`, we will end up with a more malleable API that will allow us to filter out sub-processes which aren't needed.
```rs
let oracle = None;
let sub_processes = vec![..., oracle].into_iter().flatten();
let results: Vec<Result<_, _>> = futures::future::try_join_all(sub_processes);
```
we could construct `Some` oracle future based on a config param, or return `None` otherwise

apps/src/lib/node/ledger/ethereum_node/oracle/mod.rs
```diff
@@ -359,14 +398,34 @@ mod test_oracle {
...
+    /// initializes it with a simple default configuration that is appropriate
+    /// for tests.
+    async fn start_with_default_config(
+        oracle: Oracle,
+        control_sender: control::Sender,
+        config: Config,
+    ) -> JoinHandle<()> {
+        let handle = std::thread::spawn(move || {
+            tokio_test::block_on(run_oracle_aux(oracle));
+        });
```
sug0, line 416: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

apps/src/lib/node/ledger/mod.rs
```diff
@@ -235,8 +236,21 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
...
+    if let Some(oracle_control_sender) = oracle_control_sender {
+        if let Err(error) = oracle_control_sender
+            .send(oracle::control::Command::SendConfig {
+                config: oracle::config::Config::default(),
+            })
+            .await
+        {
+            tracing::error!(?error, "Could not configure the oracle",);
+        }
+    }
```
sug0, line 253: this step can be done inside `maybe_start_ethereum_oracle()`, no need to pollute `run_aux()`, to avoid further chaos like we had before

sug0 (changes requested):

apps/src/lib/node/ledger/ethereum_node/oracle/config.rs
james-chf: They will be, this is just the struct to hold them for an `Oracle`. So these fields will be populated by the values from genesis/blockchain storage (though that isn't happening just yet in this PR)

apps/src/lib/node/ledger/mod.rs
james-chf: I've moved it into `maybe_start_ethereum_oracle` and linked the TODO to https://github.com/anoma/namada/issues/686 , note this block will be temporary in any case and the `oracle_control_sender` channel will be passed down to the `Shell` in a later PR - then this block can be remove

apps/src/lib/node/ledger/mod.rs
```diff
@@ -609,42 +610,58 @@ fn start_tendermint(
...
+
+            // TODO(namada#686): pass `oracle_control_sender` to the shell for
+            // initialization from storage, rather than using a
+            // hardcoded config
+            if let Err(error) = control_sender.blocking_send(
+                oracle::control::Command::SendConfig {
+                    config: oracle::config::Config::default(),
+                },
+            ) {
+                tracing::error!(?error, "Could not configure the oracle",);
```
batconjurer, line 650: This is an unrecoverable error. We should panic.

apps/src/lib/node/ledger/mod.rs
```diff
@@ -609,42 +610,58 @@ fn start_tendermint(
...
-/// An oracle is also returned, along with its associated channel,
-/// for receiving Ethereum events from `geth`.
-async fn maybe_start_ethereum_oracle(
+/// Potentially starts an Ethereum event oracle.
+fn maybe_start_ethereum_oracle(
     config: &config::Ledger,
     abort_sender: oneshot::Sender<()>,
-) -> (Option<mpsc::Receiver<EthereumEvent>>, task::JoinHandle<()>) {
+) -> (
+    Option<mpsc::Receiver<EthereumEvent>>,
```
batconjurer, line 618: I noticed that  we don't have a return pattern of the following `(None, Some(_), handle)`. So this return is not well typed. We should probably return an enum instead with names reflecting the dfiferent modes

batconjurer (changes requested):

apps/src/lib/node/ledger/mod.rs
james-chf: Yeah an enum is way better here

james-chf:
pls update wasm

apps/src/lib/node/ledger/ethereum_node/oracle/mod.rs
sug0: you can use something like this https://github.com/anoma/namada/blob/04cf854bafbd343ee2bd93d3436879cbe80ee5b9/apps/src/lib/node/ledger/ethereum_node/oracle.rs#L120-L142

sug0: at least until we get rid of `web30`

apps/src/lib/node/ledger/ethereum_node/oracle/config.rs
batconjurer: I have refactored these params as part of #712.

batconjurer (approved):
In general I'm good with this. I think it should be adjusted in context of PR #712.

james-chf:
I've just force pushed a rebase onto `eth-bridge-integration` including `v0.8.1`

james-chf:
pls update wasm

james-chf:
I've force pushed a rebase onto latest `eth-bridge-integration` (i.e. namada v0.9.0)

sug0 (approved):

PR #687, opened 2022-10-25
Remove staking rewards address

### LINKED ISSUE #498: Remove validator staking reward address
With a new implementation of the PoS rewards that follows the F1 Fee mechanism, the validator staking reward address and its associated public key are no longer needed.

Rather than distributing rewards every epoch from the PoS account to the staking reward address, an epoched list of reward products, with values parameterized as `(1 + fractional_reward)` will be held in storage, while the tokens will remain held in the PoS account. The rewards are applied upon withdrawal of tokens by a validator or delegator, and the number of tokens credited will take into account the accumulated rewards.

brentstone:
Closes #498


brentstone:
pls update wasm

tzemanovic (dismissed):
There were couple little left-overs in the validator pre-genesis wallet and some strings - I pushed couple more commits to remove those - besides that LGTM

brentstone:
pls update wasms

tzemanovic (approved):

Issue #641, opened 2022-10-18
`wrong Block.Header.AppHash` error after a transfer was made


james-chf:
Namada commit: `75a7e12c3d7ba3063881c4c48a23810469903edf`
Tendermint hash: `0b58342a`
Chain ID: `namada-75a7e12.69483d59a9fb174`
OS/arch: macOS 12.6 on M1 (ARM)

---

Full logs of first sync: [ledger.log.txt](https://github.com/anoma/namada/files/9812337/ledger.log.txt)

These logging related environment variables were set
```shell
export ANOMA_LOG='debug'
export ANOMA_LOG_COLOR='false'
export ANOMA_TM_STDOUT='true'

# tendermint is noisy
export TM_LOG_LEVEL='info'
```

---

Starting and syncing a normal fullnode (not genesis validator) progressed ok, around block height 1668 I submitted a faucet transfer to the local fullnode doing something like:

```shell
# local node's RPC endpoint
TENDERMINT_RPC_ADDRESS='127.0.0.1:26657'
curl "http://${TENDERMINT_RPC_ADDRESS}/status" \
  -H 'Accept: application/json' \
  -m 5 | jq

# arbitrary random address isn't in our wallet
ARBITRARY_ADDRESS='atest1d9khqw36xv6yxsesg5mnsdekxaznxs3ngs6yyd3cg9pnyw2rxqmnxsf4gyeyydjyg3qnzwzpu06kv7'

# generate an implicit account in our wallet
ALIAS="alias-$(openssl rand -hex 6)"
namadaw address gen \
  --alias ${ALIAS} \
  --unsafe-dont-encrypt

# get XAN from faucet for the random address (we can initiate this using any signer)
namadac transfer \
  --ledger-address $TENDERMINT_RPC_ADDRESS \
  --token XAN \
  --amount 1000 \
  --source faucet \
  --target ${ARBITRARY_ADDRESS} \
  --signer ${ALIAS}

namadac balance
```

These are the node logs from the first occurence of `height=1667` to first occurrence of `height=1669`. Block 1668 seems to be committed with hash `A291A73F43D04648C69A79BEAEF83066C8A57E6E87ADF36CE6DF06AE6A96E0DA`.

```
I[2022-10-18|16:16:28.658] Timed out                                    module=consensus dur=921.969ms height=1667 round=0 step=RoundStepNewHeight
[...]

PR #689, opened 2022-10-26
Backport shim fix to eth-bridge-integration

james-chf:
Fixes https://github.com/anoma/namada/issues/641 for `eth-bridge-integration`

This is a backport of https://github.com/anoma/namada/pull/656 which is in namada v0.8.1 (backporting now so that merging v0.8.1 into `eth-bridge-integration` is easier - 
Relates to https://github.com/anoma/namada/issues/673)

We had removed `Shell::process_txs` in `eth-bridge-integration`, which is now being used by the shim, so had to add something like it back - `Shell::process_proposed_txs`.


apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -131,12 +131,23 @@ impl AbcippShim {
...
+                    self.delivered_txs.push(tx.tx);
+                    Ok(Resp::DeliverTx(Default::default()))
+                }
                 #[cfg(not(feature = "abcipp"))]
                 Req::EndBlock(_) => {
-                    use namada::types::storage::BlockHash;
-                    let mut txs = vec![];
-                    std::mem::swap(&mut txs, &mut self.processed_txs);
+                    let (processing_results, _) =
+                        self.service.process_proposed_txs(&self.delivered_txs);
```
james-chf, line 141: This is the only sort of irrelevant line that should differ between `eth-bridge-integration` and `v0.8.1` now - in `eth-bridge-integration` we need to return a `DigestCounters` as part of `process_txs`/`process_proposed_txs`

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -128,6 +115,31 @@ where
         }
     }
 
+    /// Calculates what the [`TxResult`]s would be for the transactions in a
+    /// proposed block, as well as counting the number of digest transactions
+    /// present. `ProcessProposal` should be able to make a decision on whether
+    /// a proposed block is acceptable or not based solely on what this function
+    /// returns.
+    pub fn process_proposed_txs(
```
james-chf, line 123: Suggest we take the opportunity to rename this fn (and `process_single_tx`), since we are changing the signatures of these two fns anyway

james-chf:
pls update wasm

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -128,6 +115,31 @@ where
         }
     }
 
+    /// Calculates what the [`TxResult`]s would be for the transactions in a
+    /// proposed block, as well as counting the number of digest transactions
+    /// present. `ProcessProposal` should be able to make a decision on whether
+    /// a proposed block is acceptable or not based solely on what this function
+    /// returns.
```
sug0, line 122:
```suggestion
    /// Calculates what the [`TxResult`]s would be for the transactions in a
    /// proposed block, and counts the number of digest transactions
    /// present. 
```
leave the last part to `vote_on_proposal`, with a bit of clean up:
```
    /// Decides whether `ProcessProposal` should accept or reject
    /// a block proposed by some Tendermint round's leader.
    ///
    /// We first call `check_tx_status()`, to check for non-recoverable
    /// transaction errors.
```

sug0 (changes requested):

apps/src/lib/node/ledger/shell/process_proposal.rs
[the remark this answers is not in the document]
james-chf: Agreed, technically we could "process" txs during `ProcessProposal` but in our ABCI application we don't do this which keeps things simpler. `ProcessProposal` to us is really more like `CheckProposal`

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -43,27 +43,14 @@ where
...
-                    tx_bytes,
-                    &mut tx_queue_iter,
-                    &mut counters,
-                )
-            })
-            .collect();
+        let (tx_results, counters) = self.process_proposed_txs(&req.txs);
 
         // We should not have more than one `ethereum_events::VextDigest` in
         // a proposal from some round's leader.
```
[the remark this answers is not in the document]
james-chf: agree though it ends being a bit more of a refactor so would rather not do it in this PR. Specifically we currently log the reason for rejecting a proposed block, with context from the `req` struct, which we wouldn't be able to do in a `fn vote_on_proposal` without also passing in `req` or returning the reason (e.g. via an enum), I think we should go with something like the latter

apps/src/lib/node/ledger/shell/process_proposal.rs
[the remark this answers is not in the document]
james-chf: renamed to `check_proposal` and `check_proposal_tx`

apps/src/lib/node/ledger/shell/process_proposal.rs
james-chf: unsure about this suggestion, if we don't create a `vote_on_proposal` fn in this PR

batconjurer (approved):

sug0 (approved):

PR #690, opened 2022-10-26
Refactor `get_voting_powers` when applying protocol tx state changes

sug0:
Based on #684

This PR refactors `get_voting_powers` to accept different kinds of items that have been voted on (e.g. ethereum events and validator set updates).

### DIFF
--- a/apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
@@ -18,12 +18,20 @@ use namada::types::vote_extensions::ethereum_events::MultiSignedEthEvent;
 use namada::types::voting_power::FractionalVotingPower;
 
 use super::ChangedKeys;
-use crate::node::ledger::protocol::transactions::utils::{
-    self, get_active_validators,
-};
+use crate::node::ledger::protocol::transactions::utils;
 use crate::node::ledger::protocol::transactions::votes::{
     calculate_new, calculate_updated, write, Votes,
 };
+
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -18,13 +18,21 @@ use namada::types::vote_extensions::ethereum_events::MultiSignedEthEvent;
...
 use super::ChangedKeys;
-use crate::node::ledger::protocol::transactions::utils::{
-    self, get_active_validators,
-};
+use crate::node::ledger::protocol::transactions::utils;
 use crate::node::ledger::protocol::transactions::votes::{
     calculate_new, calculate_updated, write, Votes,
 };
 
+impl utils::GetVoters for [MultiSignedEthEvent] {
```
batconjurer, line 26: Do we prefer `HashSet` over `BTreeSet`?

batconjurer (approved):
One question of little importance.

apps/src/lib/node/ledger/protocol/transactions/ethereum_events/mod.rs
sug0: I generally prefer hashsets, unless there's an obvious reason to pick btreesets. here iteration order doesn't matter, so I went with what james was already using (a hashset)

james-chf: Agree `HashSet` is preferable where ordering isn't important as it communicates that

james-chf (approved):

PR #692, opened 2022-10-26
Add checks that ERC20 tokes are escrowed

batconjurer:
When a transfer is add to the Ethereum bridge pool, it should escrow the relevant ERC20 tokens in the bridge pools storage. This PR adds checks to the VP that this is done correctly.


shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -116,11 +118,29 @@ where
...
+        // check that transfer is not already in the pool
+        match (&self.ctx).read_pre(&pending_key) {
+            Ok(Some(_)) => {
+                tracing::debug!(
+                    "Rejecting transaction as the transfer is already in the \
+                     Ethereum bridge pool."
+                );
+                return Ok(false);
+            }
+            Err(_) => {
```
james-chf, line 130: can the returned error wrap this error, or can we log it if we will not propagate it (at least at debug?)

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -164,14 +184,50 @@ where
...
+                (&escrow_key).try_into().expect("This should not fail"),
+                (&owner_key).try_into().expect("This should not fail"),
+            ) {
+                Ok(Some(delta))
+                    if delta
+                        == (
+                            transfer.transfer.sender,
+                            transfer.transfer.amount,
+                        ) => {}
+                _ => {
```
james-chf, line 215: It could be really useful to log this error

james-chf (approved):
lgtm, there are some potential errors that could happen that would be useful to log or otherwise handle

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -116,11 +118,30 @@ where
...
+            }
+            Err(e) => {
+                return Err(eyre!(
+                    "Could not read the storage key associated with the \
+                     transfer: {:?}",
+                    e
+                )
+                .into());
+            }
+            _ => {}
```
james-chf, line 138: Noticed this while reviewing https://github.com/anoma/namada/pull/715 but the change is coming from this PR I guess

We would match here only for the `Ok(None)` case which means the `pending_key` got deleted(?), I think we should return an `Ok(false)` here if someone tries to do this via a wasm transaction. We should be deleting pending keys from the pool in derived transactions only, once we receive a `TransfersToEthereum` event, if I understand correctly

sug0 (changes requested):

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
batconjurer: I replied to this comment on #715.

sug0:
the heck is wrong with the diff?

batconjurer:
> the heck is wrong with the diff?

IKR? I think it's because it depends on another PR. Once the base PR is merged, I'm hoping a rebase will clean it up.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
batconjurer: Well, the reply disappeared. This means that the pending key was not used in storage. It may have been deleted (the transfer timed out), or not. We actually don't care.

PR #694, opened 2022-10-26
`TestTxEnv::spawn_accounts` should ignore internal addresses

james-chf:
This test helper writes an empty wasm VP to the `?` subkey for any passed in addresses. We shouldn't need to do this for internal (and maybe also implicit?) addresses, since they shouldn't have a VP in storage. The motivation for this PR is that in `eth-bridge-integration`, there will be an internal account (`EthBridgePool`) which uses a different Merkle tree to other accounts and cannot have arbitrary bytes written to its `?` subkey.


james-chf:
@tzemanovic does this change make sense in terms of accounts? Maybe this fn should ignore implicit accounts as well, in which case it could just accept `EstablishedAddress`es. All wasm tests are passing with this change.

tzemanovic:
> does this change make sense in terms of accounts? Maybe this fn should ignore implicit accounts as well, in which case it could just accept `EstablishedAddress`es. All wasm tests are passing with this change.

Yeah, I think this is fine and we can also ignore implicit accounts. I prefer this over changing type to `EstablishedAddress` as it's simpler ergonomics on caller side for a test helper and I don't think there's much danger to it.

The reason we have this helper is for wasm is `fn check_address_existence` (`shared/src/vm/host_env.rs`) which already ignores implicit and internal accounts. On `Storage` there's also a method `fn exists` which should do the same, but it only works with established address and it's not being used at the moment.

tests/src/vm_host_env/tx.rs
```diff
@@ -111,14 +111,20 @@ impl TestTxEnv {
...
-    /// pass account existence check in `tx_write` function.
+    /// pass account existence check in `tx_write` function. Any internal
+    /// addresses ([`Address::Internal`]) passed in are ignored, as those
+    /// should not have wasm VPs in storage in any case.
     pub fn spawn_accounts(
         &mut self,
         addresses: impl IntoIterator<Item = impl Borrow<Address>>,
     ) {
         for address in addresses {
+            if matches!(address.borrow(), Address::Internal(_)) {
```
tzemanovic, line 124:
```suggestion
            if matches!(address.borrow(), Address::Internal(_) | Address::Implicit(_)) {
```

tzemanovic: (with docs updates : )

james-chf: :+1:

james-chf:
Great, have updated to ignore implicit accounts while still accepting `Address`

tzemanovic (approved):

PR #695, opened 2022-10-26
Validator commission rates

brentstone:


brentstone:
pls update wasm

brentstone:
pls update wasm

brentstone:
pls update wasm

brentstone:
pls update wasms

brentstone:
pls update wasms

apps/src/lib/client/rpc.rs
```diff
@@ -975,6 +976,63 @@ pub async fn query_voting_power(ctx: Context, args: args::QueryVotingPower) {
...
+    let epoch = match args.epoch {
+        Some(epoch) => epoch,
+        None => query_epoch(args.query.clone()).await,
+    };
+    let client = HttpClient::new(args.query.ledger_address).unwrap();
+
+    match args.validator {
+        Some(validator) => {
+            let validator = ctx.get(&validator);
+            let validator_commission_key =
```
tzemanovic, line 993: I think we could add a call to `is_validator` here to check first for a bit nicer error handling when the given address is not a validator

apps/src/lib/client/rpc.rs
```diff
@@ -975,6 +976,63 @@ pub async fn query_voting_power(ctx: Context, args: args::QueryVotingPower) {
...
+                    println!(
+                        "No commission rate found for {} in epoch {}",
+                        validator.encode(),
+                        epoch
+                    )
+                }
+            }
+        }
+        None => {
+            println!("No validator found from the args")
```
tzemanovic, line 1031: I think better to make the `args.validator` required instead of `Option`

proof_of_stake/src/epoched.rs
```diff
@@ -128,13 +128,16 @@ pub enum DynEpochOffset {
     PipelineLen,
     /// Offset at unbonding length.
     UnbondingLen,
+    /// Offset at pipeline length - 1.
+    PipelineLenMinusOne,
```
tzemanovic, line 132: It's not a biggie, but maybe we could just call this case `One` and set it to `1`? Pipeline is always going to stay at `2`

proof_of_stake/src/parameters.rs
```diff
@@ -177,4 +178,11 @@ pub mod testing {
             }
         }
     }
+
+    /// Get an arbitrary rate - a Decimal value between 0 and 1 inclusive, with
+    /// some fixed precision
+    pub fn arb_rate() -> impl Strategy<Value = Decimal> {
+        (0..100_001_u64)
```
tzemanovic, line 185:
```suggestion
        (0..=100_000_u64)
```
special syntax for upper bound inclusive

wasm/wasm_source/src/tx_change_validator_commission.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        init_pos(&genesis_validators[..], &pos_params, Epoch(0));
+
+        let tx_code = vec![];
+        let tx_data = commission_change.try_to_vec().unwrap();
+        let tx = Tx::new(tx_code, Some(tx_data));
+        let signed_tx = tx.sign(&key);
+        let tx_data = signed_tx.data.unwrap();
+
+        println!("\ndbg0\n");
```
tzemanovic, line 84: let's clear up some of these unless they're useful for later

proof_of_stake/src/validation.rs
```diff
@@ -1469,6 +1494,62 @@ where
...
+    fn validator_commission_rate(
+        constants: &Constants,
+        errors: &mut Vec<Error<Address, TokenChange, PublicKey>>,
+        new_validators: &mut HashMap<Address, NewValidator<PublicKey>>,
+        address: Address,
+        data: Data<CommissionRates>,
+    ) {
+        match (data.pre, data.post) {
+            // Should this case ever happen since commission rate should be init
+            // at genesis? (same w consensus key tho)
```
tzemanovic, line 1506: yeah, only genesis validators will have this set - validators created later via a tx will go through this validation

shared/src/ledger/pos/vp.rs
```diff
@@ -259,6 +262,18 @@ where
...
                 });
                 changes.push(TotalVotingPower(Data { pre, post }));
+            } else if let Some(address) = is_validator_commission_rate_key(key)
+            {
+                let pre = self.ctx.pre().read_bytes(key)?.and_then(|bytes| {
+                    CommissionRates::try_from_slice(&bytes[..]).ok()
+                });
+                let post = self.ctx.post().read_bytes(key)?.and_then(|bytes| {
+                    CommissionRates::try_from_slice(&bytes[..]).ok()
+                });
```
tzemanovic, line 272: for validation, we can look-up `max_commission_rate_change` here (similarly to bond/unbond looking up the slashes here)

proof_of_stake/src/epoched.rs
brentstone: The only counterpoint to this would be in tests where we choose arbitrary PoS params - if in those tests we want to ultimately test txs to change the commission rate, then this should stay the same. But otherwise, I think this works too.

apps/src/lib/client/rpc.rs
brentstone: Thx, makes sense.

apps/src/lib/client/rpc.rs
brentstone: Thx, sounds good.

proof_of_stake/src/parameters.rs
brentstone: Thx!

proof_of_stake/src/epoched.rs
tzemanovic: ah, good point, let's keep it as is then!

proof_of_stake/src/validation.rs
```diff
@@ -1469,6 +1494,62 @@ where
         }
     }
 
+    fn validator_commission_rate(
```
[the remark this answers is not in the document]
brentstone: Thx, btw is there also some way that we should ensure that this tx only comes from the validator address to which it applies? Is this already done by construction with a signature, etc?

shared/src/ledger/pos/vp.rs
brentstone: Should we look this up from `pre` and `post` and ensure that it is unchanged as well?

apps/src/lib/client/tx.rs
```diff
@@ -975,6 +976,90 @@ pub async fn submit_withdraw(ctx: Context, args: args::Withdraw) {
...
+    let epoch = rpc::query_epoch(args::Query {
+        ledger_address: args.tx.ledger_address.clone(),
+    })
+    .await;
+
+    let tx_code = ctx.read_wasm(TX_CHANGE_COMMISSION_WASM);
+    let client = HttpClient::new(args.tx.ledger_address.clone()).unwrap();
+
+    // Check that the submitter of the tx is a validator
+    match ctx.wallet.get_validator_data() {
```
tzemanovic, line 992: This works, but I think it would be little better to use `let validator = ctx.get(&args.validator); if rpc::is_validator(&validator) { ...` instead to respect the provided argument. In vast majority of cases (if not all?) they will be the same address though

tests/src/native_vp/pos.rs
```diff
@@ -401,7 +401,10 @@ mod tests {
                         address,
                         consensus_key,
                         commission_rate: _,
+                        max_commission_rate_change: _,
                     } => {
+                        // TODO: should there be preconditions for commission
+                        // rates here?
```
tzemanovic, line 407: I don't think we need any - a new validator can choose any rates in `0..=1` and we're only generating values in this range. 

We can however add a `Transition` for `ValidPosAction::ChangeCommissionRate`, for which the model would need to have a pre-condition that ensures that the new rate is less than or equal to the validator's `max_commission_rate_change` with respect to its last epoch's rate. I think this can be done in in the abstract SM by e.g. looking through `self.all_valid_actions().into_iter().rev()` for the last rate value (either from a previous `ValidPosAction::ChangeCommissionRate` or `ValidPosAction::InitValidator` if it's not been changed before) before the last epoch switch. However, the epoch switch is currently not preserved in the abstract state machine - we only have the current `epoch`, so we might need to change that first.

wasm/wasm_source/src/vp_validator.rs
```diff
@@ -0,0 +1,759 @@
...
+                        bond_id.source != addr || *valid_sig
+                    }
+                    None => {
+                        // Any other PoS changes are allowed without signature
+                        true
+                    }
+                };
+                let comm =
+                    proof_of_stake::is_validator_commission_rate_key(key);
+                let valid_commission_rate_change = match comm {
```
tzemanovic, line 140:
```suggestion
                // Validator's commission rate change must be signed
                let valid_commission_rate_change = match comm {
```

apps/src/lib/client/tx.rs
brentstone: Ok, yeah I was also second-guessing whether we even want `validator` to be an arg in `args::TxCommissionRateChange` since in all cases, this address should be the same as the one of the tx submitter. I will change, though.

tzemanovic:
pls update wasm

tzemanovic (approved):

PR #698, opened 2022-10-27
documentation changes

bengtlofgren:


documentation/docs/src/quick-start.md
```diff
@@ -34,26 +41,37 @@ namada ledger
...
 
-Generate a local key on disk
+Accounts on Namada are divided into two subcategories:
+* Implicit accounts (no attached Validity Predicate)
+* Established accounts (attached Validity Predicate)
 
-```shell
-# first, we make a keypair and the implicit account associated with it
-# namadaw address gen instead of key gen. Preferred because they both make a keypair but the former stores the implicit address for it too 
+In order to make transactions on Namada, an implicit account is required.
```
james-chf, line 58:
```suggestion
In order to make transactions on Namada, an account is required, and an implicit account can be created easily.
```
someone can make transactions using an established account as well (afaik?)

documentation/docs/src/quick-start.md
```diff
@@ -34,26 +41,37 @@ namada ledger
...
   --alias example-implicit
 
 ➜ Enter encryption password: 
 Successfully added a key and an address with alias: "example-implicit"
 
-# use it to make an established account
 ```
+An implicit account with alias "example-implicit" has now been successfully generated and stored locally in the wallet.
+
+It is possible to "upgrade" this implicit account to become an Established Account through submitting a transaction to do so. Note that this transaction, as with any transaction on Namada, will consume "gas", and will require a positive balance of NAM tokens.
```
james-chf, line 72: idk if "upgrade" would be the best term here, you can create an established account using an implicit account, but the implicit account still remains in existence in your wallet, there is no replacement of it.

james-chf, line 72:
```suggestion
It is possible to "upgrade" this implicit account to become an Established Account through submitting a transaction to do so. Note that this transaction, as with any transaction on Namada, will consume "gas", and will require a positive balance of NAM tokens for the implicit account which is being used to create the established account.
```

james-chf: btw do you know if implicit accounts are charged gas now? i.e. when an implicit account makes a transaction, is NAM actually deducted from its balance?

documentation/docs/src/user-guide/ledger/governance.md
```diff
@@ -43,9 +43,9 @@ Now, we need to create a json file `proposal.json` holding the content of our pr
 You should change the value of:
 
 - `Author` field with the address of `my-new-acc`.
-- `voting_start_epoch` with future epoch (must be a multiple of 3)
-- `voting_end_epoch` with an epoch greater of `voting_start_epoch`, multiple of 3.
-- `grace_epoch` with an epoch greater of `voting_end_epoch` + 6
+- `voting_start_epoch` with a future epoch (must be a multiple of 3) for which you want the voting to begin
+- `voting_end_epoch` with an epoch greater of `voting_start_epoch`, a multiple of 3, and by which no further votes will be accepted
```
james-chf, line 47:
```suggestion
- `voting_end_epoch` with an epoch greater than `voting_start_epoch`, a multiple of 3, and by which no further votes will be accepted
```

documentation/docs/src/user-guide/ledger/governance.md
```diff
@@ -43,9 +43,9 @@ Now, we need to create a json file `proposal.json` holding the content of our pr
...
+- `grace_epoch` with an epoch greater of `voting_end_epoch` + 6, in which the proposal, if passed, will come into effect
```
james-chf, line 48:
```suggestion
- `grace_epoch` with an epoch greater than `voting_end_epoch` + 6, in which the proposal, if passed, will come into effect
```

documentation/docs/src/user-guide/wallet.md
```diff
@@ -2,6 +2,9 @@
 
 This document describes the different wallet concepts and options that are available to users of Namada who want to be able to [send, receive and interact](#send-and-receive-nam-tokens) with NAM tokens on the Namada blockchain.
 
+
+<!-- I want to either hyperlink or delete the below. I don't understand-->
+
```
james-chf, line 7: I would remove it as right now only the file wallet is used (afaik?) but would double check. Otherwise we have things in our documentation that are not in a release of namada (yet)

james-chf (changes requested):

documentation/docs/src/quick-start.md
bengtlofgren: Very good point actually

documentation/docs/src/user-guide/getting-started.md
```diff
@@ -15,9 +15,15 @@ The main binary `namada` has sub-commands for all of the other binaries:
 
 To explore the command-line interface, add `--help` argument at any sub-command level to find out any possible sub-commands and/or arguments.
 
+```admonish tip title="Adding binaries to path" collapsible=true
+The binaries should be added to `$PATH` from the `make install` command. However, if this for some reason did not work, a solution may be to copy the binaries from `namada/target/release` to `home/$USER/.local/bin/` for example:
```
james-chf, line 19: imo we should just link to the single source of truth installation guide, rather than including more installation instructions inline on this page. Same applies to the "Quickstart" guide which also mentions some installation instructions.

(These specific instructions here are OS dependent (this wouldn't work on macOS, I don't think) and require the user to know to `make build-release`/how to build release Rust binaries or similar first.)

james-chf (dismissed):

documentation/docs/src/quick-start.md
```diff
@@ -34,26 +41,37 @@ namada ledger
...
+```
+
 This should sync your node to the ledger and will take a while (depending on your computer). Subsequent commands (generating an account, etc.)  are unlikely to work until it is fully synced. Enquire the current block height with other participants to make sure you are synced in order to proceed.
 
 ## Account
 
-Generate a local key on disk
+Accounts on Namada are divided into two subcategories:
+* Implicit accounts (no attached Validity Predicate)
+* Established accounts (attached Validity Predicate)
```
tzemanovic, line 56:
```suggestion
* Implicit accounts (all use the same Validity Predicate)
* Established accounts (can use a chosen Validity Predicate)
```

tzemanovic (approved):

PR #702, opened 2022-10-27
fix: namadan should log at info by default

james-chf:


apps/src/bin/anoma-node/main.rs
```diff
@@ -11,7 +11,7 @@ fn main() -> Result<()> {
     color_eyre::install()?;
 
     // init logging
-    let default_directive = Directive::from_str("anoma=info")?;
+    let default_directive = Directive::from_str("namada=info")?;
```
james-chf, line 14: alternatively we could just do

```rust
    logging::init_from_env_or(LevelFilter::INFO)?;
```

like `namadac` and `namada` do

james-chf: :+1: have changed this to be `logging::init_from_env_or(LevelFilter::INFO)?` as well

tzemanovic:
strange flakiness in CI - for the e2e test https://github.com/anoma/namada/actions/runs/3346594039/jobs/5543588172, it failed on 

```
    0: Reached a timeout for expect type of command
       Needle: Tendermint node started
```
but I can see it in the logs?!

tzemanovic (approved):
thanks!

PR #703, opened 2022-10-27
Bridge Pool VP allows minting wNam

batconjurer:
In order to mint wNam on Ethereum, a transfer added to the bridge pool must escrow Nam under the Ethereum bridge VP. The bridge pool VP should check this.

Depends on #692


shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -77,6 +80,44 @@ where
...
+            }
+        } else {
+            tracing::debug!(
+                "The Ethereum bridge pool's escrow was not credited from \
+                 account {}.",
+                payer_account
+            );
+            return false;
+        }
+        true
```
james-chf, line 119: If any `self.account_balance_delta` return `None`s here, we would reach this `true`

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -131,47 +190,89 @@ where
...
+                &transfer.gas_fee.payer,
+                &BRIDGE_POOL_ADDRESS,
+                transfer.gas_fee.amount + transfer.transfer.amount,
+                transfer.gas_fee.amount,
+            ) || !self.check_nam_escrowed(
+                &transfer.transfer.sender,
+                &Address::Internal(InternalAddress::EthBridge),
+                transfer.gas_fee.amount + transfer.transfer.amount,
+                transfer.transfer.amount,
+            ) {
```
james-chf, line 220: If someone adds a `PendingTransfer` to the pool, the gas fee payer should have their account debited by `transfer.gas_fee.amount + transfer.transfer.amount`, of which `transfer.gas_fee.amount` goes to the bridge pool. The transfer sender should also have their account debited by `transfer.gas_fee.amount + transfer.transfer.amount`, and `transfer.transfer.amount` should go to the Eth bridge VP (which is acting as the escrow amount here). Shouldn't we be checking `transfer.gas_fee.amount` is debited from the `transfer.gas_fee.payer`, and `transfer.transfer.amount` is debited from `transfer.transfer.sender`, separately?

james-chf (changes requested):

james-chf:
This PR looks like it is based off https://github.com/anoma/namada/pull/692

apps/src/lib/config/ethereum_bridge/params.rs
```diff
@@ -42,7 +43,7 @@ pub struct GenesisConfig {
 pub struct Contracts {
     /// The Ethereum address of the ERC20 contract that represents this chain's
     /// native token.
-    pub native_erc20: Address,
+    pub native_erc20: EthAddress,
```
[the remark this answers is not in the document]
batconjurer: It's done in #712

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
batconjurer: Ooh, yeah, there is an issue here. Need to check if fee payer and gas payer are the same are not. Good catch.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
batconjurer: true, easy fix

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,56 @@ where
...
+where
+    DB: 'static + ledger_storage::DB + for<'iter> ledger_storage::DBIter<'iter>,
+    H: 'static + StorageHasher,
+    CA: 'static + WasmCacheAccess,
+{
+    /// If the bridge's escrow key was changed, we check
+    /// that the balance increased and that the bridge pool
+    /// VP has been triggered. The bridge pool VP will carry
+    /// out the rest of the checks.
+    fn check_escrow(&self, verifiers: &BTreeSet<Address>) -> bool {
```
james-chf, line 41: Would suggest return `Result<bool>` here as well. Specifically where we're currently doing `.expect`, we should propagate an `Err`, also where we can't retrieve the `escrow_key` value.

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,56 @@ where
...
+                "Could not retrieve the Ethereum bridge VP's balance from \
+                 storage"
+            );
+            return false;
+        };
+        let escrow_post: Amount = if let Ok(Some(bytes)) =
+            self.ctx.read_pre(&escrow_key)
+        {
+            BorshDeserialize::try_from_slice(bytes.as_slice())
+                .expect("Deserializing a balance from storage shouldn't fail")
```
james-chf, line 59: I guess someone could try to write garbage to this key, and it would cause the VP to panic

james-chf (changes requested):

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,56 @@ where
...
+        } else {
+            tracing::debug!(
+                "Could not retrieve the Ethereum bridge VP's balance from \
+                 storage"
+            );
+            return false;
+        };
+        let escrow_post: Amount = if let Ok(Some(bytes)) =
+            self.ctx.read_pre(&escrow_key)
```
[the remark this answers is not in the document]
batconjurer: I wrote a test that caught this in a branch I'm currently working on.

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,56 @@ where
...
+
+        // The amount escrowed should increase.
+        if escrow_pre < escrow_post {
+            verifiers.contains(&storage::bridge_pool::BRIDGE_POOL_ADDRESS)
+        } else {
+            tracing::info!(
+                "A normal tx cannot decrease the amount of Nam escrowed in \
+                 the Ethereum bridge"
+            );
+            false
```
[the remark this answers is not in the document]
batconjurer: In theory some one could overwrite the same value and it would then show up in pre and post reads. Escrowing 0 tokens serves no purpose, so I decided to not allow it.

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -67,11 +119,18 @@ where
             verifiers_len = verifiers.len(),
             "Ethereum Bridge VP triggered",
         );
+
+        // first check if Nam is being escrowed
+        if keys_changed.contains(&balance_key(&xan(), &super::ADDRESS)) {
+            return Ok(self.check_escrow(verifiers));
```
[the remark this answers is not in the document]
batconjurer: Then escrowing will never pass since it doesn't perform a transfer between erc20s. The `extract_valid_keys_changed` will fail since the changed keys will be empty.  A tx should either escrow nam or transfer erc20s, not both at the same time imo.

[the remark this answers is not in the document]
batconjurer: I've started adding tests in `bat/ethbridge/eth-bridge-vp-escrow-nam`

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -77,6 +80,58 @@ where
...
+        payer_account: &Address,
+        escrow_account: &Address,
+        expected_debit: Amount,
+        expected_credit: Amount,
+    ) -> bool {
+        let debited = self.account_balance_delta(payer_account);
+        let credited = self.account_balance_delta(escrow_account);
+        if debited.is_none() && credited.is_none() {
+            return false;
+        }
```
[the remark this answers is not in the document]
batconjurer: an artifact from a previous implementation. Will fix.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -77,6 +80,58 @@ where
...
+                false
+            }
+            (_, Some(SignedAmount::Negative(_))) => {
+                tracing::debug!(
+                    "The Ethereum bridge pool's escrow was not credited from \
+                     account {}.",
+                    payer_account
+                );
+                false
+            }
```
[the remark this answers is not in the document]
batconjurer: Well, not getting a delta isn't an error condition. It just means that the value in the key wasn't written to, so it doesn't show up in the writelog. Failure to read the backing store is an error, but we expect that here since if that happens, the node is in big trouble.

shared/src/ledger/eth_bridge/vp/mod.rs
batconjurer: The error in this case means we can't read our DB. Is that a recoverable error? I think it's better to panic in such a case.

shared/src/ledger/eth_bridge/vp/mod.rs
batconjurer: I guess we could propogate this error then.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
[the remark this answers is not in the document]
batconjurer: I guess it's not expected though. I can add it.

batconjurer:
pls update wasm

james-chf:
Taking a look again at this PR

shared/src/ledger/eth_bridge/vp/mod.rs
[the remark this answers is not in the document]
james-chf: Yeah it should be one or the other, escrowing nam or transferring ERC20s. But this VP as it is now looks like it would accept a transaction that sent some NAM to the Eth bridge, regardless of what else the transaction is changing. We should have something like the below:

```rust

    // we aren't concerned with keys that changed outside of our account
    let keys_changed: HashSet<_> = keys_changed
        .iter()
        .filter(|key| storage::is_eth_bridge_key(key))
        .collect();
```

moved out of `extract_valid_keys_changed` into `validate_tx`. We could dispatch the correct VP logic depending on the contents of `keys_changed`

shared/src/ledger/eth_bridge/vp/mod.rs
james-chf: It's not recoverable but I don't think panicking here is the best way to stop the process, we could propagate the error upwards, and then take a decision somewhere to shut down gracefully based on that unrecoverable error. We don't have functionality like that right now in the ledger though so have to `.unwrap()` at some point in our call path. I guess it is a bigger issue than this PR.

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -67,6 +124,12 @@ where
             verifiers_len = verifiers.len(),
             "Ethereum Bridge VP triggered",
         );
+
+        // first check if Nam is being escrowed
+        if keys_changed.contains(&balance_key(&xan(), &super::ADDRESS)) {
+            return self.check_escrow(verifiers);
+        }
```
james-chf, line 131:
```suggestion
        if keys_changed.contains(&balance_key(&xan(), &super::ADDRESS))
            && !keys_changed
                .iter()
                .any(|key| storage::is_eth_bridge_key(key))
        {
            return self.check_escrow(verifiers);
        }
```

Re. https://github.com/anoma/namada/pull/703#discussion_r1013022043, looking again at this, if NAM is sent to the Eth bridge, the Eth bridge VP passes iff the bridge pool is in the verifiers list. We want to check no Eth bridge keys changed at all as well in that case, maybe something like this.

james-chf (changes requested):

shared/src/ledger/eth_bridge/vp/mod.rs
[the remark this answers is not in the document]
batconjurer: I thought about that earlier, so yeah, I can do that.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -169,36 +209,62 @@ where
...
+        // if we are going to mint wNam on Ethereum, the appropriate
+        // amount of Nam must be escrowed in the Ethereum bridge VP's storage.
+        // TODO: We should look this address up from storage
+        if transfer.transfer.asset == wnam() {
+            // check that correct amount of Nam was put into escrow.
+            return if transfer.gas_fee.payer == transfer.transfer.sender {
+                if !self.check_nam_escrowed(
+                    &transfer.gas_fee.payer,
+                    &BRIDGE_POOL_ADDRESS,
+                    transfer.gas_fee.amount + transfer.transfer.amount,
```
sug0, line 221: might be worth adding overflow checks here. in `release` builds, these are simply ignored by the compiler, just like in C/C++.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -169,36 +209,62 @@ where
...
+                    transfer.gas_fee.amount,
+                )? || !self.check_nam_escrowed(
+                    &transfer.transfer.sender,
+                    &Address::Internal(InternalAddress::EthBridge),
+                    transfer.gas_fee.amount + transfer.transfer.amount,
```
sug0, line 226: same

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
```diff
@@ -169,36 +209,62 @@ where
...
+                transfer.transfer.amount,
+            )? {
+                Ok(false)
+            } else {
+                tracing::info!(
+                    "The Ethereum bridge pool VP accepted the transfer {:?}.",
+                    transfer
+                );
+                Ok(true)
+            };
```
sug0, line 256: I found this code a bit confusing to read, took me a while to see what was happening; maybe it was because of the repetition? it's fine for now, but I think it should be refactored later

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,61 @@ where
...
+                "Could not retrieve the Ethereum bridge VP's balance from \
+                 storage"
+            );
+            return Ok(false);
+        };
+        let escrow_post: Amount =
+            if let Ok(Some(bytes)) = self.ctx.read_bytes_post(&escrow_key) {
+                BorshDeserialize::try_from_slice(bytes.as_slice()).expect(
+                    "Deserializing the balance of the Ethereum bridge VP from \
+                     storage shouldn't fail",
```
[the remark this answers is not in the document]
batconjurer: Good point.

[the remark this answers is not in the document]
sug0: yeah, any keys in post should be checked, really. maybe it would be better to disallow reading raw bytes with `read_bytes_post`, since it isn't aware of any concrete types `T` we might be trying to deserialize user data into

shared/src/ledger/eth_bridge/vp/mod.rs
batconjurer: I actually changed the extract keys function to make the dual functions of the VP explicit.

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -27,6 +28,61 @@ where
...
+                )
+            } else {
+                tracing::debug!(
+                    "Could not retrieve the modified Ethereum bridge VP's \
+                     balance after applying tx"
+                );
+                return Ok(false);
+            };
```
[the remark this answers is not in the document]
batconjurer: Doesn't really help because there is the still the Ok(None) situation. So there will still be an else.

shared/src/ledger/eth_bridge/bridge_pool_vp.rs
batconjurer: Nope, I'll solve it now.

sug0 (approved):
nice refactor. cheers!

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -107,9 +177,9 @@ where
 
 /// If `keys_changed` represents a valid set of changed keys, return them,
 /// otherwise return `None`.
```
james-chf, line 179: docstring needs to be updated here

shared/src/ledger/eth_bridge/vp/mod.rs
```diff
@@ -51,6 +51,73 @@ where
...
+    }
+}
+
+/// One of the the two types of checks
+/// this VP must perform.
+#[derive(Debug)]
+enum CheckType {
+    Escrow,
+    Erc20Transfer(wrapped_erc20s::Key, wrapped_erc20s::Key),
+}
```
james-chf, line 119: This is a nice way of dispatching which check to do!

shared/src/types/token.rs
```diff
@@ -82,6 +82,13 @@ impl Amount {
             micro: change as u64,
         }
     }
+
+    /// Checked addition on amounts
+    pub fn checked_add(&self, amount: &Amount) -> Option<Self> {
+        self.micro
+            .checked_add(amount.micro)
+            .map(|micro| Self { micro })
+    }
```
james-chf, line 91: +1, I need to start using this as well, have opened https://github.com/anoma/namada/issues/748 to expand and push back to `main`

james-chf (approved):

shared/src/ledger/eth_bridge/vp/mod.rs
batconjurer: Oh right!

PR #706, opened 2022-10-27
fix storage height bugs and rocksdb read from arb. height and delete's diffs

### LINKED ISSUE #705: various storage height issues
in `shared/src/ledger/storage/mod.rs`:
- `fn read_with_height` should use `self.last_height` instead of `self.get_block_height()` - we should not read from uncommitted block (this causes spurious e2e failures, but it only occurs when Query request comes in between EndBlock (shimmed to FinalizeBlock) and Commit, which is quite rare)
- `fn write` and `fn delete` should use `self.block.height` instead of `self.last_height` - we should be writing into the current block's height (this only affects DB diffs, currently unused)

in `apps/src/lib/node/ledger/storage/rocksdb.rs`:
- `fn read_subspace_val_with_height` is all wrong :D affects reads on non-latest height
- `fn delete_subspace_val` should attempt to read prev value from `subspace_key`, not `key`, which is lacking the expected "subspace" prefix. Because of this it currently won't write a diff

tzemanovic:
fixes #705


apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -614,31 +614,77 @@ impl DB for RocksDB {
...
+        let key_prefix = Key::from(height.to_db_key())
+            .push(&"diffs".to_owned())
+            .map_err(Error::KeyError)?;
+        let new_val_key = key_prefix
+            .push(&"new".to_owned())
+            .map_err(Error::KeyError)?
+            .join(key)
+            .to_string();
+        // If it has a "new" val, it was written at this height
+        if let Some(new_val) = self
```
yito88, line 629: Why not use match?

yito88 (approved):

apps/src/lib/node/ledger/storage/rocksdb.rs
tzemanovic: done

sezox81:
fix storage height bugs and rocksdb read from arb. height and delete's diffs

PR #388, opened 2022-08-26
Inflation system and PoS rewards

brentstone:
Implementation of the general inflation system and the rewards system for proof-of-stake in particular.

Closes #340 and #353.

Based on #379, which touches lots of PoS-related PRs.

PR #707, opened 2022-10-28
Refactor out `VotingPower`

### LINKED ISSUE #455: Phase out use of `VotingPower` struct
The `VotingPower` struct is just a wrapper around a `u64`, and with the decision to adopt `votes_per_token = 1` per namnam, there isn't much use for passing around `VotingPower` in the code. If anything, this may still be used only in direct communication with tendermint, but otherwise should be replaced with an `Amount` or just a `u64` itself. This will result in cleaner and more succinct code.

brentstone:
Closes #455.

Currently based on #695 and being split out from #388.


brentstone:
pls update wasm

apps/src/lib/client/rpc.rs
```diff
@@ -923,14 +931,19 @@ pub async fn query_voting_power(ctx: Context, args: args::QueryVotingPower) {
...
                     println!(
-                        "Validator {} is {}, voting power: {}",
+                        "Validator {} is {}, bonded stake: {}, voting power: \
+                         {}",
                         validator.encode(),
                         if is_active { "active" } else { "inactive" },
-                        voting_power
+                        bonded_stake,
+                        (Decimal::from(bonded_stake)
+                            * pos_params.tm_votes_per_token)
```
tzemanovic, line 940: I think perhaps for most part, we could use the `bonded_stake` value as is, because `tm_votes_per_token` will only be used in tendermint validator set

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -292,10 +292,9 @@ where
             let (consensus_key, power) = match update {
                 ValidatorSetUpdate::Active(ActiveValidator {
                     consensus_key,
-                    voting_power,
+                    bonded_stake,
                 }) => {
-                    let power: u64 = voting_power.into();
-                    let power: i64 = power
+                    let power: i64 = bonded_stake
```
tzemanovic, line 297: I think this is the only place where we should apply `tm_votes_per_token`

apps/src/lib/client/rpc.rs
brentstone: Here I included the voting power in the output because `namada_tests::e2e::helpers::find_voting_power` works by looking for this string in std::out. I guess we could maybe change `find_voting_power` in favor of `find_bonded_stake` or something, but I figured this would be useful from the client-side.

apps/src/lib/node/ledger/shell/finalize_block.rs
brentstone: why here, since a validator set update is dependent on the bonded stake and not voting power? I realize I should probably change the comments and var names here if we do want to keep it as just `bonded_stake`.

apps/src/lib/client/rpc.rs
tzemanovic: I see - I think we could replace it with `find_bonded_stake` as in majority of places we'll be using that, `client voting-power` command might only be confusing

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: This is the validator set that's passed to tendermint (added to the `response.validator_updates` below)

brentstone: Ah I see. Ok, there happens to be one other location where `into_tm_votes_per_token` is currently used, and that is to set a `ferveo_common::TendermintValidator` inside `namada/apps/src/lib/node/ledger/shell/queries.rs`. Should it remain there?

wasm/wasm_source/src/tx_bond.rs
```diff
@@ -108,41 +107,86 @@ mod tests {
...
-            );
+        //     - `#{PoS}/validator/#{validator}/deltas`
+        //     - `#{PoS}/total_deltas`
+        //     - `#{PoS}/validator_set`
+
+        // Check that the validator set and deltas are unchanged before pipeline
+        // length and that they are updated between the pipeline and
+        // unbonding lengths
+        //
+        // TODO: should end be pipeline + unbonding now?
```
tzemanovic, line 135: not in here, I think, in tx_unbond once we make the change (let's do that when we switch to LazyMap based epoched)

wasm/wasm_source/src/tx_unbond.rs
```diff
@@ -167,40 +169,77 @@ mod tests {
...
+                validator_sets_post.get(epoch),
+                "Validator set at and after pipeline offset should be the \
+                 same since we are before the unbonding offset - checking \
+                 epoch {epoch}"
             );
         }
 
         {
+            // TODO: should this loop over epochs after this one as well? Are
+            // there any?
```
tzemanovic, line 221: it could yeah - it should just return the same state as on the last one known

wasm/wasm_source/src/tx_unbond.rs
```diff
@@ -257,6 +298,7 @@ mod tests {
             );
         }
         {
+            // TODO: checl logic here
```
tzemanovic, line 301: I think this does correspond to the current implementation, but not yet to the latest spec updates to unbonding offsets

brentstone:
pls update wasms

tzemanovic (dismissed):

Issue #478, opened 2022-09-20
Update PoS parameter values


brentstone:
During the August retreat in Italy, we decided upon some parameter values for the PoS system. These are:

- pipeline_offset = 2
- unbonding_offset = 21
- slash rate = 0.001 / namnam (fundamental token unit)
- votes per token = 1 / namnam (fundamental token unit)
- max active validators = 100
- block proposer reward = 1% minimum
- PoS inflation = 10% maximum

Furthermore, subsequent conversations established:
- unit denomination = 10^6 (fundamental tokens per whole NAM)

brentstone:
@cwgoes @adrianbrink @tzemanovic - do we want the same slash rate for duplicate vote attacks and light client attacks?

cwgoes:
> @cwgoes @adrianbrink @tzemanovic - do we want the same slash rate for duplicate vote attacks and light client attacks?

Yes - both varying from 0 to 1 (100%) according to the cubic slashing function.

PR #708, opened 2022-10-28
Update PoS parameters

brentstone:
Includes new parameters needed for the inflation mechanism as well as the values for PoS parameters as most recently decided (largely from the August retreat).


brentstone:
pls update wasm

brentstone:
pls update wasms

apps/src/lib/config/genesis.rs
```diff
@@ -233,6 +232,16 @@ pub mod genesis_config {
...
+        /// Expected number of epochs per year
+        pub epochs_per_year: u64,
+        /// PoS gain p
+        pub pos_gain_p: Decimal,
+        /// PoS gain d
+        pub pos_gain_d: Decimal,
+        /// PoS staked ratio
+        pub staked_ratio: Decimal,
+        /// PoS inflation rate last epoch
+        pub pos_inflation_amount: u64,
```
tzemanovic, line 244: I'm not sure if `staked_ratio` and `pos_inflation_amount` should be set in genesis config. I think `staked_ratio` should be calculated from the genesis total supply and validator stakes. `pos_inflation_amount` should probably be omitted on genesis as there is no last epoch

apps/src/lib/config/genesis.rs
```diff
@@ -233,6 +232,16 @@ pub mod genesis_config {
         // Hashes of whitelisted txs array. `None` value or an empty array
         // disables whitelisting.
         pub tx_whitelist: Option<Vec<String>>,
+        /// Expected number of epochs per year
+        pub epochs_per_year: u64,
```
tzemanovic, line 236: I think since we already have `min_duration`, we could either use that to estimate `epochs_per_year` (or vise versa)

tzemanovic:
we'll also need to update anoma-network-configs for this and the values from #478

apps/src/lib/config/genesis.rs
brentstone: Ok makes sense. FWIW, `staked_ratio` is still calculated from the genesis supply in `init_chain.rs` in branch `brent/pos-inflation-rewards` (#714)

apps/src/lib/config/genesis.rs
brentstone: Maybe vice-versa? Since I'm pretty sure we are targeting 365 epochs per year (discussed at retreat).

tzemanovic: vice-versa sounds good!

tzemanovic: this is still open right? Let's drop `min_duration` from genesis config and instead set it from something like `let min_duration = 60 * 60 * 24 * 365 / epochs_per_year`

tzemanovic (dismissed):

PR #712, opened 2022-10-28
Initialize Ethereum bridge parameters at genesis

batconjurer:
The genesis params for the Ethereum bridge need to be added to storage when init_chain is called. This PR changes the storage to hold these values and write them at genesis.


shared/src/ledger/storage/mod.rs
```diff
@@ -68,6 +72,16 @@ where
...
+    /// Minimum number of confirmations needed for the
+    /// Ethereum bridge to consider an event final
+    pub min_confirmations: Option<MinimumConfirmations>,
+    /// The Ethereum address of the ERC20 contract that represents this chain's
+    /// native token.
+    pub native_erc20: Option<EthAddress>,
+    /// The Ethereum address of the bridge contract.
+    pub bridge_contract: Option<UpgradeableContract>,
+    /// The Ethereum address of the governance contract.
+    pub governance_contract: Option<UpgradeableContract>,
```
james-chf, line 84: Do we have to put these on the `Storage` struct? Can't we store these values similarly to however other values set at genesis are stored (e.g. treasury params, gov params, pos params)

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -5,7 +5,13 @@
...
 //! - `tx_queue`: txs to be decrypted in the next block
 //! - `pred`: predecessor values of the top-level keys of the same name
-//!   - `tx_queue`
+//! - `tx_queue`
+//! - `min_confirmations`: Minimum number of confirmations needed for the
+//!   Ethereum bridge to consider an event final.
+//! - `native_erc20`: The Ethereum address of the ERC20 contract that represents
+//!   this chain's native token.
+//! - `bridge_contract`:  The Ethereum address of the bridge contract.
+//! - `governance_contract`: The Ethereum address of the governance contract.
```
james-chf, line 14: Similarly could we store these values underneath some other key?

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -21,12 +23,13 @@ where
     /// Create a new genesis for the chain with specified id. This includes
     /// 1. A set of initial users and tokens
     /// 2. Setting up the validity predicates for both users and tokens
-    /// 3. A matchmaker
+    /// 3. Validators
+    /// 4. The PoS system
+    /// 5. The Ethereum bridge parameters
```
james-chf, line 28: +1 for splitting up `init_chain`, think it was long overdue, though we should push this back to `main` as well

shared/src/ledger/eth_bridge/storage/mod.rs
```diff
@@ -4,7 +4,16 @@ pub mod vote_tallies;
 pub mod wrapped_erc20s;
 
 use super::ADDRESS;
-use crate::types::storage::{Key, KeySeg};
+use crate::types::storage::{DbKeySeg, Key, KeySeg};
```
james-chf, line 7: Yeah I think under the #EthBridge account is an appropriate place to store these parameters. In any case they should only be modifiable via a governance vote/hard fork.

apps/src/lib/config/ethereum_bridge/mod.rs
```diff
@@ -1,2 +1,37 @@
...
+        ContractVersion, Contracts, EthereumBridgeConfig, MinimumConfirmations,
+        UpgradeableContract,
+    };
+    use namada::types::ethereum_events::EthAddress;
+
+    /// Ensure we can serialize and deserialize a [`Config`] struct to and from
+    /// TOML. This can fail if complex fields are ordered before simple fields
+    /// in any of the config structs.
+    #[test]
+    fn test_round_trip_toml_serde() -> Result<()> {
```
james-chf, line 16: this test should be moved along with the `EthereumBridgeConfig`?

james-chf (approved):
lgtm!

apps/src/lib/node/ledger/shell/init_chain.rs
batconjurer: I was hoping that the functions would secretly all be really similar and we could do something generic. Didn't seem to work out that way though :cry:

apps/src/lib/config/ethereum_bridge/mod.rs
batconjurer: Makes sense. Will do.

batconjurer:
pls update wasm

sug0 (approved):

PR #719, opened 2022-10-31
Governance: refactor

Fraccaman:


Fraccaman:
pls update wasm

shared/src/ledger/parameters/mod.rs
```diff
@@ -62,7 +63,13 @@ where
                 KeyType::PARAMETER => {
                     let proposal_id = u64::try_from_slice(tx_data).ok();
                     match proposal_id {
-                        Some(id) => is_proposal_accepted(&self.ctx, id),
+                        Some(id) => {
```
tzemanovic, line 66: I liked `is_proposal_accepted` tbh! You could maybe re-implement it with `StorageRead::has_key` and then you should be able to pass `self.ctx.pre()` for it from here

tzemanovic (approved):

PR #721, opened 2022-11-01
Refactor Validator Sets

brentstone:
As specified in #283, the validator sets should be refactored from `active` and `inactive` to `consenses`, `below_capacity`, and `below_threshold`. The `active` and `consensus` validator sets are the same (just renaming).

The `inactive` validator set is further refactored into `below_capacity` and `below_threshold`. The "threshold" refers to a level of bonded stake below which we do not iterate over the validators in various PoS actions. The goal is to prevent unbounded iteration over validator sets.

Currently based on #714


cwgoes:
Is this PR still relevant? @brentstone

brentstone:
> Is this PR still relevant? @brentstone

Yes still relevant, held out on further development atm because future changes will now largely depend on #740, which is still yet to be completed.

adrianbrink:
Closing this for now, since it's stale. @brentstone we still want this but it needs to rebased and then it can be reopened.

Issue #497, opened 2022-09-21
Fix compiling with `abcipp` and set up CI to check it


james-chf:
ABCI 0.19 (formerly known as ABCI++) is coming in Tendermint v0.38 and will provide us with vote extensions, which was used by the original design of the Ethereum bridge. Code to do with it is guarded behind the `abcipp` flag, and `--no-default-features` should be used also. We should run CI to make sure that our ABCI 0.19 code continues to compile/pass tests.

Compiling with `abcipp` currently doesn't work on `main` or `eth-bridge-integration` branches (as of 2022-11-04). We should fix up in `main` first and have a CI job to at least `cargo check` it, then do the same for `eth-bridge-integration`.

- [ ] fix in `main` 
  - https://github.com/anoma/namada/pull/754
  - https://github.com/anoma/namada/pull/869
- [ ] fix in `eth-bridge-integration`
  - https://github.com/anoma/namada/pull/871
- [ ] CI in `main`
  - https://github.com/anoma/namada/pull/824
- [ ] CI in `eth-bridge-integration`

james-chf:
This has broken a couple of times

cwgoes:
@james-chf is this relevant any more?

PR #723, opened 2022-11-01
Merge 0.9.0 into eth-bridge-integration

juped:
wip, still needs rpc reimplemented

### DIFF
--- a/shared/src/ledger/events.rs
+++ b/shared/src/ledger/events.rs
@@ -7,10 +7,10 @@ use std::fmt::{self, Display};
 use std::ops::{Index, IndexMut};
 
 use borsh::{BorshDeserialize, BorshSerialize};
-use tendermint_proto::abci::EventAttribute;
 use thiserror::Error;
 
 use crate::ledger::governance::utils::ProposalEvent;
+use crate::tendermint_proto::abci::EventAttribute;
 use crate::types::ibc::IbcEvent;
 #[cfg(feature = "ferveo-tpke")]
 use crate::types::transaction::{hash_tx, TxType};
@@ -162,7 +162,7 @@ impl From<ProposalEvent> for Event {
 }
 
 /// Convert our custom event into the necessary tendermint proto type
-impl From<Event> for tendermint_proto::abci::Event {
+impl From<Event> for crate::tendermint_proto::abci::Event {
     fn from(event: Event) -> Self {
         Self {
             r#type: event.event_type.to_string(),


james-chf:
pls update wasm

james-chf:
Merging this PR should close https://github.com/anoma/namada/issues/730

james-chf:
Current status: refactoring of bridge pool query endpoints to work with the new RPC router is WIP, some tests are still broken. There are still some conflicts with latest `eth-bridge-integration` that need to be merged in, after those tests are fixed up.

james-chf:
Compiling with `abcipp` is also broken, though this is also the case on `main` - have reappropiated https://github.com/anoma/namada/issues/497 as an issue to fix this up in both the branches

james-chf:
pls update wasm

shared/src/ledger/events.rs
```diff
@@ -153,7 +162,7 @@ impl From<ProposalEvent> for Event {
 }
 
 /// Convert our custom event into the necessary tendermint proto type
-impl From<Event> for crate::facade::tendermint_proto::abci::Event {
+impl From<Event> for tendermint_proto::abci::Event {
```
sug0, line 165: this should be `crate::facade::tendermint_proto::abci::Event`, right?

sug0 (approved):

apps/src/lib/client/rpc.rs
```diff
@@ -1369,100 +1320,77 @@ pub async fn query_storage_value<T>(
...
-                    "Error in the query {} (error code {})",
-                    response.info, err
-                )
-            }
-        }
+    // In case `T` is a unit (only thing that encodes to 0 bytes), we have to
+    // use `storage_has_key` instead of `storage_value`, because `storage_value`
+    // returns 0 bytes when the key is not found.
+    let maybe_unit = T::try_from_slice(&[]);
+    if let Ok(unit) = maybe_unit {
```
batconjurer, line 1327: wow. what a hack :laughing:

shared/src/ledger/events.rs
james-chf: There is no `facade` module in `shared` (currently)`, though there is something similar in `shared/lib.rs`

https://github.com/anoma/namada/blob/6cbce4a4dcdd3811679f5eb4aa7c82f51dff74e6/shared/src/lib.rs#L9-L16

sug0: oh right, this isn't in `apps`... yeah, we can use `crate::<dep>` inside of `shared/`

james-chf: Oh yeah, this should be using that, will update

shared/src/types/keccak.rs
```diff
@@ -103,13 +104,48 @@ pub fn keccak_hash(bytes: &[u8]) -> KeccakHash {
 
 /// This module defines encoding methods compatible with Ethereum
 /// smart contracts.
+// TODO: relocate this sub-module to a new home
```
batconjurer, line 107: I've been thinking about this for a while too. Maybe we should just have an ABI mod which has both encode and keccak submods

shared/src/types/keccak.rs
```diff
@@ -103,13 +104,48 @@ pub fn keccak_hash(bytes: &[u8]) -> KeccakHash {
...
+        encoded_data: Vec<u8>,
+        /// Indicate we do not own values of type `T`.
+        ///
+        /// Passing `PhantomData<T>` here would trigger the drop checker,
+        /// which is not the desired behavior, since we own an encoded value
+        /// of `T`, not a value of `T` itself.
+        _marker: PhantomData<*const T>,
+    }
+
+    impl<T> EncodeCell<T> {
```
batconjurer, line 130: This is interesting. Should we change the return type of the `Encode::encode` method to return an instance of this struct? @sug0 ?

tests/src/e2e/helpers.rs
```diff
@@ -148,6 +149,63 @@ pub fn get_epoch(test: &Test, ledger_address: &str) -> Result<Epoch> {
     Ok(Epoch(epoch))
 }
 
+/// Get the last committed block height.
+pub fn get_height(test: &Test, ledger_address: &str) -> Result<u64> {
```
batconjurer, line 153: :frowning_face:

batconjurer (approved):
One question really.

shared/src/types/keccak.rs
batconjurer: I like it.

Issue #599, opened 2022-10-14
Fix block height in storage after epoch change


sug0:
The block height of a new epoch updated in storage, after we call `finalize_block()`, is wrong. We need to store the previous height, rather than the next height.

The fix is fairly simple. We just need to apply [this patch](https://github.com/anoma/namada/commit/c9e16e726b5aa5f97f03db5fc5c30a6febcd8860.patch) in reverse.

We also need to fix a related unit test (`test_can_send_validator_set_update`), or simply `#[ignore]` it for now.

sug0:
Linked PR https://github.com/anoma/namada/pull/594

Issue #600, opened 2022-10-14
Change `can_send_validator_set_update` logic


sug0:
Related with https://github.com/anoma/namada/issues/599

Instead of sending a validator set update at the first block of an epoch, we should do this at the second block of an epoch. This is because we don't know, in advance, the first block of a new epoch until we call `finalize_block()`. The only exception is, of course, at genesis, where we know the height of the first block.

This change only affects ABCI++ builds of the ledger.

PR #725, opened 2022-11-02
Send a validator set update only once per epoch

sug0:
Instead of spamming the chain with validator set update protocol txs, now we only send these once per epoch, at the second block height offset within the epoch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,13 @@ test-unit:
 			--skip e2e \
 			-Z unstable-options --report-time
 
+test-unit-debug:
+	$(debug-cargo) test \
+			$(TEST_FILTER) -- \
+			--skip e2e \
+			--nocapture \
+			-Z unstable-options --report-time
+
 test-wasm:
 	make -C $(wasms) test
 
--- a/apps/src/lib/node/ledger/shell/queries.rs
+++ b/apps/src/lib/node/ledger/shell/queries.rs
@@ -728,16 +728,6 @@ mod test_queries {
             /// Test if [`QueriesExt::can_send_validator_set_update`] behaves as
             /// expected.
             #[test]
-            // TODO: we should fix this test to cope with epoch changes only
-            // happening at the first block of a new epoch. an erroneous change
-            // was introduced to the ledger, that updated the epoch correctly
-            // at the first block of the new epoch, but recorded `height + 1`
-            // instead of the actual height of the epoch change. since this
-            // test depended on that erroneous logic to pass, it's busted.
-            //
-            // linked issues:
-            // - <https://github.com/anoma/namada/issues/599>
-            // - <https://github.com/anoma/namada/issues/600>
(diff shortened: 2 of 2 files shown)


batconjurer (approved):

sug0:
Closes #599 and #600

apps/src/lib/node/ledger/shell/queries.rs
```diff
@@ -604,41 +596,43 @@ where
...
 
-        // handle genesis block corner case
-        if height == BlockHeight(1) {
+    #[cfg(not(feature = "abcipp"))]
+    fn can_send_validator_set_update(&self, can_send: SendValsetUpd) -> bool {
+        // when checking vote extensions in Prepare
+        // and ProcessProposal, we simply return true
+        if matches!(can_send, SendValsetUpd::AtPrevHeight) {
             return true;
         }
```
james-chf, line 611: We can get rid of the `SendValsetUpd` now, we no longer need this context from the caller? `can_send_validator_set_update` can reason by itself using only the current decision height.

james-chf (changes requested):

Makefile
```diff
@@ -133,6 +133,13 @@ test-unit:
 			--skip e2e \
 			-Z unstable-options --report-time
 
+test-unit-debug:
+	$(debug-cargo) test \
+			$(TEST_FILTER) -- \
+			--skip e2e \
+			--nocapture \
+			-Z unstable-options --report-time
+
```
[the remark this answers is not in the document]
sug0: yeah I can create a small PR for main

apps/src/lib/node/ledger/shell/queries.rs
sug0: we call this from two places:

* when we create a new vote extension at the ABCI commit phase, where we pass `SendValsetUpd::Now` as arg
* when we verify vote extensions e.g. at prepare and process proposal, where we pass `SendValsetUpd::AtPrevHeight` as arg

we still need to distinguish between these two cases; a `BlockHeight` parameter wouldn't be enough, since we need `SendValsetUpd::AtPrevHeight` to always return true.

sug0: you could remove it altogether if we refactored prepare and process proposal to always accept validator set update protocol txs without a predicate check, but then it'd be kind of annoying to write this code again for ABCI++ if/when that happens...

james-chf: > you could remove it altogether if we refactored prepare and process proposal to always accept validator set update protocol txs without a predicate check, but then it'd be kind of annoying to write this code again for ABCI++ if/when that happens...

I would actually do this but since we're going to maintain our current `abcipp` code I guess it's fine

james-chf (approved):

james-chf:
plz merge only after https://github.com/anoma/namada/pull/723

PR #715, opened 2022-10-31
Read wnam address from Storage

batconjurer:
This PR makes the bridge pool VP lookup the Ethereum address for wNam from storage instead of using a hard-coded constant. The tests are also adjusted appropriately.

This PR depends on #703 and #712

PR #729, opened 2022-11-03
Bridge pool cli

### LINKED ISSUE #425: Make the Ethereum bridge pool queryable

batconjurer:
This PR partially closes #425. It depends on #715. It sets up CLI commands for the following:
 * Adding a new transfer to the bridge pool
 * Querying the current contents of the pool
 * Getting a merkle proof of a selected set of transfers
 
 It also generates a new binary, `namadar` which can handle these tasks. Eventually, this binary should become a more robust relayer. It refactors the queries in the shell  a bit for better UX.
 
 This is not covered by tests yet as this PR is large enough. The wasm code needs to be updated and then we can test. That will be the next PR.


apps/src/lib/client/eth_bridge_pool.rs
```diff
@@ -0,0 +1,88 @@
...
+    let path = Path::EthereumBridgePool(BridgePoolSubpath::Proof);
+    let data = args.transfers.try_to_vec().unwrap();
+    let response = client
+        .abci_query(Some(path.into()), data, None, false)
+        .await
+        .unwrap();
+    if response.code != Code::Ok {
+        println!("{}", response.info);
+    } else {
+        println!("ABI Encoded Proof:\n {:#?}", response.value);
```
sug0, line 62:
```suggestion
        println!("Ethereum ABI-encoded proof:\n {:#?}", response.value);
```

sug0 (changes requested):
there's no relaying happening yet, right? couldn't these sub-commands still be a part of `namadac`?

also, we should hold off on merging until #723 is in `eth-bridge-integration`

batconjurer:
> there's no relaying happening yet, right? couldn't these sub-commands still be a part of `namadac`?
> 
> also, we should hold off on merging until #723 is in `eth-bridge-integration`

Yeah, it could be part of the client, but I'm imagining in the future when we want to craft ethereum txs and calculate gas prices vs fees and all that shit, this will all be in this other binary. It was also a nicer ux.

batconjurer:
Although submitting to the pool does probably make more sense as a client command.

sug0 (approved):

sug0:
need rebasing against 0.9.0 I guess

james-chf:
I'll wait til after rebase on 0.9.0 to review, since the client side RPC logic changed in that release

shared/src/ledger/queries/shell.rs
```diff
@@ -374,12 +375,28 @@ where
...
+                }
+                _ => None,
+            })
+            .collect();
+        if values.len() != keys.len() {
+            return Err(storage_api::Error::SimpleMessage(
+                "One or more of the provided hashes had no corresponding \
+                 transfer in storage.",
+            ));
+        }
```
james-chf, line 395: we should revisit this later, it should be possible to forward a more specific error message to the caller that would be helpful when debugging (e.g. which hash(es) were missing transfers in storage)

james-chf:
I haven't looked why but running the `namadar` binary shows some wallet subcommands?
```
 > target/debug/namadar
Namada v0.9.0-1671-gb6112b4
Anoma Ethereum bridge pool command line interface.

USAGE:
    namadar [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --base-dir <base-dir>    The base directory is where the nodes, client and wallet
                                 configuration and state is stored. This value can also be set via
                                 `ANOMA_BASE_DIR` environment variable, but the argument takes
                                 precedence, if specified. Defaults to `.anoma`.
        --chain-id <chain-id>    The chain ID.
        --mode <mode>            The mode in which to run Anoma. Options are
                                         * Validator (default)
                                         * Full
                                         * Seed
        --wasm-dir <wasm-dir>    Directory with built WASM validity predicates, transactions. This
                                 value can also be set via `ANOMA_WASM_DIR` environment variable,
                                 but the argument takes precedence, if specified.

SUBCOMMANDS:
    construct-proof    Construct a merkle proof that the given transfer is in the pool.
    query              Get the contents of the Ethereum bridge pool.
    address            Address management, including methods to generate and look-up addresses.
    help               Prints this message or the help of the given subcommand(s)
    key                Keypair management, including methods to generate and look-up keys.
```

apps/src/lib/cli.rs
```diff
@@ -1201,21 +1211,130 @@ pub mod cmds {
...
+                .subcommand(ConstructProof::def().display_order(1))
+                .subcommand(QueryEthBridgePool::def().display_order(1))
+        }
+    }
+
+    #[derive(Clone, Debug)]
+    pub struct AddToEthBridgePool(pub args::EthereumBridgePool);
+
+    impl SubCmd for AddToEthBridgePool {
+        const CMD: &'static str = "add-transfer";
```
james-chf, line 1264: Would call this something like `add-erc20-transfer` or otherwise put it under some subcommand of `namadac` - otherwise it is just `namadac add-transfer`. This command isn't showing in the help string for some reason when you run `namadac`

james-chf (changes requested):

batconjurer:
> .

I added methods to look up addresses from the wallet, but it looks like I messed up how it looks in the ui.

apps/src/lib/cli.rs
batconjurer: I renamed the command. It shows up for me in the help menu.

shared/src/ledger/queries/shell.rs
batconjurer: I added this.

batconjurer:
> I haven't looked why but running the `namadar` binary shows some wallet subcommands?
> 
> ```
>  > target/debug/namadar
> Namada v0.9.0-1671-gb6112b4
> Anoma Ethereum bridge pool command line interface.
> 
> USAGE:
>     namadar [OPTIONS] <SUBCOMMAND>
> 
> FLAGS:
>     -h, --help       Prints help information
>     -V, --version    Prints version information
> 
> OPTIONS:
>         --base-dir <base-dir>    The base directory is where the nodes, client and wallet
>                                  configuration and state is stored. This value can also be set via
>                                  `ANOMA_BASE_DIR` environment variable, but the argument takes
>                                  precedence, if specified. Defaults to `.anoma`.
>         --chain-id <chain-id>    The chain ID.
>         --mode <mode>            The mode in which to run Anoma. Options are
>                                          * Validator (default)
>                                          * Full
>                                          * Seed
>         --wasm-dir <wasm-dir>    Directory with built WASM validity predicates, transactions. This
>                                  value can also be set via `ANOMA_WASM_DIR` environment variable,
>                                  but the argument takes precedence, if specified.
> 
> SUBCOMMANDS:
>     construct-proof    Construct a merkle proof that the given transfer is in the pool.
>     query              Get the contents of the Ethereum bridge pool.
>     address            Address management, including methods to generate and look-up addresses.
>     help               Prints this message or the help of the given subcommand(s)
>     key                Keypair management, including methods to generate and look-up keys.
> ```

Bad copy pasta. Fixed

james-chf (approved):
lgtm!

PR #731, opened 2022-11-03
Aggregate votes for validator set updates in storage

sug0:
Based on #683

Count votes on validator set updates signed off for some epoch. Currently, nothing happens when we reach a complete proof (i.e. more than 2/3 of the voting power) on some set of validators, but we could, e.g. emit some kind of event which would be consumed by a relayer.

### DIFF
--- a/apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
@@ -94,7 +94,7 @@ where
         }
     }
 
-    let (tally, changed) = if !exists_in_storage {
+    let (tally, changed, confirmed) = if !exists_in_storage {
         tracing::debug!(
             %valset_upd_keys.prefix,
             ?ext.voting_powers,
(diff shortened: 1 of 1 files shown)


apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
```diff
@@ -31,29 +34,116 @@ where
...
+            .get_epoch(height)
+            .expect("The epoch of the given block height should be known")
+    };
+
+    let valset_upd_keys = vote_tallies::Keys::from(&epoch);
+    let (exists_in_storage, _) = storage.has_key(&valset_upd_keys.seen())?;
+
+    let mut seen_by = Votes::default();
+    for (address, block_height) in ext.signatures.into_keys() {
+        // TODO: more deterministic deduplication
```
batconjurer, line 91: what's wrong we non-deterministic de-duplication?

batconjurer (approved):

apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
sug0: this arbitrarily picks the first vote we iterate over in the `HashMap` of signatures. we could pick, e.g. the vote at the earliest block height instead.

james-chf: I will update this to use the earliest one by block beight (https://github.com/anoma/namada/pull/517) (sorry should have linked the TODO here to an issue)

james-chf (approved):

PR #738, opened 2022-11-04
Allow testing wasm with Native VPs

batconjurer:
This makes some small changes to the `TestTxEnv`. 
 * It enables crediting multitokens
 * It allows use to execute wasm on it. This is useful for making sure wasm works with native vps.


tests/src/vm_host_env/tx.rs
```diff
@@ -162,6 +180,22 @@ impl TestTxEnv {
             .write(&storage_key, public_key.try_to_vec().unwrap())
             .unwrap();
     }
+
+    /// Apply the tx changes to the write log and return
+    /// the set of verifiers.
```
tzemanovic, line 185:
```suggestion
    /// Apply the tx changes to the write log.
```

tzemanovic (approved):

PR #739, opened 2022-11-04
Remove `'static` lifetime constraint from `Keys`

sug0:
Based on #731

### DIFF
--- a/apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
@@ -94,7 +94,7 @@ where
         }
     }
 
-    let (tally, changed) = if !exists_in_storage {
+    let (tally, changed, confirmed) = if !exists_in_storage {
         tracing::debug!(
             %valset_upd_keys.prefix,
             ?ext.voting_powers,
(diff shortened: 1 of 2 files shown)


batconjurer (approved):

apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
```diff
@@ -31,29 +34,116 @@ where
...
+    let epoch = {
+        // all votes we gathered are for the same epoch, so
+        // we can just fetch the block height from the first
+        // signature we iterate over, and calculate its cor-
+        // responding epoch
+        let height = ext
+            .signatures
+            .keys()
+            .map(|(_, height)| *height)
+            .by_ref()
```
james-chf, line 74: do we need this `.by_ref()`?

james-chf (approved):

apps/src/lib/node/ledger/protocol/transactions/validator_set_update/mod.rs
sug0: probably not. it creates a `&mut` reference. I thought we would need it to manually call `next()` on the iterator, without previously giving it some `let` binding

PR #751, opened 2022-11-07
Don't re-export items from the traits submodule in the storage module

james-chf:
This PR will be redone after https://github.com/anoma/namada/pull/733

The motivation for this PR is to make it so that `Sha256Hasher` and `StorageHasher` are only importable in exactly one way outside of the `crate::ledger::storage` module. Currently, these two types are importable directly from `crate::ledger::storage::traits` but also re-exported in `crate::ledger::storage` as well, meaning they are arbitrarily imported throughout the codebase in different ways which can lead to merge conflicts. i.e.

https://github.com/anoma/namada/blob/6cbce4a4dcdd3811679f5eb4aa7c82f51dff74e6/shared/src/ledger/storage/mod.rs#L27

This PR is currently changing the imports to work like so:
- `Sha256Hasher` moved from `crate::ledger::storage::traits` to `crate::ledger::storage` (so now only importable via `crate::ledger::storage`)
- `StorageHasher` is now only importable via `crate::ledger::storage::traits`


james-chf:
pls update wasm

james-chf:
pls update wasm

PR #754, opened 2022-11-07
Fix compiling with the `abcipp` feature flag

james-chf:


tzemanovic (approved):

james-chf:
I've force pushed a rebase onto namada v0.10.1. The `namada_tests` crate is not compiling with `abcipp` and looks like it needs a shim, so I'm excluding getting that crate compiling from this PR as it may be more involved. Will attempt to try this PR out against the v0.11 draft branch (https://github.com/anoma/namada/pull/793)

Issue #515, opened 2022-09-27
Don't apply duplicate vote extensions in FinalizeBlock


james-chf:
Relates to https://github.com/anoma/namada/issues/511

When we apply a derived transaction based on a vote extension digest in FinalizeBlock, if the digest contains multiple vote extensions from the same validator, apply only the earliest one and ignore the rest.

- [ ] fix the code
- [ ] update the specs

PR #760, opened 2022-11-08
Always use the earliest vote by a validator, when they have multiple votes within the same digest

james-chf:
(redone based off of PR https://github.com/anoma/namada/pull/517 since v0.9.0 and PR https://github.com/anoma/namada/pull/676 merged)

Relates to https://github.com/anoma/namada/issues/515

When applying votes from an Ethereum event digest for an Ethereum event to storage during `FinalizeBlock`, we take the earliest vote by block height we encounter for each validator, and use that as the validator's vote. Before, we were deterministically deduping but were using the *latest* vote by block height.


shared/src/ledger/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -157,17 +157,9 @@ where
     // is a less arbitrary way to do this
     let (exists_in_storage, _) = storage.has_key(&eth_msg_keys.seen())?;
 
-    let mut seen_by = Votes::default();
-    for (address, block_height) in update.seen_by.into_iter() {
-        // TODO: more deterministic deduplication
-        if let Some(present) = seen_by.insert(address, block_height) {
-            tracing::warn!(?present, "Duplicate vote in digest");
-        }
-    }
```
james-chf, line 166: This block actually wasn't necessary since we switched `seen_by` to be a `BTreeMap<Address, BlockHeight>` already

shared/src/ledger/protocol/transactions/ethereum_events/eth_msgs.rs
```diff
@@ -31,7 +31,7 @@ impl From<MultiSignedEthEvent> for EthMsgUpdate {
     ) -> Self {
         Self {
             body: event,
-            seen_by: signers.into_iter().collect(),
```
james-chf, line 34: This was already deterministically deduping though taking the latest vote rather than the earliest one.

shared/src/ledger/protocol/transactions/votes.rs
```diff
@@ -121,3 +121,104 @@ where
     storage.write(&keys.voting_power(), &tally.voting_power.try_to_vec()?)?;
     Ok(())
 }
+
+/// Deterministically constructs a [`Votes`] map from a set of validator
+/// addresses and the block heights they signed something at. We arbitrarily
+/// take the earliest block height for each validator address encountered.
+pub fn dedupe(signers: BTreeSet<(Address, BlockHeight)>) -> Votes {
+    signers.into_iter().rev().collect()
```
james-chf, line 129: I had some involved chained iterators here then realized we can just use the natural `signers.into_iter().rev().collect()`.

As I understand it, it shouldn't make a difference whether we use the earliest or latest vote by a validator for an event within a digest (especially since we don't guarantee that we will take the globally earliest/latest vote in any case) - the important thing is that we only use exactly one vote per validator per event all in all. I am guessing validators should prefer their earlier votes to be used, since in an ideal world (i.e. ABCI++) their first vote for an event would always be recorded and they wouldn't have to submit multiple votes at all.

sug0:
https://github.com/anoma/namada/pull/683#discussion_r1016494322

james-chf:
Will add the test and also rebase/resolve conflicts with latetst `eth-bridge-integration` (e2e tests should start running again also 🤞)

james-chf:
I change `GetVoters` to be implemented for `HashSet<EthMsgUpdate>` instead of `[MultiSignedEthEvent]`, since we do the deduping when we convert from `MultiSignedEthEvent` -> `EthMsgUpdate`

james-chf:
pls update wasm

batconjurer (approved):

shared/src/ledger/protocol/transactions/ethereum_events/mod.rs
sug0: I think this is also present in the validator set update code. could you update it in this PR, as well? I basically just copied and pasted your eth events code and modified it a bit

sug0 (changes requested):

shared/src/ledger/protocol/transactions/ethereum_events/mod.rs
james-chf: Leaving this as is for now but we should revisit later the `ValidatorSetUpdateVextDigest` as we should always be able to construct it without duplicates anyway

james-chf: Will quickly update this TODO

sug0 (approved):

PR #765, opened 2022-11-09
Fix up `build-and-test-bridge.yml`

james-chf:
This PR reduces the diff between the `build-and-test.yml` (which runs for `main`) and `build-and-test-bridge.yml` (which runs for `eth-bridge-integration`). There was a duplicate @ that was fixed in `build-and-test.yml` that is also fixed in `build-and-test-bridge.yml` as part of this PR.


james-chf:
One of the CI jobs failed here but not sure why - it shouldn't be related to this PR - https://github.com/anoma/namada/actions/runs/3427417380/jobs/5710398811

PR #774, opened 2022-11-11
Use `BlockSpaceAllocator` to construct new blocks during PrepareProposal

sug0:
Based on #791

Plugs the block space allocator implementation into `PrepareProposal`.


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -56,21 +63,37 @@ where
...
-            let mut mempool_txs = self.build_mempool_txs(req.txs);
+            let (mut mempool_txs, alloc) =
+                self.build_mempool_txs(alloc, &req.txs);
             txs.append(&mut mempool_txs);
 
-            // decrypt the wrapper txs included in the previous block
-            let mut decrypted_txs = self.build_decrypted_txs();
-            txs.append(&mut decrypted_txs);
+            // fill up the remaining block space with
+            // arbitrary transactions that can fit in
```
batconjurer, line 92: again, this should not be arbitrary txs, but protocol.

Makefile
```diff
@@ -25,6 +25,9 @@ build-test:
 build-release:
 	ANOMA_DEV=false $(cargo) build --release --package namada_apps --manifest-path Cargo.toml
 
+build-debug:
+	ANOMA_DEV=false $(cargo) build --package namada_apps --manifest-path Cargo.toml
+
```
james-chf, line 30: This is useful though would separate it out from this PR, it should go towards `main` (maybe `build-apps-release` and `build-apps-debug` would make more sense as target names but it's separate from this PR)

shared/src/hints.rs
```diff
@@ -0,0 +1,44 @@
...
+///     // most common path to take
+/// }
+/// ```
+#[inline]
+pub fn unlikely(b: bool) -> bool {
+    if b {
+        cold()
+    }
+    b
+}
```
james-chf, line 44: Should go towards `main`, idk much about this. I'd be slightly hesitant to have optimizations like this at this stage (we should be able to do this sort of thing by profiling anyway - https://doc.rust-lang.org/rustc/profile-guided-optimization.html ?).

shared/src/types/index_set.rs
```diff
@@ -0,0 +1,99 @@
+//! Set data structure optimized to store [`usize`] values.
+
+use std::collections::BTreeMap;
+
+/// The storage unit for the bits in an [`IndexSet`].
+#[allow(dead_code)]
```
james-chf, line 6: this module is annotated throughout with `dead_code`, is it replaced by https://github.com/heliaxdev/index-set now?

shared/src/ledger/storage_api/queries.rs
```diff
@@ -126,10 +126,12 @@ pub trait QueriesExt {
 
     /// Determines if it is possible to send a validator set update vote
     /// extension at the provided [`BlockHeight`] in [`SendValsetUpd`].
-    ///
-    /// This is done by checking if we are at the second block of a new epoch.
     fn can_send_validator_set_update(&self, can_send: SendValsetUpd) -> bool;
 
+    /// Check if we are at a given [`BlockHeight`] offset, `height_off`, within
+    /// the current [`Epoch`].
+    fn is_deciding_offset_within_epoch(&self, height_off: u64) -> bool;
```
james-chf, line 133: I think `height_offset` is clearer as a parameter name, it's OK that it is a bit longer

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -38,15 +45,15 @@ where
...
     /// Begin a new block.
     ///
-    /// We include half of the new wrapper txs given to us from the mempool
-    /// by tendermint. The rest of the block is filled with decryptions
-    /// of the wrapper txs from the previously committed block.
+    /// Block space is roughly divided into three equal parts, which we call
+    /// allotments. In the following order, each allotted area is home to:
+    /// decrypted, protocol and encrypted transactions, respectively. During
+    /// some protocol phases, we might omit encrypted and decrypted transactions
+    /// entirely, such as when we negotiate DKG parameters.
```
james-chf, line 52: nit: would instead link to the documentation for the `block_space_alloc` module, as a single source of truth currently for how this works

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -204,10 +357,107 @@ where
...
+                    |()| true,
+                )
+            })
             .collect()
     }
 }
 
+/// Return a list of the protocol transactions that haven't
+/// been marked for inclusion in the block, yet.
+fn get_remaining_txs(
```
james-chf, line 438: nit: could be named `get_remaining_protocol_txs` if it only relates to protocol txs? (similarly, the `tx_indices` param could be named `protocol_tx_indices`)

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -143,58 +168,186 @@ where
...
     }
 
     /// Builds a batch of vote extension transactions, comprised of Ethereum
     /// events and, optionally, a validator set update
     #[cfg(not(feature = "abcipp"))]
-    fn build_vote_extension_txs(&mut self, txs: &[TxBytes]) -> Vec<TxBytes> {
-        if self.storage.last_height != BlockHeight(0) {
-            deserialize_vote_extensions(txs).collect()
-        } else {
+    fn build_vote_extension_txs(
```
james-chf, line 189: I think it would be more consistent for this to be named `build_protocol_txs` now

james-chf (changes requested):
lgtm

shared/src/hints.rs
sug0: I admit I was having way too much fun when I added this module haha

shared/src/types/index_set.rs
sug0: ohhh shit, yeah we should get rid of this!

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -56,21 +63,37 @@ where
         // proposal is accepted
         self.gas_meter.reset();
         let txs = if let ShellMode::Validator { .. } = self.mode {
-            // TODO: add some info logging?
+            // start counting allotted space for txs
+            let alloc = BlockSpaceAllocator::from(&req);
+            let mut tx_indices = VecIndexSet::default();
```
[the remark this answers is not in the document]
sug0: yeah that's true. makes more sense to restrict the domain a bit

apps/src/lib/node/ledger/shell/prepare_proposal.rs
sug0: sure

sug0: note to self: probably also rename `build_mempool_txs` to `build_encrypted_txs`, since we also fetch protocol txs from the mempool

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -143,58 +168,186 @@ where
...
-            vec![]
+            return (vec![], self.get_encrypted_txs_allocator(alloc));
         }
+
+        let txs = deserialize_vote_extensions(txs, tx_indices).take_while(|tx_bytes|
+            alloc.try_alloc(&*tx_bytes)
+                .map_or_else(
+                    |status| match status {
+                        AllocFailure::Rejected { bin_space_left } => {
+                            // TODO: maybe we should find a way to include
```
batconjurer, line 205: We could also add prioritization to protocol txs

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,505 @@
...
+//! of bytes utilized by some Tendermint consensus round's proposal.
+//!
+//! This is important, because Tendermint places an upper bound
+//! on the size of a block, rejecting blocks whose size exceeds
+//! the limit stated in [`RequestPrepareProposal`].
+//!
+//! The code in this module doesn't perform any deserializing to
+//! verify if we are, in fact, allocating space for the correct
+//! kind of tx for the current [`BlockSpaceAllocator`] state. It
+//! is up to the user to dispatch the correct kind of tx into the
```
batconjurer, line 11: The term `user` is confusing here. Maybe just say this logic is handled in the prepare proposal call.

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
sug0: yeah that's what I was thinking of doing, at the `CheckTx` level, but that's for another PR

james-chf (approved):

batconjurer (approved):

PR #778, opened 2022-11-14
Implement `BlockSpaceAllocator` state machine

sug0:
Separated from #774

Allocate space for the different `TxBin` instances depending on the current phase of a type-level state machine. This will be useful to allow, e.g. decrypted txs to use as much space as needed, or to disallow including user encrypted txs in blocks when space is a critical resource needed for protocol level txs.

Unit tests will be added in a new PR, to simplify the review process.


batconjurer (commented):
Just one question and a couple of tinies.

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/decrypted_txs.rs
```diff
@@ -0,0 +1,44 @@
...
+impl NextStateImpl for BlockSpaceAllocator<BuildingDecryptedTxBatch> {
+    type Next = BlockSpaceAllocator<BuildingProtocolTxBatch>;
+
+    #[inline]
+    fn next_state_impl(mut self) -> Self::Next {
+        self.decrypted_txs.shrink();
+
+        // reserve space for protocol txs
+        let remaining_free_space = self.uninitialized_space_in_bytes();
+        self.protocol_txs =
```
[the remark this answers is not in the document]
sug0: yeah, but then we need another 1/3 for mempool txs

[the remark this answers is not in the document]
batconjurer: What I'm confused about is that is this 1/3 of the total block (which is what I expected), or 1/3 of the block minus the space for decrypted txs. If the latter, it would seem that decrypted gets 1/3, protocal 2/9 and mempool 4/9.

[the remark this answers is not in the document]
sug0: 1/3 of the total block. during the previous block, we assume we allocated about 1/3 of the space for encrypted txs, therefore decrypted txs used up 1/3 of the space. now, the remaining space should be .... fuck I see what you mean. craaap

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,254 @@
...
+//! current state of the allocator.
+//!
+//! # How space is allocated
+//!
+//! In the current implementation, we allocate space for transactions
+//! in the following order of preference:
+//!
+//! - First, we allocate space for DKG decrypted txs.
+//! - Next, we allocate space for protocol txs. Protocol txs get 1/3 of the
+//!   block space allotted to them.
```
james-chf, line 21:
```suggestion
//! - Next, we allocate space for protocol txs. 
```
"Protocol txs get 1/3 of the block space allotted to them" - this is a bit ambiguous to me? Should it be that protocol txs get allocated 1/3 of the total block space?

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,254 @@
...
+//! - Finally, we allocate space for DKG encrypted txs.
+//! - If any space remains, we try to fit other smaller txs in the block.
+//!
+//! Since decrypted txs will utilize at most as much space as
+//! encrypted txs will utilize, and we allocate 1/3 of space
```
james-chf, line 26: "Since decrypted txs will utilize at most as much space as encrypted txs will utilize" - encrypted txs being included in this block, or encrypted txs from the previous block height (or both)?

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,254 @@
...
+// this is that then we may not have enough space for decrypted txs
+
+// TODO: panic if we don't have enough space reserved for a
+// decrypted tx; in theory, we should always have enough space
+// reserved for decrypted txs, given the invariants of the state
+// machine
+
+// TODO: refactor our measure of space to also reflect gas costs.
+// the total gas of all chosen txs cannot exceed the configured max
+// gas per block, otherwise a proposal will be rejected!
```
james-chf, line 51: these could have GH issues

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,254 @@
...
+            let space_left = self.space_left_in_bytes();
+            AllocStatus::Rejected { tx, space_left }
+        }
+    }
+
+    /// Try to dump a new batch of transactions into this [`TxBin`].
+    ///
+    /// If an allocation fails, rollback the state of the [`TxBin`],
+    /// and return the respective status of the failure.
+    fn try_dump_all<'tx, T>(&mut self, txs: T) -> AllocStatus<'tx>
```
james-chf, line 222: Honestly I think it would be better not to try to share the same return type `AllocStatus` between this `try_dump_all` and `try_dump`, they are different enough. Returning `&tx` in `try_dump` isn't necessary, for `try_dump_all` it would be useful to know how far along the iteration we got (i.e. the "index" of the failing tx). I guess the caller can work this out given `&tx`, if they can reconstruct the iterator. I can suggest a code change here later maybe

james-chf (commented):
The PRs are in draft

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
sug0: `try_dump_all` will be removed. it was only necessary to handle protocol txs, whose digests and vote extensions had to be allocated as a single unit. we're getting rid of digests, though.

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
sug0:
```
allotted  -> tendermint block space chunk
allocated -> bin space chunk
```
we allot tendermint block space chunks to tx bins, and allocate tx bin chunks for txs

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
sug0: encrypted txs from the previous block

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,245 @@
...
+    #[inline]
+    fn shrink_to_fit(&mut self) {
+        self.allotted_space_in_bytes = self.occupied_space_in_bytes;
+    }
+
+    /// Try to dump a new transaction into this [`TxBin`].
+    ///
+    /// Signal the caller if the tx is larger than its max
+    /// allotted bin space.
+    fn try_dump<'tx>(&mut self, tx: &'tx [u8]) -> AllocStatus<'tx> {
```
james-chf, line 200: We don't need to return the `tx` field in `AllocStatus` anymore, we'd just be passing back to the caller what they gave us

james-chf: Is dumping the same as allocating a tx? Otherwise could this be called `try_alloc` instead of `try_dump` (especially since it returns `AllocStatus`)? Or will this eventually put the transaction bytes somewhere?

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.encrypted_txs = TxBin::default();
+    }
+}
+
+/// Allotted space for a batch of transactions of the same kind in some
+/// proposed block, measured in bytes.
+#[derive(Debug, Copy, Clone, Default)]
+struct TxBin {
+    /// The current space utilized by the batch of transactions.
+    occupied_space_in_bytes: u64,
```
james-chf, line 157: wrt. this PR it looks like `occupied_space_in_bytes` == `allocated_space_in_bytes`, is there a reason to use the term "occupied" here? (it would be easier imo to use the same terms wherever we can if there is no difference)

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
james-chf: it would be worth being explicit here I think that it is the encrypted txs from the previous block

james-chf (changes requested):
will look at https://github.com/anoma/namada/pull/774 to see how this PR is used

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
james-chf: sort of related to this comment - https://github.com/anoma/namada/pull/778#discussion_r1033597447

sug0: well, I called this `used_space` before, but you suggested I changed this to `occupied_space` :joy:

sug0: but yeah, `allocated_space` is better

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,249 @@
...
+    #[inline]
+    fn shrink_to_fit(&mut self) {
+        self.allotted_space_in_bytes = self.occupied_space_in_bytes;
+    }
+
+    /// Try to dump a new transaction into this [`TxBin`].
+    ///
+    /// Signal the caller if the tx is larger than its max
+    /// allotted bin space.
+    fn try_dump(&mut self, tx: &[u8]) -> Result<(), AllocFailure> {
```
batconjurer, line 204: Could rename this error as `Constipation`. :hankey:

sug0: legit lol'd at this

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states.rs
```diff
@@ -0,0 +1,168 @@
...
+//!    this state, we populate a block with protocol txs.
+//! 3. [`BuildingEncryptedTxBatch`] - the third state. In
+//!    this state, we populate a block with DKG encrypted txs.
+//!    This state supports two modes of operation, which you can
+//!    think of as two states diverging from [`BuildingProtocolTxBatch`]:
+//!   * [`WithoutEncryptedTxs`] - When this mode is active, no encrypted txs are
+//!     included in a block proposal.
+//!   * [`WithEncryptedTxs`] - When this mode is active, we are able to include
+//!     encrypted txs in a block proposal.
+//! 4. [`FillingRemainingSpace`] - the fourth and final state.
```
batconjurer, line 21: We shouldn't include arbitrary transactions, as per our previous discussions. If we put in encrypted txs, we could accidentally fill more then 1/3 of the block with encrypted txs. We should only take extra protocol txs. And if there is space left, that's life.

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,249 @@
...
+//! In the current implementation, we allocate space for transactions
+//! in the following order of preference:
+//!
+//! - First, we allot space for DKG decrypted txs. Decrypted txs take up as much
+//!   space as needed. We will see, shortly, why in practice this is fine.
+//! - Next, we allot space for protocol txs. Protocol txs get half of the
+//!   remaining block space allotted to them.
+//! - Finally, we allot space for DKG encrypted txs. We allow DKG encrypted txs
+//!   to take up at most 1/3 of the total block space.
+//! - If any space remains, we try to fit other smaller txs in the block.
```
batconjurer, line 25: See my detailed comment below on this point.

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// The leader of the current Tendermint round is populating
+/// all remaining space in a block proposal with arbitrary
+/// transactions.
+///
+/// For more info, read the module docs of
+/// [`crate::node::ledger::shell::prepare_proposal::block_space_alloc::states`].
+pub struct FillingRemainingSpace<Mode> {
+    /// One of [`WithEncryptedTxs`] and [`WithoutEncryptedTxs`].
```
batconjurer, line 66: This should probably be renamed to `WithProtocolTxs`, etc.

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states.rs
sug0: ah, this is being enforced in #774 - https://github.com/anoma/namada/blob/fb92431636a2f4ecbc22ef5e33992c300c7e4fbc/apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/remaining_txs.rs#L27-L47

sug0: albeit now that I think about, there's still a bug here. we could be violating the 1/3 ratio

sug0: alright fair enough, we'll limit the last state to only include new protocol txs

james-chf (approved):

PR #779, opened 2022-11-14
Murisi/faster masp e2e1

murisi:
Cut down the duration of the E2E tests by storing precomputed MASP transactions. A potential temporary fix while the MASP unit tests are being made. The advantages of this approach are as follows:
* The tests exercise the code in the MASP validity predicate
* The tests fully simulate (including full ZK-proofs) typical usage of the MASP client
* No changes to the client/stubs were required to enable it to be tested
* The acceptance of actual specific binary encodings of MASP transactions is tested

The disadvantages of this approach are as follows:
* The addresses of Albert, Bertha, and Christel must be fixed in `e2e-tests-single-node.toml` (as they occur in binary)
* The tests depend on a particular sequence of conversion tree states/anchors (i.e. 7 assets ...)
* If shielded transaction formats or reward function is changed, updating MASP tx data binaries is slow/cumbersome


cwgoes:
Superseded by merged work.

PR #782, opened 2022-11-15
CI: general improvements

Fraccaman:
- [x] parallel e2e
- [x] update mold to 1.7.0
- [x] added timing report to gh artifacts
- [x] added coverage report to gh artifacts

### DIFF
--- a/.github/workflows/build-and-test-bridge.yml
+++ b/.github/workflows/build-and-test-bridge.yml
@@ -29,12 +29,13 @@ jobs:
     timeout-minutes: 30
     runs-on: ${{ matrix.os }}
     container: 
-      image: ghcr.io/anoma/namada:wasm-0.6.1
+      image: ghcr.io/anoma/namada:wasm-0.8.0
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        wasm_cache_version: ["v1"]
+        wasm_cache_version: ["v2"]
+        mold_version: [1.7.0]
 
     steps:
       - name: Checkout repo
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test-bridge.yml
```diff
@@ -66,13 +65,49 @@ jobs:
...
+        with:
+          name: wasm-${{ github.event.pull_request.head.sha|| github.sha }}
+          path: ./wasm
+        env:
+          RUSTFLAGS: "-C linker=clang -C link-arg=-fuse-ld=/usr/local/bin/mold"
       - name: Test Wasm
         run: make test-wasm
       - name: Check wasm up-to-date
-        run: cmp -- wasm/checksums.json wasm/original-checksums.json
+        run: cmp -- wasm/checksums.json wasm/original-checksums.json || true
```
tzemanovic, line 107: will this still catch `if: failure()` below?

.github/workflows/scripts/e2e.json
```diff
@@ -0,0 +1,23 @@
+{
```
tzemanovic, line 1: what's the numbers here? I see it's being used as 'total_time' below, but I don't follow why and where it comes from

.github/workflows/build-and-test-bridge.yml
Fraccaman: oh my bad here, the bridge team asked me to let the CI pass when if the checksum is not up-to-date. But this is the wrong workflow file, ill fix this! Thanks

.github/workflows/scripts/e2e.json
Fraccaman: they are taken from `cargo test` output, an example here https://github.com/anoma/namada/actions/runs/3469076888/jobs/5795628173

Fraccaman: this means that we will have to update this file everytime we add a new e2e test, which I thinks its reasonable and can be automated later

tzemanovic (approved):

Issue #767, opened 2022-11-09
End-to-end test failures in CI when more than one validator is involved in `eth-bridge-integration`


james-chf:
e.g. `test_node_connectivity`, `double_signing_gets_slashed`, `test_genesis_validators` - these tests can pass locally but may fail in CI.

We see error logs in a validator about not being able to verify a vote extension signature, at least for `double_signing_gets_slashed`. It's not clear though why this should fail.

```
2022-11-09T13:48:56.175159Z  WARN namada_apps::node::ledger::shell::vote_extensions::eth_events: Overwrote old signature from validator while constructing ethereum_events::VextDigest sig=Ed25519(Signature(Signature { R_bytes: "a63c20465b3f790c96bc229930f3681d406d0a5852427f4466d2244f975090e7", s_bytes: "775648ee071427e93b60fd3cba31426fdbf07fc90b06cf589d96272ee0a20e06" })) validator_addr=Established: atest1v4ehgw36g3rrgs3j8quy233ng3znv33cx4znss3cg4zrysecgymy2wzzxsmrxwp5xcergdenmzdz0m
2022-11-09T13:48:56.177936Z  INFO namada_apps::node::ledger::shell::prepare_proposal: Proposing block height=3 tx_records=7
I[2022-11-09|13:48:56.182] received proposal                            module=consensus proposal="Proposal{3/2 (C47E47F3853024BAD30AAB52A0A47731D0A00D92DA1E5F204C96C7EC982AA1FA:2:589FFDBCAF38, -1) AE531B8A6EE1 @ 2022-11-09T13:48:56.179599527Z}"
I[2022-11-09|13:48:56.185] received complete proposal block             module=consensus height=3 hash=C47E47F3853024BAD30AAB52A0A47731D0A00D92DA1E5F204C96C7EC982AA1FA
2022-11-09T13:48:56.186226Z  INFO namada_apps::node::ledger::shell::process_proposal: Received block proposal proposer="2F31AF119302FB303932BBE22CC3E24DC70D25FB" height=3 hash="C47E47F3853024BAD30AAB52A0A47731D0A00D92DA1E5F204C96C7EC982AA1FA" n_txs=7
[...]

PR #786, opened 2022-11-15
Fix end-to-end test flakes involving more than one validator

james-chf:
Closes https://github.com/anoma/namada/issues/767 (current e2e flakes in CI)

The issue was that `VotingPowersMap::get_abi_encoded` wasn't completely deterministic if there were validators present in the `HashMap` with the same voting power. Interestingly, this issue only occurred in tests run with the `--release` profile.

This PR:
- makes `VotingPowersMap::get_abi_encoded` deterministic
- adds some test coverage regarding the ordering of `VotingPowersMap` items
- updates some docstrings to mention requirements around determinism

### DIFF
--- a/shared/src/proto/types.rs
+++ b/shared/src/proto/types.rs
@@ -52,8 +52,10 @@ pub trait SignedSerialize<T> {
     /// A byte vector containing the serialized data.
     type Output: AsRef<[u8]>;
 
-    /// Encodes `data` as a byte vector,
-    /// with some arbitrary serialization method.
+    /// Encodes `data` as a byte vector, with some arbitrary serialization
+    /// method. The returned output *must* be deterministic based on
+    /// `data`, so that two callers signing the same `data` will be
+    /// signing the same `Self::Output`.
     fn serialize(data: &T) -> Self::Output;
 }
 
--- a/shared/src/proto/types.rs
+++ b/shared/src/proto/types.rs
@@ -53,7 +53,9 @@ pub trait SignedSerialize<T> {
     type Output: AsRef<[u8]>;
 
     /// Encodes `data` as a byte vector, with some arbitrary serialization
-    /// method. The returned output *must* be deterministic based on
+    /// method.
+    ///
+    /// The returned output *must* be deterministic based on
     /// `data`, so that two callers signing the same `data` will be
     /// signing the same `Self::Output`.
     fn serialize(data: &T) -> Self::Output;
--- a/shared/src/types/vote_extensions/validator_set_update.rs
+++ b/shared/src/types/vote_extensions/validator_set_update.rs
@@ -1,5 +1,6 @@
 //! Contains types necessary for processing validator set updates
 //! in vote extensions.
+use std::cmp::Ordering;
 use std::collections::HashMap;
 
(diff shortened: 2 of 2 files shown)


james-chf:
Changing VotingPowersMap to use a BTreeMap instead of a HashMap (https://github.com/anoma/namada/pull/786/commits/1108d757f9d96b3c0969abbc5186be809e115b35) seemed to result in a couple good CI runs, but this might have been by chance. Our valset upd signatures are dependent on the ABI encoding of `VotingPowersMap`, which should be deterministic regardless of map type used - https://github.com/anoma/namada/blob/1616b662d70099c4f331b375b8ab79d9c9535c3f/shared/src/types/vote_extensions/validator_set_update.rs#L196

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -370,4 +387,89 @@ mod tests {
...
+        voting_powers_1.insert(validator_b.clone(), same_voting_power);
+
+        let mut voting_powers_2 = VotingPowersMap::default();
+        voting_powers_2.insert(validator_b, same_voting_power);
+        voting_powers_2.insert(validator_a, same_voting_power);
+
+        let x = voting_powers_1.get_abi_encoded();
+        let y = voting_powers_2.get_abi_encoded();
+        assert_eq!(x, y);
+    }
```
james-chf, line 474: Having this test that would only fail in `--release` mode if broken is not a great solution. I think we should split out `get_abi_encoded` into a fn that works with `&mut [(&EthAddrBook, &VotingPower)]` directly, then we can easily test that with vectors directly, rather than relying on `HashMap::iter`

shared/src/proto/types.rs
```diff
@@ -52,8 +52,10 @@ pub trait SignedSerialize<T> {
     /// A byte vector containing the serialized data.
     type Output: AsRef<[u8]>;
 
-    /// Encodes `data` as a byte vector,
-    /// with some arbitrary serialization method.
+    /// Encodes `data` as a byte vector, with some arbitrary serialization
+    /// method. The returned output *must* be deterministic based on
+    /// `data`, so that two callers signing the same `data` will be
+    /// signing the same `Self::Output`.
```
sug0, line 58: I would start a new paragraph for the portion of the docstring you added

shared/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -192,17 +197,29 @@ pub trait VotingPowersMapExt {
...
+fn compare_voting_powers_map_items(
+    first: &(&EthAddrBook, &VotingPower),
+    second: &(&EthAddrBook, &VotingPower),
+) -> Ordering {
+    let (first_power, second_power) = (first.1, second.1);
+    let (first_addr, second_addr) = (first.0, second.0);
+    match second_power.cmp(first_power) {
+        Ordering::Less => Ordering::Less,
+        Ordering::Equal => first_addr.cmp(second_addr),
+        Ordering::Greater => Ordering::Greater,
```
sug0, line 211:
```suggestion
        Ordering::Equal => first_addr.cmp(second_addr),
        ordering => ordering,
```

sug0 (approved):

batconjurer (approved):

PR #788, opened 2022-11-15
PoS spec fixes

brentstone:
Upgrading the cubic slashing description in the Namada specs. Cleaning some other lazy text fragments in other parts of the PoS specs.

TODO:
**Thinking of leaving the below tasks for another PR so as not to have this one open for too long while other Spec-related PRs are opened. Probably good to just try to merge what is here now soon.**

- may need to think more precisely about how we handle a validator who commits two infraction within an epoch of each other, as the sum total voting power fraction could exceed 1/3 if the validator voting power is large enough.
- #815 
- make changes to `bonding-mechanism.md` following #740, other things


cwgoes (approved):
ACK - thanks

drewstone:
hell yea

cwgoes:
> hell yea

what's this drive-by commenting? :smile:

PR #791, opened 2022-11-16
Add `BlockSpaceAllocator` unit tests

sug0:
Based on #778

Add unit tests for the `BlockSpaceAllocator` state machine and `TxBin` implementations.


apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+}
+
+impl<State> BlockSpaceAllocator<State> {
+    /// Return uninitialized space in tx bins, resulting from ratio conversions.
+    ///
+    /// This method should not be used outside of [`BlockSpaceAllocator`]
+    /// instance construction or unit testing.
+    #[inline]
+    fn uninitialized_space_in_bytes(&self) -> u64 {
```
james-chf, line 129: What about `free_space_in_bytes` or `remaining_space_in_bytes` instead of `uninitialized_space_in_bytes`? I'm not clear on the differences between these terms: allotted, allocating, initialized - maybe there is some overlap between them

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/decrypted_txs.rs
```diff
@@ -0,0 +1,51 @@
...
+impl NextStateImpl for BlockSpaceAllocator<BuildingDecryptedTxBatch> {
+    type Next = BlockSpaceAllocator<BuildingProtocolTxBatch>;
+
+    #[inline]
+    fn next_state_impl(mut self) -> Self::Next {
+        self.decrypted_txs.shrink();
+
+        // reserve space for protocol txs
+        let uninit = self.uninitialized_space_in_bytes();
+        self.protocol_txs = TxBin::init_over_ratio(uninit, thres::ONE_THIRD);
```
james-chf, line 32: It looks like we give protocol txs 1/3 of the remaining space after DKG decrypted txs have been allocated, but then DKG decrypted txs could take up the whole block potentially? Should we reserve space upfront for protocol txs instead?

https://github.com/anoma/namada/blob/dc86779f41b8a17d0d8734a911154b0123c585d4/apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs#L108-L120

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/protocol_txs.rs
```diff
@@ -0,0 +1,87 @@
...
+    WithEncryptedTxs, WithoutEncryptedTxs,
+};
+
+impl State for BlockSpaceAllocator<BuildingProtocolTxBatch> {
+    #[inline]
+    fn try_alloc(&mut self, tx: &[u8]) -> AllocStatus {
+        // TODO: prioritize certain kinds of protocol txs;
+        // this can be done at the `CheckTx` level,
+        // we don't need the `TxBin`s to be aware
+        // of different prioriy hints for protocol txs
```
james-chf, line 15: Yeah this will be it's own issue. I don't think we can prioritize at the `CheckTx` level in Tendermint v0.37, so we can only prioritize the protocol txs given to use by Tendermint within the context of an individual `RequestPrepareProposal`.

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/protocol_txs.rs
```diff
@@ -0,0 +1,87 @@
...
+    for BlockSpaceAllocator<BuildingProtocolTxBatch>
+{
+    type Next = BlockSpaceAllocator<BuildingEncryptedTxBatch<WithEncryptedTxs>>;
+
+    #[inline]
+    fn next_state_impl(mut self) -> Self::Next {
+        self.protocol_txs.shrink();
+
+        // reserve space for encrypted txs
+        let free_space = self.uninitialized_space_in_bytes();
```
james-chf, line 38: `free_space` as used here makes more sense to me than `uninit_space` as used in `decrypted_txs.rs`

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/encrypted_txs.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn try_alloc_batch<'tx, T>(&mut self, _txs: T) -> AllocStatus
+    where
+        T: IntoIterator<Item = &'tx [u8]> + 'tx,
+    {
+        AllocStatus::Rejected {
+            // arbitrary `tx_len` value; doesn't really matter what we
+            // choose here, as long as it's greater than zero
+            tx_len: u64::MAX,
+            space_left: 0,
+        }
```
james-chf, line 56: This is kind of a code smell though, relating to the comment at https://github.com/anoma/namada/pull/791/files#diff-91ec5a00b45ee4a93117e944d2cdd7c12774c3e0ae852489ea296a6055a77836R231 i.e. `AllocStatus` isn't an appropriate return type for this `try_alloc_batch` fn (maybe it would be cleaner to just let callers do iteration if they need to)

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,428 @@
...
+        self.encrypted_txs.current_space_in_bytes = 0;
+    }
+}
+
+/// Allotted space for a batch of transactions of the same kind in some
+/// proposed block, measured in bytes.
+#[derive(Debug, Copy, Clone, Default)]
+struct TxBin {
+    /// The current space utilized by the batch of transactions.
+    current_space_in_bytes: u64,
```
james-chf, line 157: nit: `current_space` is a bit ambiguous still, as if it could refer to the remaining space rather than the space already being used. Could name it `utilized_space`, `occupied_space`

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states.rs
```diff
@@ -0,0 +1,173 @@
...
+//! 1. [`BuildingDecryptedTxBatch`] - the initial state. In
+//!    this state, we populate a block with DKG decrypted txs.
+//! 2. [`BuildingProtocolTxBatch`] - the second state. In
+//!    this state, we populate a block with protocol txs.
+//! 3. [`BuildingEncryptedTxBatch`] - the third state. In
+//!    this state, we populate a block with DKG encrypted txs.
+//!    This state supports two modes of operation, which you can
+//!    think of as two states diverging from [`BuildingProtocolTxBatch`]:
+//!   * [`WithoutEncryptedTxs`] - When this mode is active, no encrypted txs are
+//!     included in a block proposal.
```
james-chf, line 18: When would we expect to use this `WithoutEncryptedTxs` mode? Since currently all regular txs given to us by Tendermint are "encrypted"

james-chf (changes requested):
Taken a first look, my understanding of things might not be completely correct

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc/states/protocol_txs.rs
james-chf: or `remaining_free_space` even

sug0:
Code review suggestions applied to https://github.com/anoma/namada/pull/778

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
sug0: let's say `B` is the total space Tendermint has given the application for a certain block proposal. an individual allotment `A` is a fraction of `B`, which is the total space ceded to some tx bin. for a given allocation `a`, `a <= A`, for some tx bin `T`

apps/src/lib/node/ledger/shell/prepare_proposal/block_space_alloc.rs
```diff
@@ -0,0 +1,430 @@
...
+            decrypted_txs,
+        } = args;
+
+        // produce new txs until the moment we would have
+        // filled up the bins.
+        //
+        // iterate over the produced txs to make sure we can keep
+        // dumping new txs without filling up the bins
+
+        let bins = RefCell::new(BlockSpaceAllocator::init(
```
batconjurer, line 337: Why is the RefCell needed?

sug0: the iterator borrows the allocator immutably, then while iterating we borrow it mutably. without the `RefCell`, we'd get the borrow checker yelling at us

batconjurer (changes requested):
I would love tests that show we can fill up unused space. Also we should have some hand-coded unit tests to check that all 1/3 scenario. These tests won't catch a bug if the ratios of bin allocation are wrong (say 1/3, 2/9, 4/9).

james-chf (approved):

Issue #710, opened 2022-10-28
`namadac utils join-network` should be able to use a release .tar.gz from the command line


james-chf:
Currently when joining a network, the release .tar.gz needs to be downloadable from some server (defaults to `https://github.com/heliaxdev/anoma-network-config/releases/download`, set via the `ANOMA_NETWORK_CONFIGS_SERVER` envvar). It should be possible to instead specify the path to a local release .tar.gz to use instead.

More than anything else, this would hugely simplify using a locally generated release during local development, as currently it's necessary to spin up a fake network config server to join a chain.

james-chf:
This would be a nice to have feature, as part of the new genesis flow

PR #795, opened 2022-11-16
Let `namadac utils join-network` use a local release archive

james-chf:
Closes https://github.com/anoma/namada/issues/710

This PR would make it so that a local .tar.gz can be used for joining a network, rather than having to download it from a remote server at the same time that you run `namadac utils join-network`. e.g.

```shell
namadac utils join-network --release-archive ~/Downloads/local.8d5fac7ce84a1a7a8b8c48e2.tar.gz
```

TODO
- [ ] test with a real release downloaded from https://github.com/heliaxdev/anoma-network-config
- [ ] prereview

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -3216,7 +3216,7 @@ pub mod args {
 
     #[derive(Clone, Debug)]
     pub struct JoinNetwork {
-        pub chain_id: ChainId,
+        pub chain_id: Option<ChainId>,
         pub genesis_validator: Option<String>,
         pub pre_genesis_path: Option<PathBuf>,
         pub dont_prefetch_wasm: bool,
@@ -3224,7 +3224,7 @@ pub mod args {
 
     impl Args for JoinNetwork {
         fn parse(matches: &ArgMatches) -> Self {
-            let chain_id = CHAIN_ID.parse(matches);
+            let chain_id = CHAIN_ID_OPT.parse(matches);
             let genesis_validator = GENESIS_VALIDATOR.parse(matches);
             let pre_genesis_path = PRE_GENESIS_PATH.parse(matches);
             let dont_prefetch_wasm = DONT_PREFETCH_WASM.parse(matches);
@@ -3237,7 +3237,7 @@ pub mod args {
         }
 
         fn def(app: App) -> App {
-            app.arg(CHAIN_ID.def().about("The chain ID. The chain must be known in the https://github.com/heliaxdev/anoma-network-config repository."))
+            app.arg(CHAIN_ID_OPT.def().about("The chain ID. The chain must be known in the https://github.com/heliaxdev/anoma-network-config repository."))
                 .arg(GENESIS_VALIDATOR.def().about("The alias of the genesis validator that you want to set up as, if any."))
                 .arg(PRE_GENESIS_PATH.def().about("The path to the pre-genesis directory for genesis validator, if any. Defaults to \"{base-dir}/pre-genesis/{genesis-validator}\"."))
             .arg(DONT_PREFETCH_WASM.def().about(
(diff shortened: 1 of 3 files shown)


apps/src/lib/client/utils.rs
```diff
@@ -58,22 +59,18 @@ pub async fn join_network(
...
+                fs::canonicalize(&base_dir).await.unwrap()
+            } else {
+                eprintln!("Could not canonicalize base dir: {:?}", err);
+                cli::safe_exit(1)
+            }
         }
-    }
-    let base_dir_full = fs::canonicalize(&base_dir).await.unwrap();
-    let chain_dir = base_dir_full.join(chain_id.as_str());
+    };
```
james-chf, line 73: We do the base dir existence check a bit earlier and separately from the chain dir existence check, which has to happen later on now as we may need to look at the release archive to work out what the chain ID is

apps/src/lib/client/utils.rs
```diff
@@ -154,7 +121,69 @@ pub async fn join_network(
...
-    archive.unpack(&unpack_dir).unwrap();
+
+    let chain_id = match (chain_id, release_archive) {
+        (None, None) | (Some(_), Some(_)) => {
+            eprintln!(
+                "Exactly one of --chain-id or --release-archive should have \
+                 been provided"
+            );
+            cli::safe_exit(1)
+        }
```
james-chf, line 132: Not great. It would be good if we could use an enum instead of two `Option` types, of which we know exactly one will be not `None`.

apps/src/lib/cli.rs
```diff
@@ -2439,33 +2440,42 @@ pub mod args {
...
                 genesis_validator,
                 pre_genesis_path,
                 dont_prefetch_wasm,
             }
         }
 
         fn def(app: App) -> App {
-            app.arg(CHAIN_ID.def().about("The chain ID. The chain must be known in the https://github.com/heliaxdev/anoma-network-config repository."))
+            app.arg(CHAIN_ID_OPT.def().about("The chain ID. The chain must be known in the https://github.com/heliaxdev/anoma-network-config repository.").conflicts_with(RELEASE_ARCHIVE.name))
+                .arg(RELEASE_ARCHIVE.def().about("The path to a release archive (.tar.gz)."))
```
james-chf, line 2468: Referring to the `.tar.gz` you download from https://github.com/heliaxdev/anoma-network-config/releases as a "release archive", but maybe there is a better term

apps/src/lib/client/utils.rs
```diff
@@ -1068,3 +1106,59 @@ pub fn validator_pre_genesis_file(pre_genesis_path: &Path) -> PathBuf {
...
+/// Returns the [`ChainId`] of the chain for a release archive unpacked at
+/// `unpacked_dir`.
+fn extract_chain_id(release_dir: &Path) -> eyre::Result<ChainId> {
+    let global_config_path = release_dir
+        .join(config::DEFAULT_BASE_DIR)
+        .join(config::global::FILENAME);
+    let global_config_str = std::fs::read_to_string(global_config_path)?;
+    let global_config: GlobalConfig = toml::from_str(&global_config_str)?;
+    Ok(global_config.default_chain_id)
+}
```
james-chf, line 1135: Maybe there is a better way to work out the chain ID of a release archive than looking at the `global-config.toml` inside it. This is also assuming that a release archive will contain exactly one chain (I think that is the case?)

apps/src/lib/client/utils.rs
sug0: https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.xor ?

james-chf: the `Option` types are different in this case (`Option<ChainId>` vs `Option<PathBuf>`)

PR #803, opened 2022-11-17
Implement a bit set to store PrepareProposal tx indices

sug0:
Based on #791

Implementation of a lazy bitset, to store tx indices that have already been considered for inclusion in a block. This will be used to infer which txs can be included in a block, after the `BlockSpaceAllocator` transitions to its last state.

---

**EDIT:** Drafted for removal. The implementation now lives here: https://github.com/heliaxdev/index-set


batconjurer (commented):
Is there a reason we aren't using BitVec which is already imported in shared?

sug0:
@batconjurer I did look at this crate beforehand. unfortunately, it doesn't support arbitrary indexing; you can only store indices up to the max width of the underlying integer that is used as the bitset. so, `63` for a `u64`, for instance. this data structure (the `IndexSet`) allows arbitrary indexing, e.g. `u64::MAX` is a valid index over the bitset

batconjurer:
> @batconjurer I did look at this crate beforehand. unfortunately, it doesn't support arbitrary indexing; you can only store indices up to the max width of the underlying integer that is used as the bitset. so, `63` for a `u64`, for instance. this data structure (the `IndexSet`) allows arbitrary indexing, e.g. `u64::MAX` is a valid index over the bitset

Actually, the crate already imported if bit_vec, not bitvec. But checking the documentation again, it can hold up to a usize of indices.
https://docs.rs/bit-vec/latest/bit_vec/struct.BitVec.html#method.from_elem

sug0:
> > @batconjurer I did look at this crate beforehand. unfortunately, it doesn't support arbitrary indexing; you can only store indices up to the max width of the underlying integer that is used as the bitset. so, `63` for a `u64`, for instance. this data structure (the `IndexSet`) allows arbitrary indexing, e.g. `u64::MAX` is a valid index over the bitset
> 
> Actually, the crate already imported if bit_vec, not bitvec. But checking the documentation again, it can hold up to a usize of indices. https://docs.rs/bit-vec/latest/bit_vec/struct.BitVec.html#method.from_elem

I see. That implementation doesn't have sparse storage of individual bit sets, though. The one in this PR saves a bit of space if we are iterating over many indices https://docs.rs/bit-vec/latest/src/bit_vec/lib.rs.html#212-217

Idk, which do you prefer?

batconjurer:
> > > @batconjurer I did look at this crate beforehand. unfortunately, it doesn't support arbitrary indexing; you can only store indices up to the max width of the underlying integer that is used as the bitset. so, `63` for a `u64`, for instance. this data structure (the `IndexSet`) allows arbitrary indexing, e.g. `u64::MAX` is a valid index over the bitset
> > 
> > 
> > Actually, the crate already imported if bit_vec, not bitvec. But checking the documentation again, it can hold up to a usize of indices. https://docs.rs/bit-vec/latest/bit_vec/struct.BitVec.html#method.from_elem
> 
> I see. That implementation doesn't have sparse storage of individual bit sets, though. The one in this PR saves a bit of space if we are iterating over many indices https://docs.rs/bit-vec/latest/src/bit_vec/lib.rs.html#212-217
> 
> Idk, which do you prefer?

This implementation may be better, but I think I'm going to err on the side of not adding things to the codebase unless we have to. If it somehow becomes a bottleneck, we can revisit the issue.

sug0:
alright nw

Issue #686, opened 2022-10-25
The Ethereum event oracle should initially be configured using values from blockchain storage


james-chf:
The Ethereum event oracle currently uses hardcoded configuration. It should instead use values from storage as set at genesis/by governance (added in https://github.com/anoma/namada/pull/575).

To do this, when creating the oracle, we will pass in a channel for controlling it. The `Shell` should read from storage the values of the appropriate chain parameters and configure the oracle via this channel. The oracle should not do anything until it has been configured. If no Ethereum bridge parameters have been set in storage (i.e. the Ethereum bridge hasn't been bootstrapped), then the oracle doesn't do anything.

PR #810, opened 2022-11-18
Configure the Ethereum oracle using values from storage

james-chf:
Relates to https://github.com/anoma/namada/issues/686

This PR contains code for making it so that the Ethereum oracle is configured from values in blockchain storage, rather than using a default hardcoded config.

The motivation for this PR is to implement the minimal functionality such that we can easily start running the ledger with a real Ethereum bridge deployment. Limitations of this initial implementation:
- initial configuration of the oracle is unnecessarily tied to the ABCI Commit phase - at ledger start up, if Ethereum bridge configuration values are already in local block storage and we are synced up to the latest block, we shouldn't need to wait for another block to be committed before configuring our oracle. Initial oracle configuration could be done separately out of band from ABCI but this could be a later optimization.
- only an initial configuration is sent - if values change in block storage while the ledger is running, the oracle won't be updated unless the ledger node is restarted
- the Shell currently does not know for sure whether the Ethereum oracle successfully started running or not, we could fix this by passing a `oneshot::Sender<_>` to the Ethereum oracle inside the command we use to start it, so that it can tell us it successfully started running (in a later PR)

End-to-end test coming in https://github.com/anoma/namada/pull/1032


apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -691,10 +721,67 @@ where
...
+                Some(config) => config,
+                None => {
+                    tracing::warn!(
+                        "An Ethereum oracle task appears to be running, but \
+                         there are no Ethereum bridge parameters configured \
+                         in block storage, so this oracle will do nothing"
+                    );
+                    return;
+                }
+            };
```
sug0, line 750: if no config is present, we will spam logs with this warning, every commit. is this the desired behavior?

james-chf: oh yeah good catch, we don't want to spam the logs - I've replaced this with a comment but now need to double check this is correct - if the Ethereum bridge will be launched by hard fork, then configuration will always be written during `InitChain` (so before any block is committed). In which case, validators should turn the oracle on depending on whether the Ethereum bridge is configured for the chain, and if they turn the oracle on when it's not needed, maybe that should be a misconfiguration error for which we could just stop the node

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -661,6 +688,9 @@ where
                 e
             )
         });
+        // TODO(namada#1041): we check the Ethereum oracle is started on every
```
batconjurer, line 691: Why can we just do this check in the `Shell::new` method if block height > 0 and also call it in `init_chain`?

batconjurer (approved):
Generally ok with it. Just had one question.

sug0 (approved):

apps/src/lib/node/ledger/shell/mod.rs
james-chf: Yeah this sounds like the best solution! I've updated issue https://github.com/anoma/namada/issues/1041 - will merge this PR and open a new PR to optimize this.

For Ethereum bridge parameters that may be allowed to be changed by governance vote (at the moment, that's just the `min_confirmations` parameter) - we will need to check at the beginning of each epoch if that's changed, will open a separate issue for that.

james-chf: We can do a check in `Shell::new` something like:

```
        // we may be able to start the Ethereum oracle right away if we have
        // already synced at least one block
        if shell.storage.last_height >= BlockHeight(1) {
            shell.ensure_ethereum_oracle_started()
        }
```

james-chf:
Had to push https://github.com/anoma/namada/pull/810/commits/1178d8b6f8d9b3d035219f087c4bd2883ebceb3f to fix some unit tests which were failing now that shell actually tries to configure the oracle by sending something over the control channel - we need to keep this channel alive for the duration of tests involving the `TestShell`. We have to do something similar for the broadcast receiver channel in some tests as well it looks like, e.g.

```rust
        let (mut shell, _broadcaster, _, _oracle_control_recv) =
            setup_at_height(3u64);
```

It's a bit ugly returning this (now a) 4-tuple, we can clean it up I think by changing to returning a struct which can be selectively unpacked instead.

PR #812, opened 2022-11-21
Remove abort channel between Ethereum oracle and geth subprocess

james-chf:
This PR removes logic to do with having a channel from the Ethereum oracle task to the `geth` task so that `geth` could be shut down if the Ethereum oracle (unexpectedly) terminated. Shutting down of both of these components is now managed from above by `AbortableSpawner`. If either one terminates for some reason (or any other task like Tendermint, the ABCI server, etc.), then both will get shut down by the `AbortableSpawner` in any case.


apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -28,15 +28,27 @@ pub enum Error {
...
-/// Monitor the Ethereum fullnode. If it stops or an abort
-/// signal is sent, the subprocess is halted.
+/// Monitor the Ethereum fullnode subprocess, returning only once it has
+/// stopped running. If a signal is sent on `abort_recv`, the fullnode
+/// subprocess will be killed.
 pub async fn monitor(
     mut node: eth_fullnode::EthereumNode,
     abort_recv: Receiver<Sender<()>>,
-) -> Result<()> {
+) {
```
james-chf, line 37: The send half of the `abort_recv` channel here belongs to `AbortableSpawner`, and is what is actually used to shut down the Ethereum fullnode if needed. Made this fn return `()` instead of `Result<()>` since it was always returning `Ok(())` or panicking anyway.

apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -28,15 +28,27 @@ pub enum Error {
...
+        // wait for the Ethereum fullnode to naturally exit
+        exit_status = node.wait() => {
+            match exit_status {
+                Ok(exit_status) => {
+                    tracing::info!(%exit_status, "Ethereum fullnode exited");
+                },
+                Err(err) => {
+                    tracing::warn!("Error while waiting for the Ethereum fullnode to exit: {}", err);
+                    tracing::info!("Ensuring Ethereum fullnode is shut down...");
+                    node.kill().await;
```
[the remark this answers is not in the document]
james-chf: Even if `wait()` returned an error? The docstring doesn't actually mention anything about when it will return an error, so I'm not sure

apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -28,15 +28,31 @@ pub enum Error {
...
+            match exit_status {
+                Ok(exit_status) => {
+                    if exit_status.success() {
+                        tracing::info!(%exit_status, "Ethereum fullnode exited");
+                    } else {
+                        tracing::warn!(%exit_status, "Ethereum fullnode exited with nonzero exit code");
+                    }
+                },
+                Err(err) => {
+                    tracing::warn!("Error while waiting for the Ethereum fullnode to exit: {}", err);
```
sug0, line 50:
```suggestion
                    tracing::warn!("Error while waiting for the Ethereum fullnode to exit: {err}");
```

sug0 (approved):
just a nit

batconjurer (approved):

PR #817, opened 2022-11-23
Remove protocol tx vote extension digests

### LINKED ISSUE #816: Remove protocol tx vote extension digests
The Ethereum bridge design is being revised for `abciplus` builds of the ledger. Now we don't include eth events or validator set update digest protocol txs in blocks, since those were mostly an artifact of `abcipp` logic. To comply with the new design, we should remove these types of transactions from the code.

sug0:
Closes #816

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -358,9 +358,9 @@ mod test_finalize_block {
     use namada::types::ethereum_events::EthAddress;
     use namada::types::storage::Epoch;
     use namada::types::transaction::{EncryptionKey, Fee};
-    use namada::types::vote_extensions::ethereum_events::{
-        self, MultiSignedEthEvent,
-    };
+    use namada::types::vote_extensions::ethereum_events;
+    #[cfg(feature = "abcipp")]
+    use namada::types::vote_extensions::ethereum_events::MultiSignedEthEvent;
 
     use super::*;
     use crate::node::ledger::shell::test_utils::*;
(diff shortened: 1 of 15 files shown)


sug0:
I tested the ledger manually. Seems to be working fine.

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -516,7 +623,7 @@ mod test_process_proposal {
         let tx = ProtocolTxType::EthereumEvents(vote_extension_digest)
             .sign(&protocol_key)
             .to_bytes();
-        #[allow(clippy::redundant_clone)]
+        //#[allow(clippy::redundant_clone)]
```
batconjurer, line 626: if it's not needed, just remove it.

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -49,23 +56,16 @@ where
 
             // add ethereum events and validator set updates as protocol txs
             #[cfg(feature = "abcipp")]
-            let txs = self.build_vote_extensions_txs(req.local_last_commit);
+            let txs = self.build_vote_extension_txs(req.local_last_commit);
             #[cfg(not(feature = "abcipp"))]
-            let mut txs = self.build_vote_extensions_txs(&req.txs);
-            #[cfg(feature = "abcipp")]
-            let mut txs: Vec<TxRecord> =
```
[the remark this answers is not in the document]
sug0: ABCI++ tests? well, it's not like the Tendermint folk will bring records back, so we'd have to remove them eventually either way

batconjurer (approved):

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -149,6 +149,18 @@ where
                     continue;
                 }
                 TxType::Protocol(protocol_tx) => match protocol_tx.tx {
+                    #[cfg(not(feature = "abcipp"))]
+                    ProtocolTxType::EthEventsVext(ref ext) => {
+                        for event in ext.data.ethereum_events.iter() {
+                            self.mode.deque_eth_event(event);
```
james-chf, line 155: Should we be doing this here? The vext could have come from another validator. We only want to dequeue Ethereum events that we (the validator) have the successfully included in a block, so we should have an additional check here that it is actually our vext.

This issue was already present I think, so should be a separate PR/issue (if my understanding above is correct).

james-chf (changes requested):
I feel we need to figure out a way to manage the `abcipp` feature code complexity better somehow. To note, `abcipp` does not even compile currently in our branch (or `main`) - https://github.com/anoma/namada/issues/497 - so we have a lot of code now which is not compiling. We also don't have a fully functional version of Tendermint with ABCI++ to test our current `abcipp` code against, so we don't necessarily know that it is (still) working. I'm not sure what the best solution is here but it feels like the `abcipp` feature flags are really proliferating again.

apps/src/lib/node/ledger/shell/finalize_block.rs
sug0: https://github.com/anoma/namada/issues/818 opened this issue

shared/src/ledger/protocol/mod.rs
```diff
@@ -199,16 +199,37 @@ where
...
+        }
+        #[cfg(feature = "abcipp")]
         ProtocolTxType::EthereumEvents(ext) => {
             let ethereum_events::VextDigest { events, .. } = ext;
             self::transactions::ethereum_events::apply_derived_tx(
                 storage, events,
             )
             .map_err(Error::ProtocolTxError)
         }
+        #[cfg(feature = "abcipp")]
```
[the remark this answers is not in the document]
sug0: ahh I see it, sorry

james-chf (approved):

PR #836, opened 2022-11-28
Review CPoS article

brentstone:


documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+- Anything else?
+
+This article is divided into three main sections. First, we describe the mechanics and configuration of the epoched PoS system. Then we describe Namada's novel approach to slashing for validator misbehavior. Finally, we describe the inflation and rewards system within the PoS module.
+
+
+# Epoched Proof of Stake
+
+The native staking token in Namada is identified as *NAM*. We call the smallest denomination of this token *namnam*, where 1 million *namnam* is equal to 1 *NAM*.
+
+Namada's epoched PoS model is influenced by that of [Cosmos](https://github.com/cosmos/cosmos-sdk/blob/019444ae4328beaca32f2f8416ee5edbac2ef30b/docs/architecture/adr-039-epoched-staking.md#pipelining-the-epochs). Sets of blocks are batched into epochs, which are periods of time that have some fixed configuration of the PoS system. For example, the set of validators who participate in the consensus protocol and their associated consensus voting powers can only change at the boundary of two epochs and are constant throughout the duration of one.
```
cwgoes, line 16: Cosmos has not actually implemented epoched staking. I think we could say "influenced by prior discussions on epoch staking such as this one" (with the same link) but otherwise it's best to just describe what an epoch is

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+Namada employs epoched staking for two main reasons:
+- Light clients become more efficient because there are fewer state changes (particularly validator set changes) to store, resulting in fewer headers needed to verify state
```
cwgoes, line 19: primarily because validator set changes are less frequent, so light clients don't need to download as many of them, and bisection becomes more efficient

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+	- TODO: maybe need to reword
+	- Q: does this light client efficiency apply only to IBC or also for Eth?
+- Eventually, Namada (and the larger Anoma ecosystem) will rely on the Ferveo DKG protocol to encrypt transactions in the mempool. Ferveo must know the validator set for a near-future epoch in order to have sufficient time to compute the DKG in advance of the given epoch.
+
+Thus, changes to the state of the PoS system are enqueued to take effect at some epoch in the future, the lengths of which we refer to as *offsets* relative to the epoch in which the transaction is submitted. Most state changes, including validator set updates, take effect at the *pipeline offset* - 2 epochs into the future. The other important offset in the model is the *unbonding offset* of 21 epochs, primarily the length of time before an unbond becomes withdrawable.
```
cwgoes, line 24: wherever we list values, we should make it clear that they are parameters - e.g. "unbonding offset (by default, 21 epochs)"

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+Below, we will review the rules around important PoS actions
+- Bonding tx submitted at epoch `n` initiates changes at epoch `n + pipeline_offset (n+2)`
+	- Bond becomes active, validator bonded stake and validator set is updated
+- Unbonding tx submitted at epoch `n`
+	- Updates validator set with decreased validator voting power at epoch `n + pipeline_offset (n+2)`
+	- Amount of the unbond is withdrawable at epoch `n + unbonding_offset + pipeline_offset (n+23)`
+		- TODO: check over this w Tomas
+- Redelegation tx submitted at epoch `n`
+	- Only support 100% token amount redelegated
+	- TODO: more details needed here perhaps?
```
cwgoes, line 41: we should explain the voting power changes

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+- `min_validator_stake` = 1 NAM
+
+# Cubic Slashing
+
+As in many other PoS blockchains, slashing is applied as a way to punish misbehaving validators. If a validator is detected to have misbehaved then portions of the validator's bonded stake will be slashed or seized. This propagates down to the stake of delegators associated to the misbehaving validator as well. Validator infractions that are subject to slashing include:
+
+- proposing an invalid block
+- signing an invalid block
+- proposing two different blocks at the same height
+- double-signing a block
```
cwgoes, line 63: this is the same as the bullet above, no? I think we also slash for amnesia attacks, whatever Tendermint can detect.

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+Ultimately, slashed tokens are sent to a slash pool fund, which...
```
cwgoes, line 65: the fund could be spent by governance to refund in the case of accidental misbehaviour

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+$$\text{sum}~ = \sum_{i \in \text{infractions}} \frac{\text{vp}_{i}}{\text{vp}_{tot}}$$
+
+Note that the voting power in the above expression is the voting power of the validator *when the infraction was committed*.
+
+3. The slash rate is then a function of the square of this sum, bounded below by some configurable minimum slash rate $r_{\text{min}}$:
+
+$$\max (~r_{\text{min}}~,~9*\text{sum}^2 ~)$$
+
+The factor of $9$ is included such that the slash rate maxes out at $100\%$ for total voting power of $1/3$. This is a critical point in the BFT consensus model, as a total colluding voting power of greater than $1/3$ can prevent the chain from achieving consensus.
```
cwgoes, line 84: it would be nice to include the graph here, I think

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+Once the slash rate is determined, the voting power of the jailed misbehaving validator can be immediately slashed and updated. This validator can then submit a transaction to become unjailed, which takes effect at the pipeline offset relative to the epoch of submission. Once the validator has been reinstated into the appropriate validator set depending on its new voting power, it and its delegators can resume bonding, unbonding, and redelegating.
+
+Namada's slashing system also guarantees that only delegators whose stake were used in a misbehaving validator's voting power during the epoch of infraction are slashed. If a delegator that contributed to a misbehaving validator's voting power at the time of infraction has since redelegated to a new validator by the time the infraction is discovered, then the protocol will still properly slash the delegator. This is an improved PoS guarantee offered by Namada, whereas in other chains, such as Cosmos, there is no guarantee that all delegators contributing voting power to an infraction (and only these delegators) will be properly slashed for the infraction.
```
cwgoes, line 88: Cosmos isn't a chain - "in other chains using the Cosmos SDK's proof-of-stake module"

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+$$\prod_{e=0}^e\big(1 + (1 - c_V(e))\frac{r_V(e)}{s_V(e)} \big),$$
+
+where $c_V(e)$ is a validator's commission rate for delegations in epoch $e$.
+
+**Check this next bit**
+There are several nice advantages of using the rewards products:
+- The storage of the rewards products allows the validators' voting powers to be updated immediately by considering the rewards products along with their bonded staked tokens.
+- The number of tokens to be credited to a delegator (or validator for a self-bond) only needs to be calculated at the moment of withdrawal by considering the original amount in the bond and the rewards products over the course of the bond's lifetime.
+- Lots of unneeded iteration in storage can be avoided
+- ???
```
cwgoes, line 146: Delegators don't need to continuously withdraw rewards and redelegate

cwgoes (commented):
Thanks! see comments

documentation/cubic-pos-article.md
brentstone: I thought L62 is specific to a block proposer but L63 can also apply to other validators who are just signing. Also took this list from https://specs.namada.net/economics/proof-of-stake/bonding-mechanism.html#slashing, though it's possible that this may need to be updated instead.

Not sure if Tendermint detects amnesia attacks yet, from https://github.com/tendermint/tendermint/issues/5270 it looks like it may be put into v0.37. As it stands, the only kinds of `MisbehaviorType` in tendermint are `DuplicateVote` and `LightClientAttack` (there is also an `Unknown` variant in this enum).

cwgoes: hmm. we should definitely support all evidence that Tendermint can detect, and for now we can consider any sort of safety attack to be of the same sort of severity (meriting cubic slashing) (though independent parameters is fine). 

I see what you mean but I don't think there's a difference in proposing vs signing from a safety perspective, safety is < 1/3 byzantine regardless

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,156 @@
...
+
+Once the slash rate is determined, the voting power of the jailed misbehaving validator can be immediately slashed and updated. This validator can then submit a transaction to become unjailed, which takes effect at the pipeline offset relative to the epoch of submission. Once the validator has been reinstated into the appropriate validator set depending on its new voting power, it and its delegators can resume bonding, unbonding, and redelegating.
+
+Namada's slashing system also guarantees that only delegators whose stake were used in a misbehaving validator's voting power during the epoch of infraction are slashed. If a delegator that contributed to a misbehaving validator's voting power at the time of infraction has since redelegated to a new validator by the time the infraction is discovered, then the protocol will still properly slash the delegator. This is an improved PoS guarantee offered by Namada, whereas in other chains using the Cosmos SDK's PoS module, there is no guarantee that all delegators contributing voting power to an infraction (and only these delegators) will be properly slashed for the infraction.
```
cwgoes, line 87:
```suggestion
Namada's slashing system also guarantees that only delegators whose stake were used in a misbehaving validator's voting power during the epoch of infraction are slashed. If a delegator that contributed to a misbehaving validator's voting power at the time of infraction has since redelegated to a new validator by the time the infraction is discovered, then the protocol will still properly slash the delegator. This is an improved PoS guarantee offered by Namada, whereas in other chains using the Cosmos SDK's PoS module, there is no guarantee that all delegators contributing voting power to an infraction (and only these delegators) will be properly slashed for the infraction. For further context, see [this issue](https://github.com/cosmos/cosmos-sdk/issues/1440).
```

cwgoes (commented):
A few more minor comments

documentation/cubic-pos-article.md
brentstone: Aside from the amnesia attack, we currently do support all Tendermint evidences. I suppose it may depend on the Tendermint version we use as well, but https://github.com/tendermint/tendermint/issues/5270 is still open as well atm. Let me know if I should include the amnesia attack in the article.

Gotcha regarding proposing vs signing. In this case, should these 4 lines be condensed into:
```
- Signing an invalid block
- Signing two different blocks at the same height
```
?

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,156 @@
...
+
+Namada's PoS module is inspired in large part by the PoS system of the Cosmos Hub and by [prior discussions](https://github.com/cosmos/cosmos-sdk/blob/019444ae4328beaca32f2f8416ee5edbac2ef30b/docs/architecture/adr-039-epoched-staking.md#pipelining-the-epochs) on its future improvements, which have not yet been implemented. The ["nothing at stake" problem](https://blog.ethereum.org/2014/01/15/slasher-a-punitive-proof-of-stake-algorithm) is solved by slashing misbehaving validators, which amounts primarily to confiscating their bonded tokens.
+
+In the PoS system, both users and validators are incentivized with rewards to contribute to the security of the network by bonding tokens in the PoS module. Namada has a regular schedule for minting new tokens (inflation), many of which are paid out to accounts that have bonded tokens. Non-valdating accounts can participate by *delegating* their tokens to a particular validator: the tokens are locked in the PoS system and the target validator can then use those tokens as voting power in the consensus mechanism.
```
tzemanovic, line 9: "paid out to accounts that have bonded tokens" - they don't go to the accounts directly as they are auto-bonded, but maybe that's important

tzemanovic, line 9:
```suggestion
In the PoS system, both users and validators are incentivized with rewards to contribute to the security of the network by bonding tokens in the PoS module. Namada has a regular schedule for minting new tokens (inflation), many of which are paid out to accounts that have bonded tokens. Non-validating accounts can participate by *delegating* their tokens to a particular validator: the tokens are locked in the PoS system and the target validator can then use those tokens as voting power in the consensus mechanism.
```

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,156 @@
...
+	- Amount of the unbond is withdrawable at epoch `n + unbonding_offset + pipeline_offset (n+23)`
+- Redelegation tx submitted at epoch `n`
+	- Updates validator set with voting powers for old and new validators at epoch `n + pipeline_offset (n+2)`
+	- Only support redelegating 100% of the token amount of a bond
+	- A redelegated bond cannot be redelegated again until epoch `n + unbonding_offset + pipeline_offset (n+23)`
+	- Slashes are properly handled to ensure that only tokens that contributed to a validator's voting power at the time of infraction are affected
+
+**Validator Sets**
+
+Namada validators are classified into one of three sets:
```
tzemanovic, line 44: I think it's important to distinguish here that this is only in relation to consensus, due to the consensus validator set size limit. These are not relevant to e.g. governance

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,156 @@
...
+
+- proposing an invalid block
+- signing an invalid block
+- proposing two different blocks at the same height
+- double-signing a block
+
+Ultimately, slashed tokens are sent to a slash pool fund, from which the funds can be spent by governance to refund slashed accounts in the event of accidental misbehavior.
+
+Typically, the slash amount for an infraction is proportional to a validator's voting power, however  Namada employs a slashing scheme that has more severe punishments for correlated validator misbehavior: so-called [*cubic slashing*](https://specs.namada.net/economics/proof-of-stake/cubic-slashing.html), wherein the slashed token amount can be proportional to the cube of a validator's voting power (the slash rate is quadratic). Cubic slashing employs more severe punishments to validators who commit infractions close to each other in time, making it riskier for an entity to operate larger or multiple similarly configured validators.
```
tzemanovic, line 66: > making it riskier for an entity to operate larger or multiple similarly configured validators

I'd say that rather it is to have very little punishment for accidental (e.g. operational) misbehavior, but much more severe for deliberate repeated attacks

documentation/cubic-pos-article.md
brentstone: thx, changed to "owed to accounts that have bonded tokens"

documentation/cubic-pos-article.md
brentstone: thx, added the following sentence to the end of this section:
"Note that this construction of the validator sets is only important for the consensus mechanism; all validators can participate with their voting power in governance."

documentation/cubic-pos-article.md
brentstone: thx, changed the sentence to:
"Cubic slashing is designed to impose more severe punishments for deliberate, repeated validator attacks but much milder punishments for accidental or non-malicious misbehaviors."

awasunyin:
Not yet! But I have it on my notes. First I need to program a publication calendar.

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Cubic Slashing
+
+As in many other PoS blockchains, slashing is applied as a way to punish misbehaving validators. If a validator is detected to have misbehaved then portions of the validator's bonded stake will be slashed or seized. This propagates down to the stake of delegators associated to the misbehaving validator as well. Validator infractions that are subject to slashing include:
+
+- signing an invalid block
+- signing two different blocks at the same height
+- amnesia attacks (once they are supported by tendermint, see [this issue](https://github.com/tendermint/tendermint/issues/5270))
+
+Ultimately, slashed tokens are sent to a slash pool fund, from which the funds can be spent by governance to refund slashed accounts in the event of accidental misbehavior.
```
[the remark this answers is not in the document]
brentstone: It means the former, not something more nuanced

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
+# Introduction
+
+Namada is a L1 proof-of-stake blockchain that provides interchain asset-agnostic privacy. Namada is based on a byzantine fault-tolerant (BFT) consensus mechanism widely used in the Cosmos ecosystem, provided by Tendermint ([Go implementation]((https://github.com/tendermint/tendermint)) with the [rust bindings]((https://github.com/informalsystems/tendermint-rs)))., that enables fast block finality times and much lower computational overhead than in proof-of-work (PoW) consensus mechanisms, as in Bitcoin. For consensus to be achieved in a BFT system, at least a 2/3 majority of the nodes (called validators) must be honest and behave according to the protocol.
```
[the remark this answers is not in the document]
brentstone: I feel like "much lower complexity" is preferable here to quoting some hash rate, especially without some context. And I don't think it is important to the content of the article to delve into the context of hash rates here. This is just an introductory paragraph after all.

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+Namada uses proof-of-stake (PoS) to provide security against Sybil attacks on the network, wherein a malicious entity achieves enough voting power to disrupt consensus of the true chain. Rather than miners competing to create the next block by hashing with lots of computing power in PoW systems, the block proposer is selected from a set of validators with probability proportional to the validator's bonded stake (or voting power). This voting power is then used by the validators to determine if the 2/3 majority of consensus has been achieved. Furthermore, the block proposer mechanism also allows PoS chains to be much greener than PoW chains that need to consume massive amounts of energy to reach consensus.
```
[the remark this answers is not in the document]
brentstone: Removing "true." As for "much greener," what kind of stat do you recommend? As before, I don't think providing actual numbers matter much here (unless we could say precisely some percentage decrease in energy grid usage, which I'm not sure we can), and after all this is just the introduction rather than the focus of the article.

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+
+This article is divided into three main sections. First, we describe the mechanics and configuration of the epoched PoS system. Then we describe Namada's novel approach to slashing for validator misbehavior. Finally, we describe the inflation and rewards system within the PoS module.
+
+
+# Epoched Proof of Stake
+
+The native staking token in Namada is identified as *NAM*. We call the smallest denomination of this token *namnam*, where 1 million *namnam* is equal to 1 *NAM*. An epoch defines a grouped set of consecutive blocks and a period of time that have some fixed configuration of the PoS system. For example, the set of validators who participate in the consensus protocol and their associated consensus voting powers can only change at the boundary of two epochs and are constant throughout the duration of one.
+
+Namada employs epoched staking for two main reasons:
+- Light clients become more efficient because there are fewer state changes (particularly validator set changes) to keep track of, resulting in quicker state verification
```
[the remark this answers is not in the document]
brentstone: Which part, the bit about light clients? @cwgoes do you think this bullet point should be more detailed, and if so, what other information should be included?

documentation/cubic-pos-article.md
```diff
@@ -0,0 +1,155 @@
...
+- `r_max`: the maximum reward rate, expressed in annual percentage (or APR)
+- `R_target`: the targeted staked token ratio (total staked supply / total token supply)
+- `epochs_per_yr`: the expected number of epochs in a year
+- `KP_nom`: the proportional gain factor (P in PD controller)
+- `KD_nom`: the differential gain factor (D in PD controller)
+
+Then there are some protocol values dependent on the epoch:
+- `S`: the total supply of tokens
+- `L`: the total number of staked tokens (locked in the PoS system)
+- `I`: the most recent inflation amount (in units of tokens per epoch)
```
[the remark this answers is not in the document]
brentstone: Thanks for the note! Good to know actually. Gonna try to put a π inside of the code-like text.

cwgoes:
Closing as this has been published.

Issue #825, opened 2022-11-25
Update to Tendermint v0.37.0-rc1


james-chf:
Currently our Tendermint v0.37.x fork (`0.1.4-abciplus`) is based off of this commit - https://github.com/heliaxdev/tendermint/commit/0b58342a46b1917a40d506e48e87448dcc8b3131 (from 2022-08-31). We need to update our fork to be based on the more recent Tendermint v0.37-rc1 (https://github.com/tendermint/tendermint/releases/tag/v0.37.0-rc1). The way we do patching for our Tendermint dependencies is off as well, this can be fixed at the same time, and CI for our forks can also be cleaned up a bit (https://github.com/anoma/namada/issues/537).

First, making the patching saner:
- [x] create a `ledger-main` branch for `tendermint-rs`, using the `tendermint-rs` version used in namada v0.10.1
- [x] create a `ledger-main` branch for `ibc-rs` and pin it to `tendermint-rs` version used in namada v0.10.1 (https://github.com/heliaxdev/ibc-rs/pull/11)
- [x] create a `ledger-main` branch for `tower-abci` and pin it to `tendermint-rs` version used in namada v0.10.1 (https://github.com/heliaxdev/tower-abci/pull/6)
- [x] update `tendermint-v0.37-integration` to use revisions from the `ledger-main` branches of tendermint-rs/ibc-rs/tower-abci, for `0.1.4-abciplus` (https://github.com/anoma/namada/pull/827)

Then
- [x] integrate v0.37.0-rc1 changes for `tendermint-rs`(https://github.com/heliaxdev/tendermint-rs/pull/10) into `ledger-main`
- [x] integrate v0.37.0-rc1 changes for `ibc-rs`(https://github.com/heliaxdev/ibc-rs/pull/9) into `ledger-main`
- [x] integrate v0.37.0-rc1 changes for `tower-abci`(https://github.com/heliaxdev/tower-abci/pull/5) into `ledger-main`
- [x] update https://github.com/heliaxdev/tendermint `ledger-main` to be based on upstream `v0.37.0-rc1` (https://github.com/heliaxdev/tendermint/pull/213)

- [ ] update Namada `main` to use newer revisions from the `ledger-main` branches of tendermint-rs/ibc-rs/tower-abci, for `0.1.4-abciplus + v0.37.0-rc1` (being done directly in https://github.com/anoma/namada/pull/838)
[...]

PR #839, opened 2022-11-29
Build Tendermint v0.37-rc1 binaries in CI

james-chf:
Relates to https://github.com/anoma/namada/issues/825

Builds Tendermint v0.37.0-rc1 in CI, alongside the current one. We don't run tests against this new version of Tendermint yet.


tzemanovic:
will go to cometBFT

PR #843, opened 2022-11-30
Add Tendermint max block size param to genesis config file

sug0:
We allow validators to configure the max Tendermint block size at genesis.


apps/src/lib/config/genesis.rs
```diff
@@ -226,6 +226,10 @@ pub mod genesis_config {
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
     pub struct ParametersConfig {
+        // Max Tendermint block size in bytes.
+        // XXX: we use i64 becauase that's what tower-abci
+        // and tendermint-rs expect
+        pub max_bytes_per_block: i64,
```
james-chf, line 232: Is this meant to be the same as `max_tx_bytes` from `PrepareProposal`? https://github.com/tendermint/tendermint/blob/v0.37.x/spec/abci/abci++_methods.md#prepareproposal

sug0: yeah, it's the same type. btw, sorry, I didn't mean to request a review yet

apps/src/lib/config/genesis.rs
```diff
@@ -226,6 +227,8 @@ pub mod genesis_config {
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
     pub struct ParametersConfig {
+        // Max payload size, in bytes, for a tx batch proposal.
+        pub max_proposal_bytes: ProposalBytes,
```
james-chf, line 231: Would suggest we have a dedicated section for Tendermint consensus params (the other parameters in this `ParametersConfig` are specific to our application, e.g. to do with epochs, whitelisting, etc.)

I can't find a documentation page for Tendermint consensus params, but I'm guessing it is the parameters of this struct - https://github.com/tendermint/tendermint/blob/v0.37.x/spec/abci/abci%2B%2B_methods.md#consensusparams - we could even have a struct which matches the subset of these consensus params we care about customizing e.g.

```rust
    #[derive(Clone, Debug, Deserialize, Serialize)]
    pub struct ConsensusParamsConfig {
        pub block: BlockParamsConfig,
    }

    #[derive(Clone, Debug, Deserialize, Serialize)]
    pub struct BlockParamsConfig {
        // Maximum size of a block, in bytes.
        pub max_bytes: ProposalBytes,
    }
```

This would make it super clear this is a parameter we feed back to Tendermint at the end of ABCI rounds (we could even have `From`/`Into` impls for this consensus params struct, and the actual `ConsensusParams` we need to return in `ResponseEndBlock`

james-chf (changes requested):
lgtm though I think we should mark out Tendermint consensus params in their own config struct (either in this PR, or some later one). This PR should be targeting `main`

sug0:
@james-chf can't target `main` yet, because the block space allocator is a hard dependency of this PR. without it, blocks are always configured to be 100 MiB in size, which isn't ideal. I guess I forgot to mention that in the description :sweat_smile: btw, I'm still missing the evidence config. will add that tomorrow

sug0:
so, actually, it looks like we do not configure evidence params right now, and they seem to be using 1 MiB of space at most, by default. that should be plenty of space, so I'm not going to touch that config.

apps/src/lib/config/genesis.rs
sug0: the max tx size param isn't specific to Tendermint, it is specific to Namada. Tendermint is actually being configured with a hard-coded value of 100 MiB per block, at most. the value we are configuring in `ParametersConfig` doesn't get consumed by Tendermint

james-chf: I mean here the `max_proposal_bytes` param, which corresponds to the `max_bytes` parameter in Tendermint - https://github.com/tendermint/tendermint/blob/v0.37.x/spec/core/data_structures.md#blockparams - this is a parameter we would update on chain and then feed back to Tendermint (whereas the other parameters like `min_num_of_blocks`, `max_expected_time_per_block`, etc., Tendermint doesn't need to know about).

james-chf: Ok I think I understand how it is meant to work now in this PR. There is a Tendermint consensus parameter `consensus_params.block.max_bytes` which is the maximum size a block may be, we set this at genesis to be 100 MiB. We never update this after genesis. Instead we have a different `max_proposal_bytes` on chain parameter which is not related to Tendermint, and is always guaranteed to be lower than the hardcoded `consensus_params.block.max_bytes` we set at genesis (i.e. max 90 MiB). 

The way I thought it might work is that `consensus_params.block.max_bytes` and this `max_proposal_bytes` would be the same value, i.e. both should start at 100 MiB. Then if validators wanted to, they could effect a governance vote on chain to update the on chain parameter (i.e. `max_proposal_bytes`), which would then be fed back to Tendermint in `ResponseEndBlock`.

james-chf:
> @james-chf can't target `main` yet, because the block space allocator is a hard dependency of this PR. without it, blocks are always configured to be 100 MiB in size, which isn't ideal. I guess I forgot to mention that in the description 😅 btw, I'm still missing the evidence config. will add that tomorrow

Wouldn't it just be setting a max cap of 100 MiB rather than having blocks always be 100 MiB in size? (I would have thought applying this PR to `main`, there would be no functional change, except the max possible block size would be different, but I could be wrong). In that case, the changes in this PR don't look like they depend on the block space allocator PRs.

sug0:
> > @james-chf can't target `main` yet, because the block space allocator is a hard dependency of this PR. without it, blocks are always configured to be 100 MiB in size, which isn't ideal. I guess I forgot to mention that in the description 😅 btw, I'm still missing the evidence config. will add that tomorrow
> 
> Wouldn't it just be setting a max cap of 100 MiB rather than having blocks always be 100 MiB in size? (I would have thought applying this PR to `main`, there would be no functional change, except the max possible block size would be different, but I could be wrong). In that case, the changes in this PR don't look like they depend on the block space allocator PRs.

yeah, it'd just set a max block size of 100 MiB, but I'm not sure if it's a good idea to include this PR in `main` without a way to throttle the num of txs included in a block

core/src/ledger/parameters/storage.rs
```diff
@@ -20,10 +21,19 @@ pub fn is_parameter_key(key: &Key) -> bool {
...
         || is_max_expected_time_per_block_key(key)
         || is_tx_whitelist_key(key)
         || is_vp_whitelist_key(key)
+        || is_implicit_vp_key(key)
+        || is_epochs_per_year_key(key)
+        || is_pos_gain_p_key(key)
+        || is_pos_gain_d_key(key)
+        || is_staked_ratio_key(key)
+        || is_pos_inflation_amount_key(key)
+        || is_max_proposal_bytes_key(key)
```
james-chf, line 36: oic the efficiency issue now. It would be good I guess if all protocol parameters were under some subtree e.g. `/parameters/protocol/...` or something, then we could easily recognize them

apps/src/lib/config/genesis.rs
```diff
@@ -232,6 +234,8 @@ pub mod genesis_config {
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
     pub struct ParametersConfig {
+        // Max payload size, in bytes, for a tx batch proposal.
```
james-chf, line 237: nit: would suggest we make this docstring more specific, e.g. something like

> /// Max size in bytes of `txs` returned in `response::PrepareProposal`, which must be strictly less than the `block.max_bytes` consensus parameter

If we have some spec/doc eventually, we could link out to that in this docstring. Somewhere that has the context on why we use our own app side max tx bytes parameter, rather than just using Tendermint's one (i.e. `request::PrepareProposal::max_tx_bytes`)

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -454,6 +454,14 @@ async fn write_tm_genesis(
     genesis.genesis_time = genesis_time
         .try_into()
         .expect("Couldn't convert DateTimeUtc to Tendermint Time");
+    genesis.consensus_params.block = Some(block::Size {
+        // maximum size of a serialized Tendermint block
+        // cannot go over 100 MiB
+        max_bytes: 100 << 20,
+        // gas is metered app-side, so we disable it
```
james-chf, line 461: would make this comment a TODO and link it to a GH issue on metering gas app side

core/src/ledger/parameters/mod.rs
```diff
@@ -111,6 +115,16 @@ impl Parameters {
             pos_inflation_amount,
         } = self;
 
+        // write max proposal bytes parameter
```
batconjurer, line 118: We should add validation somewhere that the input isn't insane. Here perhaps?

core/src/types/chain.rs
```diff
@@ -16,6 +17,159 @@ pub const CHAIN_ID_PREFIX_MAX_LEN: usize = 19;
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct Visitor;
+
+        impl<'de> serde::de::Visitor<'de> for Visitor {
+            type Value = ProposalBytes;
+
+            fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(f, "a u64 in the range 1 - 94371840")
```
james-chf, line 58: Instead of `94371840` we should interpolate a const which has a docstring, e.g. `ProposalBytes::RAW_MAX`?

core/src/ledger/parameters/mod.rs
batconjurer: Or maybe it' crazy to validate genesis input. But we will need some mechanism for validation if it is changed by governance. Tomas mentioned something about guarding it with VPs. Maybe a separate pr.

batconjurer (approved):

core/src/ledger/parameters/mod.rs
sug0: we're grabbing this parameter from the toml config, whose deserialization already has some logic that checks if it is in the range [1 B, 90 MiB]

sug0: I'm not sure if that's enough?

core/src/types/chain.rs
```diff
@@ -16,6 +17,159 @@ pub const CHAIN_ID_PREFIX_MAX_LEN: usize = 19;
...
+            where
+                E: serde::de::Error,
+            {
+                ProposalBytes::new(size as u64).ok_or_else(|| {
+                    serde::de::Error::invalid_value(
+                        serde::de::Unexpected::Signed(size),
+                        &self,
+                    )
+                })
+            }
```
james-chf, line 83: Do we need this `visit_i64`, would we ever expect it to get called? `inner` should always be de/serialized as a u64?

core/src/ledger/parameters/mod.rs
batconjurer: Ah, I missed that. So good enough for now. For governance, I'm not so sure as it doesn't require a hard fork but just a change in storage. So we might need this VP idea. But that can be a separate PR.

james-chf (changes requested):

core/src/types/chain.rs
batconjurer: Crappy problem with the toml library.

sug0: deserialization fails without an `i64` visitor 😓

core/src/types/chain.rs
sug0: sure

core/src/ledger/parameters/storage.rs
sug0: I was playing around with trie codegen over the weekend for this specific use-case, but I still need to improve the generated code. https://gist.github.com/sug0/1bc632eb19ebdeb2d596c1617fba4138

apps/src/lib/node/ledger/tendermint_node.rs
sug0: I think we're already metering gas app-side
https://github.com/anoma/namada/blob/926f47bf80d86a12c5ba2b044c113a31bc77deaf/apps/src/lib/node/ledger/shell/prepare_proposal.rs#L45-L52

apps/src/lib/config/genesis.rs
sug0: working on a spec right now, actually

core/src/ledger/parameters/storage.rs
sug0: btw, we do have the same internal address as a prefix in each of these keys

apps/src/lib/node/ledger/tendermint_node.rs
james-chf: oops yeah, I got this mixed up with setting a max gas cap on blocks (coming after the block allocator I guess)

james-chf (approved):
lgtm!

apps/src/lib/node/ledger/tendermint_node.rs
sug0: yeah, that's still a TODO

PR #844, opened 2022-11-30
Directly salt the code hash in signatures

juped:
The code hash is not handled on its own anywhere besides where the full code blob is, so it doesn't need to be individually timestamped. However, a proposed change to transaction encryption is to encrypt each field piecemeal, in which case this change would be required in order to avoid blatantly revealing code via hash.


cwgoes:
Decided not to do this for now, so @juped I think we can close this?

PR #845, opened 2022-11-30
 Removed MerkleValue type.

batconjurer:
The design of the merkle tree is that each subtree should handle it's own particulars and the top level storage should be unaware of these details. The MerkleValue type violated this principal by requiring top level storage to correctly cast values to the right type which would then be checked by the sub-trees. This caused bugs and made strained the storage API. This PR removes this type.

PR #846 is the corresponding PR to main.

### DIFF
--- a/shared/src/ledger/storage/merkle_tree.rs
+++ b/shared/src/ledger/storage/merkle_tree.rs
@@ -55,6 +55,9 @@ pub enum Error {
 /// Result for functions that may fail
 type Result<T> = std::result::Result<T, Error>;
 
+/// Type alias for bytes to be put into the Merkle storage
+pub(super) type StorageBytes = Vec<u8>;
+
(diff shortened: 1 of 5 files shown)


shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -376,7 +373,7 @@ impl<H: StorageHasher + Default> MerkleTree<H> {
     pub fn get_sub_tree_existence_proof(
         &self,
         keys: &[Key],
-        values: Vec<MerkleValue>,
+        values: Vec<Vec<u8>>,
```
sug0, line 376: lgtm, just a small suggestion: maybe we could give a type alias to `Vec<u8>`? sort of like `TxBytes`, for txs

sug0 (approved):

james-chf (approved):
lgtm!

Issue #302, opened 2022-08-09
Shared MASP Transaction Code


murisi:
Right now, the MASP transaction construction code (i.e. code that fetches unspent nodes, selects unspent notes for a transaction, queries available asset conversions, supplies the transaction gas, ...) is all in the `anoma_apps` crate under `tx.rs`. This presents integration difficulties for those working on the MASP web client since they cannot access this crate, and hence they end up duplicating or recreating the necessary transaction creation code.

One solution to this problem might be to somehow move the transaction construction code/logic from `tx.rs` into the `anoma` shared crate since this crate would be accessible from the web client. Perhaps further clarify what the requirements of the web client team (@memasdeligeorgakis ) are.

memasdeligeorgakis:
In my view to avoid duplication, the logic of constructing transactions and querying balances should not be coupled with:
- terminal CLI context (very specific to the desktop CLI)
- certain ways of fetching past transactions (specific to the desktop environment for now as we do not pass sys calls to WASM for networking)
- caching past transactions (specific to the desktop environment for now as we do not pass sys calls to WASM for file virtual system)
So ideally those can be passed to the modules from the consuming code utilising the MASP actions. Or the actions could be pure functions and all the above listed are completely defined by the consumer. Otherwise, the WASM/web environment does not expose requirements. And true, that code should likely live in a place that is exposed to the outside of `anoma`.

When I created the MASP code for the client, my intention was to do it so that it could be used for both, but now it has some outdated dependencies and it does not exactly match the functionality of your latest MASP features, so I think it is not good at least as-is.

cwgoes:
@murisi will the currently in-progress SDK work fix this?

murisi:
@cwgoes Yes, the current in-progress SDK will fix this. (#848 was also fixing this same issue, but it has been superseded by the current SDK work.) To address @memasdeligeorgakis 's points, the following has been done:
- Terminal CLI context has been separated from the client functionality
- Transaction fetching is now parameterized by any implementation of `Client` (rather than depending on specific syscalls)
- The caching system is now parameterized by generic reading/saving code (supplied by `ShieldedUtils`)
- The client code (including the MASP client) is now exposed to other crates since it's now in `shared`
- The current `shared` MASP code is more up to date since it's essentially the same as what's in `namada/main`

cwgoes:
I believe that this is completed, but please re-open if not.

PR #848, opened 2022-11-30
Murisi/shared masp2

murisi:
Factored the platform specific components out of `ShieldedContext` and then moved `ShieldedContext` into the `shared` crate. Essentially addressing: https://github.com/anoma/namada/issues/302 .


shared/src/ledger/masp.rs
```diff
@@ -200,3 +246,1044 @@ pub fn get_params_dir() -> PathBuf {
...
+    async fn query_storage_value<T: Send>(
+        &self,
+        key: &storage::Key,
+    ) -> Option<T>
+    where
+        T: BorshDeserialize;
+
+    async fn query_epoch(&self) -> Epoch;
+
+    fn local_tx_prover(&self) -> LocalTxProver;
```
memasdeligeorgakis, line 261: Is this specific to the consumer? Something that would differ from what is being done on desktop/CLI?

shared/src/ledger/masp.rs
```diff
@@ -200,3 +246,1044 @@ pub fn get_params_dir() -> PathBuf {
...
+
+    async fn query_conversion(
+        &self,
+        asset_type: AssetType,
+    ) -> Option<(
+        Address,
+        Epoch,
+        masp_primitives::transaction::components::Amount,
+        MerklePath<Node>,
+    )>;
```
memasdeligeorgakis, line 275: Also here, does this mean that we should define this in a web/WASM specific way?

```rust
    async fn query_conversion(
        &self,
        asset_type: AssetType,
    ) -> Option<(
        Address,
        Epoch,
        masp_primitives::transaction::components::Amount,
        MerklePath<Node>,
    )>;
```

cwgoes:
@murisi @mariari is this PR still applicable?

mariari:
> @murisi @mariari is this PR still applicable?

The work in #925 (integration branch) / #921 (branch proper)  follows this work up given the commit log @murisi would know more about the work committed after I went on break.

murisi:
> @murisi @mariari is this PR still applicable?

@mariari is correct. The initial scope of my work was to put the MASP transaction code into an SDK. After completing that, the scope then broadened into putting CLI subcommands into an SDK.  Hence this branch is followed up/superseded by #925 and #921 .

PR #851, opened 2022-12-01
 Changed the bytes passed into merkle proof to be references

batconjurer:
This syncs up our branch with a change I made from comments on PR #846 .

We got a regression test failure in the wasm, and I have cherry picked the fix in from PR #813


james-chf (approved):

batconjurer:
The wasm for `tx_unbond` failed, likely because of PR #813

james-chf (approved):

shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -56,7 +56,7 @@ pub enum Error {
 type Result<T> = std::result::Result<T, Error>;
 
 /// Type alias for bytes to be put into the Merkle storage
-pub(super) type StorageBytes = Vec<u8>;
+pub(super) type StorageBytes<'a> = &'a [u8];
```
sug0, line 59: since values of this type can either be owned or borrowed, it could also be changed to:
```suggestion
pub(super) type StorageBytes<'a> = std::borrow::Cow<'a, [u8]>;
```

batconjurer: So I implemented this, and I got a lot of added verbosity and I'm not sure what I won.

sug0: hmmm. not all battles can be won. sorry xD

batconjurer: Well, let me fiddle with this a bit longer

batconjurer: Yeah...., to get a a benefit from this is major refactor. And the benefit is that I save a clone in a single method. Just doesn't seem worth it at the moment.

sug0 (approved):
CI is failing but it's not a code issue. just some random intermediate command that ran into an error it seems

PR #853, opened 2022-12-01
do not try to reveal MASP sentinel key used for shielded txs

tzemanovic:
Implicit accounts auto-reveal is interacting badly with the masp sentinel key (masp_tx_key) - a tx from shielded address tries to auto-reveal the PK for this key when it calls `fn tx_signer`, which fails with e.g.:

```
Submitting a tx to reveal the public key for address atest1d9khqw36xvcnswzzx4p5zvpkxfp5vd3hxaznz33jxv6yyvfcx9rrjwp5gse5gd348ycny3zrt9hd7k...
Unable to load the keypair from the wallet for the implicit address atest1d9khqw36xvcnswzzx4p5zvpkxfp5vd3hxaznz33jxv6yyvfcx9rrjwp5gse5gd348ycny3zrt9hd7k. Failed with: No matching key found
```

This PR skips auto-reveal for the sentinel key.

### DIFF
--- a/apps/src/lib/client/signing.rs
+++ b/apps/src/lib/client/signing.rs
@@ -3,7 +3,7 @@
 
 use borsh::BorshSerialize;
 use namada::proto::Tx;
-use namada::types::address::{Address, ImplicitAddress};
+use namada::types::address::{masp_tx_key, Address, ImplicitAddress};
 use namada::types::key::*;
 use namada::types::storage::Epoch;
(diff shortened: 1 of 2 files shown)


tzemanovic:
closing this as an explicit `--signer` is needed for the wrapper tx - we should however improve the client to catch when `--signer` or `--signing-key` is not specified, otherwise it fails on trying to reveal the sentinel PK, which is confusing

PR #854, opened 2022-12-01
docs: update PoS user docs for 0.11.0

tzemanovic:
- replace `voting-power` cmd with `bonded-stake`
- added 2 new args for `init-validator` for their commission rate
- removed staking reward accounts - the rewards will be instead auto-bonded in the validator account


brentstone (approved):
LGTM

Issue #598, opened 2022-10-14
The minimum number of block confirmations required to bridge ERC20 tokens from Ethereum to Namada is not enforced (audit issue #6)


james-chf:
This was finding 6 from the audit of `eth-bridge-integration` branch (commit 57fc202ea33df5072c9286f5ccb339537b1be885)

> Severity: Major
> Transferring ERC20 tokens to Namada by using the transferToERC function in the contracts/Bridge.sol contract emits the TransferToNamada event. This event includes the number of block confirmations that must be reached on the Ethereum chain before validators vote to include them on Namada. This number is provided by the function parameter uint256 confirmations and must be at least the protocol-specified minimum number of confirmations, initially set to 100. However, this aforementioned invariant is not enforced in the RawTransfersToNamada::decode function in apps/src/lib/node/ledger/ethereum_node/events.rs:288. As a result, a user can circumvent the protocol-specified minimum number of confirmations, which might be exploited to double-spend funds.
> Recommendation
> We recommend validating the number of confirmations required for the TransferToNamada event to assert the required minimum.

james-chf:
This issue is valid (though the `TransfersToNamada` event is emitted by `transferToNamada` contract function rather than `transferToERC`). A relevant part of code in the audited commit is https://github.com/anoma/namada/blob/57fc202ea33df5072c9286f5ccb339537b1be885/apps/src/lib/node/ledger/ethereum_node/events.rs#L127-L136

It is still the case in latest `eth-bridge-integration`

https://github.com/anoma/namada/blob/12ca4da092c4f5f54061783237d9ca7f30b3db51/apps/src/lib/node/ledger/ethereum_node/events.rs#L126-L135

This issue also looked like it was affecting the `RawTransfersToEthereum` events as well, but the signature for that event is out of date in the ledger code, the smart contract doesn't emit custom confirmations for that event, we will fix this as part of https://github.com/anoma/namada/issues/561. For all other events we currently use the hardcoded `MIN_CONFIRMATIONS` since those events are emitted by the smart contracts themselves and don't involve any external user of the bridge, so the confirmations for those events are not customizable.

We should take the recommendation for `TransferToNamada` events and more generally - make sure a decoded `PendingEvent::confirmations` is always at least the configured minimum confirmations.

[...]

PR #855, opened 2022-12-01
Ensure minimum confirmations are enforced for pending Ethereum events

james-chf:
Closes https://github.com/anoma/namada/issues/598

This PR makes sure that when building a `PendingTransfer` from a`RawTransfersToNamada` or`RawTransfersToEthereum` event, if the custom confirmations specified is less than (the protocol specified) `min_confirmations`, then `min_confirmations` is used instead.

### DIFF
--- a/apps/src/lib/node/ledger/ethereum_node/events.rs
+++ b/apps/src/lib/node/ledger/ethereum_node/events.rs
@@ -126,7 +126,8 @@ pub mod eth_events {
             match signature {
                 signatures::TRANSFER_TO_NAMADA_SIG => {
                     RawTransfersToNamada::decode(data).map(|txs| PendingEvent {
-                        confirmations: txs.confirmations.into(),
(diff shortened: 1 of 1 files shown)


batconjurer (approved):

apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -798,6 +803,104 @@ pub mod eth_events {
                 }]
             )
         }
+
+        /// Test that for Ethereum events for which a custom number of
+        /// confirmations may be specified, if a value lower than the
+        /// protocol-specified minimum confirmations is attempted to be used,
+        /// then the protcol-specified minimum confirmations is used instead.
```
sug0, line 810:
```suggestion
        /// then the protocol-specified minimum confirmations is used instead.
```

sug0 (changes requested):

Issue #603, opened 2022-10-14
Include a check to ensure that a transfer is sent only if the slashed_amount is greater than 0 (audit issue #14)


james-chf:
This was finding 14 from the audit of `eth-bridge-integration` branch (commit 57fc202ea33df5072c9286f5ccb339537b1be885)

> Severity: Informational
> In apps/proof_of_stake/src/lib.rs:881 the transfer method is called regardless of the amount being slashed.
> Recommendation
> We recommend checking if the slashed_amount is greater than 0 before calling transfer in apps/proof_of_stake/src/lib.rs:881.

james-chf:
Relevant code in the `slash` fn which the Ethereum bridge will use once Ethereum bridge slashing logic is implemented: https://github.com/anoma/namada/blob/57fc202ea33df5072c9286f5ccb339537b1be885/proof_of_stake/src/lib.rs#L870-L886

The `slashed_amount` variable goes through being a `u64` before being converted to `Self::TokenAmount` and passed to the `transfer` fn. This is still the case in the latest code - https://github.com/anoma/namada/blob/12ca4da092c4f5f54061783237d9ca7f30b3db51/proof_of_stake/src/lib.rs#L937-L953

tzemanovic:
With change to ceiling mul in #1830, slash rate and slashable token amount being greater than 0, the slash amount that's being transferred should always be non-zero. However, we're currently not transferring slashes, so we should check on this again with #1508

cwgoes:
Not relevant.

PR #856, opened 2022-12-01
Return early in `PosBase::transfer` if an attempt is made to transfer zero tokens

james-chf:
Relates to https://github.com/anoma/namada/issues/603

This potential PR would add a helper fn for checking if an `Amount::is_zero`, and use it to short circuit `PosBase::transfer` if somehow a zero amount was attempted to be transferred.


james-chf:
pls update wasm

james-chf:
@tzemanovic not sure if this is worth adding, to resolve this audit issue https://github.com/anoma/namada/issues/603 , if not this PR can be closed

tzemanovic:
> @tzemanovic not sure if this is worth adding, to resolve audit issue #603, if not this PR can be close

If I understand the issue correctly, it is that we're not checking that slash amount is greater than 0, which is probably a good thing to verify with cubic slashing. I think we should somehow ensure this from slashing logic, maybe having some lower bound param or something. 

The transfer with 0 is almost no-op (it might write `0` into target balance, if there was none before), but I think that might be besides the point. I'm happy to merge this as the early return is better, but I don't think it solves the issue (if I grokked it right). Let's please the leave the issue open for now as we still have to change the slashing logic substantially.

james-chf:
> If I understand the issue correctly, it is that we're not checking that slash amount is greater than 0
yep, it is almost a noop like you say, we may initialize a balance to be explicitly 0 where there was none before (but no balance and 0 balance should be being treated functionally equivalently)

tzemanovic (approved):

PR #859, opened 2022-12-02
Reduce serialized size of `BlockResults`

sug0:
We replace `bit-vec` with the index set implementations in https://github.com/heliaxdev/index-set, with the goal of reducing the serialized size (in bytes) of `BlockResults` instances. The representation of `index_set::vec::VecIndexSet` is more compact than the bit vector implementation we previously depended on.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2948,6 +2948,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ce23b50ad8242c51a442f3ff322d56b02f08852c77e4c0b4d3fd684abc89c683"
 
 [[package]]
+name = "index-set"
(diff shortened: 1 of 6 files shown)


tzemanovic (approved):
nice!

murisi:
Thanks! Looks correct to me. Assuming that transaction rejections are sparse enough, this should be a major improvement. (Conversely if rejections occur frequently in production, then we would have to be paying for the additional `BTreeMap` indicies.)

murisi (approved):
Thanks! Looks correct to me. Assuming that transaction rejections are sparse enough, this should be a major improvement. (Conversely if rejections occur frequently in production, then we would have to be paying for the additional `BTreeMap` indicies.)

PR #865, opened 2022-12-06
Bridge Pool CLI tests

batconjurer:
Adds E2E tests for the bridge pool cli. The test added checks that a transfer can be submitted to the pool and that when querying the pool, the submitted transfer is present.

Since the logic for signing bridge pool roots is not in place yet, a test for constructing proofs will be a later PR.


apps/src/lib/client/eth_bridge_pool.rs
```diff
@@ -67,11 +70,19 @@ pub async fn construct_bridge_pool_proof(args: args::BridgePoolProof) {
...
-
-    println!("{:#?}", serde_json::to_string_pretty(&response));
+    let pool_contents: HashMap<String, PendingTransfer> = response
+        .into_iter()
+        .map(|transfer| (transfer.keccak256().to_string(), transfer))
+        .collect();
+    if pool_contents.is_empty() {
+        println!("Bridge pool is empty.");
+    } else {
+        println!("Bridge pool contents: ");
```
sug0, line 85:
```suggestion
        print!("Bridge pool contents: ");
```
nit: maybe this is the desired output here? `Bridge pool contents: { ...`

apps/src/lib/client/eth_bridge_pool.rs
```diff
@@ -67,11 +70,19 @@ pub async fn construct_bridge_pool_proof(args: args::BridgePoolProof) {
...
         .await
         .unwrap();
-
-    println!("{:#?}", serde_json::to_string_pretty(&response));
+    let pool_contents: HashMap<String, PendingTransfer> = response
+        .into_iter()
+        .map(|transfer| (transfer.keccak256().to_string(), transfer))
+        .collect();
+    if pool_contents.is_empty() {
+        println!("Bridge pool is empty.");
```
sug0, line 83:
```suggestion
        println!("Bridge pool is empty.");
        return;
```

tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -125,3 +131,106 @@ fn run_ledger_with_ethereum_events_endpoint() -> Result<()> {
...
+    ];
+
+    let mut namadac_tx = run!(
+        test,
+        Bin::Client,
+        tx_args,
+        Some(CLIENT_COMMAND_TIMEOUT_SECONDS)
+    )
+    .unwrap();
+    namadac_tx.exp_string("Transaction applied").unwrap();
```
sug0, line 224:
```suggestion
    namadac_tx.exp_string("Transaction accepted").unwrap();
    namadac_tx.exp_string("Transaction applied").unwrap();
```
should we perhaps add an accepted check?

sug0: also check if the RPC server is running beforehand

apps/src/lib/client/eth_bridge_pool.rs
batconjurer: This I'm not sure about until I see it on the screen. Will get back to it when I have time.

james-chf (changes requested):
I've just looked at

apps/src/lib/client/eth_bridge_pool.rs
```diff
@@ -67,11 +70,20 @@ pub async fn construct_bridge_pool_proof(args: args::BridgePoolProof) {
...
+        return;
+    } else {
+        println!("Bridge pool contents: ");
+    }
+    println!("{}", serde_json::to_string_pretty(&pool_contents).unwrap());
```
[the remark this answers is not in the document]
batconjurer: That's what I had at first, but it doesn't grep properly. So I couldn't get the e2e test to pass.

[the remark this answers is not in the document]
batconjurer: I could make json with one key "bridge_pool_contents" or something. And then the key is rest.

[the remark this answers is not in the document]
batconjurer: Yeah, that's what I'll do.

apps/src/lib/client/eth_bridge_pool.rs
batconjurer: It made the whole thing a json payload, which is what I really wanted anyway.

james-chf (approved):
lgtm, formatting needs to be fixed for CI

sug0 (approved):
run `make fmt`, otherwise lgtm!

Issue #367, opened 2022-08-21
`CheckTx` should reject transactions which cannot be charged


james-chf:
When Tendermint receives a transaction from via RPC or gossiped from another Tendermint node, it should call `CheckTx` on the ledger and only accept the transaction into its mempool if it passes. This code path is currently around the `mempool_validate` method of the `Shell`.

https://github.com/anoma/namada/blob/d1a06a26abcfc8c3c0fd8904c020a79fac1ac52d/apps/src/lib/node/ledger/shell/mod.rs#L552-L566

We must not allow transactions into the mempool for which we could not incur some cost to the sender if the transaction were to be included in a block. This is to mitigate against mempool spamming. For example, we should check that the sender account of a regular transaction has sufficient `NAM` to at least pay some base cost if the transaction should fail, or a protocol transaction was crafted and signed by an active validator to prevent fake protocol transactions bloating the mempool. This should be specced out in more detail.

As a side effect, `CheckTx` will become more computationally expensive for nodes to run as it will involve e.g. verifying signatures. 

This issue is important for the Ethereum bridge as there will be various protocol transactions to be included in a block in a timely fashion, which mempool spamming may prevent. There is potentially mempool prioritization functionalityi n Tendermint that we could use, that may or may not be WIP.

sug0:
Partially solved by https://github.com/anoma/namada/pull/817, which:

1) Checks if a protocol is indeed a `ProtocolTx`.
2) When 1 verifies, checks the validity of the protocol tx, including validating its signature against the public key of the validator in the protocol tx.

cwgoes:
@grarco I think this falls under the scope of the fee system implementation?

grarco:
Sort of. Speaking with @sug0 and @james-chf in the last few days we thought this issue might be a bit broader and non-trivial. It might be worth a discussion. For the moment, though, I think it actually falls under the fee system 👍🏻

PR #866, opened 2022-12-06
Add extra `CheckTx` validation

sug0:
Partially resolve https://github.com/anoma/namada/issues/367

We check if:

1) A transaction is successfully validated by `process_tx`.
2) A transaction is not a decrypted tx, because these can only be crafted by block proposers.

We are not charging base fees to invalid encrypted txs, yet. This needs to be specced out, first.

### DIFF
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -695,6 +695,8 @@ where
     /// Validate a transaction request. On success, the transaction will
     /// included in the mempool and propagated to peers, otherwise it will be
     /// rejected.
+    // TODO: move this to another file after 0.11 merges,
+    // since this method has become fairly large at this point
     pub fn mempool_validate(
         &self,
         tx_bytes: &[u8],
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -737,6 +737,9 @@ where
                             self.storage.last_height,
                         ) {
                             response.log = String::from(VALID_MSG);
+                            // validator set update votes should be decided
+                            // as soon as possible
+                            response.priority = i64::MAX;
                         } else {
                             response.code = 1;
                             response.log = String::from(
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -724,8 +724,9 @@ where
                             response.log = String::from(VALID_MSG);
                         } else {
                             response.code = 1;
-                            response.log = String::from(
-                                "Invalid Ethereum events vote extension",
+                            response.log = format!(
+                                "{INVALID_MSG}: Invalid Ethereum events vote \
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -737,28 +740,43 @@ where
...
+                            "{INVALID_MSG}: The following protocol tx cannot \
+                             be added to the mempool: {tx:?}"
                         );
                     }
-                    // `process_tx` errors are handled by
-                    // `Shell::finalize_block`
-                    _ => response.log = String::from(VALID_MSG),
+                    Ok(TxType::Wrapper(_) | TxType::Raw(_)) => {
+                        response.log = String::from(VALID_MSG);
+                    }
```
james-chf, line 763: Should we be allowing `TxType::Raw` into the mempool? I thought all user txs would have to be of `TxType::Wrapper`

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -737,28 +740,43 @@ where
...
                             );
                         }
                     }
                     Ok(TxType::Protocol(ProtocolTx { tx, .. })) => {
                         response.code = 1;
                         response.log = format!(
-                            "The following protocol tx cannot be added to the \
-                             mempool: {tx:?}"
+                            "{INVALID_MSG}: The following protocol tx cannot \
+                             be added to the mempool: {tx:?}"
```
james-chf, line 758: nit: it was already like this before this PR, but I'm unsure if we should debug stringify the `tx` and put it in the response message because it could be huge, "{INVALID_MSG}: Forbidden protocol transaction" or something similar might be ok

james-chf (changes requested):

apps/src/lib/node/ledger/shell/mod.rs
sug0: we were allowing raw transactions to be included before, but yeah, they're not even affecting state in FinalizeBlock

batconjurer: No, we don't allow TxType::Raw

batconjurer: Or, I should say, they will be rejected by process proposal. So we shouldn't include them

batconjurer (approved):

apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
```diff
@@ -33,6 +33,7 @@ where
     ///  * There are no duplicate Ethereum events in this vote extension, and
     ///    the events are sorted in ascending order.
     #[inline]
+    #[allow(dead_code)]
```
james-chf, line 36: I want to suggest we remove dead code from our codebase, but that's a separate code style issue. e.g. we would now remove `validate_eth_events_vext` and `validate_valset_upd_vext` in this PR if they aren't needed anymore

james-chf (approved):
lgtm

Issue #198, opened 2022-07-19
Switch to ABCI++


james-chf:
Depends on https://github.com/anoma/namada/issues/153

Currently the ledger supports running Tendermint with ABCI (Tendermint v0.34.13) or with ABCI++ (Tendermint v0.36, yet to be released). The Ethereum bridge will require using various features of ABCI++. We should eventually switch to supporting ABCI++ only, this issue is to track that switch.

If we are committed to using ABCI++ in Namada in order to support the Ethereum bridge, it may be worth switching to earlier than a v0.36 release e.g. once a v0.36 release candidate or alpha is available, or even a specific Tendermint commit that is stable enough and supports ABCI++ (like the Tendermint commit we use currently - `29e5fbcc648510e4763bd0af0b461aed92c21f30`).

james-chf:
We will be using ABCI v0.18.0 (i.e. parts of the former ABCI++) for initial Namada release but would eventually like to switch to ABCI v0.19.0 (full ABCI++) later on to support things like Ferveo and a better Ethereum bridge implementation

PR #871, opened 2022-12-08
Fix the abcipp feature flag

batconjurer:
This PR fixes up the `abcipp` feature flag. It should not compile, pass clippy, and tests. Quite a few places where there were flags before have been removed to keep the code it guarded under active maintenance. The shims were reworked to get rid to the tx records. Validating val set updates has been refactored to be switched off for the moment.


apps/src/lib/node/ledger/shims/abcipp_shim_types.rs
```diff
@@ -273,6 +273,24 @@ pub mod shim {
...
 
+        #[derive(Debug, Default)]
+        pub struct PrepareProposal {
+            pub txs: Vec<super::TxBytes>,
+        }
+
+        #[cfg(feature = "abcipp")]
+        impl From<PrepareProposal> for ResponsePrepareProposal {
+            fn from(_: PrepareProposal) -> Self {
+                Self::default()
```
james-chf, line 284: should there be a `TODO`/`unimplemented!` here? We have https://github.com/anoma/namada/issues/198 as a catch all issue we can link to

james-chf (approved):

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -267,13 +266,8 @@ where
                     );
                     false
                 })
-            })
-            .unwrap_or({
-                // NOTE: if we're not supposed to send a validator set update
-                // vote extension at a particular block height, we will
-                // just return true as the validation result
-                true
-            })
```
sug0, line 276: can you leave the original `NOTE` comment in?

shared/src/ledger/queries_ext.rs
```diff
@@ -315,7 +315,7 @@ where
     fn can_send_validator_set_update(&self, _can_send: SendValsetUpd) -> bool {
         // TODO: implement this method for ABCI++; should only be able to send
         // a validator set update at the second block of an epoch
-        true
+        false
```
sug0, line 318: :joy:

sug0 (changes requested):

PR #880, opened 2022-12-09
Don't configure AWS credentials when running tests in CI

james-chf:
Some CI jobs where we run tests don't appear to require AWS to be configured, if we remove this then these jobs can succeed in CI for forks of Namada


Fraccaman:
If we do this, `sccache` will not be able to store/pull the cacheed artifacts from s3. Same for the other PR.

james-chf:
oh right, I'll close these PRs, I am doing something like this in my fork to get these CI jobs to run. Maybe we can make sccache optional somehow, so the job can potentially succeed without needing AWS configured?

PR #887, opened 2022-12-13
Initialize `BlockSpaceAllocator` with `max_proposal_bytes`

sug0:
Based on https://github.com/anoma/namada/pull/771

Initialize `BlockSpaceAllocator` with the `max_proposal_bytes` genesis parameter, instead of `max_tx_bytes` (retrieved from `RequestPrepareProposal`).


shared/src/ledger/queries_ext.rs
```diff
@@ -165,6 +168,9 @@ pub trait QueriesExt {
         &'db self,
         epoch: Option<Epoch>,
     ) -> Box<dyn Iterator<Item = (EthAddrBook, Address, token::Amount)> + 'db>;
+
+    /// Retrieve the `max_proposal_bytes` consensus parameter from storage.
+    fn get_max_proposal_bytes(&self) -> ProposalBytes;
```
james-chf, line 173: for sure `QueriesExt` can be broken up/refactored/done away with soon! The API of the `Storage` struct grows further...

shared/src/ledger/queries_ext.rs
```diff
@@ -433,4 +439,14 @@ where
...
+
+    fn get_max_proposal_bytes(&self) -> ProposalBytes {
+        let key = get_max_proposal_bytes_key();
+        let (maybe_value, _gas) = self
+            .read(&key)
+            .expect("Must be able to read ProposalBytes from storage");
+        let value =
+            maybe_value.expect("ProposalBytes must be present in storage");
+        decode(value).expect("Must be able to decode ProposalBytes in storage")
+    }
```
james-chf, line 451: I still think ideally we should be returning errors rather than panicking here. The issue is we do not have a nice way to gracefully shut down within Shell ABCI methods upon encountering an unrecoverable error, so we will have to panic somewhere up the stack right now anyway.

james-chf (approved):
lgtm!

shared/src/ledger/queries_ext.rs
sug0: lol yeah

batconjurer: I'm actually deleting it. It will become `PosQueries`. Any funcitonality that isn't that will have to live elsewhere.

shared/src/ledger/queries_ext.rs
batconjurer: A panic in the shell triggers the abortable spawner which shuts down all the other processes. I think we should panic here.

PR #877, opened 2022-12-09
fix test cov

tzemanovic:
merge #869

PR #888, opened 2022-12-13
Fix some CI in `eth-bridge-integration`

james-chf:
- make sure there are no `cargo check` warnings when `dev` feature isn't enabled (this isn't yet checked in CI I don't think)
- make sure abcipp compiles
- update nightly version in CI for eth bridge (this is fixed in `main` already)

### DIFF
--- a/apps/src/lib/node/ledger/tendermint_node.rs
+++ b/apps/src/lib/node/ledger/tendermint_node.rs
@@ -458,6 +458,12 @@ async fn write_tm_genesis(
         .expect("Couldn't convert DateTimeUtc to Tendermint Time");
     #[cfg(not(feature = "abcipp"))]
     {
+        // NOTE(feature = "abcipp"): this setting of the block consensus
+        // parameters is guarded for the time being as it is different for
+        // Tendermint v0.37 to the ABCI++-compatible tendermint-rs we were using
+        // previously. Once ABCI++ is finished, hopefully the layout of
+        // these block consensus parameters will be similar if not the
+        // same.
         genesis.consensus_params.block = Some(block::Size {
             // maximum size of a serialized Tendermint block
             // cannot go over 100 MiB
--- a/apps/src/lib/node/ledger/tendermint_node.rs
+++ b/apps/src/lib/node/ledger/tendermint_node.rs
@@ -19,7 +19,9 @@ use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::process::Command;
 
 use crate::config;
-use crate::facade::tendermint::{block, Genesis};
+#[cfg(not(feature = "abcipp"))]
+use crate::facade::tendermint::block;
+use crate::facade::tendermint::Genesis;
 use crate::facade::tendermint_config::net::Address as TendermintAddress;
 use crate::facade::tendermint_config::{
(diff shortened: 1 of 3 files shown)


james-chf:
One CI job is failing, but that will be fixed separately by https://github.com/anoma/namada/pull/877

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -454,16 +456,21 @@ async fn write_tm_genesis(
...
+            max_bytes: (100 << 20) - 1, /* unsure if we are dealing with an
+                                         * open
+                                         * range, so it's better to subtract
+                                         * one,
+                                         * here */
+            // gas is metered app-side, so we disable it
+            // at the Tendermint level
+            max_gas: -1,
+        });
+    }
```
sug0, line 473: hm why does this not work for `abcipp`? btw we should add some kind of `NOTE(feature = "abcipp")` here, in case people in the future try to nuke this config param from the code

sug0 (changes requested):

apps/src/lib/node/ledger/tendermint_node.rs
james-chf: there is this separate block below, I just guarded this to get `make check-abcipp` etc to pass

```rust
    #[cfg(feature = "abcipp")]
    {
        genesis.consensus_params.timeout.commit =
            config.consensus_timeout_commit.into();
    }
```

james-chf: This was the error

```
error[E0308]: mismatched types
   --> apps/src/lib/node/ledger/tendermint_node.rs:457:38
    |
457 |       genesis.consensus_params.block = Some(block::Size {
    |  _____------------------------------___^
    | |     |
    | |     expected due to the type of this binding
458 | |         // maximum size of a serialized Tendermint block
459 | |         // cannot go over 100 MiB
460 | |         max_bytes: (100 << 20) - 1, /* unsure if we are dealing with an open
...   |
465 | |         max_gas: -1,
466 | |     });
    | |______^ expected struct `Size`, found enum `std::option::Option`
    |
    = note: expected struct `Size`
                 found enum `std::option::Option<Size>`
```

james-chf: I didn't look into it further than this, but I'd hope `genesis.consensus_params.block` can stay the same for ABCI++ (once ABCI++ is finished, and `tendermint-rs` updated for it fully)

sug0: oh looks like it's just a matter of wrapping the value in an `Option`. you can add a separate let binding like so:
```rs
let size = block::Size {
    // maximum size of a serialized Tendermint block
    // cannot go over 100 MiB
    max_bytes: (100 << 20) - 1, /* unsure if we are dealing with an open
                                 * range, so it's better to subtract one,
                                 * here */
    // gas is metered app-side, so we disable it
    // at the Tendermint level
    max_gas: -1,
});
let size = Some(size);
genesis.consensus_params.block = size;
```

james-chf: that works!

sug0 (approved):

Issue #69, opened 2022-02-04
simplified gas accounting


tzemanovic:
Depends on anoma/anoma#875, anoma/anoma#876, anoma/namada#8

For the limited set of whitelisted VPs and txs that can be used in a network, we'd like to implement a simplified gas accounting system, where each VP and tx WASM code will have (almost) constant gas consumption.

This has at least 2 high-level components:
 - determining the gas consumed by a tx
 - turning the gas into tx fee and charging it from the tx source account

grarco:
Closed in favor of #1010

PR #889, opened 2022-12-13
Fee specs

grarco:
Adds specs for gas and fee.

Addresses #69

### DIFF
--- a/documentation/specs/src/economics/fee-system.md
+++ b/documentation/specs/src/economics/fee-system.md
@@ -1,7 +1,143 @@
-## Fee system
+# Fee system
 
-In order to be accepted by the Namada ledger, transactions must pay fees. Transaction fees serve two purposes: first, the efficient allocation of block space given permissionless transaction submission and varying demand, and second, incentive-compatibility to encourage block producers to add transactions to the blocks which they create and publish.
+In order to be accepted by the Namada ledger, transactions must pay fees. Transaction fees serve two purposes: first, the efficient allocation of block space and gas (which are scarce resources) given permissionless transaction submission and varying demand, and second, incentive-compatibility to encourage block producers to add transactions to the blocks which they create and publish.
 
-Namada transaction fees can be paid in any fungible token which is a member of a whitelist controlled by Namada governance. Governance also sets minimum fee rates (which can be periodically updated so that they are usually sufficient) which transactions must pay in order to be accepted (but they can always pay more to encourage the proposer to prioritise them). When using the shielded pool, transactions can also unshield tokens in order to pay the required fees.
+Namada transaction fees can be paid in any fungible token which is a member of a whitelist controlled by Namada governance. Governance also sets minimum fee rates (which can be periodically updated so that they are usually sufficient) which transactions must pay in order to be accepted (but they can always pay more to encourage the proposer to prioritize them). When using the shielded pool, transactions can also unshield tokens in order to pay the required fees.
 
-The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a token identifier and $GP_{min}$ is the minimum price per unit gas which must be paid by a transaction paying fees using that asset. This whitelist can be updated with a standard governance proposal. All fees collected are paid directly to the block proposer (incentive-compatible, so that side payments are no more profitable).
+The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a token identifier and $GP_{min}$ is the minimum (base) price per unit gas which must be paid by a transaction paying fees using that asset. This whitelist can be updated with a standard governance proposal. All fees collected are paid directly to the block proposer (incentive-compatible, so that side payments are no more profitable).
+
+Fees are distributed among the delegators with the mechanism explained in the [POS](./proof-of-stake/reward-distribution.md) specs.
+
+Fees are only meant for `InnerTx` transactions: `WrapperTx`s are not subject to them.
+
+## Fee payment
+
+The `WrapperTx` struct holds all the data necessary for the payment of fees in the form of the types: `Fee`, `GasLimit` and the `PublicKey` used to derive the address of the fee payer which coincides with the signer of the wrapper transaction itself.
+
+Since fees have a purpose in allocating scarce block resources (space and gas limit) they have to be paid upfront, as soon as the transaction is deemed valid and accepted into a block (refer to [replay protection](../base-ledger/replay-protection.md) specs for more details on transactions' validity). Moreover, for the same reasons, the fee payer will pay for the entire `GasLimit` allocated and not the actual gas consumed for the transaction: this will incentivize fee payers to stick to a reasonable gas limit for their transactions allowing for the inclusion of more transactions into a block. Since the gas used by a transaction leaks a bit of information about the transaction itself: a submitter may want to obfuscate this value a bit by increasing the gas limit of the wrapper transaction but he will be charged for this (refer to section 2.1.3 of the Ferveo [documentation](https://eprint.iacr.org/2022/898.pdf)).
+
+Fees are not distributed among the validators who actively participate in the block validation process. This is because a tx submitter could be side-paying the block proposer for tx inclusion which would prevent the correct distribution of fees among validators. The fair distribution of fees is enforced by the block proposer rotation policy of Tendermint.
+
+By requesting an upfront payment, fees also serve as prevention against DOS attacks since the signer needs to pay for all the submitted transactions. More specifically, to serve as a denial-of-service and spam prevention mechanism, fee payment needs to enforce:
+
+1. **Succesful** payment at block inclusion time (implying the ability to check the good outcome at block creation time)
+2. Minimal payment overhead in terms of computation/memory requirements (otherwise fee payment itself could be exploited as a DOS vector)
+
+Given that transactions are executed in the same order they appear in the block this will lead to a common behavior across all the block proposers: they'll tend to place all the wrapper transactions before the decrypted transactions coming from the previous block. By doing this, they will make sure to prevent inner transactions from draining the addresses of the funds needed to pay fees. The proposer will be able to check in advance that fee payers have enough unshielded funds and, if this is not the case, exclude the transaction from the block and leave it in the mempool for future inclusion. This behavior ultimately leads to more resource-optimized blocks. 
+
+As a drawback, this behavior could cause some inner txs coming from the previous block to fail (in case they involve an unshielded transfer) because funds have been moved to the block proposer as a fee payment for a `WrapperTx` included in the same block. This is somehow undesirable since inner transactions' execution should have priority over the wrapper. There are two ways to overcome this issue:
+
+1. Users are responsible for correctly timing/funding their transactions with the help of the wallet
+2. We force in protocol that a block should list the wrappers after the decrypted transactions
+
+If we follow the second option the block proposers will no more be able to optimize the block (this would require running the inner transactions to calculate the possibly new unshielded balance) and, inevitably, some wrapper transactions for which fees cannot be paid will end up in the block. These will be deemed invalid during validation so that the corresponding inner transaction will not be executed, preserving the correctness of the state machine, but it represents a slight underoptimization of the block and a potential vector for DOS attacks since the invalid wrapper has allocated space and gas in the block without being charged due to the lack of funds. Because of this, we stick to the first option by not imposing any specific order in procotol.
+
+Fees are collected via protocol, in the `finalize_block` function, for `WrapperTx`s which have been processed with success: this is to prevent a malicious block proposer from including transactions that are known in advance to be invalid just to collect more fees. Given the two-block execution model of Namada (wrapper and inner tx) and the need to collect fees for the allocated resources, nothing can be done in case the inner transaction fails: by that point, fees have already been collected and no refunds will be issued, meaning that the inner tx signer is responsible for submitting a semantically valid transaction for the state of the application (importance on the lifetime parameter of the tx here). If enough funds are available, these are deducted from the unshielded storage balances of the fee payers and directed to the balance of the block proposer: payers (wrapper tx signers) are responsible for keeping enough unshielded funds for their transactions. If instead, the balance is not enough to cover fees, then the corresponding inner transaction gets discarded and no funds are moved from the payer address to the block proposer. This is due to the following reasons:
+
+- It would penalize the tx submitter who might not be responsible for the lack of funds at that moment
+- Moving insufficient funds would incentivize the block proposer to include transactions for which fees cannot be paid. Since the block proposer knows the balances of the involved addresses at block creation time and given the strategy of placing wrapper txs first, this would constitute malicious behavior by the proposer
+
+By discarding the transaction without paying fees instead we avoid these pitfalls. This logic implies that the strategy of placing wrapper transactions before any decrypted tx in the block will be reinforced. It might look like a contradiction to what was said before, a wrapper transaction included in a block will not pay fees for the resources it acquired, but this is not true:
+
+- An address might have no funds at all making it impossible to perform any payment
+- The only actor with an interest in managing block resources is the block proposer who's aware of the balances of the involved addresses
+
+To support the in-protocol fee payment mechanism we need to update the `Header` struct to carry the `ProposerAddress`:
+
+```rust
+/// The data from Tendermint header
+/// relevant for Anoma storage
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct Header {
+    /// Merkle root hash of block
+    pub hash: Hash,
+    /// Timestamp associated to block
+    pub time: DateTimeUtc,
+    /// Hash of the addresses of the next validator set
+    pub next_validators_hash: Hash,
+    /// Address of the block proposer
+    pub proposer_address: Address
+}
+```
+
+From this address, it is then possible to derive the relative Namada address for the payment.
+
+The `Fee` field of `WrapperTx` is defined as follows:
+
+```rust
+pub struct Fee {
+    /// amount of the fee
+    pub amount: Amount,
+    /// address of the token
+    pub token: Address,
+}
+```
+
+The signer of the wrapper transaction defines the token in which fees must be paid among those available in the token whitelist. At the same time, he also sets the amount which must meet the minimum price per gas unit for that token $GP_{min}$ (also defined in the whitelist). The difference between the minimum and the actual value set by the submitter represents the incentive for the block proposer to prefer the inclusion of this transaction over other ones.
+
+The block proposer can check the validity of these two parameters while constructing the block. These validity checks are also replicated in `process_proposal` and `mempool_check`. In case a transaction with invalid parameters ended up in a block, it would be discarded without paying any fee (as already explained earlier in this document). The block proposer and the `process_proposal` function should also cache the available funds of every fee-paying address involved in the block: this is because a signer might submit more than one transaction per block and the check on funds should take into consideration the updated value of the unshielded balance.
+
+Since the whitelist can be changed via governance, transactions could fail these checks in the block where the whitelist change happens. For `mempool_check`, the checks could reject transactions that may become valid in the future or vice-versa: since we can assume a slow rate of change for these parameters and mempool and block space optimizations are a priority, it is up to the clients to track any changes of these parameters and act accordingly.
+
+## Gas accounting
+
+We provide a mapping between all the whitelisted transactions and VPs to their cost in gas units. Being the cost hardcoded, it is guaranteed that the same transaction will always require the same amount of gas: since the price per gas is controlled via governance, though, the price for the same transaction may vary in time. A transaction is also charged with the gas required by the validity predicates that it triggers.
+
+Gas accounting is about preventing a transaction from exceeding two gas limits:
+
+1. Its own `GasLimit` (declared in the wrapper transaction)
+2. The gas limit of the entire block
+
+### Wrapper GasLimit
+
+The protocol injects a gas counter in each transaction and VP to be executed which allows monitoring of the exact amount of gas utilized. To do so, the gas meter simply checks the hash of the transaction or VP against the table in storage to determine which one it is and, from there, derives the amount of gas required.
+
+To perform the check we need the limit which was declared by the corresponding wrapper transaction: this limit should be saved in storage together with the queue of encrypted transactions for easy access.
+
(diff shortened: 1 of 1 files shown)


documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+### Block GasLimit
+
+This constraint is given by the following two:
+
+- The compliance of each inner transaction with the `WrapperTx` gas limit explained in the previous section
+- The compliance of the cumulative wrapper transactions' `GasLimit` with the maximum gas allowed for a block
+
+Tendermint doesn't provide more than the `BlockSize.MaxGas` parameter, leaving the validation step to the application (see [tendermint spec](https://github.com/tendermint/tendermint/blob/29e5fbcc648510e4763bd0af0b461aed92c21f30/spec/core/data_structures.md#consensusparams) and [issue](https://github.com/tendermint/tendermint/issues/2310)): therefore, instead of using the Tendermint provided param, Namada introduces a `MaxBlockGas` protocol parameter.
+This limit is checked during block validation, in `process_proposal`: if the block exceeds the maximum amount of gas allowed, the validators will still accept the block and discard only the excess transactions.
```
grarco, line 111: To my understanding, Tendermint validators prevote for `nil` when a block exceeds the size limit (which is a related topic). In here I propose instead to use the same logic we apply to invalid txs, discarding only the invalid ones and still accept the block (at the moment we reject the entire block only when the order of the decrypted txs doesn't match the order of the corresponding wrappers committed in the previous block).

We may think about rejecting the entire block for certain cases (in generale for those in which the proposer can check at block construction time whether a tx is valid or not) or for some sort of punishment in the form of slashing

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
 
-The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a token identifier and $GP_{min}$ is the minimum price per unit gas which must be paid by a transaction paying fees using that asset. This whitelist can be updated with a standard governance proposal. All fees collected are paid directly to the block proposer (incentive-compatible, so that side payments are no more profitable).
\ No newline at end of file
+The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a token identifier and $GP_{min}$ is the minimum (base) price per unit gas which must be paid by a transaction paying fees using that asset. This whitelist can be updated with a standard governance proposal. All fees collected are paid directly to the block proposer (incentive-compatible, so that side payments are no more profitable).
+
+Fees are distributed among the delegators with the mechanism explained in the [POS](./proof-of-stake/reward-distribution.md) specs.
```
cwgoes, line 9:
```suggestion
Fees are distributed among the delegators with the mechanism explained in the [proof-of-stake reward distribution specs](./proof-of-stake/reward-distribution.md).
```

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+Fees are only meant for `InnerTx` transactions: `WrapperTx`s are not subject to them.
```
cwgoes, line 11: not sure I follow this - don't the fees need to be included in the outermost transaction? we have to check fees before accepting any transaction. also this seems to contradict the below

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+## Fee payment
+
+The `WrapperTx` struct holds all the data necessary for the payment of fees in the form of the types: `Fee`, `GasLimit` and the `PublicKey` used to derive the address of the fee payer which coincides with the signer of the wrapper transaction itself.
+
+Since fees have a purpose in allocating scarce block resources (space and gas limit) they have to be paid upfront, as soon as the transaction is deemed valid and accepted into a block (refer to [replay protection](../base-ledger/replay-protection.md) specs for more details on transactions' validity). Moreover, for the same reasons, the fee payer will pay for the entire `GasLimit` allocated and not the actual gas consumed for the transaction: this will incentivize fee payers to stick to a reasonable gas limit for their transactions allowing for the inclusion of more transactions into a block. Since the gas used by a transaction leaks a bit of information about the transaction itself: a submitter may want to obfuscate this value a bit by increasing the gas limit of the wrapper transaction but he will be charged for this (refer to section 2.1.3 of the Ferveo [documentation](https://eprint.iacr.org/2022/898.pdf)).
```
cwgoes, line 17:
```suggestion
Since fees have a purpose in allocating scarce block resources (space and gas limit) they have to be paid upfront, as soon as the transaction is deemed valid and accepted into a block (refer to [replay protection](../base-ledger/replay-protection.md) specs for more details on transactions' validity). Moreover, for the same reasons, the fee payer will pay for the entire `GasLimit` allocated and not the actual gas consumed for the transaction: this will incentivize fee payers to stick to a reasonable gas limit for their transactions allowing for the inclusion of more transactions into a block. Since the gas used by a transaction leaks a bit of information about the transaction itself: a submitter may want to obfuscate this value a bit by increasing the gas limit of the wrapper transaction but they will be charged for this (refer to section 2.1.3 of the Ferveo [documentation](https://eprint.iacr.org/2022/898.pdf)).
```

(fee payer may be a person or another sort of entity, we really don't know)

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+- It would penalize the tx submitter who might not be responsible for the lack of funds at that moment
+- Moving insufficient funds would incentivize the block proposer to include transactions for which fees cannot be paid. Since the block proposer knows the balances of the involved addresses at block creation time and given the strategy of placing wrapper txs first, this would constitute malicious behavior by the proposer
+
+By discarding the transaction without paying fees instead we avoid these pitfalls. This logic implies that the strategy of placing wrapper transactions before any decrypted tx in the block will be reinforced. It might look like a contradiction to what was said before, a wrapper transaction included in a block will not pay fees for the resources it acquired, but this is not true:
```
cwgoes, line 40: including a transaction which cannot pay fees should be considered an invalid block proposal

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+    /// address of the token
+    pub token: Address,
+}
+```
+
+The signer of the wrapper transaction defines the token in which fees must be paid among those available in the token whitelist. At the same time, he also sets the amount which must meet the minimum price per gas unit for that token $GP_{min}$ (also defined in the whitelist). The difference between the minimum and the actual value set by the submitter represents the incentive for the block proposer to prefer the inclusion of this transaction over other ones.
+
+The block proposer can check the validity of these two parameters while constructing the block. These validity checks are also replicated in `process_proposal` and `mempool_check`. In case a transaction with invalid parameters ended up in a block, it would be discarded without paying any fee (as already explained earlier in this document). The block proposer and the `process_proposal` function should also cache the available funds of every fee-paying address involved in the block: this is because a signer might submit more than one transaction per block and the check on funds should take into consideration the updated value of the unshielded balance.
```
cwgoes, line 78: correct proposers should never create blocks with invalid transactions, and such block proposals should be rejected, if possible

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+Since the whitelist can be changed via governance, transactions could fail these checks in the block where the whitelist change happens. For `mempool_check`, the checks could reject transactions that may become valid in the future or vice-versa: since we can assume a slow rate of change for these parameters and mempool and block space optimizations are a priority, it is up to the clients to track any changes of these parameters and act accordingly.
+
+## Gas accounting
+
+We provide a mapping between all the whitelisted transactions and VPs to their cost in gas units. Being the cost hardcoded, it is guaranteed that the same transaction will always require the same amount of gas: since the price per gas is controlled via governance, though, the price for the same transaction may vary in time. A transaction is also charged with the gas required by the validity predicates that it triggers.
```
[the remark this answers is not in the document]
cwgoes: here we need to define how exactly gas is measured, it's not clear to me from this part

documentation/specs/src/economics/fee-system.md
cwgoes: What is the benefit of accepting the block anyways? Is there any case where a correct proposer would accidentally create an invalid block?

cwgoes (changes requested):
Thanks for the spec! Left some comments, and a general one: no unshielded wrapper transaction fee payments is a pretty serious drawback :frowning: - I don't think we would need a full `Transfer` transaction, but could we special-case a very simple MASP transaction part involving one unshielding action, perhaps?

documentation/specs/src/economics/fee-system.md
grarco: Yes this is bad phrasing, what I meant is that even if fees are specified in the outer transactions they are relative to the inner one, meaning that the outer per se is not subject to fees. I'll rephrase this

documentation/specs/src/economics/fee-system.md
grarco: The only benefit is that we don't need to go through another Tendermint round for that specific block height, so we have an higher TPS in these cases.

A block proposer is able to determine whether a transaction is valid or not in most cases, expect for the inner tx wasm code (like trying to perform a transfer with not enough funds). At the moment, if not enough valid txs are available to fill the block, the proposer could place some invalid txs in there since he doesn't have a disincentive to do so. These invalid txs would be correctly rejected by the validators and the proposer could not redeem fees for these but there's still a small damage being done: since these invalid txs have been inserted in a block they will be removed from mempool once the block gets committed, forcing the submitters to resend them.

So I guess we can reject blocks containing invalid txs (for which validity can be checked at `process_proposal` time)

documentation/specs/src/economics/fee-system.md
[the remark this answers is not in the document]
grarco: Correct, that part is missing. For this first implementation we could go with something as basic as the wasm size for a specific transaction/vp. At a later time we can come up with a better system taking into account the specific machine instructions executed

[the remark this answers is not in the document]
tzemanovic: alternatively, I think we could get some average run time for each whitelisted tx and use that until we have more fine-grained gas metering per ops

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,143 @@
...
+
+To perform the check we need the limit which was declared by the corresponding wrapper transaction: this limit should be saved in storage together with the queue of encrypted transactions for easy access.
+
+Since the hash can be retrieved as soon as the transaction gets decrypted, we can immediately check whether the `GasLimit` set in the corresponding wrapper is enough to cover this amount. This check, though, is weak because we also need to keep in account the gas required for the involved VPs which is hard to determine ahead of time: this is just an optimization to short-circuit the execution of transactions whose gas limit is not enough to cover for even the tx itself.
+
+When executing the VPs the procedure is the same and, again, since we know ahead of time the gas required by each VP we can immediately terminate the execution if it overshoots the limit.
+
+In any case, if the gas limit is exceeded, the transaction is considered invalid and all the modifications applied to the WAL get discarded. This doesn't affect the other transactions which can be executed normally (see the following section).
```
tzemanovic, line 101: I would just explicitly add here that the gas used for the tx that exceed its gas limit is still added to the used gas in the block

documentation/specs/src/economics/fee-system.md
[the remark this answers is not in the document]
grarco: Yes, the run time looks like a better proxy for cost estimation

documentation/specs/src/economics/fee-system.md
grarco: Actually, since the block gas limit has to be compared against the the `GasLimit` declared in the `WrapperTx` we don't need to monitor the actual gas usage of an `InnerTx` against the block gas limit but only against the declared gas limit of the corresponding wrapper. So the gas meter could just check this constraint and avoid checking for the block gas limit since that is indirectly checked by the combination of these:
- checking in `process_proposal` that the cumulated `GasLimit` of alle the wrappers does not exceed the block gas limit
- preventing every inner from exceeding the gas limit allocated by its wrapper

documentation/specs/src/economics/fee-system.md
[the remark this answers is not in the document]
cwgoes: if runtime is not particularly input dependent then average runtime is OK but if it is input dependent then I'm not sure that would suffice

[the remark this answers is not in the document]
grarco: We can expect the runtime to depend on the specific input, so yes this solution would just be an approximation (not necessarily close to real). This is just temporary while we develop a proper gas meter able to track the actual execution cost at runtime

cwgoes (dismissed):
mostly LGTM, just one nit

sug0:
@grarco For reference: also related to #367

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
-The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a token identifier and $GP_{min}$ is the minimum price per unit gas which must be paid by a transaction paying fees using that asset. This whitelist can be updated with a standard governance proposal. All fees collected are paid directly to the block proposer (incentive-compatible, so that side payments are no more profitable).
\ No newline at end of file
+The token whitelist consists of a list of $(T, GP_{min})$ pairs, where $T$ is a
+token identifier and $GP_{min}$ is the minimum (base) price per unit gas which
+must be paid by a transaction paying fees using that asset. This whitelist can
+be updated with a standard governance proposal. All fees collected are paid
+directly to the block proposer (incentive-compatible, so that side payments are
+no more profitable).
+
+Fees are distributed among the delegators with the mechanism explained in the
```
cwgoes, line 23: No, they aren't - fees go only to the proposer. Proof-of-stake _rewards_ are distributed to the delegators.

(otherwise it's profitable for the proposer to accept side payments instead)

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
+for their transactions allowing for the inclusion of more transactions into a
+block. Since the gas used by a transaction leaks a bit of information about the
+transaction itself, a submitter may want to obfuscate this value a bit by
+increasing the gas limit of the wrapper transaction (refer to section 2.1.3 of
+the Ferveo [documentation](https://eprint.iacr.org/2022/898.pdf)).
+
+Fees are not distributed among the validators who actively participate in the
+block validation process. This is because a tx submitter could be side-paying
+the block proposer for tx inclusion which would prevent the correct distribution
+of fees among validators. The fair distribution of fees is enforced by the block
```
cwgoes, line 52:
```suggestion
of fees among validators. The fair distribution of fees is enforced by the stake-proportional block
```

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
+proposer rotation policy of Tendermint.
+
+By requesting an upfront payment, fees also serve as prevention against DOS
+attacks since the signer needs to pay for all the submitted transactions. More
+specifically, to serve as a denial-of-service and spam prevention mechanism, the
+fee system needs to enforce:
+
+1. **Succesful** payment at block inclusion time (implying the ability to check
```
cwgoes, line 60:
```suggestion
1. **Successful** payment at block inclusion time (implying the ability to check
```

documentation/specs/src/economics/fee-system.md
cwgoes: ah as we have a `GP_{min}` we could actually pay `gas * GP_min` to everyone if we wanted, and the difference to the proposer, that would also be incentive-compatible

cwgoes: do you think that this would be easy to do? it is slightly preferable if possible

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
+implies that:
+
+1. The unshielding provides just the right amount of funds to pay fees
+2. The actual wasm execution runs successfully
+
+The first condition can be tested statically and requires that:
+
+1. The tx encodes a `Transfer`
+2. The `shielded` field must be set to `Some`
+3. The `source` address must be the masp4. The `target` address matches that of
```
cwgoes, line 221: what's the masp4?

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
+5. The `amount`, added to the already available unshielded balance for that
+   token, is just enough to cover the fees, i.e. the value given by
+   $Fee.amount * GasLimit$ (to prevent leveraging this transfer for other
+   purposes)
+
+The spending key associated with this operation could be relative to any address
+as long as the signature of the transfer itself is valid.
+
+If checks 1 to 5 fail, the transaction can be safely discarded, while if the
+check fails at point 6 the transaction could be kept in mempool for future
```
cwgoes, line 233: what is point 6? there are only 5 points above

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,378 @@
...
+### Protocol transactions
+
+Protocol transactions can only be correctly crafted by validators and serve a
+role in allowing the chain to function properly. Given these, they are not
+subject to fees and do not charge gas.
+
+## Gas accounting
+
+We provide a mapping between all the whitelisted transactions and VPs to their
+cost in gas units: more specifically, the cost of a tx/VP is given by the run
```
cwgoes, line 263: but this is a fixed mapping, right? the actual run-time execution speed could vary

cwgoes (changes requested):
Thanks! A few comments/questions.

documentation/specs/src/economics/fee-system.md
grarco: Typo

documentation/specs/src/economics/fee-system.md
grarco: Yes, for the moment it will be a fixed mapping. After this first implementation is done, we'll start working on a proper run-time gas meter

documentation/specs/src/economics/fee-system.md
grarco: It is technically feasible but very inefficient. I think it would be better to keep it as is for the moment and then implement the distribution once #1115 has been completed

cwgoes: wait why would this be inefficient?

grarco: Because at the moment we don't have an easy way to get all the delegators for a specific validator. To do so we need to iterate over all the bonds deltas

cwgoes (changes requested):

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,391 @@
...
+on transactions' validity). Moreover, for the same reasons, the fee payer will
+pay for the entire `GasLimit` allocated and not the actual gas consumed for the
+transaction: this will incentivize fee payers to stick to a reasonable gas limit
+for their transactions allowing for the inclusion of more transactions into a
+block. Since the gas used by a transaction leaks a bit of information about the
+transaction itself, a submitter may want to obfuscate this value a bit by
+increasing the gas limit of the wrapper transaction (refer to section 2.1.3 of
+the Ferveo [documentation](https://eprint.iacr.org/2022/898.pdf)).
+
+Fees are not distributed among the validators who actively participate in the
```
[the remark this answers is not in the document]
grarco: We were discussing this [here](https://github.com/anoma/namada/pull/889#discussion_r1098311749)

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,391 @@
...
+unshielding transfer attached to the wrapper. More specifically the correctness
+implies that:
+
+1. The unshielding provides just the right amount of funds to pay fees
+2. The actual wasm execution runs successfully
+
+The first condition can be tested statically and requires that:
+
+1. The tx encodes a `Transfer`
+2. The `shielded` field must be set to `Some`
```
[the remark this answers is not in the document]
grarco: Either that or we could place a gas limit on the unshielding transaction itself, what do you think is the best approach?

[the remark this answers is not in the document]
cwgoes: Perhaps both, to be honest. This is a potential DoS vector if it's non-trivial to check.

documentation/specs/src/economics/fee-system.md
cwgoes: Ah, but staking fee distribution should be able to accomplish this (I don't suggest iterating). We can save this for a future upgrade, though.

documentation/specs/src/economics/fee-system.md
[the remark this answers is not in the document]
cwgoes: Ah - replied there - let's save this for a future upgrade, there's a way but I think the engineering work is too much to justify right now.

adrianbrink:
@grarco to finish this pr in the next couple of days and merge latest by mid next week

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,376 @@
...
+```rust
+pub struct WrapperTx {
+  /// The fee to be paid for including the tx
+  pub fee: Fee,
+  /// Used to determine an implicit account of the fee payer
+  pub pk: common::PublicKey,
+  /// Max amount of gas that can be used when executing the inner tx
+  pub gas_limit: GasLimit,
+  /// The optional unshielding transaction for fee payment
+  pub unshield: Option<Transaction>,
```
cwgoes, line 174: This should be only a shielded tx, not a full `Transaction`, right?

documentation/specs/src/economics/fee-system.md
```diff
@@ -1,7 +1,376 @@
...
+  /// the encrypted payload
+  pub inner_tx: EncryptedTx,
+  /// sha-2 hash of the inner transaction acting as a commitment
+  /// the contents of the encrypted payload
+  pub tx_hash: Hash,
+}
+```
+
+The new `unshield` field carries an optional masp transaction struct. The
```
cwgoes, line 183: what is the type of a "MASP transaction struct"? presumably we should use that in the field above

cwgoes (approved):
ACK modulo minor comments

documentation/specs/src/economics/fee-system.md
grarco: Exactly, but in code the shielded (masp) part of a transfer is a struct called `Transaction`

PR #891, opened 2022-12-13
Move `wasm_for_tests/` into the wasm workspace

james-chf:
Closes https://github.com/anoma/namada/issues/102

This PR attempts to move our test wasms (`wasm_for_tests/wasm_source`) into the `wasm` workspace (`wasm/test_wasms`). Previously test wasms were prebuilt directly under `wasm_for_tests/*.wasm` - since `wasm/*.wasm` globs our production wasms, the test wasms are now built under `wasm/test_wasms/built/*.wasm`. This is just to get things working, not actually sure where is best to put them.

Low priority, may be better to tidy up references to the `wasm_for_tests` directory first in our code, as it's referenced in many places.


james-chf:
Moving this works straightforwardly enough but the current `wasm_for_tests` path is hardcoded/references in disparate places, we should probably clean that up a bit first before doing this move (e.g. as in https://github.com/anoma/namada/pull/893)

PR #893, opened 2022-12-14
Add `namada_test_utils` crate enumerating test wasms

james-chf:
We use `wasm_for_tests` wasms in tests across multiple crates (`namada_apps`, `namada`, `namada_tests`, `namada_wasm`) - we duplicate logic for getting the relative paths to and reading test wasms. This PR tries to consolidate that logic to one place.

Added a `namada_test_utils` crate that can be pulled in in `[dev-dependencies]` as needed, but alternatively, the `TestWasms` enum could go somewhere like `namada_core` and be guarded by the "testing" flag. I think it would make sense to have a test utils crate of some sort as 

This PR only changes very few existing usages to check it works, there are more usages that should be updated e.g. as in https://github.com/anoma/namada/pull/891).


test_utils/src/lib.rs
```diff
@@ -1 +1,96 @@
...
+        let repo_root = cwd
+            .ancestors()
+            .find(|path| path.join("CHANGELOG.md").exists())
+            .unwrap_or_else(|| {
+                panic!(
+                    "Couldn't find the root of the repository for the current \
+                     working directory {}",
+                    cwd.to_string_lossy()
+                )
+            });
```
james-chf, line 64: We don't have the `.git` directory when running tests in CI, so this code crudely looks for the ancestory directory which has a `CHANGELOG.md` file, there is probably a better way to do this.

james-chf:
pls update wasm

james-chf:
I've redone this PR on top of 0.14

james-chf:
pls update wasm

PR #894, opened 2022-12-14
Update the test wasm `tx_write_storage_key` to be able to write any arbitrary value

james-chf:
This PR changes `tx_write_storage_key.wasm` to be able to write any specified value, rather than always writing `b"arbitrary value"`. Also renames `tx_write_storage_key.wasm` to `tx_write.wasm`

Motivation for this PR is to use it in https://github.com/anoma/namada/pull/886, for writing arbitrary `token::Amount` values to balance keys in e2e tests.


core/src/types/tx_data.rs
```diff
@@ -0,0 +1,26 @@
...
+    Hash,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+pub struct TxWriteData {
+    /// The storage key to be written to.
+    pub key: storage::Key,
+    /// The bytes to be written.
+    pub value: Vec<u8>,
+}
```
james-chf, line 26: I'm not sure where best to put this, it needs to be importable into both the `namada_wasm_for_tests` and `namada_tests` crates. Maybe a `namada_test_utils`  crate like in https://github.com/anoma/namada/pull/893, but have put this here for now.

james-chf:
pls update wasm

core/src/types/tx_data.rs
tzemanovic: the crate looks like a good idea! In the interim, I think you could put it behind `cfg(feature = "testing")`, but it's not enabled from `wasm_tests` yet

james-chf: I tried this and having to use `namada_core/testing` in the `namada_wasm_for_tests` crate seems to break the wasm build somehow - probably via the pulling in of the `rand` crate. e.g.

```
 > make -C wasm_for_tests/wasm_source tx_write
cargo build --release --target wasm32-unknown-unknown --target-dir 'target' --features tx_write && \
        cp "./target/wasm32-unknown-unknown/release/namada_wasm_for_tests.wasm" ../tx_write.wasm
   Compiling num-traits v0.2.15
   Compiling tempfile v3.3.0
   Compiling wait-timeout v0.2.0
error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /Users/main/.cargo/registry/src/github.com-1ecc6299db9ec823/wait-timeout-0.2.0/src/lib.rs:66:9
   |
66 |         imp::wait_timeout(self, dur)
   |         ^^^ use of undeclared crate or module `imp`

For more information about this error, try `rustc --explain E0433`.
```

I can try out putting this in a `namada_test_utils` crate now since that should ultimately be less hacky.

james-chf: I've put this in a `namada_test_utils` crate now

PR #896, opened 2022-12-14
Murisi/parameterized tx

murisi:
Working on moving functions from `tx.rs` and `rpc.rs` into the shared crate. Plan to do this in 3 steps:
* Make the command line argument structs independent of a `Context` object
* Remove the dependence of `tx.rs` and `rpc.rs` functions on the `Context` object
* Move as many functions from `tx.rs` and `rpc.rs` into the `shared` crate as possible


mariari:
Closed see #921 for a successor and #925 for its integration branch into v0.12.0

Issue #882, opened 2022-12-09
Insufficient key matching in ledger params


sug0:
Currently, the ledger takes the following parameters:
https://github.com/anoma/namada/blob/f9d72f0e850b793af246afb9477ab029ae716b5b/core/src/ledger/parameters/storage.rs#L5-L14
However, when matching parameter storage keys, we only match against a subset of them:
https://github.com/anoma/namada/blob/f9d72f0e850b793af246afb9477ab029ae716b5b/core/src/ledger/parameters/storage.rs#L21-L27
We should add the remaining key matches to `is_protocol_parameter_key`, or better yet, create some sort of `KeySet` abstraction using, e.g. prefix trees, which will allow us to efficiently match against a large set of storage keys.

PR #903, opened 2022-12-16
Fix key matching in ledger params

sug0:
Proposed fix for #882

Instead of comparing storage keys in linear time, this PR proposes a binary search over the storage keys that have been declared. This solution requires us to maintain a sorted array of all keys, though, which might still be quite error prone.

### DIFF
--- a/core/src/ledger/parameters/storage.rs
+++ b/core/src/ledger/parameters/storage.rs
@@ -13,6 +13,21 @@ const POS_GAIN_D_KEY: &str = "pos_gain_d";
 const STAKED_RATIO_KEY: &str = "staked_ratio_key";
 const POS_INFLATION_AMOUNT_KEY: &str = "pos_inflation_amount_key";
 
+// keep these keys in sync with the defs above;
+// make sure to store them in sorted order!
+const ALL_KEYS: &[&str] = &[
+    EPOCH_DURATION_KEY,
+    EPOCHS_PER_YEAR_KEY,
+    IMPLICIT_VP_KEY,
+    MAX_EXPECTED_TIME_PER_BLOCK_KEY,
+    POS_GAIN_D_KEY,
+    POS_GAIN_P_KEY,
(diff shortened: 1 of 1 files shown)


core/src/ledger/parameters/storage.rs
```diff
@@ -13,17 +13,37 @@ const POS_GAIN_D_KEY: &str = "pos_gain_d";
 const STAKED_RATIO_KEY: &str = "staked_ratio_key";
 const POS_INFLATION_AMOUNT_KEY: &str = "pos_inflation_amount_key";
 
+// keep these keys in sync with the defs above;
+// make sure to store them in sorted order!
+const ALL_KEYS: &[&str] = &[
```
tzemanovic, line 18: maybe we could add a test case to check that they are indeed sorted

sug0: yeah, we can't test if all the keys were included in the slice though :(

sug0: but I can write a test to check if the keys are sorted!

tzemanovic: Yeah, destructuring a struct pattern is great for this. We can do something like:

```rust
struct Keys {
    epoch_duration: &'static str,
    vp_whitelist: &'static str,
    ...
}

const KEYS: Keys = Keys {
    epoch_duration: "epoch_duration",
    vp_whitelist: "vp_whitelist",
};

const ALL_KEYS: &[&str] = {
    // if something is missing, we get compiler error here
    let Keys {
        epoch_duration,
        vp_whitelist,
    } = KEYS;
    &[epoch_duration, vp_whitelist]
};
```

sug0: I found a neat crate that creates an array with all enum variants https://crates.io/crates/enum_array

I'll push a commit using this crate in a bit

PR #912, opened 2022-12-16
remove prefix iter

### LINKED ISSUE #902: remove reverse prefix iterator
the reverse prefix iterator is currently relying on rocksdb and its API is not working as expected (setting `Direction::Reverse` in rocksdb `iterator_opt` doesn't reverse the order of the matched items), see https://github.com/rust-rocksdb/rust-rocksdb/blob/8710105cb5811dfbf36f37029747a038cebd732b/tests/test_iterator.rs

tzemanovic:
closes #902


tzemanovic:
pls update wasm

tzemanovic:
we can also remove `reverse_order` from mockdb.rs

tzemanovic:
> `reverse_order`

done

apps/src/lib/node/ledger/storage/mod.rs
```diff
@@ -344,4 +347,45 @@ mod tests {
...
+        let mismatched_prefix = storage::Key::parse("different").unwrap();
+        // We'll write sub-key in some random order to check prefix iter's order
+        let sub_keys = [2_i32, 1, i32::MAX, -1, 260, -2, i32::MIN, 5, 0];
+
+        for i in sub_keys.iter() {
+            let key = prefix.push(i).unwrap();
+            let value = i.try_to_vec().unwrap();
+            storage.write(&key, value).unwrap();
+
+            let key = mismatched_prefix.push(i).unwrap();
```
brentstone, line 373: Should we still have the `mismatched_prefix` in here? Doesn't look like it's used for much.

brentstone (approved):
Aside from the `mismatched_prefix` usage in `test_persistent_storage_prefix_iter`, LGTM

apps/src/lib/node/ledger/storage/mod.rs
tzemanovic: let's keep it to cover that the `mismatched_prefix` is not returned from `iter_prefix`

PR #913, opened 2022-12-16
fix iter prefix

### LINKED ISSUE #165: Tx/VP prefix iter should include keys from write log
The prefix iterator for Tx and VP posterior state should also iterate over the storage keys from write log. Currently, they only iterate over the keys committed to the storage, but replace any overwritten values from write log.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763005834348) by [Unito](https://www.unito.io)

tzemanovic:
fixes #165


tzemanovic:
pls update wasm

brentstone (approved):
LGTM

PR #914, opened 2022-12-16
Update environment-setup.md

renatodelgaudio:
Namada version should be the same version as specified in NAMADA_TAG


bengtlofgren:
Already changed in [PR 955](https://github.com/anoma/namada/pull/955)

PR #915, opened 2022-12-16
Add `EthereumBridgeConfig::read` method

james-chf:
Needed for https://github.com/anoma/namada/pull/810

This PR adds an `EthereumBridgeConfig::read` method and makes some other miscellaneous changes to the structs in `parameters.rs`

This is the only file being changed in PR https://github.com/anoma/namada/pull/810 that is under the `shared` directory, and so may move to `core`

### DIFF
--- a/ethereum_bridge/src/parameters.rs
+++ b/ethereum_bridge/src/parameters.rs
@@ -5,8 +5,8 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use namada_core::ledger::storage;
 use namada_core::ledger::storage::types::encode;
 use namada_core::ledger::storage::Storage;
-use namada_core::types::ethereum_events::EthAddress;
 use namada_core::ledger::storage_api::StorageRead;
+use namada_core::types::ethereum_events::EthAddress;
 use namada_core::types::storage::Key;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


batconjurer (approved):

ethereum_bridge/src/parameters.rs
```diff
@@ -157,13 +174,78 @@ impl EthereumBridgeConfig {
...
+        let bridge_contract_key = bridge_storage::bridge_contract_key();
+        let governance_contract_key = bridge_storage::governance_contract_key();
+
+        let Some(min_confirmations) = StorageRead::read::<MinimumConfirmations>(
+            storage,
+            &min_confirmations_key,
+        )
+        .unwrap_or_else(|err| {
+            panic!("Could not Borsh-deserialize {min_confirmations_key}: {err:?}")
+        }) else {
```
sug0, line 198: I think the first error you encounter is an I/O error, no?

ethereum_bridge/src/parameters.rs
```diff
@@ -157,13 +174,78 @@ impl EthereumBridgeConfig {
...
+            contracts: Contracts {
+                native_erc20,
+                bridge: bridge_contract,
+                governance: governance_contract,
+            },
+        })
+    }
+}
+
+fn must_read_key<DB, H, T: BorshDeserialize>(
```
sug0, line 221: add a docstring

sug0 (changes requested):

ethereum_bridge/src/parameters.rs
james-chf: I'm not sure, you may be right though, just going by the API of `StorageRead::read`, it doesn't specify the causes of errors that might be returned (beyond the wrapper types at https://github.com/anoma/namada/blob/1f8fd9d751e5754eb39018fdc2fd1a7ef7e71b37/core/src/ledger/storage_api/error.rs#L6-L15), so I've made this error message more generic (the `{err:?}` should give more context anyway).

ethereum_bridge/src/parameters.rs
james-chf: added a docstring

PR #916, opened 2022-12-17
parameterized tx

mariari:
Followup of #896, accomplishing the same goals.


mariari:
Supper seeded by #921

PR #918, opened 2022-12-18
Murisi/parameterized tx2

murisi:
An attempt to move functionality from `tx.rs`, `rpc.rs`, and `signing.rs` by means of parameterization by concrete clients. More concretely, the following changes were made:
* Parameterized the `tx.rs` and `rpc.rs` argument types so that they can be used with or without a `Context`
* Separated the pure from the impure part of `Wallet` and moved the pure part into the `shared` crate
* Moved the function implementations in `rpc.rs` and `tx.rs` into the `shared` crate leaving behind function wrappers in `apps` (that could then be used to capture `Result`s and print them out in the CLI)
* Simplified the `ShieldedContext` trait by removing those functions in it that have been moved into the `shared` crate

With this approach, for each platform `ShieldedContext`, `WalletUtils`, and just the `perform` function of Tendermint's `Client` trait would have to be implemented (since the rest of `Client` has default implementations specified in terms of `perform`) for each client.

The only dependencies added to the `shared` crate are `toml`, `bimap`, `orion`, `tokio`, and `serde`. Hopefully this should not cause issues when building WASM targets.


mariari:
Superseded by #921

PR #921, opened 2022-12-19
Shared Sdk 

mariari:
Followup of #918 and succeeds #916

An attempt to move functionality from `tx.rs`, `rpc.rs`, and `signing.rs` by means of parameterization by concrete clients. More concretely, the following changes were made:
* Parameterized the `tx.rs` and `rpc.rs` argument types so that they can be used with or without a `Context`
* Separated the pure from the impure part of `Wallet` and moved the pure part into the `shared` crate
* Moved the function implementations in `rpc.rs` and `tx.rs` into the `shared` crate leaving behind function wrappers in `apps` (that could then be used to capture `Result`s and print them out in the CLI)
* Simplified the `ShieldedContext` trait by removing those functions in it that have been moved into the `shared` crate

With this approach, for each platform `ShieldedContext`, `WalletUtils`, and just the `perform` function of Tendermint's `Client` trait would have to be implemented (since the rest of `Client` has default implementations specified in terms of `perform`) for each client.

The only dependencies added to the `shared` crate are `toml`, `bimap`, `orion`, `tokio`, and `serde`. Hopefully this should not cause issues when building WASM targets.

The integration branch corresponding to this branch is #925


adrianbrink:
@memasdeligeorgakis @mateuszjasiuk @jurevans Can you verify that the last sentence `The only dependencies added to the shared crate are toml, bimap, orion, tokio, and serde. Hopefully this should not cause issues when building WASM targets.` is true?

mariari:
> Tokio with features = ["full'] does not compile. It looks like support for wasm is still in progress. Some features are working since v1.21.0 [PR](https://github.com/tokio-rs/tokio/pull/4967) sync,macros,io-util,rt,time discussion [here](https://github.com/tokio-rs/tokio/issues/1597) and [here](https://github.com/tokio-rs/tokio/pull/4716)
> 
> Rest of the dependencies are fine from what I can tell.

Thank you for linking the issue, it seems to be a non problem as the only feature that is needed to `make build` is `time`.

adrianbrink:
Can we compile this without `tokio = full`?

@mariari @cwgoes

adrianbrink:
I'm closing this in favor of #925

PR #926, opened 2022-12-19
Add `StorageKeys` derive macro

sug0:
Allows us to derive storage key helpers from a struct comprised solely of `&'static str` fields.

For instance:

```rs
#[derive(StorageKeys)]
struct Keys {
  param1: &'static str,
  param2: &'static str,
}
```

Would generate a `VALUES` field with `param1` set to `"param1"` and `param2` set to `"param2"`. It would also generate a sorted slice `ALL`, whose contents are `"param1"` and `"param2"`, which can be used to efficiently match a storage key, with a binary search over `ALL`.


macros/src/lib.rs
```diff
@@ -149,3 +151,91 @@ pub fn validity_predicate(
...
+
+            const VALUES: #struct_def_ident = Self {
+                #values_list
+            };
+        }
+    }
+    .into()
+}
+
+fn create_ponctuated<F, M>(
```
tzemanovic, line 230: just a wee typo here
```suggestion
fn create_punctuated<F, M>(
```

macros/src/lib.rs
```diff
@@ -149,3 +151,91 @@ pub fn validity_predicate(
...
+                 but got {field_type} instead"
+            );
+        }
+        idents.push(field.ident.clone().expect("Expected a named field"));
+    }
+
+    idents.sort();
+
+    let ident_list = create_ponctuated(&idents, |ident| ident.clone());
+    let values_list = create_ponctuated(&idents, |ident| {
```
tzemanovic, line 188:
```suggestion
    let ident_list = create_punctuated(&idents, |ident| ident.clone());
    let values_list = create_punctuated(&idents, |ident| {
```
(for below)

tzemanovic (approved):

batconjurer (changes requested):
You need to add unit tests. I made a PR with a happy flow unit test. But obviously many unhappy flows should be checked. Parameters with the same name, types other than &'static string, etc.

sug0:
For reference, added changes from #937

macros/src/lib.rs
```diff
@@ -149,3 +151,134 @@ pub fn validity_predicate(
...
+
+                &[ #ident_list ]
+            };
+
+            const VALUES: #struct_def_ident = Self {
+                #values_list
+            };
+        }
+    }
+    .to_string()
```
tzemanovic, line 228: Thanks for adding the test @batconjurer ! Just a wee nit, it's a shame to turn this to string just to parse it again in `derive_storage_keys`. Let's just return the TokenStream and turn it into a string in the test where it's needed

batconjurer: I'll try, but usually this level of indirection is needed. Otherwise you get yelled at for using the proc-macro API in a tests (because it produces a binary)

batconjurer: Oh, I'm dumb. TokenStream2 was designed exactly for this reason. Will fix.

macros/src/lib.rs
```diff
@@ -149,3 +151,209 @@ pub fn validity_predicate(
...
+    /// Test if the `ALL` slice generated in `StorageKeys` macro
+    /// derives is sorted in ascending order.
+    #[test]
+    fn test_storage_keys_derive_sorted_slice() {
+        let test_struct = quote! {
+            struct Keys {
+                word: &'static str,
+                is: &'static str,
+                bird: &'static str,
+                the: &'static str,
```
tzemanovic, line 262: 🐦

tzemanovic (dismissed):

batconjurer (approved):

batconjurer (approved):

macros/src/lib.rs
```diff
@@ -222,6 +223,7 @@ fn derive_storage_keys_inner(struct_def: TokenStream2) -> TokenStream2 {
                 &[ #ident_list ]
             };
 
+            #[allow(dead_code)]
```
tzemanovic, line 221: could we leave `VALUES` without `#[allow(dead_code)]`? It can help to spot if any of the keys are not being used

sug0: yeah, sure, at least `VALUES` should be used

batconjurer (dismissed):

tzemanovic (approved):

tzemanovic:
@brentstone there's some `test_tx_change_validator_commissions` case failing in https://github.com/anoma/namada/actions/runs/3750000986/jobs/6369272997#step:8:37077, it's got the seed there

brentstone:
> 

See https://github.com/anoma/namada/pull/965

tzemanovic:
> @brentstone there's some `test_tx_change_validator_commissions` case failing in https://github.com/anoma/namada/actions/runs/3750000986/jobs/6369272997#step:8:37077, it's got the seed there

fixed in https://github.com/anoma/namada/pull/965

tzemanovic (approved):

PR #936, opened 2022-12-20
Update from-docker.md

Romainua:
Base on hub.docker.com/r/heliaxde account there are not namada image


Fraccaman:
Hi @Romainua! Sorry for the misleading docs, we have docker images here: https://github.com/anoma/namada/pkgs/container/namada! We will update the docs right away :)

PR #960, opened 2022-12-23
Add JSON as an option for logging by env var

DaveWK:


apps/src/lib/logging.rs
```diff
@@ -30,13 +32,27 @@ pub fn set_subscriber(filter: EnvFilter) -> Result<()> {
...
         true
     };
-
-    let my_collector = Subscriber::builder()
-        .with_ansi(with_color)
-        .with_env_filter(filter)
-        .finish();
-    tracing::subscriber::set_global_default(my_collector)
-        .wrap_err("Failed to set log subscriber")
+    let json_format = if let Ok(val) = env::var(JSON_ENV_KEY) {
```
tzemanovic, line 35: this makes the json fmt the default

tzemanovic:
thanks @DaveWK ! I've added couple more changes to keep the "full" fmt as the default and also to allow to use "pretty" fmt in #968. Will merge from there

PR #961, opened 2022-12-23
faucet PoW

tzemanovic:
closes #959

depends on #926 for storage keys macro, #966 for avoiding lifetime issues and #962 for tx fees

first commit from this PR: [f7b6378](https://github.com/anoma/namada/pull/961/commits/f7b63784490d0b119fa63f7d52c53441bbcec031)

This is a breaking change for client - to be able to withdraw from token, client must request PoW challenge from the faucet, solve it and submit `tx_testnet_faucet_transfer.wasm` tx instead of `tx_transfer.wasm`.

This is non-breaking change for genesis as the added `faucet_pow_difficulty` is optional (defaults to `0` - no challenge).

TODOs:
- [x] integrate in vp_faucet
- [x] initialize faucet account with difficulty if used in `init_chain.rs`
- [x] set `faucet_pow_difficulty` from genesis
- [x] add a `tx_faucet_transfer` wasm to send `token::Transfer` wrapped in `faucet::Solution` and increment counter
- [x] add RPC endpoint to obtain faucet challenge for a `token::Transfer`
- [x] integrate in the client transfer command (detect faucet source)
- [x] allow to configure faucet withdrawal limit from genesis
- [ ] add the PoW into the wrapper tx (with conditional compilation when `not(mainnet)` to allow fee-less tx (with minimum tx fees from #962)

PR #962, opened 2022-12-23
Testnet fees

### LINKED ISSUE #958: Implement a simple fee system

grarco:
Closes #958 

Implements a simplified fee system to mitigate DOS

### DIFF
--- a/apps/src/lib/client/signing.rs
+++ b/apps/src/lib/client/signing.rs
@@ -6,6 +6,7 @@ use namada::proto::Tx;
 use namada::types::address::{Address, ImplicitAddress};
 use namada::types::key::*;
 use namada::types::storage::Epoch;
+use namada::types::token::Amount;
 use namada::types::transaction::{hash_tx, Fee, WrapperTx};
 
 use super::rpc;
@@ -174,7 +175,7 @@ pub async fn sign_wrapper(
     let tx = {
         WrapperTx::new(
             Fee {
-                amount: args.fee_amount,
+                amount: Amount::from(100),
                 token: ctx.get(&args.fee_token),
             },
             keypair,
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -2,7 +2,10 @@
 
 use namada::ledger::pos::types::into_tm_voting_power;
 use namada::ledger::protocol;
+use namada::ledger::storage::write_log::StorageModification;
+use namada::ledger::storage_api::StorageRead;
 use namada::types::storage::{BlockHash, BlockResults, Header};
+use namada::types::token::Amount;
 
 use super::governance::execute_governance_proposals;
 use super::*;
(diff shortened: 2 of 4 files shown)


tzemanovic (approved):

tzemanovic:
On a second thought, I think this will prevent from withdrawing from faucet, unless you already have some balance. Not sure if that's what we want

grarco:
That's correct, at the moment this PR needs two things:

1. Define a sensible amount of the fee: it's currently at 100 micro NAMs but faucet releases up to 1000 NAM per request so we should probably raise this
2. Find a way to allow withdrawal from faucet

Regarding the second point, an idea could be this: in the three checks that we perform, when trying to access the balance of an account the first time the key is not in storage and we default to 0. For the testnet only we could change this default to be a minimum amount to allow the user to run the first transaction (ideally withdrawing from faucet)

tzemanovic:
>     1. Define a sensible amount of the fee: it's currently at 100 micro NAMs but faucet releases up to 1000 NAM per request so we should probably raise this
 
With the faucet withdrawal PoW (#961), we can set `faucet_pow_difficulty` in genesis - it takes around a minute with difficulty `3` to solve. I'd say we can safely allow up to ~100 txs from that. Let me make the faucet limit configurable from genesis too, so we can control this with it and we can set it to 10_000 micro NAM.

>     2. Find a way to allow withdrawal from faucet
> 
> 
> Regarding the second point, an idea could be this: in the three checks that we perform, when trying to access the balance of an account the first time the key is not in storage and we default to 0. For the testnet only we could change this default to be a minimum amount to allow the user to run the first transaction (ideally withdrawing from faucet)

Yeah, some along this line sounds good - I think we can introduce a "mainnet" feature, which we'll need for other things and allow faucet withdrawal with PoW without fees when `not(feature = "mainnet")`

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -128,9 +131,80 @@ where
...
+                    // Charge fee
+                    let fee_payer =
+                        if wrapper.pk != address::masp_tx_key().ref_to() {
+                            wrapper.fee_payer()
+                        } else {
+                            address::masp()
+                        };
+
+                    let balance_key = token::balance_key(
+                        &self.storage.native_token,
```
tzemanovic, line 146: I think we should be using `wrapper.fee_token` here (it defaults to `NAM`)

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -575,7 +576,30 @@ where
...
+                // Check balance for fee
+                if let Ok(TxType::Wrapper(wrapper)) = process_tx(tx) {
+                    let fee_payer = if wrapper.pk != masp_tx_key().ref_to() {
+                        wrapper.fee_payer()
+                    } else {
+                        masp()
+                    };
+                    // check that the fee payer has sufficient balance
+                    let balance = self
+                        .get_balance(&self.storage.native_token, &fee_payer);
```
tzemanovic, line 589: same here

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -398,6 +400,15 @@ mod test_process_proposal {
             ..Default::default()
         });
         let keypair = crate::wallet::defaults::daewon_keypair();
+        // reduce address balance to match the 100 token fee
+        let balance_key = token::balance_key(
+            &shell.storage.native_token,
```
tzemanovic, line 405: and here

apps/src/lib/client/signing.rs
```diff
@@ -174,7 +175,7 @@ pub async fn sign_wrapper(
     let tx = {
         WrapperTx::new(
             Fee {
-                amount: args.fee_amount,
+                amount: Amount::from(100),
```
tzemanovic, line 178: let's add a `const MIN_FEE` for this somewhere in core and re-use it in other places

PR #964, opened 2022-12-26
added comments

bengtlofgren:
Started with some comments

### DIFF
--- a/core/src/ledger/faucet.rs
+++ b/core/src/ledger/faucet.rs
@@ -43,7 +43,12 @@ pub struct Challenge {
     /// PoW difficulty
     pub difficulty: u8,
     /// The counter value of the `transfer.target`
+    /// TODO: What does this mean though?
     pub counter: Counter,
+
+    // TODO: 
+    // What about the public key? I.e the initial thing that is hashed? 
+    // I'm assuming we don't hash the transaction data
 }
 
 /// One must find a value for this type to solve a [`Challenge`] that is at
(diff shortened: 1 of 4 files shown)


core/src/ledger/faucet.rs
```diff
@@ -43,7 +43,12 @@ pub struct Challenge {
     /// PoW difficulty
     pub difficulty: u8,
     /// The counter value of the `transfer.target`
+    /// TODO: What does this mean though?
```
tzemanovic, line 46: The current counter value is read from storage for the given `transfer.target`. The VP then checks that this matches what's in pre-storage and incremented in post-storage

core/src/ledger/faucet.rs
```diff
@@ -43,7 +43,12 @@ pub struct Challenge {
...
     pub counter: Counter,
+
+    // TODO: 
+    // What about the public key? I.e the initial thing that is hashed? 
+    // I'm assuming we don't hash the transaction data
```
tzemanovic, line 51: we're hashing `Challenge` that includes the transfer with the target address

core/src/ledger/faucet.rs
```diff
@@ -82,6 +87,7 @@ impl Challenge {
     }
 
     /// Try to find a solution to the [`Challenge`].
+    /// TODO: This loop can continue on forever, should probably put some limit
```
tzemanovic, line 90: it keeps printing while it's looping, so it shouldn't appear stuck

core/src/ledger/faucet.rs
```diff
@@ -103,6 +109,7 @@ impl Challenge {
         {
             *old = *new;
         }
+        // This is the nonce that is hashed with the PK
```
tzemanovic, line 112: it's hashed with `challenge_bytes`

PR #965, opened 2022-12-26
Fix validator commission change wasm test

brentstone:
Resolving the wasm test error seen in https://github.com/anoma/namada/pull/926#issuecomment-1363805447. This was due to an arbitrary value for a new commission rate that was the same as the previous rate, which threw an error from within `PosActions::change_validator_commission_rate`.

The arb values for a new rate within `tx_change_validator_commission.rs` are now prevented from being the same as the previous rate.

Additionally, the previous `ChangeIsZero` error is removed, and if a tx to change the commission rate to the same as the previous one is submitted, the code instead returns early rather than attempt to make storage changes. The last commit can be removed if we prefer to keep the `ChangeIsZero` error.

### DIFF
--- a/proof_of_stake/src/lib.rs
+++ b/proof_of_stake/src/lib.rs
@@ -563,15 +563,14 @@ pub trait PosActions: PosReadOnly {
                 }
             };
         let params = self.read_pos_params()?;
-
         let rate_at_pipeline = *commission_rates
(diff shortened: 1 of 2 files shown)


tzemanovic (dismissed):
LGTM, thanks! I'll just re-target this to main

tzemanovic (approved):

Issue #959, opened 2022-12-23
Implement PoW based faucet


adrianbrink:
- generate public key
- brute-force a nonce
- hash nonce and public key until you have something with a couple of leading zeros
- faucet tx takes pk + nonce - faucet vp verifies that pow is valid
- faucet vp stores a mapping from pk to [nonces] - enforces that every pk can only ever submit a PoW puzzle once

PR #966, opened 2022-12-27
core/storage_api: use GATs to hide lifetime in StorageRead trait

tzemanovic:
fixes #828

PR #969, opened 2022-12-30
testnet PoW

tzemanovic:
This is alternative approach to #961, re-use most of the logic from `faucet_pow` module, but instead of adding a special tx for faucet withdrawal (the inner tx) to be executed and checked as a regular tx, we add an option PoW field to the `WrapperTx` (`#[cfg(not(feature = "mainnet"))] pow_solution: Option<crate::ledger::testnet_pow::Solution>`). This is needed because the fee needs to be checked in the mempool at which point we don't have access to the inner tx.

- added `"mainnet"` feature (with `make check-mainnet`, `make clippy-mainnet` and `make test-unit-mainnet`) - for the "mainnet" build, the `WrapperTx` doesn't contain the `pow_solution` field)
- added a check to client/signing `sign_wrapper` that the address derived from wrapper tx PK has enough balance to pay the minimum tx fee (100 namnam)
- only in `cfg!(not(feature = "mainnet"))`:
  - added `faucet_pow_difficulty` and `faucet_withdrawal_limit` fields to genesis (optional, defaults 0 PoW difficulty if not set)
  - when a tx is withdrawal from faucet, the client automatically requests PoW challenge, attempts to find a solution and attach it to the wrapper tx
  - when an address derived from wrapper tx PK doesn't have enough balance to pay for the tx fee (100 namnam), it will also automatically request PoW challenge, attempts to find a solution and attach it to the wrapper tx

closes https://github.com/anoma/namada/issues/959

depends on https://github.com/anoma/namada/pull/926 for storage keys macro, https://github.com/anoma/namada/pull/966 for avoiding lifetime issues and https://github.com/anoma/namada/pull/962 for tx fees

first commit from this PR: f7b63784490d0b119fa63f7d52c53441bbcec031

This is a breaking change for client - to be able to withdraw tokens from faucet without having enough balance in the signing key to cover the minimum tx fee (set in #962 to 100 namnam), client must request PoW challenge from the faucet, solve it and attach it to the wrapper tx (@memasdeligeorgakis).

This is non-breaking change for genesis as the added faucet_pow_difficulty is optional (defaults to 0 - no challenge), and `faucet_withdrawal_limit`.


tzemanovic:
pls update wasm

tzemanovic:
pls update wasm

tzemanovic:
pls spawn devnet [anoma-devnet-faucet-and-fees,3,heliaxdev@1be9883,ON]

tzemanovic:
pls spawn devnet [anoma-devnet-faucet-and-fees,3,heliaxdev@1be9883,ON]

PR #974, opened 2023-01-03
Reduces block size

grarco:
Reduces temporary block size from 16 to 5 MiB


adrianbrink (approved):

grarco:
We should prefer #976 over this one

PR #975, opened 2023-01-03
Add `max_proposal_bytes` ledger param

sug0:
Allows us to configure the ledger with a `max_proposal_bytes` param, that is used to limit the maximum size of a batch of transactions proposed by some Tendermint consensus round leader.


sug0:
CI might fail in `apps/src/lib/node/ledger/tendermint_node.rs` if it's checking ABCI++. it's fixed in #976

batconjurer (approved):

PR #978, opened 2023-01-04
Including the PoW faucet and fees into docs

bengtlofgren:


documentation/docs/src/testnets/pow.md
```diff
@@ -0,0 +1,20 @@
+# The PoW Solution on Namada
+
+In order to combat the "costlessness of blockspace" problem with a testnet, and hence prevent DOS attacks as a result, when a Namada account does not have the required fees to pay for a transaction, the user must complete a Proof of Work challenge. The difficulty of this challenge is a parameter set by governance, and will dictate the (average) computational expenditure needed in order to complete the challenge.
```
adrianbrink, line 3: This PoW challenge only applies to the faucet. All other transactions require fees as usual.

adrianbrink (changes requested):
Only PoW challenge for faucet txs not every tx.

documentation/docs/src/testnets/pow.md
Fraccaman: if Im not mistake, the current implementation requires either to pay for fees or to attach a valid PoW, not only for the faucet. This is because there are problems with the wrapper tx (if im not mistake). Can you confirm @tzemanovic ?

tzemanovic: that's right - you can either pay the fee or attach PoW to any tx type and only for faucet withdrawal it is always required - the client handles this

adrianbrink: can we change this though. 

only the faucet is a special case. normal transactions should accept a pow proof.

adrianbrink: Couldn't we do exactly the same thing that we do for the MASP?

Treat it as a special tx where there is no fee payer for the wrapper tx and then have the fee be conducted afterwards in the form of a PoW proof?

tzemanovic: I think the use of "sentinel key" currently on masp txs is temp solution - for the fee's we're checking the PK of the wrapper, I don't think we can do the same for faucet unless we also add some faucet "sentinel" key, but what is to gain anyway if this is testnet-only anyway?

Fraccaman:
@adrianbrink can we merge this?

bengtlofgren:
Yes please

adrianbrink (approved):

PR #981, opened 2023-01-04
Liveness protection for testnet

brentstone:
Currently based on v0.12.2

In the absence of rewards in the current testnet, we would like to ensure liveness by providing a mechanism for fully unbonding validators who go offline for extended periods of time.


cwgoes:
Is this still relevant?

brentstone:
> Is this still relevant?

I'm not sure actually, do you think we will still want some feature like this? Even for mainnet?

cwgoes:
I don't think we need to unbond them, but perhaps if a validator doesn't vote in some number of blocks we should jail them (so they aren't voting)?

brentstone:
> I don't think we need to unbond them, but perhaps if a validator doesn't vote in some number of blocks we should jail them (so they aren't voting)?

I'm in favor of this, so let's keep this open. I'd like to come back to it once we're good with slashing, redelegation, etc if that is fine.

adrianbrink:
> > I don't think we need to unbond them, but perhaps if a validator doesn't vote in some number of blocks we should jail them (so they aren't voting)?
> 
> I'm in favor of this, so let's keep this open. I'd like to come back to it once we're good with slashing, redelegation, etc if that is fine.

I'll close this for now. This still needs to be done and here is the corresponding issue: https://github.com/anoma/namada/issues/1360

PR #995, opened 2023-01-10
Fix `cargo doc` and the makefile

sug0:
Minor PR that renames `ANOMA_DEV` to `NAMADA_DEV` in the makefile, and fixes the build of rust docs.

### DIFF
--- a/apps/src/lib/node/ledger/shell/prepare_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/prepare_proposal.rs
@@ -26,6 +26,8 @@ use super::block_space_alloc::{AllocFailure, BlockSpaceAllocator};
 #[cfg(feature = "abcipp")]
 use crate::facade::tendermint_proto::abci::ExtendedCommitInfo;
 use crate::facade::tendermint_proto::abci::RequestPrepareProposal;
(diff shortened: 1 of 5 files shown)


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -26,6 +26,8 @@ use super::block_space_alloc::{AllocFailure, BlockSpaceAllocator};
 #[cfg(feature = "abcipp")]
 use crate::facade::tendermint_proto::abci::ExtendedCommitInfo;
 use crate::facade::tendermint_proto::abci::RequestPrepareProposal;
+#[allow(unused_imports)]
+use crate::node::ledger::shell::block_space_alloc;
```
james-chf, line 30: Is it needed in this file?

james-chf (approved):

batconjurer (approved):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
sug0: the docs reference `block_space_alloc`

PR #996, opened 2023-01-10
Write proofs to storage along with their respective validator set updates

sug0:
The complete proof of a validator set update (comprised of the signatures reflecting >2/3 voting power) is written to storage, along with the validator set update itself. The relayer will query this proof from storage, which it will use to call the Ethereum smart contracts with.


sug0:
seems to work fine, judging by the ledger's debug logs, so I'm opening this up for review

ethereum_bridge/src/protocol/transactions/validator_set_update/mod.rs
```diff
@@ -141,3 +152,86 @@ where
 
     Ok(changed)
 }
+
+#[cfg(test)]
+mod test_valset_upd_state_changes {
```
batconjurer, line 157: I know its more difficult, but I would love a test when there are votes but not a quorum.

batconjurer (approved):

batconjurer (approved):
LGTM!

ethereum_bridge/src/test_utils.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// Set up a [`TestStorage`] initialized at genesis with the given
+/// validators.
+pub fn setup_storage_with_validators(
+    active_validators: HashMap<Address, token::Amount>,
+) -> (TestStorage, HashMap<Address, TestValidatorKeys>) {
+    // set last height to a reasonable value;
+    // it should allow vote extensions to be cast
+    let mut storage = TestStorage {
+        last_height: 3.into(),
```
james-chf, line 54: nit: this `3` could be a const somewhere with the something similar to the comment above as a docstring - what's the minimum block height for which vote extensions could be cast during the consensus round? (2 or 3?)

ethereum_bridge/src/storage/mod.rs
```diff
@@ -1,4 +1,6 @@
 //! Functionality for accessing the storage subspace
-pub use namada_core::ledger::eth_bridge::storage::bridge_pool;
+pub mod proof;
 pub mod vote_tallies;
-pub use namada_core::ledger::eth_bridge::storage::{wrapped_erc20s, *};
+pub use namada_core::ledger::eth_bridge::storage::{
+    bridge_pool, wrapped_erc20s, *,
+};
```
james-chf, line 6:
```suggestion
// TODO: the old location should be deprecated and eventually the modules below should be moved here
pub use namada_core::ledger::eth_bridge::storage::*;
```
nit: we don't need the specific module names if we are using `*`, we should use one or the other?
we should probably have a TODO/issue to eventually move those modules (once all imports using the old path have been updated)

ethereum_bridge/src/storage/proof.rs
```diff
@@ -0,0 +1,82 @@
...
+            signatures: BTreeMap::new(),
+        }
+    }
+
+    /// Add a new signature to this [`EthereumProof`].
+    pub fn attach_signature(
+        &mut self,
+        addr: Address,
+        height: BlockHeight,
+        signature: common::Signature,
```
james-chf, line 38: We should accept `secp256k1::Signature` directly instead of `common::Signature` since the other variant of `common::Signature` isn't valid for this API, what do you think? Would prevent accidental misuse

ethereum_bridge/src/storage/proof.rs
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+        if let common::Signature::Secp256k1(sig) = signature {
+            self.signatures.insert((addr, height), sig);
+        }
+    }
+
+    /// Add a new batch of signatures to this [`EthereumProof`].
+    pub fn attach_signature_batch<I>(&mut self, batch: I)
+    where
+        I: IntoIterator<Item = ((Address, BlockHeight), common::Signature)>,
```
james-chf, line 48: similarly here - think we should be using `secp256k1::Signature` instead of `common::Signature`

james-chf (changes requested):

ethereum_bridge/src/storage/proof.rs
sug0: there's a pattern match on the signature. only secp sigs are stored internally

sug0: it's mostly a convenience for the caller. otherwise the pattern matching would be the caller's responsibility

ethereum_bridge/src/test_utils.rs
sug0: 3 was chosen somewhat arbitrarily. you could have already sent a vote extension at `last_height = 1`, even

james-chf: ah right, I thought there might have been something specific about the `last_height` value being `3` (or rather, not being `1`)

james-chf (approved):

ethereum_bridge/src/storage/mod.rs
sug0: we shouldn't even need the star import here. I might make a PR some time later to get rid of it

PR #1001, opened 2023-01-11
Remove extra dependency installation instructions

### LINKED ISSUE #841: Update Tendermint version in the project README
The project README file instructs users to install Tendermint 0.34.x, which is no longer used. We should point users to our fork of Tendermint instead.

adrianbrink:
Fixes #841


cwgoes:
I'm not sure we should remove these instructions entirely. Even if we install from our Tendermint fork automatically, folks might occasionally want to use modified Tendermint versions and we should probably use the binary if it is there already (perhaps with a warning for incorrect version).

PR #1002, opened 2023-01-11
eth bridge clippy-abcipp

juped:


batconjurer (approved):
I mean, this is all fine, but are there actually any abci++ feature flags in vm env?

PR #1012, opened 2023-01-11
eth-bridge-integration: sync with 0.13.0

juped:
The thing to look out for here is the faucet POW integration, which touches everything.

Incorporates #1002 for some easy conflict resolution, even though that's properly its own change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4269,7 +4269,7 @@ dependencies = [
 
 [[package]]
 name = "namada_ethereum_bridge"
-version = "0.11.0"
+version = "0.13.0"
 dependencies = [
  "assert_matches",
  "borsh",
--- a/ethereum_bridge/Cargo.toml
+++ b/ethereum_bridge/Cargo.toml
@@ -4,7 +4,7 @@ edition = "2021"
 license = "GPL-3.0"
 name = "namada_ethereum_bridge"
 resolver = "2"
-version = "0.11.0"
+version = "0.13.0"
 
 [features]
 default = ["abciplus"]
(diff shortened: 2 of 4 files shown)


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -130,10 +130,13 @@ where
...
             .iter()
-            .map(|tx| {
-                Tx::from(match tx.decrypt(privkey) {
-                    Ok(tx) => DecryptedTx::Decrypted(tx),
-                    _ => DecryptedTx::Undecryptable(tx.clone()),
+            .map(|queued_tx| {
+                Tx::from(match queued_tx.tx.decrypt(privkey) {
+                    Ok(decrypted_tx) => DecryptedTx::Decrypted {
+                        tx: decrypted_tx,
+                        has_valid_pow: queued_tx.has_valid_pow,
```
batconjurer, line 137: Doesn't this need a feature flag over the `has_valid_pow` field?

shared/src/ledger/queries/router.rs
```diff
@@ -250,7 +250,7 @@ macro_rules! try_match_segments {
         $end = $request.path.len();
         match $request.path[$start..$end].parse::<$arg_ty>() {
             Ok(parsed) => {
-                println!("Parsed {}", parsed);
+                // println!("Parsed {}", parsed);
```
batconjurer, line 253: What happened here?

batconjurer (changes requested):

shared/src/ledger/queries/router.rs
batconjurer: caveman debugging :hammer_and_pick:

batconjurer:
> is this still relevant?

No, we merge it a later version from main already.

adrianbrink:
> > is this still relevant?
> 
> No, we merge it a later version from main already.

Closing it then

PR #1013, opened 2023-01-11
Add multitoken spec

yito88:


cwgoes (changes requested):
Thanks! Can you add:
- The conditions under which state changes (e.g. transfers) are allowed, both basic transfers and this sub_prefix special case
- A table of all the storage keys & values

cwgoes (approved):
let's merge this

PR #1014, opened 2023-01-12
The FAQ and a bit more

bengtlofgren:


awasunyin (commented):
documentation/docs/src/SUMMARY.md

FAQ should be at the same level as Troubleshooting? and not a separate section?

awasunyin (dismissed):

Issue #1009, opened 2023-01-11
Replay protection


grarco:
Replay protection implementation following the [specs](https://specs.namada.net/base-ledger/replay-protection.html)

# TODO

- [ ] Hash-based replay protection
   - internal address
   - internal VP
   - `mempool_validate` check
   - `process_proposal` check
   - `finalize_block` removal of inner hash in case of out-of-gas
   - client hash check?
- [ ] Introduce `ChainId` field in tx
    - `mempool_validate` check
    - `process_proposal` check
    - client update
- [ ] Introduce `expiration` field in tx
    - `mempool_validate` check
    - `process_proposal` check
    - client update

PR #1017, opened 2023-01-12
Replay protection

grarco:
Addresses #1009 

Implements hash-based replay protection

- Updates specs
- Adds an internal address for replay protection
- Adds a native VP to prevent any writing to the hash subspace
- Change `WrapperTx‘ commit to be computed on the hash of the **unsigned** inner tx 
- Adds `mempool_validate` checks
- Adds `process_proposal` checks and rejects block on most errors
- Updates `finalize_block` to remove the hash of the tx when it fails for out-of-gas (to allow rewrapping)
- Unit tests


grarco:
pls update wasm

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -319,6 +325,23 @@ where
                         msg
                     );
                     stats.increment_errored_txs();
+                    // FIXME: unit test
```
tzemanovic, line 328: is this for the last part of `test_remove_tx_hash` that's commented out?

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -319,6 +325,23 @@ where
...
+
+                    // If transaction type is Decrypted and failed because of
+                    // out of gas, remove its hash from storage to allow
+                    // repwrapping it
```
tzemanovic, line 332:
```suggestion
                    // rewrapping it
```

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -42,11 +50,20 @@ where
     }
 
     /// Check all the given txs.
-    pub fn process_txs(&self, txs: &[Vec<u8>]) -> Vec<TxResult> {
-        let mut tx_queue_iter = self.storage.tx_queue.iter();
+    pub fn process_txs(&mut self, txs: &[Vec<u8>]) -> Vec<TxResult> {
```
tzemanovic, line 53: Unfortunately, we cannot make `self` mutable here because non-validator nodes are currently not required to run prepare/process proposal requests in TM and also we don't know if it's from an accepted proposal yet, so we should only mutate state once we're finalizing it

core/src/types/transaction/mod.rs
```diff
@@ -39,6 +40,18 @@ pub fn hash_tx(tx_bytes: &[u8]) -> Hash {
     Hash(*digest.as_ref())
 }
 
+/// Get the hash of the unsigned transaction (if signed), otherwise the hash of
+/// entire tx.
+pub fn unsigned_hash_tx(tx_bytes: &[u8]) -> Hash {
+    match SignedTxData::try_from_slice(tx_bytes) {
```
tzemanovic, line 46: I don't know if this quite right - I think for both wrappers and inner txs, the `tx_bytes` is encoded `Tx`, which has the `SignedTxData` in the `data` field. To hash unsigned tx, you can try to decode `SignedTxData` from the `tx.data` and if it's valid, replace the `tx.data` with the `SignedTxData.data` (this is the unsigned data) field to reconstruct unsigned tx.

apps/src/lib/node/ledger/shell/process_proposal.rs
grarco: I'll add a local cache in here and run the checks again in finalize block 👍

grarco:
pls update wasm

grarco:
pls update wasm

PR #1025, opened 2023-01-16
Shielded Inflation rewards

mariari:
A followup pr of #714 that adds the inflation rewards for the shielded pool.

The formulas follow from:

https://specs.namada.net/economics/inflation-system.html

### DIFF
--- a/core/src/ledger/storage/masp_conversions.rs
+++ b/core/src/ledger/storage/masp_conversions.rs
@@ -16,7 +16,7 @@ use crate::types::address::Address;
 use crate::types::dec::Dec;
 use crate::types::storage::{Epoch, Key};
 use crate::types::token::MaspDenom;
-use crate::types::uint::I256;
+use crate::types::uint::{I256, Uint};
 use crate::types::{address, token};
 
 /// A representation of the conversion state
@@ -39,6 +39,7 @@ pub struct ConversionState {
     >,
 }
 
+#[cfg(feature = "wasm-runtime")]
 fn calculate_masp_rewards<D, H>(
     wl_storage: &mut super::WlStorage<D, H>,
     addr: &Address,
@@ -74,7 +75,6 @@ where
     let denomination_base =
         read_denom(wl_storage, &wl_storage.get_native_token().unwrap(), None).unwrap().unwrap();
 
-    // TODO replace 6 with native token call
     let denomination_offset = 10u64.pow((denomination.0 - denomination_base.0) as u32);
     let conversion = |amt| amt /  denomination_offset;
     let total_tokens = conversion(total_tokens) ;
@@ -195,6 +195,8 @@ where
 
     println!("----------------------------");
     println!("TIME TO DEBUG");
+    println!("noterized_inflation: {:?}", noterized_inflation);
+    println!("denomination_offset: {}", denomination_offset);
     println!("TIME TO DEBUG");
     println!("----------------------------");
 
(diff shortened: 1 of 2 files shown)


core/src/ledger/storage/mod.rs
```diff
@@ -1282,6 +1285,54 @@ pub mod testing {
...
+
+                self.write(
+                    &token::total_supply_key(addr),
+                    total_supply.try_to_vec().unwrap(),
+                )
+                .expect("Should not fail to put a test total supply");
+                self.write(
+                    &token::balance_key(addr, &masp_addr),
+                    total_supply_in_masp.try_to_vec().unwrap(),
+                )
```
mariari, line 1328: I'm not sure how this gets initialized on chain from the genesis config @brentstone @murisi would either of you know?

mariari:
It seems that currently the `e2e` tests fail, I have to investigate why still

mariari:
Currently `19` e2e tests fail:

```lisp
ifailures:
    e2e::eth_bridge_tests::everything
    e2e::ibc_tests::run_ledger_ibc
    e2e::ledger_tests::double_signing_gets_slashed ; timeout
    e2e::ledger_tests::implicit_account_reveal_pk
    e2e::ledger_tests::invalid_transactions        ; timeout
    e2e::ledger_tests::ledger_many_txs_in_a_block
    e2e::ledger_tests::ledger_txs_and_queries

    ;; Must be updated to the new rewards
    ;; "Needle: NAM: 0.2", in the temp file "NAM: 467977.4\r\n"
    e2e::ledger_tests::masp_incentives

    e2e::ledger_tests::masp_pinned_txs       ; passes on rerun (timeout?)
    e2e::ledger_tests::masp_txs_and_queries  ; passes on rerun (timeout?)
    e2e::ledger_tests::pos_bonds
    e2e::ledger_tests::pos_init_validator
    e2e::ledger_tests::pos_rewards
    e2e::ledger_tests::proposal_offline

    -- trying to connect: tcp connect error: OS error 111
    -- Figure this might be due to a timeout somewhere
    e2e::ledger_tests::proposal_submission

    e2e::ledger_tests::run_ledger_load_state_and_reset
    e2e::ledger_tests::test_genesis_validators      -- timeout
    e2e::ledger_tests::test_node_connectivity_and_consensus
    e2e::wallet_tests::wallet_encrypted_key_cmds

```

Most of them are due to timeout, rerunning two of them locally has made it pass (likely again due to timeout). With two interesting `e2e` tests failing

the first is `masp_incentives` which should be updated to the new formula, however the contrast between the expected value and what was inflated seems a bit much. It goes from `0.2 NAM` to `467977.4 NAM` if I'm reading the tmp files correctly.

The other one is `e2e::ledger_tests::proposal_submission` which gives a tcp connection error, which may be due to timeouts on epoch, however that is unconfirmed.

Overall, I think the pr should be fit to merge outside of fine tuning `masp_incentives`, with timing issues handled in further prs

mariari:
pls update wasm

murisi (changes requested):
Thanks for the changes! Overall the code looks good. I'm mainly concerned about two and a half things:
* Is the inflation that is minted exactly equal to what is stored in `token::last_inflation(addr)`? If it is, then maybe we can make this equivalence clearer in the code... If not, we need to fix the discrepancy lest it subtly affects the PID controller.
* The conversion tree code was written under the assumption that no rewards are given for shielded NAM (and hence that no compounding takes place). This was my mistake. Since your code can potentially give rewards for locked NAM (meaning that compounding necessarily happens), the conversion tree code needs to be updated.
* Now that compounding of rewards can take place, we need to add an E2E test for this compounding behavior.

mariari:
rerolled it on brent/inflation-reward that was properly merged

mariari:
Last commit before merging v0.16.0. Note that the e2e incetives fails, but everything else works

mariari:
pls update wasms

genesis/dev.toml
```diff
@@ -178,10 +218,10 @@ max_inflation_rate = "0.1"
 target_staked_ratio = "0.6667"
 # Portion of a validator's stake that should be slashed on a duplicate
 # vote.
-duplicate_vote_min_slash_rate = "0.001"
+duplicate_vote_min_slash_rate = "0.1"
 # Portion of a validator's stake that should be slashed on a light
 # client attack.
-light_client_attack_min_slash_rate = "0.001"
+light_client_attack_min_slash_rate = "0.1"
```
bengtlofgren, line 224: Why has this been changed in this PR? Not sure I follow?

core/src/types/token/parameters.rs
```diff
@@ -0,0 +1,129 @@
...
+pub const KP_SP_GAIN_KEY: &str = "proptional_gain";
+
+/// The key for the nominal derivative gain of a shielded pool for a given asset
+pub const KD_SP_GAIN_KEY: &str = "derivative_gain";
+
+/// The key for the locked ratio target for a given asset
+pub const LOCKED_RATIO_TARGET_KEY: &str = "locked_ratio_target";
+
+/// The key for the max reward rate for a given asset
+pub const MAX_REWARD_RATE: &str = "max_reward_rate";
```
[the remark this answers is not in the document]
bengtlofgren: I agree with @murisi here

```suggestion
pub const MAX_REWARD_RATE_KEY: &str = "max_reward_rate";
```

PR #1027, opened 2023-01-16
Doc strings for Issue 5 of Eth Bridge Audit

batconjurer:
Added some small clarifying doc strings


ethereum_bridge/src/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -33,7 +33,7 @@ impl utils::GetVoters for HashSet<EthMsgUpdate> {
 
 /// Applies derived state changes to storage, based on Ethereum `events` which
 /// were newly seen by some active validator(s) in the last epoch. For `events`
```
sug0, line 35: last epoch is slightly confusing here. should it be "current epoch"?

PR #1032, opened 2023-01-16
End-to-end test for configuring the Ethereum oracle

james-chf:
Closes https://github.com/anoma/namada/issues/686

This PR adds an end-to-end test checking that the Ethereum event oracle uses the configured values from blockchain storage (implementation already merged in https://github.com/anoma/namada/pull/810). To do this, an oracle-related `Config` struct had to be moved from `namada_apps` to `namada_ethereum_bridge`, to be accessible in `namada_tests`.


core/src/lib.rs
```diff
@@ -12,7 +12,7 @@ pub mod ledger;
 pub mod proto;
 pub mod types;
 
-#[cfg(feature = "abciplus")]
+#[cfg(not(feature = "abcipp"))]
```
james-chf, line 15: Had to change this otherwise `make check-abcipp` stopped working

tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -237,3 +240,69 @@ fn test_add_to_bridge_pool() {
...
+            genesis
+        },
+        None,
+    )?;
+
+    // start the ledger with the real oracle and wait for a block to be
+    // committed
+
+    // TODO(namada#1061): need to start up a fake Ethereum node here for the
+    // oracle to connect to, to avoid errors in the ledger logs
```
james-chf, line 281: At the moment, the `namadan` logs gets spammed with error messages about the oracle not being able to connect to an Ethereum RPC endpoint, for this test it doesn't matter as we are just testing the oracle is initially configured correctly, but when the linked issue [namada#1061](https://github.com/anoma/namada/issues/1061) is done we can change this here also)

core/src/lib.rs
sug0: for the sake of consistency, I would even rename every feature flag to `#[cfg(not(feature = "abcipp"))]`. obviously that's out of the scope of this PR, though

sug0 (approved):
lgtm

apps/Cargo.toml
```diff
@@ -74,6 +74,7 @@ abciplus = [
 
 [dependencies]
 namada = {path = "../shared", default-features = false, features = ["wasm-runtime", "ferveo-tpke"]}
+namada_ethereum_bridge = {path = "../ethereum_bridge"}
```
batconjurer, line 77: The pattern here is to re-export needed deps inside of the shared sub-crate. You'll notice that PoS does the same.

batconjurer (changes requested):

apps/Cargo.toml
james-chf: oic, have updated to use `namada::eth_bridge` instead

james-chf: updated in `tests` crate also now

batconjurer (approved):

PR #1036, opened 2023-01-17
Changing the inflation system documentation to be a bit more clear

mariari:
This pr is a minor change on the document file:

https://specs.namada.net/economics/inflation-system.html

I hope it makes the intent a bit more clear.

### DIFF
--- a/documentation/specs/src/economics/inflation-system.md
+++ b/documentation/specs/src/economics/inflation-system.md
@@ -35,17 +35,17 @@ Second, we take as input the following state values:
 
 - $S_{NAM}$ is the current supply of NAM
 - $L_{PoS}$ is the current amount of NAM locked in proof-of-stake
-- $I_{PoS}$ is the current proof-of-stake inflation amount, in units of tokens per epoch
+- $I_{PoS-last}$ is the proof-of-stake inflation amount from the previous epoch, in units of tokens per epoch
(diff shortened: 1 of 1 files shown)


mariari:
Should I also make a comment about what if `L_{SP_A}` is also 0. If no tokens are in the shielded pool, then the `I_{SP_A}` should also be 0, as there is nothing to give out, meaning the last_inflation should be 0. The same should be the same for staking

cwgoes:
> Should I also make a comment about what if `L_{SP_A}` is also 0. If no tokens are in the shielded pool, then the `I_{SP_A}` should also be 0, as there is nothing to give out, meaning the last_inflation should be 0. The same should be the same for staking

Hmm, is this necessary? If there are no tokens in the shielded pool nothing can be claimed anyways. As the PD controller generally runs the reward rate might be high in that case, the rewards will just be unclaimed.

mariari:
> > Should I also make a comment about what if `L_{SP_A}` is also 0. If no tokens are in the shielded pool, then the `I_{SP_A}` should also be 0, as there is nothing to give out, meaning the last_inflation should be 0. The same should be the same for staking
> 
> Hmm, is this necessary? If there are no tokens in the shielded pool nothing can be claimed anyways. As the PD controller generally runs the reward rate might be high in that case, the rewards will just be unclaimed.

I guess that is fine then, we can leave that part as is

brentstone (approved):
LGTM

PR #1038, opened 2023-01-17
Swap block height nonces with epoch values in validator set updates

sug0:
This change is required to be able to deterministically derive the next nonce we pass to a smart contract call in the Ethereum bridge.


sug0:
just some broken ABCI++ stuff in CI, which I'm fixing now.

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -274,34 +274,40 @@ where
...
+            // send one at the current block height.
+            // remember, we need to gather a quorum of
+            // votes, so this check is quite important!
+            //
+            // can send = true -> verify = false
+            // can send = false -> verify = true
+            //
+            // (we simply invert the can send logic)
+            return !self.storage
+                .can_send_validator_set_update(SendValsetUpd::Now);
```
sug0, line 289: todo: perhaps add some logging here

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
```diff
@@ -50,56 +55,32 @@ where
...
         ext: validator_set_update::SignedVext,
-        last_height: BlockHeight,
+        signing_epoch: Epoch,
     ) -> std::result::Result<
         (token::Amount, validator_set_update::SignedVext),
         VoteExtensionError,
     > {
-        #[cfg(feature = "abcipp")]
-        if ext.data.block_height != last_height {
+        if self.storage.last_height.0 == 0 {
```
james-chf, line 63: Shouldn't we be checking the `epoch` inside the vext rather than our local node's `last_height`? Something like `if ext.data.epoch == Epoch(0)`

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -765,7 +765,12 @@ where
                         if let Err(err) = self
                             .validate_valset_upd_vext_and_get_it_back(
                                 ext,
-                                self.storage.last_height,
+                                // n.b. only accept validator set updates
+                                // issued at
+                                // the current epoch (signing off on the
+                                // validators
+                                // of the next epoch)
+                                self.storage.get_current_epoch().0,
```
james-chf, line 773: `self.storage.get_current_epoch()` returns different values depending on the phase of ABCI++ we are at (i.e. if we are at `FinalizeBlock` or `Commit`, it will be the epoch of the block that's not yet committed, but outside of those phases, it's the epoch of the most recently committed block) - would that matter here? 

More generally, we need a better API around getting "current" block heights I think (maybe a state machine for tracking which phase of ABCI++ we are in, and implementing methods differently for certain states? e.g. `Storage<FinalizeBlock|Commit>.get_current_epoch()` and `Storage<OtherPhases>.get_current_epoch()`

james-chf (commented):
Some comments, I haven't reviewed anything guarded by abcipp

core/src/types/vote_extensions/validator_set_update.rs
```diff
@@ -31,11 +32,8 @@ pub type VextDigest = ValidatorSetUpdateVextDigest;
     Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, BorshSchema,
 )]
 pub struct ValidatorSetUpdateVextDigest {
```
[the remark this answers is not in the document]
sug0: no need to, digests should all be processed within the epoch they were crafted. it's also possible to key them by epoch, through the validator set update storage key

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
sug0: it's fine for the epoch to be 0, what isn't fine is an extension that was signed at height 0. height 1 should be the first block height being decided

sug0: if `last_height` has value 0, we are still deciding the first block, which couldn't have possibly had these protocol txs yet, because we only broadcast them at the commit phase of the first height

apps/src/lib/node/ledger/shell/mod.rs
sug0: a type level state machine wouldn't work, but an enum with the current phase is possible

core/src/types/vote_extensions/validator_set_update.rs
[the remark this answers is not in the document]
james-chf: oic, that makes sense - so the `Epoch` is implicitly the epoch of the block containing the `ValidatorSetUpdateVextDigest`

apps/src/lib/node/ledger/shell/mod.rs
james-chf: If we can, let's use `storage.last_height` (i.e. most recently committed block height) or something derived from it, as that value doesn't depend on which phase of ABCI we're in

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -275,34 +274,49 @@ where
...
+            // send one at the current block height.
+            // remember, we need to gather a quorum of
+            // votes, so this check is quite important!
+            //
+            // can send = true -> verify = false
+            // can send = false -> verify = true
+            //
+            // (we simply invert the can send logic)
+            let verify_passes = !self.storage
+                .can_send_validator_set_update(SendValsetUpd::Now);
```
batconjurer, line 289: this method name is super confusing. I appreciate the comment, but maybe a better name. Like must_send or smth

batconjurer (approved):

james-chf (commented):
There is something I'm still not sure about (that looks to have been the case even before this PR) - we implicitly depend on our local node's state in some places when validating/processing valset update vote extensions (or vote extension digests). i.e. our local node's current block height (or epoch).I'm thinking about edge cases relating to where our local node is not fully up to sync, but is being gossiped vote extension transactions from other nodes which are further ahead, so we may end up validating against our local "incorrect" epoch or block height.

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
```diff
@@ -50,56 +55,32 @@ where
...
             tracing::error!(
-                ext_height = ?ext.data.block_height,
-                ?last_height,
-                "Validator set update vote extension issued for a block \
-                 height different from the expected last height.",
+                "Dropping validator set update vote extension issued at \
+                 genesis"
```
james-chf, line 66:
```suggestion
                "Cannot validate validator set update vote extension as have \
                 not yet synced any blocks"
```
or some sort of similar message that indicates we can't validate as we haven't synced any blocks yet (rather than it being anything to do with the valset upd vext itself)

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
```diff
@@ -176,10 +157,13 @@ where
             VoteExtensionError,
         >,
     > + '_ {
+        // TODO: filter vexts whose epoch is not the current one
```
james-chf, line 160: I guess this TODO is already handled in the `validate_valset_upd_vext_and_get_it_back` calls?

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
```diff
@@ -50,56 +55,32 @@ where
...
-                "Validator set update vote extension issued for a block \
-                 height higher than the chain's last height.",
+                vext_epoch = ?ext.data.signing_epoch,
+                expected_epoch = ?signing_epoch,
+                "Validator set update vote extension issued for an epoch \
+                 different from the expected one.",
             );
-            return Err(VoteExtensionError::UnexpectedBlockHeight);
+            return Err(VoteExtensionError::UnexpectedEpoch);
         }
```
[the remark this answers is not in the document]
james-chf: If we're accepting validating a valset upd vext can be stateful, this is fine I think, re. your point as well about preventing spam of the mempool. An honest up-to-date node can reject vexts that look to be from an unexpected epoch.

james-chf (approved):

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
sug0: it's handled in an awkward place, I added it to `deserialize_vote_extensions`. it's because of the stale protocol txs, that cross epoch boundaries. might see if it's feasible to move that logic here, which makes more sense

Issue #989, opened 2023-01-09
Transfers of wNAM from Ethereum to Namada should result in release of NAM from escrow


james-chf:
When someone successfully makes a transfer of wNAM (wrapped NAM ERC20) tokens over the Ethereum bridge from Ethereum to Namada, it should result in native NAM being released from escrow on Namada to the receiver, rather than "wrapped wrapped NAM" multitoken being minted. In other words, the wNAM ERC20 address should be treated differently to other ERC20s by the Ethereum bridge when it appears as the asset in a `TransferToNamada` event.

Issue #508, opened 2022-09-23
End-to-end tests for transfers from Ethereum to Namada using the fake events endpoint


james-chf:
We should have end-to-end tests testing submission of fake Ethereum events to a Namada validator. These tests should be able to be run in CI. They should cover (at least) what the shell script test does in https://github.com/anoma/namada/pull/430 - i.e. sending wDAI to an established address on a Namada chain - but this issue is open ended, we need to come up with more test cases. Examples of things we need to cover:

- Namada chains with different numbers of validators (one, more than one) and different fractional voting powers
- transfers sent around the time of epoch transitions and validator set changes, when fractional voting powers may change

As a separate later issue - we will have equivalent end-to-end tests that use the fully deployed Ethereum smart contracts rather than the direct events submission endpoint.

To scope/plan out this issue for the moment (though perhaps this issue should become a tracking issue for individual issues for individual e2e tests)
- [ ] e2e test that sends an ERC20 transfer to an address on a Namada chain (e.g. similar to the test at https://github.com/anoma/ethereum-bridge-e2e-tests/commit/2746512b6afb32c70f768da22ea1252d6f9556ba) 
  - [ ] implicit receiver address
  - [ ] established receiver address
- [ ] extend the above e2e tests to work with chains with more than one validator (depends on https://github.com/anoma/namada/issues/1055)

cwgoes:
@james-chf Is this status up to date?

james-chf:
I've updated the checklist in the body of this issue - the previous second task (e2e tests relating to multitoken transfers with `namadac transfer`) has its own issue now so have removed it (https://github.com/anoma/namada/issues/578). Scoped this issue to be about adding e2e tests that submit fake Ethereum events to a `namada` node to test out Ethereum -> Namada transfers.

PR #1040, opened 2023-01-17
Handle wNAM transfers differently from other ERC20 transfers from Ethereum

james-chf:
Relates to https://github.com/anoma/namada/issues/989 and https://github.com/anoma/namada/issues/508

This PR 
- makes it so that transfers of the wNAM ERC20 (i.e. transfers of ERC20s with an address that matches the `native_erc20` config value in chain storage) do not result in the minting of wrapped ERC20s on Namada. For now, we no-op, in a later PR, we will release NAM from escrow
- adds some functionality for testing the Ethereum bridge using the fake events endpoint in e2e tests, and stubs out a test for wNAM transfers (which right now just greps for the log line indicating that the wNAM transfer was received)
- moves a helper fn for reading the `native_erc20` address from storage from the bridge pool VP to the `namada_ethereum_bridge` crate, so it can be used more generally (e.g. as in this PR)


ethereum_bridge/src/parameters.rs
```diff
@@ -218,6 +219,31 @@ impl EthereumBridgeConfig {
...
+    H: storage::traits::StorageHasher,
+{
+    let native_erc20 = bridge_storage::native_erc20_key();
+    match storage.read(&native_erc20) {
+        Ok((Some(bytes), _)) => {
+            Ok(EthAddress::try_from_slice(bytes.as_slice()).expect(
+                "Deserializing the Native ERC20 address from storage \
+                 shouldn't fail.",
+            ))
+        }
```
sug0, line 237: call `StorageRead::read` from `core` instead, maybe? it does a borsh deserialize automatically

ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -41,53 +46,76 @@ where
...
-        })?;
-        _ = changed_keys.insert(supply_key);
+        if asset != &wrapped_native_erc20 {
+            let mut changed =
+                mint_wrapped_erc20s(storage, asset, receiver, amount)?;
+            changed_keys.append(&mut changed)
+        } else {
+            tracing::warn!(
+                "Redemption of the wrapped native token is not yet supported"
+            )
```
sug0, line 64: maybe this is a bit silly, but we could log here the receiver and amount that would have been applied in storage

sug0 (changes requested):

ethereum_bridge/src/parameters.rs
james-chf: Yeah `StorageRead::read` is much nicer!

ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
james-chf: Sure, have updated the e2e test to check those also

sug0 (approved):

batconjurer (approved):

PR #1045, opened 2023-01-17
ci: forgot exit code i 

Fraccaman:


grarco (dismissed):

grarco (approved):

tests/src/e2e/ledger_tests.rs
```diff
@@ -126,7 +126,7 @@ fn test_node_connectivity_and_consensus() -> Result<()> {
     // 4. Check that all the nodes processed the tx with the same result
     let mut validator_0 = bg_validator_0.foreground();
     let mut validator_1 = bg_validator_1.foreground();
-    let expected_result = "all VPs accepted transaction";
+    let expected_result = "successful txs: 1";
```
tzemanovic, line 129: I think there's an issue with this (and we actually already have it in some tests) that is, 1 tx will appear twice in "successful txs", once for the wrapper and then inner tx. Some of the existing tests that are waiting for "Transaction is valid" have the same issue

tzemanovic: I think ideally the wrapper tx result should use a different type and format differently in the client (related to #20) and maybe we should distinguish these in the ledger stats log too

wasm/wasm_source/src/vp_implicit.rs
```diff
@@ -710,7 +710,8 @@ mod tests {
             std::fs::read(VP_ALWAYS_TRUE_WASM).expect("cannot load wasm");
 
         // hardcoded hash of VP_ALWAYS_TRUE_WASM
-        tx_env.init_parameters(None, None, Some(vec!["E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855".to_string()]));
+        // this must be lower case, as its written into genesis as lower case
+        tx_env.init_parameters(None, None, Some(vec!["e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855".to_string()]));
```
tzemanovic, line 714: I don't think these changes are necessary as the `init_paramaters` calls to the params update functions that apply lowercasing now

tests/src/e2e/ledger_tests.rs
Fraccaman: yeah, I agree, ill try to open a new PR to improve e2e

PR #1060, opened 2023-01-20
Add new Ethereum bridge RPC endpoints to query validator sets

sug0:
- Query an ABI encoded validator set at some `epoch` RPC param.
- Query an ABI encoded validator set update proof for some `epoch` RPC param.

Also add unit tests to check that these endpoints behave as expected, and refactor the Ethereum bridge queries in the router into a new sub-router.


core/src/types/voting_power.rs
```diff
@@ -4,9 +4,61 @@ use std::iter::Sum;
 use std::ops::{Add, AddAssign};
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use ethabi::ethereum_types as ethereum;
 use eyre::{eyre, Result};
 use num_rational::Ratio;
 
+/// Namada voting power, normalized to the range `0 - 2^32`.
```
batconjurer, line 11: How many times have we implemented this type :laughing:

sug0: no idea. I didn't find it in our code, so I made my own type lmao

james-chf (approved):

shared/src/ledger/queries/shell.rs
```diff
@@ -36,6 +29,11 @@ type Conversion = (
 );
 
 router! {SHELL,
+    // Shell provides storage read access, block metadata and can dry-run a tx
+
+    // Ethereum bridge specific queries
+    ( "eth_bridge" ) = (sub ETH_BRIDGE),
```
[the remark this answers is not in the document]
sug0: took a while to figure out how to move this to a sub-router haha, but yeah agreed

PR #1063, opened 2023-01-20
Change vext behavior if the bridge is inactive

batconjurer:
* Adds a storage key indicating the active status of the bridge. 
* Adds a quick query method to check if the bridge is active.
* Makes sure the vexts / protocol txs for eth events and bridge pool roots are not issued if the bridge is not active.
* Makes sure the vexts / protocol txs are not accepted if the the bridge is not active.

### DIFF
--- a/apps/src/lib/node/ledger/shell/vote_extensions.rs
+++ b/apps/src/lib/node/ledger/shell/vote_extensions.rs
@@ -57,6 +57,13 @@ pub enum VoteExtensionError {
     DivergesFromStorage,
     #[error("The signature of the Bridge pool root is invalid")]
     InvalidBPRootSig,
+    #[error(
+        "Received a vote extension for the Ethereum bridge which is currently \
+         not active"
+    )]
+    EthereumBridgeInactive,
+    #[error("A vote extension for the Ethereum bridge is missing.")]
+    MissingBridgeVext,
 }
 
 impl<D, H> Shell<D, H>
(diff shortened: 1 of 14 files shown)


apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -344,6 +371,104 @@ where
...
+    ) {
+        let mut eth_evs = vec![];
+        let mut bp_roots = vec![];
+        let mut valset_upds = vec![];
+        let bridge_active = self.storage.is_bridge_active();
+
+        for ext in deserialize_vote_extensions(vote_extensions) {
+            if let Some(validator_set_update) = ext.validator_set_update {
+                valset_upds.push(validator_set_update);
+            }
```
sug0, line 456: validator set updates should also be excluded from proposals, but they must be included an epoch before the bridge is started. therefore, `is_bridge_active()` needs an `Epoch` param. we should check if the bridge is active at the next epoch, to include valset upds

sug0: ah nevermind, just realized we have `check_bridge_status()`. this should be what we should be querying, then

sug0 (changes requested):
hm a couple of things:
- I think we should query the eth bridge status when validating vote extensions. this way, vote extensions issued when the bridge is inactive are rejected by `CheckTx`, but for this to work...
- we may need to refactor the way we store the activation status of the bridge. we need to query its status at a given epoch, the same when the vote extension was issued.
- it may be the case that a vote extension was issued during a period when the bridge was active, but then governance disabled the bridge before it got included, somehow? maybe this isn't an issue at all, because of pipeline lengths and so on

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -252,45 +265,59 @@ where
...
-        self.validate_eth_events_vext(
-            ext,
+        if !self.storage.is_bridge_active() {
+            ext.is_none()
+        } else {
+            if ext.is_none() {
+                return false;
+            }
+            self.validate_eth_events_vext(
+                ext.unwrap(),
```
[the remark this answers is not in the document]
batconjurer: I changed to an `if let` instead.

apps/src/lib/node/ledger/shell/vote_extensions.rs
batconjurer: That code isn't actually changed. It just registered as changed to github somehow.

sug0 (approved):
I can handle any validator set update code that is required to be changed in a later PR, so I'll approve this one

james-chf (approved):
`make check-abcipp` is failing in CI, otherwise lgtm

PR #1072, opened 2023-01-23
Handle TransferToEthereum

yito88:
- Add the inserted height to the Merkle tree of the bridge pool for the timeout checking
- Handle TransferToEthereum event
  - Delete the pending transfers contained in the event
  - Refund and delete for timed-out transfers

### DIFF
--- a/core/src/ledger/eth_bridge/storage/bridge_pool.rs
+++ b/core/src/ledger/eth_bridge/storage/bridge_pool.rs
@@ -1,6 +1,6 @@
 //! Tools for accessing the storage subspaces of the Ethereum
 //! bridge pool
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::convert::TryInto;
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
@@ -12,7 +12,7 @@ use crate::types::eth_abi::Encode;
 use crate::types::eth_bridge_pool::PendingTransfer;
 use crate::types::hash::Hash;
 use crate::types::keccak::{keccak_hash, KeccakHash};
-use crate::types::storage::{DbKeySeg, Key, KeySeg};
+use crate::types::storage::{BlockHeight, DbKeySeg, Key, KeySeg};
 
 /// The main address of the Ethereum bridge pool
 pub const BRIDGE_POOL_ADDRESS: Address =
@@ -68,6 +68,13 @@ pub fn is_bridge_pool_key(key: &Key) -> bool {
     matches!(&key.segments[0], DbKeySeg::AddressSeg(addr) if addr == &BRIDGE_POOL_ADDRESS)
 }
 
+/// Check if a key is for a pending transfer
+pub fn is_pending_transfer_key(key: &Key) -> bool {
+    is_bridge_pool_key(key)
+        && *key != get_signed_root_key()
+        && *key != get_nonce_key()
+}
+
 /// A simple Merkle tree for the Ethereum bridge pool
 ///
 /// Note that an empty tree has root [0u8; 20] by definition.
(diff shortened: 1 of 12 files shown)


ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -84,22 +103,263 @@ where
...
+    for event in transfers {
+        let pending_transfer = event.into();
+        let key = get_pending_key(&pending_transfer);
+        if storage.has_key(&key)?.0 {
+            _ = storage.delete(&key)?;
+            _ = pending_keys.remove(&key);
+        } else {
+            return Err(eyre::eyre!(
+                "The transfer doesn't exist in the bridge pool"
+            ));
```
sug0, line 136: I think the only way we could have run into this code path is if a quorum of validators decided on an event that didn't happen (that is, effectively, we would have had a Byzantine quorum of validators at that point). to transfer to Ethereum, users need to provide a proof that they sent a transfer to the bridge pool, at which point their transfer would be registered in storage.

therefore, maybe it's wiser to panic here? or use `unreachable!()`? there's also `namada_core::hints::likely()`, for the optimistic if branch.

sug0 (commented):
a few comments, but otherwise lgtm

core/src/ledger/eth_bridge/storage/bridge_pool.rs
```diff
@@ -95,16 +105,31 @@ impl BridgePoolTree {
...
+            eyre!("Could not parse key segment as a hash").into()
+        })
     }
 
     /// Update the tree with a new value.
     ///
     /// Returns the new root if successful. Will
     /// return an error if the key is malformed.
-    pub fn insert_key(&mut self, key: &Key) -> Result<Hash, Error> {
+    pub fn insert_key(
```
batconjurer, line 126: nit: The rest of the trees follow a key-value order for the insert arguments.

core/src/ledger/storage/merkle_tree.rs
```diff
@@ -324,12 +322,15 @@ impl<H: StorageHasher + Default> MerkleTree<H> {
     fn update_tree(
```
batconjurer, line 322: Let s just the make the height the value of the tree.

core/src/ledger/storage/traits.rs
```diff
@@ -60,6 +63,13 @@ impl<'a, H: StorageHasher + Default> SubTreeRead for &'a Smt<H> {
         }
     }
 
+    fn subtree_inserted_height(
+        &self,
+        _key: &Key,
+    ) -> Result<BlockHeight, Error> {
+        Err(Error::MerkleTree("SMT can't store the height".to_string()))
+    }
+
```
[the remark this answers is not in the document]
batconjurer: I think we should just tree the block height as the value type of the tree. Then we can use all the existing interfaces.

batconjurer (changes requested):

core/src/ledger/storage/merkle_tree.rs
```diff
@@ -344,10 +342,28 @@ impl<H: StorageHasher + Default> MerkleTree<H> {
...
 
+    /// Get the value in the tree
+    pub fn get(&self, key: &Key) -> Result<Vec<u8>> {
+        let (store_type, sub_key) = StoreType::sub_key(key)?;
+        self.tree(&store_type).subtree_get(&sub_key)
+    }
+
     /// Update the tree with the given key and value
-    pub fn update(&mut self, key: &Key, value: impl AsRef<[u8]>) -> Result<()> {
+    pub fn update(
```
batconjurer, line 352: can't we just pass the blockheight in as the value parameter here?

batconjurer (changes requested):

ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -84,22 +104,263 @@ where
...
+    // Remove the completed transfers from the bridge pool
+    for event in transfers {
+        let pending_transfer = event.into();
+        let key = get_pending_key(&pending_transfer);
+        if likely(storage.has_key(&key)?.0) {
+            _ = storage.delete(&key)?;
+            _ = pending_keys.remove(&key);
+        } else {
+            // The transfer should exist in the bridge pool
+            unreachable!()
```
sug0, line 136: nit: you can add the message in the comment to the panic itself
```suggestion
            unreachable!("The transfer should exist in the bridge pool");
```

sug0 (approved):
lgtm, just a small suggestion

sug0 (approved):

batconjurer (approved):

batconjurer (changes requested):
I have thought of a mild security concern that should be addressed actually. WIll leave comments in the relevant places.

batconjurer (approved):
I have thought of a mild security concern that should be addressed actually. WIll leave comments in the relevant places.

batconjurer:
> I have thought of a mild security concern that should be addressed actually. WIll leave comments in the relevant places.

Actually, it's fine.

PR #1083, opened 2023-01-25
PoS: convert tokens to TM voting power when checking for changes

### LINKED ISSUE #1082: tendermint validator set updates don't respect `tm_votes_per_token` param
this can cause tendermint to reject validator set updates when a small token change that doesn't affect TM voting power is given to it

tzemanovic:
backported from #740 to fix #1082

### DIFF
--- a/.changelog/unreleased/bug-fixes/1083-fix-tm-voting-power-check.md
+++ b/.changelog/unreleased/bug-fixes/1083-fix-tm-voting-power-check.md
@@ -1,0 +1,3 @@
+- Fixed Tendermint validator set update check to
+  respect the PoS tm_votes_per_token parameter.
+  ([#1083](https://github.com/anoma/namada/pull/1083))


proof_of_stake/src/lib.rs
```diff
@@ -805,14 +807,28 @@ pub trait PosBase {
                 if let (Some(prev_epoch), Some(prev_validators)) =
                     (previous_epoch, prev_validators)
                 {
-                    if prev_validators.active.contains(validator) {
+                    let prev_validator_stake =
+                        self.validator_stake(&validator.address, prev_epoch);
+                    if prev_validators.active.contains(validator)
+                        || into_tm_voting_power(
```
cwgoes, line 813: we should calculate `into_tm_voting_power(params.tm_votes_per_token, validator.bonded_stake)` once, store it in a variable, and reuse it in all these comparisons

cwgoes: & ditto for previous validator stake

cwgoes (changes requested):
Minor nits. Code changes otherwise look fine but haven't tested.

PR #1087, opened 2023-01-25
Ledger `--time` help

grarco:
More detailed help message

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -1702,11 +1702,14 @@ pub mod args {
         }
 
         fn def(app: App) -> App {
-            app.arg(
-                NAMADA_START_TIME
-                    .def()
-                    .about("The start time of the ledger."),
-            )
+            app.arg(NAMADA_START_TIME.def().about(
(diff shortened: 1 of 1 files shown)


apps/src/lib/cli.rs
```diff
@@ -1702,11 +1702,14 @@ pub mod args {
...
-            app.arg(
-                NAMADA_START_TIME
-                    .def()
-                    .about("The start time of the ledger."),
-            )
+            app.arg(NAMADA_START_TIME.def().about(
+                "The start time of the ledger. Accepts a relaxed form of \
+                 RFC3339. A space or a 'T' are accepted as the separator \
+                 between the date and time parts. Additional spaces are \
+                 allowed between each component.\nAll of these examples are \
```
bengtlofgren, line 1709: LGTM; only suggestion would maybe be to either use "parts" or "components" in both sentences. Something like "... the date and time **components**" Eh LGTM otherwise.

bengtlofgren (approved):

bengtlofgren (dismissed):

PR #1092, opened 2023-01-26
Bengt/pgf specs

bengtlofgren:


documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
...here, and art (conceptually - a physical painting is excludable and rivalrous, but the painting as-such is not). Namada's software stack, supporting research, and ecosystem tooling are all public goods, as are the information ecosystem and education which provide for the technology to be used safety, the hardware designs and software stacks (e.g. instruction set, OS, programming language) on which it runs, and the atmosphere and biodiverse environment which renders its operation possible. Without these things, Namada could not exist, and without their continued sustenance it will not continue to. Public goods, by their nature as non-excludable and non-rivalrous, are mis-modeled by economic systems (such as payment-for-goods) built upon the assumption of scarcity, and are usually either under-funded (relative to their public benefit) or funded in ways which require artificial scarcity and thus a public loss. For this reason, it is in the interest of Namada to help out, where possible, in funding the public goods upon which its existence depends in ways which do not require the introduction of artificial scarcity, balancing the costs of available resources and operational complexity.
```
cwgoes, line 3: we should keep this part, it's important as context

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
...ding to which justice cannot be done here. Most mechanisms fall into two categories: need-based and results-based, where need-based allocation schemes attempt to pay for particular public goods on the basis of cost-of-resources, and results-based allocation schemes attempt to pay (often retroactively) for particular public goods on the basis of expected or assessed benefits to a community and thus create incentives for the production of public goods providing substantial benefits (for a longer exposition on retroactive PGF, see [here](https://medium.com/ethereum-optimism/retroactive-public-goods-funding-33c9b7d00f0c), although the idea is [not new](https://astralcodexten.substack.com/p/lewis-carroll-invented-retroactive)). Additional constraints to consider include the cost-of-time of governance structures (which renders e.g. direct democracy on all funding proposals very inefficient), the necessity of predictable funding in order to make long-term organisational decision-making, the propensity for bike-shedding and damage to the information commons in large-scale public debate (especially without an identity layer or Sybil resistance), and the engineering costs of implementations.
```
cwgoes, line 7: we should keep this part, it's important as context

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+content: Vec<u8>,
+author: Address,
+r#type: PGFCouncil,
+votingStartEpoch: Epoch,
+votingEndEpoch: Epoch,
+graceEpoch: Epoch,
+}
+```
 
-This requires the following protocol components:
```
cwgoes, line 13: deleting this is fine :+1:

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
-- Interface support: the interface should support limited liquid democracy for delegate selection and approval voting for public goods council candidates. The interface or explorer should display past retroactive PGF winners and past/current continuous funding recipients. Proposal submission for continuous and retroactive funding will happen separately, in whatever manner the public goods council deems fit.
+### Constructing the council
+All valid PGF councils will be established multisignature account addresses. These must be created by the intdended parties that wish to create a council. The council will therefore have the discretion to decide what threshold will be required for their multisig (i.e the "k" in the "k out of n").
+
+The council will be resonsible to publish this address to voters and express their desired `spending_cap`.
```
cwgoes, line 26: how do they do that? can we allow them to publish in a transaction, just for convenience?

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+- Construction of PGF council multisigs 
+
+### Dealing with ties
+In the rare occurance of a tie, the council with the lower spending_cap will win the tiebreak.
+
+In the case of equal tiebreaks, the addresses with lower alphabetical order will be chosen. This is very arbitrary due to the expected low frequency.
+
+### Ensuring enough votes 
+
+In order for a new PGF council to be elected (and hence halting the previous council's power), $\frac{1}{3}$ of validating power must vote on the "New PGF Council Proposal". Once this condition has been met, majority vote will be decided as to which PGF council will be elected.
```
cwgoes, line 67: majority? I think you mean the (council, spending_cap) pair with the most votes, right?

this is approval voting (a pretty standard mechanism, so worth using the name)

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+
+### Electing the council
+
+Once the elected council has been decided upon, the established address corresponding to the multisig is added to the `PGF` internal address, and the `spending_cap` variable is stored.
+
+### End of Term Summary
+
+At the end of each term, the council is encouraged to submit a "summary"  which describes the funding decisions the councils have made and their reasoning for these decisions. This summary will act as an assessment of the council and will be the information point for governance to decide whether to re-elect the council. This summary should be made as transparant as pos
```
cwgoes, line 75:
```suggestion
At the end of each term, the council is encouraged to submit a "summary"  which describes the funding decisions the councils have made and their reasoning for these decisions. This summary will act as an assessment of the council and will be the information point for governance to decide whether to re-elect the council.
```

cwgoes: we can include other details later

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+
+## Mechanism
+
+The governance proposal PGFCouncil elects the multisignature account address of the PGF council.
+
+The PGF council members will (by definition) be in charge of a "k of n" multisignature account whereby the $\text{10% inflation} * \text{spending_cap}$ is allocated and spent from the `PGF` internal address.
```
cwgoes, line 81: to be clear, the multisig does not itself hold the PGF internal funds, it is just allowed to spend them

cwgoes (changes requested):
Thanks, a few requested changes and I think we should briefly discuss this sync. I realize that I didn't mention it in our last discussion but I think we should keep the original concept of a small allocation paid _to the PGF councIl_ for their service. This should be fixed by the network so to avoid bike-shedding and side payments. There can be a configurable parameters, e.g. 0.5% inflation, and this can just be paid per epoch to the PGF multisig directly, which can then choose what to do with it, but it is "intended" for the councilmembers themselves.

documentation/specs/src/economics/public-goods-funding.md
grarco: Do we need 1/3 of the voting power to vote on the proposal or to vote `Yay` on the proposal (regardless of the specific councils voted)? I'm thinking about a lot of voting power voting for `Nay` and a very small power voting for a council that ends up being elected

cwgoes: we need 1/3 to vote "yay" on the proposal, yeah - then the council with the most votes (which could be under 1/3) gets elected

documentation/specs/src/economics/public-goods-funding.md
cwgoes: also, this tx proves that the multisig can sign

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+
+In the case of equal tiebreaks, the addresses with lower alphabetical order will be chosen. This is very arbitrary due to the expected low frequency.
+
+### Ensuring enough votes 
+
+In order for a new PGF council to be elected (and hence halting the previous council's power), $\frac{1}{3}$ of validating power must vote on the "New PGF Council Proposal". Once this condition has been met, majority vote will be decided as to which PGF council will be elected.
+
+### Electing the council
+
+Once the elected council has been decided upon, the established address corresponding to the multisig is added to the `PGF` internal address, and the `spending_cap` variable is stored.
```
cwgoes, line 71: we need to track spending_cap, the pgf VP should check that some spending_amount variable (reset to 0 on election of a new council) is incremented by the amount of any transfer

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,45 +1,193 @@
...
+
+## Addresses
+Governance adds 1 internal address:
+
+`PGF` internal address
+
+The internal address VP will be allocated the 10% inflation of NAM. This will be added in addition to what was unspent by the previous council.
+It will also implement the logic required in order to allow the cPGF transactions to be made. I.e finalize-block, which will send the addresses their respective amounts each end-of-epoch.
+
+The Governance address should also allow the burning of funds by the council.
```
cwgoes, line 137: I don't think this requires special handling

cwgoes: or, is there an argument for special handling?

bengtlofgren: Regarding burning

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -183,6 +167,19 @@ pub fn(self) -> bool {
 }
 ```
 
+## Storage
+
+### Storage keys
+
+Each recipient will be listed under this storage space (for cPGF)
+- `/PGFAddress/active_projects/Address = Amount`
```
Fraccaman, line 175: add `spending_cap` and `spent_amount`

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -1,36 +1,92 @@
...
+--spending-cap "<proposed-spending-cap>"
+```
+The `--spending-cap` argument is a `float` $0 < x \leq 1$, which indicates the maximum proportion of the total funds available to the PGF council that the PGF council is able to spend during their term.
 
 A council consisting of the same members should also be able to propose multiple spending caps (with the same multisig address). These will be voted on as separate councils and votes counted separately.
 
 
 ### Voting on the council
-Once the council has been constructed, it can be voted on by governance particpants through a vote as structured below:
+Once the council has been constructed and brodcasted, it can be voted on by governance particpants. There will be a window in between which the `PgfProposal` was passed and when the final votes for the new council will be accepted. This is a governance parameter.
```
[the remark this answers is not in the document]
bengtlofgren: I meant like we let `PgfProposal` pass, which says we will elect a new council. Then there should be some "voting period" with a start and an end. That is what this [start, end] period is what I mean when I say "window".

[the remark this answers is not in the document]
grarco: Governance proposals already have a voting window, I think there's no need to state it again here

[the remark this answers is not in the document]
bengtlofgren: But there is a difference right? This governance proposal is just to say "we are voting for a new PGF council". That voting window is already implemented. We also need one for the time in which we decide *which* council is elected. Or am I missing smth?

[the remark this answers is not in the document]
grarco: I don't think we need anything more than what we already have. Someone proposes to vote for a new council. This proposal has a voting window inside which we accept incoming votes. When the voting period is over we simply tally the received votes and elect the winning council (if any)

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -83,9 +61,34 @@ The `--spending-cap` argument is a `float` $0 < x \leq 1$, which indicates the m
 
 A council consisting of the same members should also be able to propose multiple spending caps (with the same multisig address). These will be voted on as separate councils and votes counted separately.
 
+Proposing candidacy as a PGF council is something that is done at any time. This simply signals to the rest of Governance that a given established multisignature account address is willing to be voted on during a PGF council election in the future.
+
+Candidacy proposals last a default of 31 epochs. There is no limit to the number of times a council can be proposed for candidacy. This helps ensure that no PGF council is elected that does not intend to become one.
```
grarco, line 66: I think we can either use a `Map<Council, (Epoch, Url)>` or `VecDeque<(Council, Epoch, Url)>` where `Council = (Address, spending_cap)` to store the candidates in storage.

At each epoch, we can remove from this collection the expired candidacies. If we use the queue, the voters can use the indexes of the queue to specify their choices (even though this might be more error prone

bengtlofgren: What is the benefit of using the queue? What do you mean by indexes of the queue to specify choices?

grarco: We can use a `Vec` as well, with a queue we might be faster in certain operations (like pruning from the head). A voter can use the index of the council in the vector (or queue) instead of providing the tuple `(Address, spending_cap)`. At the moment I'm more concerned of what happens if some candidates are dropped in the middle of a voting period

grarco: If we remove expired candidacies during a voting period indexes would point to another council so this is not safe. At this point we can use the `Map` I'd say

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,243 @@ Namada groups public goods into four categories, with earmarked pools of funding
 
 ### Funding amounts
 
-In Namada, 10% inflation per annum of the NAM token is directed to this public goods mechanism, 5% to continuous funding and 5% to retroactive funding. This is a genesis default and can be altered by governance.
+In Namada, up to 10% inflation per annum of the NAM token is directed to this public goods mechanism. The division of these funds is entirely the discretion of the elected PGF council.
```
cwgoes, line 31:
```suggestion
In Namada, up to 10% inflation per annum of the NAM token is directed to this public goods mechanism. The further division of these funds is entirely up to the discretion of the elected PGF council.
```

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,243 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
 
 Namada encourages the public goods council to adopt a default social consensus of an equal split between categories, meaning 1.25% per annum inflation for each category (e.g. 1.25% for technical research continuous funding, 1.25% for technical research retroactive PGF). If no qualified recipients are available, funds may be redirected or burnt.
 
-Namada also pays the public goods council members themselves (in total) a default of 0.1% inflation per annum.
\ No newline at end of file
+Namada also encourages the PGF council to allocate the PGF council members up to 5% of these funds (0.5% inflation) as a payment for value added from the council itself.
```
cwgoes, line 35: This one we should hardcode, I think - it should be a protocol parameter - we don't really want councils to be competing on how much they pay themselves.

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,243 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+--alias my-multisig \
+--publickeys pk1,pk2,pk3... \
+--signers signer1,signer2,signer3... \
+--threshold k
+```
+
+### Proposing Candidacy
+The council will be responsible to publish this address to voters and express their desired `spending_cap`. This will be done directly to the ledger through the following CLI command:
+
+```bash!
```
cwgoes, line 55: Ditto, this should be in the docs

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,243 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
-Namada also pays the public goods council members themselves (in total) a default of 0.1% inflation per annum.
\ No newline at end of file
+Namada also encourages the PGF council to allocate the PGF council members up to 5% of these funds (0.5% inflation) as a payment for value added from the council itself.
+
+## Voting for the Council
+
+### Constructing the council
+All valid PGF councils will be established multisignature account addresses. These must be created by the intdended parties that wish to create a council. The council will therefore have the discretion to decide what threshold will be required for their multisig (i.e the "k" in the "k out of n").
+
+A new multisignature address will be constructed through the CLI with the following command
```
cwgoes, line 42: This is great but it should be in the docs not the specs (we don't need command examples in the specs)

cwgoes (changes requested):

documentation/specs/src/economics/public-goods-funding.md
bengtlofgren: Good idea

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -42,137 +98,65 @@ struct OnChainVote {
...
 
+### PGF council transactions
+The PGF council members will be responsible for collecting signatures offline. One member will then be responsinble for submitting a transaction containing at least $k $ out of the signatures.
 
-### cPGF (continuous PGF)
-
-<!-- These transactions will be more involved than the rPGF transactions (which will act as simple transactions).
+The collecting member of the council will then be responsible for submitting this tx through the multisig. The multisig will only accept the tx if this is true.
 
+Note that there is no cap to $n$ in $k$-out-of-$n$ apart from the limitations of the multisig account.
```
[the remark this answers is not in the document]
cwgoes: no, I don't think this is necessary

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+The governance set consists of Alice, Bob, Charlie, Dave, and Elsa. Each member has 20% voting power.
+
+The current PGF council consits of Dave and Elsa.
+
+- At epoch 42, Alice proposes the `PgfProposal` with the following struct:
+
+```rust
+struct PgfProposal{
+  id: 2
+  content: Vec<32,54,01,24,13,37>,
```
cwgoes, line 143: what is this exactly? serialised something?

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+    yay: proposalVote,
+}
+```
+
+Whereby the `proposalVote` includes
+```rust
+HashSet<(address: 0xBobCharlieMultisig, spending_cap: 1.0)>
+```
+
+- At epoch 49, Bob and Charlie submit identical transactions. Charlie also
```
cwgoes, line 171: Charlie also what?

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+Whereby the `proposalVote` includes
+```rust
+HashSet<(address: 0xBobCharlieMultisig, spending_cap: 1.0)>
+```
+
+- At epoch 49, Bob and Charlie submit identical transactions. Charlie also 
+
+- At epoch 50, Dave votes `Nay` on the proposal.
+
+- At epoch 51, Elsa votes `Yay` but on the multisigs `0xElsaAndSisterMultisig` AND `0xBobCharlieMultisig`.
```
cwgoes, line 175: With what spending caps?

documentation/specs/src/economics/public-goods-funding.md
bengtlofgren: :+1:

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+
+The collecting member of the council will then be responsible for submitting this tx through the multisig. The multisig will only accept the tx if this is true.
+
+Note that there is no cap to $n$ in $k$-out-of-$n$ apart from the limitations of the multisig account.
+
+The PGF council should be able to make both retroactive and continuous public funding transactions. Retroactive public funding transactions should be straightforward and implement no additional logic to a normal transfer.
+
+However, for continuous PGF (cPGF), the council should be able to submit a one time transaction which indicates the recipient addresses that should be eligble for receiveing cPGF. 
+
+The following data is attached to the PGF transaction and will allow the counsil to decide which projects will be continously funded. Each tuple represent the address and the respecting amount of NAM that will receive every epoch. The list of funding recipients will be stored in storage under the PGF internal address substorage space.
```
cwgoes, line 202:
```suggestion
The following data is attached to the cPGF transaction and will allow the counsil to decide which projects will be continously funded. Each tuple represent the address and the respecting amount of NAM that will receive every epoch. The list of funding recipients will be stored in storage under the PGF internal address substorage space.
```

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+
+```rust
+struct cPgfRecipients {
```
cwgoes, line 205: Probably we should have a transaction to add a recipient and a transaction to remove a recipient, just for ease of use by the council

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+}
+```
+The spending cap amount here is different from the `spending_cap` float. This is an actual NAM value that is calculated through `spending_cap * PGF_VP.balance`
+
+
+
+
+
+
+
```
cwgoes, line 257: can we delete the extra spaces

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,237 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+### Struct
+
+```rust
+struct Council {
+    address: Address,
+    spending_cap_amount: u64,
+    spent_amount: u64,
+}
+```
+The spending cap amount here is different from the `spending_cap` float. This is an actual NAM value that is calculated through `spending_cap * PGF_VP.balance`
```
cwgoes, line 250: float? where's a float? there should never be any floats anywhere ever :smile:

cwgoes: also I don't understand what this is - why would we multiply the spending cap by the balance?

cwgoes (changes requested):

documentation/specs/src/economics/public-goods-funding.md
bengtlofgren: Well I thought of spending_cap as a percentage of the balance. I guess we can make it a uint as well and it directly be the literal amount of NAM. Benefit of percentage is that it is interpretable and easy to remember whereas Amount is not (could be many zeros).

grarco: You can use the `Amount` type for the spending cap

bengtlofgren: I just don't think `Amount` is interpretable and can lead to a lot of misvotes. I'd rather use a percentage for spending cap.

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,214 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+
+
+### Proposing Candidacy
+The council will be responsible to publish this address to voters and express their desired `spending_cap`.
+
+The `--spending-cap` argument is a `u8` $0 < x \leq 100$, which indicates the maximum percentage of the total funds available to the PGF council that the PGF council is able to spend during their term.
+
+A council consisting of the same members should also be able to propose multiple spending caps (with the same multisig address). These will be voted on as separate councils and votes counted separately.
+
+Proposing candidacy as a PGF council is something that is done at any time. This simply signals to the rest of Governance that a given established multisignature account address is willing to be voted on during a PGF council election in the future.
```
cwgoes, line 50:
```suggestion
Proposing candidacy as a PGF council is something that is done at any time. This simply signals to the rest of governance that a given established multisignature account address is willing to be voted on during a PGF council election in the future.
```

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,214 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+  content: Vec<u8>,
+  author: Address,
+  r#type: PGFCouncil,
+  votingStartEpoch: Epoch,
+  votingEndEpoch: Epoch,
+  graceEpoch: Epoch,
+}
+```
+
+The above proposal type exists in order to determine *whether* a new PGF council will be elected. In order for a new PGF council to be elected (and hence halting the previous council's power), $\frac{1}{3}$ of validating power must vote on the `PgfProposal` and more than half of the votes must be in favor. If more than half of the votes are against no council is elected and the previous council's ability to spend funds (if applicable) is revoked. [Approval voting](https://en.wikipedia.org/wiki/Approval_voting#:~:text=Approval%20voting%20allows%20voters%20to,consider%20to%20be%20reasonable%20choices.) is employed in order to elect the new PGF council, *whilst* the `PgfProposal` is active. In other words, voters may vote for multiple PGF councils, and the council with the greatest proportion of votes will be elected.
```
cwgoes, line 78:
```suggestion
The above proposal type exists in order to determine *whether* a new PGF council will be elected. In order for a new PGF council to be elected (and hence halting the previous council's power), $\frac{1}{3}$ of validating power must vote on the `PgfProposal` and more than half of the votes must be in favor. If more than half of the votes are against no council is elected and the previous council's ability to spend funds (if applicable) is revoked. [Approval voting](https://en.wikipedia.org/wiki/Approval_voting#:~:text=Approval%20voting%20allows%20voters%20to,consider%20to%20be%20reasonable%20choices.) is employed in order to elect the new PGF council, *whilst* the `PgfProposal` is active. In other words, voters may vote for multiple PGF councils, and the council & spending cap pair with the greatest proportion of votes will be elected.
```

documentation/specs/src/economics/public-goods-funding.md
cwgoes: we still need to describe this, I think?

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,214 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+
+- At epoch 54, the voting period ends and the votes are tallied. Since 80% > 33% of the voting power voted on this proposal (everyone except Charlie), the intitial condition is passed and the Proposal is active. Further, because out of the total votes, most were `Yay`, (75% > 50% threshold), a new council will be elected. The council that received the most votes, in this case `0xBobCharlieMultisig` is elected the new PGF council. The Council `(address: 0xElsaAndSisterMultisig, spending_cap: 50)` actually received 0 votes because Elsa's vote included the wrong spending_cap.
+
+- At epoch 57, Bob and Charlie have the effective power to carry out Public Goods Funding transactions.
+````
+
+### End of Term Summary
+
+At the end of each term, the council is encouraged to submit a "summary"  which describes the funding decisions the councils have made and their reasoning for these decisions. This summary will act as an assessment of the council and will be the information point for governance to decide whether to re-elect the council.
```
cwgoes, line 170:
```suggestion
At the end of each term, the council is encouraged to submit a "summary"  which describes the funding decisions the councils have made and their reasoning for these decisions. This summary will act as an assessment of the council and will be the primary document on the basis of which governance should decide whether to re-elect the council.
```

documentation/specs/src/economics/public-goods-funding.md
cwgoes: bump

documentation/specs/src/economics/public-goods-funding.md
```diff
@@ -38,8 +28,214 @@ Namada groups public goods into four categories, with earmarked pools of funding
...
+```rust
+struct cPgfRecipients {
+    recipients: HashSet<(Address, u64)>
+}
+```
+The mechanism for these transfers will be implemented in `finalize-block.rs`, which will send the addresses their respective amounts each end-of-epoch.
+Further, the following transactions should be added in order to ease the management of cPGF recipients.
+
+```rust
+impl addRecipient for cPgfRecipients
```
cwgoes, line 197: this is not clear, needs to be described

cwgoes (changes requested):
Almost there! A few final minor comments

documentation/specs/src/economics/public-goods-funding.md
bengtlofgren: I think I added this

cwgoes (approved):
:fire:

PR #1093, opened 2023-01-27
Moving inner_tx from WrapperTx to Tx.

murisi:
The objective is to (with as little change as possible)  make Namada transactions signable on hardware constrained wallets by making them smaller. The approach to this taken in this branch is to replace inner transactions (amongst other objects) with their hashes, and then only attach the encrypted payloads at the top level.

More precisely, the following changes have been made:
* Replaced the transaction code field of `Tx` with a sum type that can either contain a hash or the literal code
* Removed the `inner_tx` field of `WrapperTx` leaving behind `tx_hash`
* Added an `inner_tx` field to `Tx` representing an attached encrypted inner transaction
* Added an `inner_tx_code` field to `Tx` representing an attached encrypted inner transaction code
* The ledger now decrypts and checks the attached `inner_tx` field against the `Wrapper::tx_hash` field
* The ledger now decrypts and checks the attached `inner_tx_code` field against the hash in the `Tx::code` field of the inner transaction

The implications of these changes are:
* Clients can now sign over either the full WASM code or a hash of it
* No encryption is required for/before the signing process (and hence on hardware wallets) because the signature for a `WrapperTx` is over the hash of an unencrypted `Tx`


murisi:
pls update wasm

murisi:
pls update wasm

core/src/types/internal.rs
```diff
@@ -48,12 +48,18 @@ impl From<bool> for HostEnvResult {
...
     /// A wrapper for `crate::types::transaction::WrapperTx` to conditionally
     /// add `has_valid_pow` flag for only used in testnets.
     #[derive(Debug, Clone, BorshDeserialize, BorshSerialize)]
     pub struct WrapperTxInQueue {
         /// Wrapper tx
         pub tx: crate::types::transaction::WrapperTx,
+        /// the encrypted payload
+        pub inner_tx: Option<EncryptedTx>,
+        /// the encrypted payload
+        pub inner_tx_code: Option<EncryptedTx>,
```
batconjurer, line 62: Given that this type can no also be code, maybe it should be renamed to something more generic. Maybe even just `Encrypted`.

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -921,9 +922,16 @@ mod test_utils {
         /// Add a wrapper tx to the queue of txs to be decrypted
         /// in the current block proposal
         #[cfg(test)]
-        pub fn enqueue_tx(&mut self, wrapper: WrapperTx) {
+        pub fn enqueue_tx(
+            &mut self,
+            wrapper: WrapperTx,
+            inner_tx: Option<EncryptedTx>,
```
batconjurer, line 928: It is a leak of information here to indicate whether the inner tx contains data or code or not.

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/mod.rs
murisi: If you're talking more generally about the `inner_tx: Option<EncryptedTx>` field of `Tx`, then I believe this isn't necessarily the case... For a valid transaction, the presence of a `WrapperTx` (with its mandatory `tx_hash: Hash` field) implies that the `inner_tx` field of its container is necessarily occupied by `Some(...)`. Hence for the current kinds of transactions we have, I think this information would already be known/be derivable.

The reason why `inner_tx: Option<EncryptedTx>` was given an option type was to allow hardware-constrained hardware wallets to receive abbreviated/truncated versions of transactions (without making more types) to sign over.

core/src/proto/types.rs
```diff
@@ -382,7 +471,67 @@ impl Tx {
...
+    ) -> Self {
+        let inner_tx = EncryptedTx::encrypt(&tx.to_bytes(), encryption_key);
+        self.inner_tx = Some(inner_tx);
+        self
+    }
+
+    /// Attach the given transaction code to this one. Useful when the inner_tx
+    /// field contains a Tx and its code field needs a witness.
+    #[cfg(feature = "ferveo-tpke")]
+    pub fn attach_inner_tx_code(
```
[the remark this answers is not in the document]
murisi: I checked and believe the encryption algorithm is probabilistic: running the encryption algorithm twice on the same message and key is giving different ciphertexts. Scanning through its code I see nonces and uses of `rng` in the cipher initialization.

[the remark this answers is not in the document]
batconjurer: Indeed, that seems to be the case.

apps/src/lib/node/ledger/shell/mod.rs
batconjurer: I see. Sounds reasonable

murisi:
pls restart ci

core/src/types/internal.rs
murisi: Just renamed `WrapperTxInQueue` to `TxInQueue` to represent that all 3 fields are included in  and are the main constituents of the `Tx` struct.

Fraccaman:
pls spawn devnet [anoma-devnet-0.13.0-zondax.toml,60,heliaxdev@1be9883,ON]

Fraccaman:
pls spawn devnet [anoma-devnet-0.13.0-zondax.toml,60,heliaxdev@1be9883,ON]

Fraccaman:
pls spawn devnet [anoma-devnet-0.13.0-zondax,60,heliaxdev@1be9883,ON]

Fraccaman:
pls spawn devnet [anoma-devnet-0.13.0-zondax,60,heliaxdev@1be9883,ON]

Fraccaman:
pls spawn devnet [anoma-devnet-0.13.0-zondax,60,heliaxdev@1be9883,ON]

murisi:
> @murisi would we still want to use this with the planned changes in #1255? I also see #1156, is that an alternative to this one?

Yes, that's the plan. I'm working on a branch that applies the proposal #1255 to this branch. If this idea works out, then this PR will be superseded/redundant. #1156 was the original approach to enabling hardware wallet signing and is now superseded/redundant. I've just closed it now.

PR #1098, opened 2023-01-27
Compensate bridge pool relayers

batconjurer:
This PR does the following:

 * Adds a relayer address field to`TransferToEthereum` events parsed from Ethereum
 * Adds a cli argument for addinga relayer address to bridge pool proofs
 * Adds logic for paying the relayer the gas fees of the transfers relayed.
 * Covers with tests.


apps/src/lib/node/ledger/ethereum_node/events.rs
```diff
@@ -308,7 +311,9 @@ pub mod eth_events {
                         ParamType::Uint(256),
                         ParamType::String,
                         ParamType::Uint(256),
+                        ParamType::String,
```
yito88, line 314: Does this mean that each `transfer` includes the relayer address?
Also, we have to change `TRANSFER_TO_ETHEREUM_SIG`.

batconjurer: Good catch. I thought I had removed that.  And I forgot the signature too.

sug0 (approved):

Issue #1046, opened 2023-01-17
Ethereum event oracle should track its progress across node restarts


james-chf:
The Ethereum event oracle iterates through Ethereum blocks and checks for events from the bridge smart contracts to vote for. When a validator node starts, it should continue processing blocks starting with the most recent Ethereum block that it hasn't fully processed (assuming it is currently an active validator - see https://github.com/anoma/namada/issues/589).

If a validator shuts down without having fully processed an Ethereum block (i.e. it's voted for some of the events in the block, but not all of them), when it starts up again it is slightly inefficient but OK for the validator to re-process the block from the start - a validator may broadcast more than one vote for an event as only at most one vote per validator per event will be counted in storage.

See https://github.com/anoma/namada/issues/589#issuecomment-1384242000 for some more context

PR #1101, opened 2023-01-30
Start the Ethereum event oracle from the correct Ethereum block between node restarts

james-chf:
Closes https://github.com/anoma/namada/issues/1046

This PR
- adds an explicit Borsh-serializable type for representing Ethereum block heights in storage - right now, this is underlain by `num256::Uint256`, but there may be a better type to use. Ethereum itself seems to use an arbitrary precision signed integer to represent block heights.
- makes it so that the oracle continuously publishes its most recently processed Ethereum block height to a [tokio::watch](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) channel. At every ABCI Commit, we record the most recently processed Ethereum block to local storage (i.e. RocksDB).
- makes it so that the oracle should start processing from the most recently processed Ethereum block between node restarts.

Limitations
- persistence of the most recently processed Ethereum block height is done periodically as part of consensus, rather than as and when an Ethereum block is actually processed. This means when a node starts up, it may reprocess a block and broadcasts votes for Ethereum events it has already voted on. Although ultimately only one vote per validator address per Ethereum event is counted on chain, these are unnecessary txs being broadcast into the mempool.

TODO
- [ ] e2e test


james-chf:
Needs an e2e test

core/src/types/ethereum.rs
```diff
@@ -0,0 +1,94 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+use num256::Uint256;
+
+/// This type must be able to represent any valid Ethereum block height. It must
+/// also be Borsh serializeable, so that it can be stored in blockchain storage.
+///
+/// In Ethereum, the type for block height is an arbitrary precision integer - see <https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/types/block.go#L79>.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct BlockHeight {
+    inner: Uint256,
```
james-chf, line 16: This `Uint256` type is what we (and our web3 crate) are mostly using to represent Ethereum block heights, we need something Borsh-serializeable for storage though

core/src/ledger/storage/local_node.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// What the initial value for this key should be when a chain is created.
+    fn initial_value(&self) -> T;
+    /// The storage key under which this value is stored.
+    fn key(&self) -> storage::Key;
+}
+
+/// Represents the last Ethereum block height that the Ethereum oracle fully
+/// processed.
+#[derive(Debug)]
+pub struct EthereumOracleLastProcessedBlock;
```
james-chf, line 22: There are other local node values we currently store in RocksDB (e.g. `tx_queue`)

apps/src/lib/node/ledger/ethereum_node/oracle/last_processed_block.rs
```diff
@@ -0,0 +1,14 @@
+//! Functionality to do with publishing which blocks we have processed.
```
batconjurer, line 1: This probably doesn't need to live in a separate file.

apps/src/lib/node/ledger/ethereum_node/oracle/mod.rs
```diff
@@ -349,6 +361,22 @@ fn process_queue(
     confirmed
 }
 
+/// Functionality to do with publishing which blocks we have processed.
+pub mod last_processed_block {
```
sug0, line 365: nit: module docstrings need to be moved inside the module
```suggestion
pub mod last_processed_block {
    //! Functionality to do with publishing which blocks we have processed.
```

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -778,10 +809,20 @@ where
                 );
                 return;
             };
+            let start_block = match &self.storage.ethereum_height {
+                Some(start_block) => start_block.clone(),
+                None => ethereum::BlockHeight::from(0),
+            };
```
sug0, line 815: `self.storage.ethereum_height.cloned().unwrap_or_default()`?

core/src/types/ethereum.rs
```diff
@@ -0,0 +1,94 @@
+//! Types to do with Ethereum.
+
+use std::fmt;
+use std::num::NonZeroU64;
+use std::ops::{Add, AddAssign, Deref};
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use num256::Uint256;
```
sug0, line 8: couldn't we simply use the uin256 type from `ethabi`? `num256` could be purged from our deps

core/src/types/ethereum.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// This type must be able to represent any valid Ethereum block height. It must
+/// also be Borsh serializeable, so that it can be stored in blockchain storage.
+///
+/// In Ethereum, the type for block height is an arbitrary precision integer - see <https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/types/block.go#L79>.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
sug0, line 14:
```suggestion
#[derive(Default, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```

sug0 (changes requested):

sug0:
another thing, which I forgot to mention in the review: we now have `core/src/types/ethereum_structs.rs`. we could move this `BlockHeight` def there

core/src/types/ethereum.rs
sug0: ah, that's a shame

core/src/types/ethereum_structs.rs
```diff
@@ -1,2 +1,95 @@
-//! Ethereum bridge struct re-exports.
+//! Ethereum bridge struct re-exports adn types to do with ethereum.
```
sug0, line 1:
```suggestion
//! Ethereum bridge struct re-exports and types to do with ethereum.
```
just this type here. otherwise lgtm

sug0 (approved):

Issue #56, opened 2022-01-26
MASP validity predicate


juped:
Depends on anoma/namada#54, anoma/namada#55.

Desired abilities:

* stores notes/nullifiers
* verifies (proof, notes nullifiers) sets from transactions
* exposes readable Merkle tree roots for clients in its storage
* holds assets (just balances for now?)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762709310621) by [Unito](https://www.unito.io)

murisi:
@cwgoes If I'm not mistaken, I believe @juped has completed these tasks. But I would need to review the validity predicate code to ascertain whether the validation is sufficient, i.e. to ensure that the VP does not in any circumstances accept any transactions that should be rejected.

murisi:
@juped Ah, I see. Great! Regardless of whether or not we port, I see a potential issue at https://github.com/anoma/namada/blob/main/wasm/wasm_source/src/vp_masp.rs#L75 . If the client somehow manages to construct a `Transaction` with `shielded_tx.value_balance` equal to zero (and hence having no components), the VP might crash in the process of `unwrap`ping.

More generally to prevent potential exploits, we should probably try (whether here or in the port) to enforce the 4 conditions at https://github.com/anoma/namada/blob/main/documentation/specs/src/masp/ledger-integration.md#boundary-conditions . They serve to ensure that the outer `Transfer` object is consistent with the inner `Transaction` in terms of the numbers and fields of the transparent inputs/outputs. I think we haven't really had an issue here yet because the CLI client has been well behaved so far.

murisi:
> yes; but we were discussing porting it to native, which would be so much less flaky
I'm not sure what the status of this is, but I've identified some potential further issues in the MASP VP code as it currently stands: https://github.com/anoma/namada/issues/1373 .

PR #1104, opened 2023-01-31
Addresses the boundary conditions for the Masp VP predicate

mariari:
This PR implements the boundary conditions found in [#56](https://github.com/anoma/namada/issues/56#issuecomment-1400015486).

The following checks are not included

1. If the target address is not the masp, then the public key must be
checked

2. If the source address is not the masp, then the asset type must be
derived

3. If the source address is not the masp, then the value must be
derived

The first one is not satisifed as TxOut lacks a public key to check
against, not only in code but also in the spec

https://github.com/anoma/namada/blob/83e83315a5eaec302d08ed5db242d70625cfe45f/documentation/specs/src/masp/ledger-integration.md#transparent-output-format

The second and third are not satisfied as TxIn in code lacks the
values in the spec that can be used to satisfy them

https://github.com/anoma/namada/blob/83e83315a5eaec302d08ed5db242d70625cfe45f/documentation/specs/src/masp/ledger-integration.md#transparent-input-format

I suggest a followup pr once those values are in.

### DIFF
--- a/wasm/wasm_source/src/vp_masp.rs
+++ b/wasm/wasm_source/src/vp_masp.rs
@@ -1,11 +1,57 @@
 use std::cmp::Ordering;
 
 use masp_primitives::asset_type::AssetType;
-use masp_primitives::transaction::components::Amount;
+use masp_primitives::transaction::components::{Amount, TxIn, TxOut};
 /// Multi-asset shielded pool VP.
 use namada_vp_prelude::address::masp;
 use namada_vp_prelude::storage::Epoch;
 use namada_vp_prelude::*;
+
+/// Generates the current asset type given the current epoch and an
+/// unique token address
+fn asset_type_from_epoched_address(epoch: Epoch, token: &Address) -> AssetType {
+    // Timestamp the chosen token with the current epoch
+    let token_bytes = (token, epoch.0)
+        .try_to_vec()
+        .expect("token should serialize");
+    // Generate the unique asset identifier from the unique token address
+    AssetType::new(token_bytes.as_ref()).expect("unable to create asset type")
+}
+
+/// Checks if the asset type matches the expected asset type, Adds a
+/// debug log if the values do not match.
+fn valid_asset_type(
+    asset_type: &AssetType,
+    asset_type_to_test: &AssetType,
+) -> bool {
+    let res =
+        asset_type.get_identifier() == asset_type_to_test.get_identifier();
+    if !res {
+        debug_log!(
+            "The asset type must be derived from the token address and \
+             current epoch"
+        );
+    }
+    res
+}
+
+/// Checks if the reported transparent amount and the unshielded
+/// values agree, if not adds to the debug log
+fn valid_transfer_amount(
+    reporeted_transparent_value: u64,
+    unshielded_transfer_value: u64,
+) -> bool {
+    let res = reporeted_transparent_value == unshielded_transfer_value;
+    if !res {
+        debug_log!(
+            "The unshielded amount {} disagrees with the calculated masp \
+             transparented value {}",
+            unshielded_transfer_value,
+            reporeted_transparent_value
+        )
+    }
+    res
+}
 
 /// Convert Namada amount and token type to MASP equivalents
 fn convert_amount(
@@ -13,13 +59,7 @@ fn convert_amount(
     token: &Address,
     val: token::Amount,
 ) -> (AssetType, Amount) {
-    // Timestamp the chosen token with the current epoch
-    let token_bytes = (token, epoch.0)
-        .try_to_vec()
-        .expect("token should serialize");
-    // Generate the unique asset identifier from the unique token address
-    let asset_type = AssetType::new(token_bytes.as_ref())
-        .expect("unable to create asset type");
+    let asset_type = asset_type_from_epoched_address(epoch, token);
     // Combine the value and unit into one amount
     let amount = Amount::from_nonnegative(asset_type, u64::from(val))
         .expect("invalid value or asset type for amount");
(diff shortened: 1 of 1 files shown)


mariari:
pls update wasm

cwgoes:
This looks alright to me, but @murisi knows the code better - can you take a look?

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -55,7 +97,32 @@ fn validate_tx(
...
 
         // Handle shielding/transparent input
+        // The following boundary conditions must be satisfied
+        // 1. One transparent input
+        // 2. Zero transparent output
+        // 3. the transparent transaction value pool's amount must equal the
+        // containing wrapper transaction's fee amount
         if transfer.source != masp() {
+            // Satisfies 1.
+            if shielded_tx.vin.len() != 1 {
```
murisi, line 107: Given that we no longer intend to check the amount and asset type of the transparent input, this condition is no longer necessary. The value balance checks will ensure that the user has indeed provided sufficient inputs.

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -55,7 +97,32 @@ fn validate_tx(
...
+                debug_log!(
+                    "Transparent input to a transaction from the masp must be \
+                     1 but is {}",
+                    shielded_tx.vin.len()
+                );
+                return reject();
+            }
+
+            // Satisfies 2.
+            if !shielded_tx.vout.is_empty() {
```
murisi, line 117: This check enforces that transactions with a non-MASP source must have no transparent outputs. While this check can be done, the transaction would still be valid (i.e. leaving the shielded pool and transparent pool in a consistent state) without it. Perhaps we can remove this check to simplify the logic?

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -70,9 +137,50 @@ fn validate_tx(
...
         // Handle unshielding/transparent output
+        // The following boundary conditions must be satisfied
+        // 1. Zero transparent inupt
+        // 2. One transparent output
+        // 3. Asset type must be properly derived
+        // 4. Value from the output must be the same as the containing transfer
+        // 5. Public key must be the hash of the target
         if transfer.target != masp() {
+            // Satisfies 1.
+            if !shielded_tx.vin.is_empty() {
```
murisi, line 148: This check enforces that transactions with non-MASP targets mut not have any transparent inputs. While this check can be done, the transaction would still be valid (i.e. leaving the shielded pool and transparent pool in a consistent state) without it. Perhaps we can remove this check to simplify the logic?

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -70,9 +137,50 @@ fn validate_tx(
...
+            // Satisfies 3. and 4.
+            if !(valid_asset_type(&expected_asset_type, &out.asset_type)
+                && valid_transfer_amount(out.value, u64::from(transfer.amount)))
+            {
+                return reject();
+            }
+
             // Timestamp is derived to allow unshields for older tokens
-            let atype =
+            let atype: &AssetType =
```
murisi, line 183: This use of `.components().next().unwrap()` seems to be unsafe. What guarantees that `shielded_tx.value_balance` has at least one component? And when there are multiple components, what does the first extracted component mean semantically?

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -55,7 +97,32 @@ fn validate_tx(
         transparent_tx_pool += shielded_tx.value_balance.clone();
 
         // Handle shielding/transparent input
+        // The following boundary conditions must be satisfied
+        // 1. One transparent input
+        // 2. Zero transparent output
+        // 3. the transparent transaction value pool's amount must equal the
+        // containing wrapper transaction's fee amount
```
murisi, line 104: Where do we check that condition (3) is met? It seems to me that there should be some mention of the wrapper transaction's fee somewhere in this code?

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -55,7 +97,32 @@ fn validate_tx(
...
         if transfer.source != masp() {
```
murisi, line 105: Where do we check for the for the following condition?: If the source address is the MASP validity predicate, then no transparent inputs are permitted in the shielded transaction.

It seems to me that if the source and target are MASP, then both the conditional on line 105 and line 146 are skipped. This would allow for these sorts of transactions to have unlimited transparent inputs and outputs.

wasm/wasm_source/src/vp_masp.rs
```diff
@@ -70,9 +137,50 @@ fn validate_tx(
         }
 
         // Handle unshielding/transparent output
+        // The following boundary conditions must be satisfied
+        // 1. Zero transparent inupt
+        // 2. One transparent output
+        // 3. Asset type must be properly derived
+        // 4. Value from the output must be the same as the containing transfer
+        // 5. Public key must be the hash of the target
         if transfer.target != masp() {
```
murisi, line 146: Where do we check for the following condition?: If the target address is the MASP validity predicate, then no transparent outputs are permitted in the shielded transaction.

It seems to me that if the source and target are MASP, then both the conditional on line 105 and line 146 are skipped. This would allow for these sorts of transactions to have unlimited transparent inputs and outputs.

murisi (changes requested):
Thanks for the changes. It seems that 3 checks are missing (for MASP target, for MASP source, and for transaction fees). And perhaps maybe some checks could be slightly weakened without affecting the validity of this VP?

murisi:
@cwgoes You mentioned that you found the specs at https://github.com/anoma/namada/blob/main/documentation/specs/src/masp/ledger-integration.md confusing and had (or were going to) rewritten some parts. Have you made any edits to the `Boundary Conditions` section? Any improvements there could help improve the clarity of this code...

PR #1105, opened 2023-01-31
Extra Bridge pool queries

### LINKED ISSUE #425: Make the Ethereum bridge pool queryable

batconjurer:
Closes #425


core/src/types/voting_power.rs
```diff
@@ -169,6 +178,66 @@ impl BorshSchema for FractionalVotingPower {
...
+                de::Error::custom("Expected a '/' separated pair of numbers"),
+            ))?;
+        let numer = numer
+            .trim()
+            .parse::<u64>()
+            .map_err(|e| de::Error::custom(e.to_string()))?;
+        let denom = denom
+            .trim()
+            .parse::<u64>()
+            .map_err(|e| de::Error::custom(e.to_string()))?;
```
sug0, line 219: this technically allows you to parse `1/2` as well (without the whitespace). is that the intended behavior?

sug0: actually reading through the code again, you're only making use of the serialization routine, right? deserialization doesn't seem to be used?

batconjurer: That is intended behavior. And I added the deserialization because I figured I would use it in a test e2e test later.

batconjurer: Although I could just add a `FromStr` impl later if I suppose.

sug0 (approved):

yito88 (approved):

PR #1110, opened 2023-02-01
Wallet bip39 / bip32 / bip44 support

karbyshev:
Implements HD wallet derivation / recovery from a given mnemonic code.


apps/src/lib/cli.rs
```diff
@@ -2983,6 +3038,46 @@ pub mod args {
         }
     }
 
+    /// Wallet add key and implicit address arguments
+    #[derive(Clone, Debug)]
+    pub struct KeyAndAddressRestore {
+        /// Scheme type
+        pub scheme: SchemeType,
```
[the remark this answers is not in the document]
karbyshev: Then we should forbid using mnemonic codes for key generation of the other schemes, I guess?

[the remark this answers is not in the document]
karbyshev: Done

apps/src/lib/wallet/mod.rs
```diff
@@ -513,32 +635,110 @@ impl Wallet {
...
+    std::io::stdin().read_line(&mut response)?;
+    Ok(SecStr::from(response))
+}
+
+fn read_mnemonic_code() -> Result<Mnemonic, GenRestoreKeyError> {
+    let phrase = get_secure_user_input("Input mnemonic code: ")
+        .map_err(|_| GenRestoreKeyError::MnemonicInputError)?;
+
+    Mnemonic::from_phrase(
+        unsafe { std::str::from_utf8_unchecked(phrase.unsecure()) },
```
tzemanovic, line 680: The use of the `SecStr` is not playing nicely with the `Mnemoic` :/ Even though we could use the safe variant `from_utf8`, I'd be in favor of replacing it with zeroize because we're not using anything that's not already provided from zeroize and just use plain `zeroize::Zeroizing` over a String. This can then be easily passed into `Mnemoic::from_phrase` as it implements Deref

apps/src/lib/wallet/derivation_path.rs
```diff
@@ -0,0 +1,167 @@
...
+    where
+        P: Into<Box<[ChildIndex]>>,
+    {
+        Self(DerivationPathInner::new(path))
+    }
+
+    pub fn is_compatible(&self, scheme: SchemeType) -> bool {
+        let mut it = self.0.into_iter();
+        let _ = it.next();
+        if let Some(coin_type) = it.next() {
```
tzemanovic, line 36:
```suggestion
        if let Some(coin_type) = self.0.as_ref().get(1) {
```

apps/src/lib/wallet/derivation_path.rs
```diff
@@ -0,0 +1,167 @@
...
+                SchemeType::Common => unimplemented!("not implemented"),
+            },
+        ]
+    }
+
+    fn bip44(
+        scheme: SchemeType,
+        account: Option<u32>,
+        change: Option<u32>,
+        address: Option<u32>,
```
tzemanovic, line 63: This is only ever used with all 0s. Will we ever need to call these with None?

core/src/types/key/common.rs
```diff
@@ -280,6 +280,13 @@ impl super::SigScheme for SigScheme {
         );
     }
 
+    fn from_bytes(_seed: [u8; 32]) -> Self::SecretKey {
+        panic!(
```
tzemanovic, line 284:
```suggestion
        unimplemented!(
```
this one will also panic but it conveys more info

apps/src/lib/wallet/derivation_path.rs
karbyshev: I will remove `Option` from the parameters.

apps/src/lib/wallet/mod.rs
karbyshev: SecStr is replaced with Zeroizing now.

apps/src/lib/wallet/derivation_path.rs
```diff
@@ -0,0 +1,167 @@
...
+    ) -> Self {
+        let mut indexes = Self::bip44_base_indexes_for_scheme(scheme);
+        if let Some(account) = account {
```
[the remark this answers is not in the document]
karbyshev: `Option` is removed from arguments.

tzemanovic:
pls update wasm

tzemanovic (dismissed):

tzemanovic:
pls update wasm

tzemanovic (approved):

Issue #432, opened 2022-09-02
End-to-end tests for wrapped ERC20 transfers between accounts


james-chf:
When a balance transfer is carried out, the EthBridge VP should only allow it if the sender of the transfer (i.e. owner of the balance which is being decreased) authorized the transaction.

This issue can be closed when we have tests for the following types of ERC20 transfers (similar to the generic multitoken transfer tests in https://github.com/anoma/namada/pull/886):
- [ ] implicit -> implicit
- [ ] implicit -> established
- [ ] established -> implicit
- [ ] established -> established

james-chf:
#8 is completed, so have widened the scope of this issue to cover wrapped ERC20 transfers for all accounts, not just established accounts. This issue depends on https://github.com/anoma/namada/issues/578, after which we can specifically have tests for wrapped ERC20 transfers.

PR #1112, opened 2023-02-01
Ensure wrapped ERC20 transfers are correctly authorized and add end-to-end tests for them

james-chf:
Closes https://github.com/anoma/namada/issues/432 and relates to https://github.com/anoma/namada/issues/508

This PR
- ensures we check the relevant VPs are being executed for wrapped ERC20 transfers
- adds end-to-end tests covering
  - transfers of DAI from Ethereum to implicit and established Namada accounts
  - authorized transfers of wrapped DAI between implicit <-> established accounts once on Namada
  - rejecting transactions that attempt to transfer wrapped DAI from implicit/established accounts, if the transactions aren't signed by the key which controls the sending account

NB: some helper fns added/used in this PR are also in https://github.com/anoma/namada/pull/886/files#diff-b26bbd416d9d500f8f12aee8f7b3182f35b5cacb748898d9c07fe2cdd41e69b8

### DIFF
--- a/shared/src/ledger/native_vp/ethereum_bridge/authorize.rs
+++ b/shared/src/ledger/native_vp/ethereum_bridge/authorize.rs
@@ -10,8 +10,9 @@ use namada_core::types::address::Address;
 pub(super) fn is_authorized(
     verifiers: &BTreeSet<Address>,
     sender: &Address,
+    receiver: &Address,
 ) -> bool {
-    verifiers.contains(sender)
+    verifiers.contains(sender) && verifiers.contains(receiver)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


sug0 (approved):
maybe I'd change some of the `INFO` to `DEBUG` logs, but otherwise lgtm

shared/src/ledger/native_vp/ethereum_bridge/authorize.rs
```diff
@@ -1,37 +1,43 @@
...
-) -> Result<bool> {
-    tracing::warn!(
-        "authorize::is_authorized is not implemented, so all transfers are \
-         authorized"
-    );
-    Ok(true)
+    verifiers: &BTreeSet<Address>,
+    sender: &Address,
+) -> bool {
+    verifiers.contains(sender)
```
batconjurer, line 14: honest question, I've always wondered if we trigger the receiver vp as well. Or can people not refuse money from certain accounts?

batconjurer (commented):
Generally looks good, I'm just waiting on an answer as to whether we trigger receiver vps for token transfers.

shared/src/ledger/native_vp/ethereum_bridge/authorize.rs
batconjurer: It seems we should trigger the receiver account as well.

batconjurer (changes requested):
Make sure the receiver vp also gets triggered.

shared/src/ledger/native_vp/ethereum_bridge/authorize.rs
james-chf: Will add this, will require some small refactoring.

PR #1113, opened 2023-02-01
Write Merkle tree stores only when a new epoch

yito88:
- Write Merkle tree roots and stores only when a new epoch
- Rebuild the tree with the latest saved stores and subspace diffs if needed


yito88:
pls update wasm

yito88:
Depends on #1116

yito88:
pls update wasm

PR #1115, opened 2023-02-03
Fix the ABI encoding of secp256k1 signatures

sug0:


core/src/ledger/storage/mod.rs
```diff
@@ -26,7 +26,7 @@ use rayon::iter::{
 #[cfg(feature = "wasm-runtime")]
 use rayon::prelude::ParallelSlice;
 use thiserror::Error;
-pub use traits::{Sha256Hasher, StorageHasher};
+pub use traits::{DummyHasher, KeccakHasher, Sha256Hasher, StorageHasher};
```
batconjurer, line 29: I see new imports in this file, but the diff doesn't show any uses of them.

core/src/types/key/secp256k1.rs
```diff
@@ -499,6 +551,10 @@ impl TryFrom<&[u8; 65]> for Signature {
 pub struct SigScheme;
 
 impl super::SigScheme for SigScheme {
+    #[cfg(any(test, feature = "secp256k1-sign-verify"))]
+    type Hasher = crate::ledger::storage::KeccakHasher;
+    #[cfg(not(any(test, feature = "secp256k1-sign-verify")))]
+    type Hasher = crate::ledger::storage::DummyHasher;
```
batconjurer, line 557: should this impl block exist at all if the sign-verify flag is disabled?

sug0: it exists, if you notice the code in `sign()`, we panic there

batconjurer: But *should* it? Instead of runtime panics, we could have compile time checks. Is this trait bound required elsewhere in note feature gated code?

batconjurer (approved):

core/src/types/eth_abi.rs
```diff
@@ -102,13 +101,9 @@ pub trait Encode<const N: usize>: Sized {
...
     /// keccak hash of the encoded string appended to an Ethereum
     /// signature header. This can then be signed.
-    fn signable_keccak256(&self) -> Vec<u8> {
-        let mut output = [0; 32];
-        let message = self.encode().into_inner();
-        let mut state = Keccak::v256();
-        state.update(&message);
-        state.finalize(&mut output);
-        SignableEthBytes::as_signable(&output)
+    fn signable_keccak256(&self) -> KeccakHash {
```
james-chf, line 104: way nicer now we use `KeccakHash` (`[u8; 32]`) now instead of `Vec<u8>`s throughout

james-chf (approved):
lgtm (apart from the signature stuff which is beyond me!)

PR #1123, opened 2023-02-07
`Tx` expiration

grarco:


grarco:
pls update wasm

grarco:
pls update wasm

core/src/types/transaction/wrapper.rs
```diff
@@ -167,9 +167,6 @@ pub mod wrapper_tx {
         pub fee: Fee,
         /// Used to determine an implicit account of the fee payer
         pub pk: common::PublicKey,
-        /// The epoch in which the tx is to be submitted. This determines
-        /// which decryption key will be used
-        pub epoch: Epoch,
```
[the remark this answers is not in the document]
batconjurer: If we remove it, we will have to put it back when the dkg comes online.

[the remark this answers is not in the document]
grarco: Having spoken with @batconjurer the issue is now clear to me. I wasn't aware that with dkg we'll keep also old keys  so that we have the possibility to decrypt transactions coming from previous epochs. This means that we need to keep the `epoch` field since a transaction lifetime is not inherently limited to a single epoch. I'll revert this change

apps/src/lib/cli.rs
```diff
@@ -2923,6 +2926,12 @@ pub mod args {
                     "The maximum amount of gas needed to run transaction",
                 ),
             )
+            .arg(EXPIRATION_OPT.def().about(
+                "The expiration datetime of the transaction, after which the \
+                 tx won't be accepted anymore. All of these examples are \
+                 equivalent:\n2012-12-12T12:12:12Z\n2012-12-12 \
+                 12:12:12Z\n2012-  12-12T12:  12:12Z",
```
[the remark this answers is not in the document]
grarco: Yes in theory we should be able to parse the string even if it contains random spaces

grarco:
pls update wasm

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -129,6 +132,11 @@ pub enum ErrorCodes {
     InvalidOrder = 4,
     ExtraTxs = 5,
     Undecryptable = 6,
+    ReplayTx = 7,
```
batconjurer, line 135: This is unfortunately not documented in main, but the idea of these numbers is that past a certain threshold, we reject the block. In this branch, those are error codes `>=4`. However, we can't reject an entire block if we get the `InvalidDecryptedChainId` error since we have to include all decrypted txs in order. So its number should be lowered below the critical threshold

batconjurer: I think the same is also true of `ExpiredDecryptedTx`

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -141,6 +136,34 @@ where
     }
 }
 
+fn validate_tx_bytes(
```
batconjurer, line 139: Could we change the name of this function to something like `check_expired`? If the jobs it performs in the future increases, we can move back to this vaguer name.

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -21,32 +23,65 @@ where
...
     ) -> ProcessProposal {
-        let tx_results = self.process_txs(&req.txs);
+        let tx_results =
+            self.process_txs(&req.txs, self.get_block_timestamp(req.time));
 
         ProcessProposal {
-            status: if tx_results.iter().any(|res| res.code > 3) {
-                ProposalStatus::Reject as i32
-            } else {
+            status: if tx_results.iter().all(|res| {
```
batconjurer, line 43: Ah here. If the ordering of the errors is fixed, we just need to change the `3` to a `5` in the old code

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -101,51 +144,137 @@ where
...
+                                    }
+
+                                    // Tx expiration
+                                    if let Some(exp) = tx.expiration {
+                                        if block_time > exp {
+                                            return TxResult {
+                                                code: ErrorCodes::ExpiredDecryptedTx
+                                                    .into(),
+                                                info: format!(
+                    "Dercrypted tx expired at {:#?}, block time: {:#?}",
```
batconjurer, line 218: Oh god. What did fmt do here??

batconjurer: Also, `Dercrypted`. :laughing:

batconjurer (changes requested):

apps/src/lib/node/ledger/shell/prepare_proposal.rs
grarco: This function also checks that the transaction is a `Wrapper`, that's why I chose a vague name

apps/src/lib/node/ledger/shell/process_proposal.rs
grarco: I think I'm missing something, what do you mean when you say "in the old code"?

batconjurer:
```            status: if tx_results.iter().any(|res| res.code > 3) {
                ProposalStatus::Reject as i32
            } else {
            ```

apps/src/lib/node/ledger/shell/prepare_proposal.rs
batconjurer: Missed that. Ok then.

apps/src/lib/node/ledger/shell/process_proposal.rs
grarco: Ah, I think I see what you mean. I've changed that code to use the enum variants because I was struggling to keep track of the error codes and thought that using the variants was more readable. Would you prefer to revert to using the codes instead?

batconjurer: We rely on it heaving in `eth-bridge-integration`, so I thinks it's easier for you to make this change than us.

grarco: Actually, in this PR I've also changed `process_proposal` to reject the block in more cases: in particular to reject a block even in case of an `InvalidTx` or `InvalidSig` which we were not doing before. Does this look ok to you? In theory, the block proposer should be able to verify these at block construction time. So in essence the threshold could stay at 3 like before if I reorder the variants correctly in the enum

batconjurer: Yep, sounds good

batconjurer (approved):

PR #1124, opened 2023-02-07
Murisi/masp boundary conditionals

murisi:
Removed redundant condition pertaining to transparent input count for non-MASP source. No longer force transparent outputs to be zero if transaction has non-MASP source. No longer force zero transparent inputs if target is non-MASP. Removed unsafe AssetType extraction.

Transaction inputs are now handled separately/independently from transaction outputs. The benefit of this approach is that the code is more concise whilst covering more cases (shielding, unshielding, shielded, and transparent). The drawback of this approach is that if .. { ... } else { ... }; if .. { ... } else { ... }; rather than an if-else ladder with 4 branches is used.


murisi:
pls update wasm

murisi:
pls update wasm

PR #1131, opened 2023-02-08
e2e/ledger_tests: fix run_ledger_load_state_and_reset in debug build

tzemanovic:
Follow-up to #740 which affected this test in debug build (`NAMADA_E2E_DEBUG=true`) - we have some left over `dbg!` in PoS that isn’t present in the CI as it uses release build and it fill up the process runner buffer. The `dbg!`s will be removed in another PR


brentstone (approved):
LGTM

PR #1138, opened 2023-02-09
ledger: default folder in home directory

Fraccaman:


.changelog/unreleased/improvements/1138-base-directory.md
tzemanovic, line 1: I think it would be good to add more details here as it’s a user facing change, I.e. the new paths per OS. XDG dir is linux only and when it’s not XDG there’s another possible path

PR #1139, opened 2023-02-09
refactor: read_and_confirm_pwd and new_password_prompt moved to the

mateuszjasiuk:
client


cwgoes:
@mateuszjasiuk Do we still need this, and if so should we rebase/target on @juped's SDK branch?

mateuszjasiuk:
@cwgoes 
It's not needed. @murisi already fixed the issue with different approach. I will close it. Sorry.

PR #1141, opened 2023-02-09
Sync write-log and storage prefix iterator ordering

### LINKED ISSUE #1136: write-log prefix iter is not ordered
the write-log prefix iters added in #913 are not ordered by storage keys, which causes can cause issues (discovered in #714) in the WlStorage's iter prefix logic, which combines write-log and storage prefix iters, expecting both of these to be in order of the keys

### LINKED ISSUE #1137: storage keys parsed type doesn't preserve order
The `storage::Key` type should have the same order as in its raw string/bytes form. Specifically, this is important for prefix iterators that in the DB follow the order of raw bytes, but in other places like WriteLog where we use `storage::Key`, the order doesn't match. In #714 we found that this occurs when the storage keys contain addresses, but we should ensure consistency for all the possible values.

Related to #30, I think in long term, the `storage::Key` type would better be represented as raw bytes instead of the parsed key segments in memory and we can apply segments parsing on demand, which besides other advantages would prevent this issue altogether.

brentstone:
Fixes a bug wherein the prefix iterator ordering was different between the write-log and the storage, resulting in erroneous behavior in the `WlStorage`.

Closes #1136.


brentstone:
pls update wasm

PR #1156, opened 2023-02-14
Murisi/alternative txs

murisi:
The objective is to (with as little change as possible)  make Namada transactions signable on hardware constrained wallets by making them smaller. The approach to this taken in this branch is to replace inner transactions (amongst other objects) with their hashes, and then only attach the encrypted payloads at the top level.

More precisely, the following changes have been made:
* Removed the `inner_tx` field of `WrapperTx` leaving behind `tx_hash`
* Added an `inner_tx` field to `Tx` representing an attached encrypted inner transaction
* The ledger now decrypts and checks the attached `inner_tx` field against the `Wrapper::tx_hash` field

The implications of these changes are:
* No encryption is required for/before the signing process (and hence on hardware wallets) because the signature for a `WrapperTx` is over the hash of an unencrypted `Tx`


murisi:
pls update wasm

murisi:
pls restart ci

PR #1157, opened 2023-02-14
Relay validator set updates

sug0:
Based on #1159.
Implements item 6) of #242.

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -1912,6 +1912,10 @@ pub mod args {
     const ETH_GAS: ArgOpt<u64> = arg_opt("eth-gas");
     const ETH_GAS_PRICE: ArgOpt<u64> = arg_opt("eth-gas-price");
     const ETH_ADDRESS: Arg<EthAddress> = arg("ethereum-address");
+    const ETH_RPC_ENDPOINT: ArgDefault<String> = arg_default(
+        "eth-rpc-endpoint",
+        DefaultFn(|| "http://localhost:8545".into()),
+    );
     const FEE_AMOUNT: ArgDefault<token::Amount> =
         arg_default("fee-amount", DefaultFn(|| token::Amount::from(0)));
     const FEE_PAYER: Arg<WalletAddress> = arg("fee-payer");
@@ -2226,6 +2230,8 @@ pub mod args {
     pub struct ValidatorSetUpdateRelay {
         /// The query parameters.
         pub query: Query,
+        /// The Ethereum RPC endpoint.
+        pub eth_rpc_endpoint: String,
         /// The epoch of the validator set to relay.
         pub epoch: Option<Epoch>,
         /// The Ethereum gas that can be spent during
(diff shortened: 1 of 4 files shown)


apps/src/lib/client/eth_bridge/validator_set.rs
```diff
@@ -44,3 +51,80 @@ pub async fn query_validator_set_args(args: args::ActiveValidatorSet) {
...
+            encoded_validator_set_args_fut,
+            governance_address_fut
+        )
+        .unwrap();
+
+    let (bridge_hash, gov_hash, signatures): (
+        [u8; 32],
+        [u8; 32],
+        Vec<Signature>,
+    ) = abi_decode_struct(encoded_proof);
```
batconjurer, line 88: I seems like we are doing lots of unnecessary rounds of serialization.

apps/src/lib/client/eth_bridge/validator_set.rs
```diff
@@ -44,3 +51,80 @@ pub async fn query_validator_set_args(args: args::ActiveValidatorSet) {
...
+    let pending_tx = relay_op.send().await.unwrap();
+    let transf_result = pending_tx
+        .confirmations(args.confirmations as usize)
+        .await
+        .unwrap();
+
+    println!("{transf_result:?}");
+}
+
+// NOTE: there's a bug (or feature?!) in ethers, where
```
batconjurer, line 119: Probably a feature :laughing:

apps/src/lib/cli.rs
```diff
@@ -2192,6 +2227,66 @@ pub mod args {
         }
     }
 
+    #[derive(Debug, Clone)]
+    pub struct ValidatorSetUpdateRelay {
```
batconjurer, line 2231: How does Ethereum know which address to charge the gas from?

sug0: in hardhat, it always selected an account configured by default, but you are right, there should be an option to change the default wallet to another addr

apps/src/lib/client/eth_bridge/validator_set.rs
sug0: the RPC method could return a non-abi encoded data response. right now it borsh serializes the abi encoding of the data. CLI commands requesting a proof are kind of obsolete now, but if we wanted to keep them anyway (so users could potentially implement shell scripts around this or smth?), they could be responsible for returning the ABI encoding of the data queried from our RPC endpoints

batconjurer: I think it makes more sense to borsh serialize the raw struct and down the road, if some script needs abi encoding, it can call abi encoding on the raw struct itself. I don't think that's an unreasonable burden to put on people implementing their own scripts. 

I just feel like we have a lot of artifact code that I would like to clear out. This damn repo is complicated enough as it is.

sug0: I feel the same. I'll work on this later though, right now it's out of the scope of this PR

batconjurer (approved):

Issue #1061, opened 2023-01-20
Start up an Ethereum node in CI for use in e2e tests


james-chf:
We want a semi-realistic Ethereum node implementation to work with in e2e tests. We should be able to spin it up and deploy the bridge smart contracts to it for an e2e test, then tear it down.

This doesn't necessarily have to be `geth`, we can probably use something lightweight and more development-friendly like [`npx hardhat node`](https://hardhat.org/). We can use the same version of hardhat node that is used by the tests in https://github.com/anoma/ethereum-bridge.

PR #1171, opened 2023-02-21
Roundtrip Bridge pool test

batconjurer:
This fills out an e2e test that now does the following:
 * Adds a transfer to the pool
 * Queries that it is added correctly
 * Constructs a proof to be relayed
 * Submits and ethereum event indicating successful relaying
 * Checks that the transfer is removed from the bridge pool

### DIFF
--- a/tests/src/e2e/eth_bridge_tests.rs
+++ b/tests/src/e2e/eth_bridge_tests.rs
@@ -1,6 +1,7 @@
 mod helpers;
 
 use std::num::NonZeroU64;
+use std::str::FromStr;
 
 use color_eyre::eyre::Result;
 use namada::eth_bridge::oracle;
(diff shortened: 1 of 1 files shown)


tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -193,7 +199,14 @@ fn test_add_to_bridge_pool() {
         &Who::Validator(0),
         ethereum_bridge::ledger::Mode::EventsEndpoint,
     );
-
+    // TODO(namada#1061): need to start up a fake Ethereum node here for the
+    // oracle to connect to, to avoid errors in the ledger logs
```
sug0, line 203:
```suggestion
```
remove these lines and close https://github.com/anoma/namada/issues/1061

sug0 (changes requested):

sug0 (approved):

PR #1176, opened 2023-02-22
Removed managed Ethereum fullnode

sug0:


apps/src/lib/node/ledger/ethereum_node/mod.rs
```diff
@@ -4,7 +4,6 @@ pub mod test_tools;
 use std::ffi::OsString;
 
 use thiserror::Error;
-use tokio::sync::oneshot::{Receiver, Sender};
```
batconjurer, line 7: I would like to get rid of this file and check the whole module to be `oracle`

batconjurer: You can get rid of this error type completely.

apps/src/lib/node/ledger/ethereum_node/test_tools/mod.rs
```diff
@@ -131,7 +131,7 @@ pub mod mock_web3_client {
             _: Option<Uint256>,
             _: impl Debug,
             mut events: Vec<&str>,
-        ) -> Result<Vec<Log>> {
+        ) -> Result<Vec<Log>, Error> {
```
batconjurer, line 134: We can use a bespoke error here.

batconjurer (changes requested):

batconjurer (approved):

PR #1182, opened 2023-02-23
[feat]: Dont' persist storage changes at genesis

batconjurer:
Persisting storage at genesis (init chain) can cause database corruption if a node restarts before reaching blockheight 1. The persistence of state should only occur at the time that tendermint is informed of the next block height. Thus we remove the db write call in init chain and defer persisting until the first commit block call.

### DIFF
--- a/apps/src/lib/node/ledger/shell/init_chain.rs
+++ b/apps/src/lib/node/ledger/shell/init_chain.rs
@@ -360,10 +360,6 @@ where
             response.validators.push(abci_validator);
         }
 
-        self.wl_storage
-            .commit_genesis()
-            .expect("Must be able to commit genesis state");
-
         Ok(response)
     }
 }


batconjurer:
I'm not sure why this fixes the db corruption. There are other storage writes in `init_chain` that seems like they would be persisted if the ledger were shut down prior to the first `commit_block`. I'm basing this on the fact that flush is called on rocksdb on drop. However, I don't observe this locally.

batconjurer:
> I'm not sure why this fixes the db corruption. There are other storage writes in `init_chain` that seems like they would be persisted if the ledger were shut down prior to the first `commit_block`. I'm basing this on the fact that flush is called on rocksdb on drop. However, I don't observe this locally.

Ok, I thing I've resolved my confusion. The point is that we are no longer committing a block, so the flushed data in the sst files gets ignored.

PR #1184, opened 2023-02-23
Updates RocksDB dump

grarco:
Improves the `ledger dump-db` command. RocksDB iterator doesn't actually recognize the provided prefix correctly leading to duplicated keys in the dump file.

This PR solves the duplicated keys issue and adds an `historic` flag: if the flag is set it will dump the `subspace` keys and the diff of the last height, otherwise it will only dump the `subspace` subways.

### DIFF
--- a/apps/src/lib/node/ledger/storage/rocksdb.rs
+++ b/apps/src/lib/node/ledger/storage/rocksdb.rs
@@ -242,10 +242,14 @@ impl RocksDB {
     }
 
     /// Dump last known block
-    pub fn dump_last_block(&self, out_file_path: std::path::PathBuf) {
+    pub fn dump_last_block(
+        &self,
+        out_file_path: std::path::PathBuf,
+        historic: bool,
+    ) {
         use std::io::Write;
 
-        // Fine the last block height
+        // Find the last block height
         let height: BlockHeight = types::decode(
             self.0
                 .get("height")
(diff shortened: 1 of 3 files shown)


grarco:
I've updated this PR, see description

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -273,32 +281,44 @@ impl RocksDB {
...
 
-        // Dump accounts subspace and block height data
         dump_it("subspace".to_string());
-        let block_prefix = format!("{}/", height.raw());
-        dump_it(block_prefix);
+
+        if extended_dump {
+            // Dump the keys prepended with the selected block height (includes
+            // subspace diff keys)
+            dump_it(height.raw());
```
grarco, line 313: In here I manually specify the prefixes to dump. This is a bit tedious and also not future-proof since we'd need to update this list in case we changed the well-known keys that we write to storage.

An alternative to this could be to iterate over all the keys given the prefix `"a"`: given the lexicographical order that Namada imposes on RocksDB this prefix should include all the keys but the height-prepended one. I didn't choose this solution because of two doubts that I still have:

- What happens if at a certain point the encoded value of a height starts with an `"a"`? I guess we would have duplicated keys in the dump which is what we already suffered from while debugging the testnet
- We'd still need to manage separately the `results` subkey, since this one is **followed** by the height (is there a reason why we do this and not te opposite as for the other keys?)

PR #1187, opened 2023-02-24
Wip Namada rollback

grarco:


apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -302,6 +302,69 @@ impl RocksDB {
...
+        // Revert the non-height-prepended metadata storage keys
+        for metadata_key in [
+            "next_epoch_min_start_height",
+            "next_epoch_min_start_time",
+            "tx_queue",
+        ] {
+            let key = Key::from(metadata_key.to_owned().to_db_key());
+            let previous_key =
+                Key::from(format!("pred/{}", metadata_key).to_db_key());
+            let previous_value = self.read_subspace_val(&previous_key)?.expect(
```
adrianbrink, line 320: Does this keep track of whether the key was changed in the last block; ie we need to make certain that we only revert to the `pred` key if it was actually different on block `n-1`

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -190,6 +190,26 @@ pub fn reset(tendermint_dir: impl AsRef<Path>) -> Result<()> {
     Ok(())
 }
 
+pub fn rollback(tendermint_dir: impl AsRef<Path>) -> Result<()> {
+    let tendermint_path = from_env_or_default()?;
+    let tendermint_dir = tendermint_dir.as_ref().to_string_lossy();
+
+    // Rollback tendermint state
+    std::process::Command::new(tendermint_path)
```
adrianbrink, line 198: Are there docs we can link to that explain how Tendermint rollback behaves?

apps/src/lib/node/ledger/storage/rocksdb.rs
grarco: As far as I understood these three keys are updated with every block

apps/src/lib/node/ledger/tendermint_node.rs
grarco: I wasn't able to find proper documentation for this. We could link [this](https://github.com/tendermint/tendermint/blob/64747b2b184184ecba4f4bffc54ffbcb47cfbcb0/cmd/tendermint/commands/rollback.go) file which contains the help message, or simply include it in the help message of the Namada command

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -158,6 +158,20 @@ pub fn reset(config: config::Ledger) -> Result<()> {
     Ok(())
 }
 
+pub fn rollback(config: config::Ledger) -> Result<()> {
+    // Rollback Tendermint state
+    let tendermint_block_height =
```
grarco, line 163: We could backup the Namada state here before trying the rollback to have a copy in case the procedure should fail and leave the db in an inconsistent state. From what I saw, the tendermint rollback command doesn't do that so we could do the same for the tendermint state

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -158,6 +158,22 @@ pub fn reset(config: config::Ledger) -> Result<()> {
     Ok(())
 }
 
+pub fn rollback(config: config::Ledger) -> Result<()> {
+    // Rollback Tendermint state
+    tracing::info!("Rollback Tendermint state");
```
grarco, line 163: Should we do a backup here before reverting the state in case anything went wrong? The Tendermint rollback command doesn't look to be doing so. We could also just ask the users to do the backup themselves before running this command

apps/src/lib/node/ledger/tendermint_node.rs
adrianbrink: let's include the link to the code at least

PR #892, opened 2022-12-14
Implement cubic slashing

brentstone:
Based on v0.16.0

Closes #14, closes #35, closes #124, closes #789.

Note to consider for review: in the pseudocode (Informal partnership), slashes are processed at the very end of an epoch, while in Namada, they are processed at the very beginning of an epoch. This produces slightly different bounds for iteration in several places.

In order to merge into 0.17.0, the following items have been removed from the PR and will have their own PRs to integrate after:
- fix client queries with slashing (may need to use `get_slashed_amount`)
- proper transferring of slashed tokens to the slash pool address

PR #1196, opened 2023-03-07
add lazy set collection

tzemanovic:
based on 0.14.1

related to #168 (we might also want to add key hashing variants of these later, if we have a use case for them)

### DIFF
--- a/tests/src/storage_api/collections/lazy_set.rs
+++ b/tests/src/storage_api/collections/lazy_set.rs
@@ -85,6 +85,8 @@ mod tests {
         Insert(TestKey),
         /// Remove a key-val from a [`LazySet`]
         Remove(TestKey),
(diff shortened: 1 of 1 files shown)


brentstone:
I like this simple key method you have here that holds `()` values, but do you think it could still be useful to have another lazy set variant that particularly holds values / types that are Borsh-serializable and hashable? A la what is done in https://github.com/anoma/namada/pull/892/commits/7dd6508af148711ac06c64af5a80c46cfb282f69? In this commit, the lazy set is useful for holding types that are not going to implement the `KeySeg` trait, but they can be hashed, and so the set is effectively a map from a `u64` hash value of a Borsh-serializable value to the value itself.

If this is desired, can call it `LazyHashSet` to distinguish with what is already in this PR. Then add to this PR.

PR #1197, opened 2023-03-07
PoS: ensure unique consensus keys

### LINKED ISSUE #1188: PoS: ensure that consensus keys are unique
in the new PoS system (#740) we're not yet checking that given consensus key for init_validator tx is unique, which could be potentially break once passed to tendermint (related issue #106, but checking for uniqueness is a simpler fix)

tzemanovic:
closes #1188 

based on #1196


brentstone (dismissed):
LGTM!

PR #1198, opened 2023-03-07
Handle native inflation rewards for native tokens

murisi:
Fixing the handling of positive Namada token rewards for possession of shielded Namada tokens. The approach is taken is as follows:
* Removed the `prev_root` field of `ConversionState` structure for now since it seems to never be read.
* Conversions from a previous epoch to a new epoch are done directly without reference to the native token in the zeroth epoch. It seems to have been circular references to the native token that was causing positive native inflation rewards to fail E2E tests.
* Native token rewards given for possession of other types of shielded assets are first deflated. This saves the client from having to convert native tokens from multiple epochs into the latest whilst allowing inflation to still be expressed in terms of the latest native token.


murisi:
pls update wasm

PR #1206, opened 2023-03-09
Polish vote extension code

sug0:
* Only start proposing validator set updates an epoch before the bridge is enabled through governance, or while the bridge is active.
* Reject all current vote extension protocol transactions from the mempool, if the bridge is not active.
* Change `tracing::error!()` log lines to `tracing::debug!()` log lines, so node logs aren't spammed with validation logic.
* Remove the `index-set` crate from `apps`, as that was only required by the remaining txs allocator (now defunct).
* Convert PrepareProposal into CheckTx mempool tests.

### DIFF
--- a/apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
+++ b/apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
@@ -71,7 +71,7 @@ where
             );
             return Err(VoteExtensionError::UnexpectedBlockHeight);
         }
-        if last_height.0 == 0 {
+        if ext.data.block_height.0 == 0 {
             tracing::debug!("Dropping vote extension issued at genesis");
             return Err(VoteExtensionError::UnexpectedBlockHeight);
         }
--- a/apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
+++ b/apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
@@ -53,7 +53,7 @@ where
     > {
         #[cfg(feature = "abcipp")]
         if ext.data.block_height != last_height {
-            tracing::error!(
+            tracing::debug!(
                 ext_height = ?ext.data.block_height,
                 ?last_height,
                 "Bridge pool root's vote extension issued for a block height \
@@ -63,7 +63,7 @@ where
         }
         #[cfg(not(feature = "abcipp"))]
         if ext.data.block_height > last_height {
-            tracing::error!(
+            tracing::debug!(
                 ext_height = ?ext.data.block_height,
                 ?last_height,
                 "Bridge pool root's vote extension issued for a block height \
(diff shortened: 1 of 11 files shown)


apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
```diff
@@ -88,20 +85,35 @@ where
...
-                tracing::error!(
+                tracing::debug!(
                     block_height = ?ext.data.block_height,
                     "The epoch of the Bridge pool root's vote extension's \
                      block height should always be known",
                 );
                 return Err(VoteExtensionError::UnexpectedEpoch);
             }
         };
+        if !self
```
batconjurer, line 96: This debug line should come first. It would be very strange to get one of the above errors if we expect the bridge to be inactive.

apps/src/lib/node/ledger/shell/vote_extensions/eth_events.rs
```diff
@@ -66,18 +66,47 @@ where
...
+                     block height should always be known",
+                );
+                return Err(VoteExtensionError::UnexpectedEpoch);
+            }
+        };
+        if !self
+            .wl_storage
+            .ethbridge_queries()
+            .is_bridge_active_at(ext_height_epoch)
+        {
```
batconjurer, line 102: Same here. If the bridge is off, best make sure this is the line we see.

apps/src/lib/node/ledger/shell/vote_extensions/val_set_update.rs
```diff
@@ -60,21 +60,34 @@ where
...
-            tracing::error!(
+            tracing::debug!(
                 vext_epoch = ?ext.data.signing_epoch,
                 expected_epoch = ?signing_epoch,
                 "Validator set update vote extension issued for an epoch \
                  different from the expected one.",
             );
             return Err(VoteExtensionError::UnexpectedEpoch);
         }
+        if !self
```
batconjurer, line 78: same here

batconjurer (changes requested):
Small change

apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
sug0: I kept the block height query before the epoch and bridge queries on purpose. this is because it's very cheap to read a value from memory and compare it against another value in memory. checking the epoch doesn't need to read a value from storage either, but it's slightly more expensive, because it takes linear time, based on the number of epochs we've gone through. checking the status of the bridge does read a value from storage, so it would be easy to abuse it to send invalid protocol txs.

idk, I guess we need to read values from storage anyway, to confirm who signed the protocol tx, etc. what do you think?

batconjurer: I'm going to err on the side of better (or more illuminating and expected) error messages. These kinds of small optimizations are not as important. Blockchains aren't fast.

batconjurer (approved):

Issue #1208, opened 2023-03-10
Validator set update relayer daemon


sug0:
The Namada relayer should be able to be started as a background process, to continuously update the validator set in the Ethereum state machine. The basic operation of the daemon would be to watch epoch changes, at which point the validator set update proof would be queried and relayed to Ethereum.

PR #1209, opened 2023-03-10
Synchronize with Ethereum before performing relays

sug0:
Required by https://github.com/anoma/namada/issues/1208

### DIFF
--- a/apps/src/lib/client/eth_bridge.rs
+++ b/apps/src/lib/client/eth_bridge.rs
@@ -1,2 +1,30 @@
 pub mod bridge_pool;
 pub mod validator_set;
+
+use std::ops::ControlFlow;
+
+use tokio::time::{Duration, Instant};
+use web30::client::Web3;
+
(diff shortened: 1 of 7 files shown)


apps/src/lib/client/eth_bridge.rs
```diff
@@ -1,2 +1,30 @@
...
+use std::ops::ControlFlow;
+
+use tokio::time::{Duration, Instant};
+use web30::client::Web3;
+
+use crate::node::ledger::ethereum_oracle::eth_syncing_status;
+use crate::timeouts::TimeoutStrategy;
+
+/// Block until Ethereum finishes synchronizing.
+pub async fn block_on_eth_sync(deadline: Instant, url: &str) {
```
batconjurer, line 13: Is this really the strategy we want if ethereum is not synced? If it is far behind, this method will take a long to time to resolve, hijacking the cli, forcing the user to kill it. Maybe it is just better to stop the operation with a long and maybe some helpful info about when to try again.

batconjurer (approved):

Issue #1080, opened 2023-01-24
Check whether the Ethereum oracle should be reconfigured after every epoch transition


james-chf:
After PR https://github.com/anoma/namada/pull/1079, the Ethereum event oracle will effectively only be configured once when a validator node starts up. It a parameter affecting the oracle is changed on-chain (at the moment, this is just the `min_confirmations` parameter) e.g. via a governance proposal, any running Ethereum oracle won't automatically be reconfigured - validators will need to restart their node.

If a validator is running an Ethereum oracle, they should check whether it needs reconfiguring after every epoch transition (e.g. a governance proposal may have passed to change a relevant parameter like `min_confirmations`), and do so.

kunimei:
good

sug0:
Closed by https://github.com/anoma/namada/pull/1210

PR #1210, opened 2023-03-11
Once an epoch, the ethereum oracle configuration gets updated

batconjurer:
This PR resolves issue #1080.

 * Oracle storage has been moved under the control of the `Parameters` vp. This allows it to be changed by storage.
 * Once an epoch, the shell sends the config in storage to the oracle.
 * Since the oracle has no notion of epochs, after processing every ethereum blocks, checks if a new config has been sent.

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -402,6 +402,9 @@ where
 
         if new_epoch {
             self.update_epoch(&mut response);
+            // send the latest oracle configs. These may have changed due to
+            // governance.
+            self.update_eth_oracle();
         }
 
         let _ = self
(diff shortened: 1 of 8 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -497,6 +497,9 @@ where
                 },
             )
             .expect("Must be able to update validator sets");
+        // send the latest oracle configs. These may have changed due to
+        // governance.
+        self.update_eth_oracle();
```
sug0, line 502: I would add this under `if new_epoch {` (in `finalize_block()`). `update_epoch()`, as the docstring suggests, is only for validator set and consensus param changes

sug0 (approved):

PR #1216, opened 2023-03-13
Implement a validator set update relayer daemon

sug0:
https://github.com/anoma/namada/issues/1208


sug0:
**TODO:** fix crashes in the daemon when a proof isn't available yet, for a given epoch

batconjurer (approved):

PR #1212, opened 2023-03-13
lazy collections with `Address`

tzemanovic:
based on #1256  ~~#1196~~

fixes #1211

PR #1218, opened 2023-03-14
fix nested lazy vec iter

### LINKED ISSUE #1213: prefix iter with a predicate
The `LazyVec` collection has a sub-key for length of the vec (`len`) and a sub-key for the elements (`data`), but because they start with the same storage prefix, we cannot prefx-iterate the data without also matching the length key which adds an extra unwanted iteration. This iteration fails on decoding unless the data has the same type as the length (`u64`) which is arguably even worse, because the length is not one of the data elements.

This is a more general issue and is likely it will come up again. I think we want some way to skip prefix iterations based on some predicate (in here, skipping the length key)

tzemanovic:
fixes #1213

based on #1212 

This PR:
- Adds a test that replicates the issue for LazyVec nested inside another collection (only LazyMap currently allows nesting) 
- Adds a new storage_api function `iter_prefix_with_filter`. This is preferable over the regular `iter_prefix` combined with the iterator's `filter` function as it avoids trying to decode values that don't pass the filter. For `iter_prefix_bytes`, filter works fine
- Adds a new method `is_data_sub_key` to `trait LazyCollection`
- Fixes the `fn iter` for a `LazyMap` with nested collection using the `iter_prefix_with_filter` with `is_data_sub_key` as a predicate

### DIFF
--- a/.changelog/unreleased/bug-fixes/1218-nested-lazy-vec-iter.md
+++ b/.changelog/unreleased/bug-fixes/1218-nested-lazy-vec-iter.md
@@ -1,0 +1,3 @@
+- Fixed an issue with the iterator of LazyMap with a nested LazyVec collection
+  that would match non-data keys and fail to decode those with the data decoder.
+  ([#1218](https://github.com/anoma/namada/pull/1218))


brentstone (dismissed):
LGTM! Thanks

brentstone (approved):
LGTM!

.changelog/v0.15.0/bug-fixes/1218-nested-lazy-vec-iter.md
```diff
@@ -0,0 +1,3 @@
+- Fixed an issue with the iterator of LazyMap with a nested LazyVec collection
```
tzemanovic, line 1: woops, this is supposed to be in unreleased

Issue #1090, opened 2023-01-26
Formatting token amounts should support placing the decimal point in different places


james-chf:
When displaying/interacting with token amounts in the CLI, we work with the token amount in either base units (10^1) or micros (10^-6). 

For some tokens, it makes sense to expect or display the decimal point in a different place e.g. for wrapped ERC20s, we will want to match wherever the original ERC20 puts the decimal point. This is different depending on the ERC20 (and some ERC20s do not use a decimal point at all).

For each token type (including multitoken tokens), we should store details of where a decimal point should be displayed (this could be on-chain, under the token's storage subspace, so that it could be easily queried).

From the UX side, CLI commands to do with token amounts like `namadac balance`, `namadac transfer` would need to be updated.

sug0:
Closed by https://github.com/anoma/namada/pull/1282

Issue #1089, opened 2023-01-26
`token::Amount` may not be able to fully represent some wrapped token amounts


james-chf:
On Namada, we have an `Amount` type which uses units of `micros` (10^-6).
https://github.com/anoma/namada/blob/76169dcdaf15911ac2f69dec78604e18c4c5b259/core/src/types/token.rs#L32-L34

Since we use `u64` type for `micros`, the is maximum limit for how much a single `Amount` can hold is something like `u64::MAX * 10^-6` which is roughly ~18.5bn full units. We use `Amount` type to represent amounts of native tokens on Namada as well as wrapped tokens from other chains. On Ethereum, ERC20 tokens can have supplies much larger than ~18.5bn e.g. [SHIB](https://etherscan.io/token/0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce). We want to be able to represent any wrapped token amounts in our `Amount` type. 

Multiple possible solutions including changing our `Amount` type to support a higher supply (e.g. [num256::Uint256](https://docs.rs/num256/0.1.3/num256/uint256/struct.Uint256.html) instead of u64), or using a different `Amount` type entirely for wrapped tokens.

james-chf:
Have tagged this issue https://github.com/anoma/namada/labels/IBC as well as https://github.com/anoma/namada/labels/ethereum-bridge as I presume there are types of tokens that could come over IBC, that could be affected by this issue as well

sug0:
Closed by https://github.com/anoma/namada/pull/1282

PR #1233, opened 2023-03-20
Added erc20 token amount type

### LINKED ISSUE #1085: Ensure ERC20 base units map to `token::Amount` type correctly
In the smart contracts, ERC20 amounts over the Ethereum bridge are represented in ERC20 base units as  `uint256`.

```solidity
    struct NamadaTransfer {
        address from;
        uint256 amount;
        string to;
    }

    struct ERC20Transfer {
        address from;
        address to;
        uint256 amount;
        string feeFrom;
        uint256 fee;
    }
```

On Namada, we have an `Amount` type which uses units of `micros` (10^-6).
https://github.com/anoma/namada/blob/76169dcdaf15911ac2f69dec78604e18c4c5b259/core/src/types/token.rs#L32-L34

Currently, as of `eth-bridge-integration` commit aeb3a4979390a9dd3251a0383ee769179e3329c6, 1 ERC20 unit translates to 1 micro.

We need to ensure: ERC20 base units are correctly translated to/from `Amount` type, in the sense that it is explicitly specified how 1 ERC20 base unit should translate to `Amount`. e.g. it should perhaps translate to 10^6 `micros` rather than 1 mico. Since we use `u64` type for `micros`, there is also maximum limit to how much should be sent in a transfer (something like `u64::MAX * 10^-6` which is roughly ~18.5bn) - though such transfers shouldn't be that large in practice (e.g. due to bridge token whitelisting/supply caps), we should make sure if a transfer of such an amount was attempted it is handled robustly.

batconjurer:
This PR closes #1085, #1089, and #1090.

A new type, `Erc20Amount` has been added for amounts of ERC20 tokens according to the [EIP-20 specification](https://eips.ethereum.org/EIPS/eip-20). Whitelisted ERC20 tokens must now keep a denomination in storage. These denominations must also be kept in sync with the oracle, which needs them for processing ethereum events.


ethereum_bridge/src/storage/eth_bridge_queries.rs
```diff
@@ -142,6 +144,31 @@ where
...
+    /// Get the number of decimal places of a whitelisted ERC20.
+    pub fn get_erc20_denomination(
+        self,
+        token_address: &EthAddress,
+    ) -> Option<u8> {
+        if let Some(addr) =
+            self.wl_storage.ethbridge_queries().get_wnam_address()
+        {
+            if addr == *token_address {
+                return Some(6);
```
sug0, line 162: nit: maybe implement a const fn `token::Amount::denomination() -> u8 { 6 }`?

shared/src/ledger/native_vp/ethereum_bridge/bridge_pool_vp.rs
```diff
@@ -147,33 +157,54 @@ where
...
                     },
                 }
             } else {
+                let escrow_amount = if native {
+                    EscrowAmount::Nam(
+                        Amount::try_from(transfer.transfer.amount).map_err(
+                            |e| {
+                                eyre!(
+                                    "Could not process the amount of NAM to \
+                                     be escrowed into the bridge pool: {:?}",
```
sug0, line 181: nit: assuming this implements `Error`, it's best to display the actual error string
```suggestion
                                     be escrowed into the bridge pool: {}",
```

sug0 (changes requested):
might need another review pass, since token whitelists need some changes to allow transfers to ethereum, even if a whitelist update removes a token whose balance is still greater than 0 for some namada client

apps/src/lib/cli.rs
```diff
@@ -2713,6 +2716,48 @@ pub mod args {
...
+        fn parse(matches: &ArgMatches) -> Self {
+            let erc20_amount = TOKEN_AMOUNT_OPT.parse(matches);
+            let amount = AMOUNT_OPT.parse(matches);
+            if let Some(amt) = erc20_amount {
+                Self::Erc20(amt)
+            } else if let Some(amt) = amount {
+                Self::Native(amt)
+            } else {
+                panic!("No amount provided")
+            }
```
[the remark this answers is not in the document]
sug0: https://docs.rs/clap/latest/clap/error/struct.Error.html
maybe we could also use this API instead of panicking

batconjurer:
> might need another review pass, since token whitelists need some changes to allow transfers to ethereum, even if a whitelist update removes a token whose balance is still greater than 0 for some namada client

Yes, but that will be another PR.

sug0:
This PR contains an implementation that closes https://github.com/anoma/namada/issues/1290

sug0:
Superseded by https://github.com/anoma/namada/pull/1718 and https://github.com/anoma/namada/pull/1282

PR #1238, opened 2023-03-20
SDK changes to help the interface

mateuszjasiuk:
Added `multicore` feature flag to the `shared` and `core`.
- So we can submit shielded transfers in the browser using one thread

Changed `ShieldedUtils` trait to be async.
- because load/save actions in JS are async

Before merging this, it would be nice to change `masp_proofs`, to be able to turn off the `multicore` in `librustzcash`.
Something like this:
```
-multicore = ["bellman/multicore"]
+multicore = ["bellman/multicore, zcash_proofs/multicore"]
```
With those changes shielded transfers should work in browser.


cwgoes:
What changes does this make, and do we still need them? If so, can you rebase on @juped's SDK branch?

mateuszjasiuk:
> What changes does this make, and do we still need them? If so, can you rebase on @juped's SDK branch?

So these changes are needed for shielded transfers to work. I would rather add them after Ray merges Sdk branch to main. The reasons are:
- namada-interface main targets the namada branch that Ray is merging
- in this PR namada points to local versions of masp_proofs and masp_primitives which also have few changes

I think the best approach would be for Ray to finish the merge, then change masp_primitives and masp_proofs(I think change there is only with feature flags, but need to double check), and then merge those changes.

tzemanovic (approved):

PR #1241, opened 2023-03-21
Increment the bridge pool's batch nonce

sug0:
If we detect that a transfer to ethereum event was newly confirmed, we increment the bridge pool's batch nonce.


sug0:
writing a test to make sure the bp nonce gets updated

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -355,6 +359,15 @@ where
                                 );
                             }
                         }
+                        // should we update the bridge pool nonce?
+                        if !increment_bp_nonce
```
batconjurer, line 363: This isn't quite right, it will increase the nonce too frequently. We should increase the nonce every time a `TransferToEthereum` event gets enough votes to be confirmed.

batconjurer (changes requested):

sug0:
Closed in favor of an upcoming PR

PR #1242, opened 2023-03-22
Whitelist gas accounting

grarco:


apps/src/lib/client/rpc.rs
```diff
@@ -1892,23 +1882,26 @@ pub async fn query_conversions(ctx: Context, args: args::QueryConversions) {
 }
 
 /// Query a conversion.
-pub async fn query_conversion(
-    client: HttpClient,
+pub async fn query_conversion<CLIENT>(
```
grarco, line 1885: I needed these changes for benchmarks since I wanted to reuse some client functions of masp (specifically to construct shielded txs) but without a real client nor a running ledger. I suspect the SDK guys are doing something similar to this but I haven't has the time to check

grarco:
pls update wasm

grarco:
pls update wasm

grarco:
pls update wasm
ù

apps/src/lib/cli.rs
```diff
@@ -1621,14 +1621,15 @@ pub mod args {
...
-    const GAS_AMOUNT: ArgDefault<token::Amount> =
-        arg_default("gas-amount", DefaultFn(|| token::Amount::from(0)));
-    const GAS_LIMIT: ArgDefault<token::Amount> =
-        arg_default("gas-limit", DefaultFn(|| token::Amount::from(0)));
-    const GAS_TOKEN: ArgDefaultFromCtx<WalletAddress> =
-        arg_default_from_ctx("gas-token", DefaultFn(|| "NAM".into()));
+    const FEE_AMOUNT: ArgDefault<token::Amount> =
+        arg_default("fee-amount", DefaultFn(|| token::Amount::from(0)));
+    const GAS_LIMIT: ArgDefault<GasLimit> =
+        arg_default("gas-limit", DefaultFn(|| GasLimit::from(10)));
```
grarco, line 1630: This default for the `GasLimit` doesn't really have much sense given how different the costs for different txs will be and the lack of a refunding mechanism (at the moment). It's there to ease the tx submission process in testnets but it will probably be better to remove it in future PRs or even in this one

grarco:
pls update wasm

grarco:
pls update wasm

apps/src/lib/client/rpc.rs
tzemanovic: yeah, it's exactly like that

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -707,6 +707,30 @@ where
...
+            // Tx gas limit
+            let mut gas_meter = TxGasMeter::new(u64::from(&wrapper.gas_limit));
+            if gas_meter.add_tx_size_gas(tx_bytes.len()).is_err() {
+                response.code = ErrorCodes::TxGasLimit.into();
+                response.log =
+                    "Wrapper transactions exceeds its gas limit".to_string();
+                return response;
+            }
+
+            // Max block gas
```
tzemanovic, line 719: I think in the mempool we shouldn't check max block gas as that would prevent queuing up more txs than what fits into a single block

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic, line 1: I think in here we should also add the `BlockGasMeter` to verify that a proposal didn't exceed it

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -12,7 +16,7 @@ use namada::types::transaction::{AffineCurve, DecryptedTx, EllipticCurve};
 
 use super::super::*;
 #[allow(unused_imports)]
-use super::block_space_alloc;
+pub use super::block_space_alloc;
```
tzemanovic, line 19: why `pub use` here? It is pub in the super mod

wasm/checksums.json
```diff
@@ -1,20 +1,74 @@
...
-    "tx_withdraw.wasm": "tx_withdraw.07fbd5e324fc93370c356db9c6f7eff13fc2886559370ba2dc5c713e10516918.wasm",
-    "vp_implicit.wasm": "vp_implicit.f25c93a729d23562c3a9bd79344c03199981e209e3756fd0cf72da9906b541a4.wasm",
-    "vp_masp.wasm": "vp_masp.b34240e9c3ee0914d68f4669b3ebf5eb55447b0a499954fa6cfcf8a1396df99e.wasm",
-    "vp_testnet_faucet.wasm": "vp_testnet_faucet.497a191e0ef340a723c8b01e6c051a82979a7dcae51842f8ef8e945d5903346c.wasm",
-    "vp_token.wasm": "vp_token.d556e61b8a01b8aae5fb0628ba6a598c97008c19e3cc3d8ca0c514b9c1c85ed9.wasm",
-    "vp_user.wasm": "vp_user.f18423c4e47838a9d2e020196faf37deca6d6ebc5348d595ff3466df81b5879a.wasm",
-    "vp_validator.wasm": "vp_validator.08ab66b0b6bb12a9a56a22db196ff49bc04282d73401902c02c50c63a4fc08cc.wasm"
+    "tx_bond.wasm": {
+        "gas": "160",
+        "hash": "tx_bond.8d0afeab4dc91d9294e3c1e1b5f3cd4f1e13f19e1631b894ccce4a4768d991da.wasm"
```
tzemanovic, line 4: it is cosmetic, but since we're changing the format here, could we make these fields to only contain the hash without the file name?

apps/src/lib/node/ledger/shell/mod.rs
grarco: In here I actually just check that the single tx in exam doesn't exceed the gas allowed for the entire block by itself. `prepare_proposal` will then construct the block computing the cumulated gas required by the txs that are being included

apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Actually the check on the entire block gas limit is done in `process_proposal` based on the declared `gas_limit`of the wrapper txs. Here we simply check that each tx doesn't exceed its own limit and this guarantees that the block gas limit is respected.

More specifically, regarding the check in `process_proposal`: in ABCI++ the block is rejected in case any tx exceeds the block gas limit (either because of cumulated gas coming from previous txs or because of itself) while in ABCI the signer tx(s) will be discarded.

apps/src/lib/node/ledger/shell/prepare_proposal.rs
grarco: Without the `pub` here I could not build the docs because of some unresolved links

apps/src/lib/node/ledger/shell/mod.rs
tzemanovic: I see, thx!

apps/src/lib/node/ledger/shell/prepare_proposal.rs
tzemanovic: ah I see - I think it would be better to change the docs to use `super::block_space_alloc` and remove the unused import here and in ` apps/src/lib/node/ledger/shell/block_space_alloc/states.rs` remove the `prepare_proposal::` from docs that refer to `block_space_alloc`

grarco:
pls update wasm

grarco:
pls update wasm

grarco:
pls update wasm

apps/src/lib/cli.rs
grarco: At the moment I've added the possibility to dry-run the entire tx (wrapper + inner) to estimate the gas cost manually. Were you thinking about something that doesn't require an action from the user? Like the client automatically doing the estimation right before submitting a tx?

cwgoes: Yes, the latter - at least this should be possible as a default, which can be turned off by a flag manually specifying gas.

wasm/gas.json
```diff
@@ -0,0 +1,20 @@
+{
```
cwgoes, line 1: How are we generating this file, exactly?

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -809,6 +830,53 @@ where
...
+        let balance: token::Amount = self
+            .wl_storage
+            .read(&balance_key)
+            .expect("must be able to read")
+            .unwrap_or_default();
+
+        let wrapper_fees = self.get_wrapper_tx_fees();
+        match balance.checked_sub(wrapper_fees) {
+            Some(amount) => {
+                self.wl_storage
```
cwgoes, line 852: Do we have a more... abstract version of this function, perhaps?

I feel like we often do logic with raw keys that should be abstracted.

core/src/ledger/gas.rs
```diff
@@ -15,34 +13,75 @@ pub enum Error {
...
 
-/// Gas metering in a block. Tracks the gas in a current block and a current
-/// transaction.
-#[derive(Debug, Default, Clone)]
+/// Trait to share gas operations for transactions and validity predicates
+pub trait TxVpGasMetering {
+    /// Add gas cost. It will return error when the
+    /// consumed gas exceeds the provided transaction gas limit, but the state
+    /// will still be updated
+    fn add(&mut self, gas: u64) -> Result<()>;
```
cwgoes, line 41: Might be clearer to call this "consume" ?

cwgoes (commented):
Mostly looks fine, a few comments.

wasm/gas.json
grarco: At the moment manually. We first run the benchmarks, then we go through the html report produced and compute the gas cost based on the average execution time. The conversion should be something like 1ms~10 gas units. I didn't feel like automating this because we only need to run the benchmarks from time to time

apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Yes, I've improved this in #1327

Issue #1247, opened 2023-03-23
PoS double signing e2e test - failed client balance check


tzemanovic:
There's a mysterious failure in `e2e::ledger_tests::double_signing_gets_slashed` that's hard to reproduce (I've only seen it in CI so far, but couldn't reproduce locally yet), in which the client cmd to submit a transfer tx fails on a check of the source balance. It doesn't appear to be a timing issue - I tried to delay the tx until after first block is committed and it still occurred.

Examples:
- https://github.com/anoma/namada/actions/runs/4497974699/jobs/7915031102 - failed again on retry
- https://github.com/anoma/namada/actions/runs/4477468042/jobs/7869034687 - failed again after adding a commit to try to wait for a first block
- https://github.com/anoma/namada/actions/runs/4477925567/jobs/7870059979 - from the same branch as above, passed later on on some new commits
- https://github.com/anoma/namada/actions/runs/4488665807/jobs/7893547551
- https://github.com/anoma/namada/actions/runs/4509858928/jobs/7940108986

PR #1253, opened 2023-03-27
test/e2e/double_singing: wait for first block

tzemanovic:
fixes https://github.com/anoma/namada/issues/1247

The test ensures that the node is ready by waiting for a first block before proceeding with a tx. This can be improved with https://github.com/anoma/namada/issues/94 by having client automatically handle this.

based on https://github.com/anoma/namada/pull/1246


brentstone (approved):
LGTM

brentstone:
#1258 doesn't have the change with where we call `self.slash()` for the `Shell`, which was present in this PR, though

Issue #1257, opened 2023-03-27
Handle invalid transfer to Ethereum events


sug0:
The ledger is currently processing each confirmed Ethereum event it receives in FinalizeBlock as if it were valid. Naturally, some transfers are subject to failure. We should handle such transfers, by removing them from the bridge pool and not refunding gas costs. Assets themselves need to be refunded.

sug0:
(**EDIT:** removed the gas cost refund requirements from the issue)

PR #1260, opened 2023-03-28
Handle invalid transfer to Ethereum events

sug0:
Addresses #1257

### DIFF
--- a/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
+++ b/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
@@ -252,9 +252,10 @@ where
             } else {
                 tracing::debug!(
                     ?pending_transfer,
-                    "Invalid transfer to Ethereum detected, but the relayer \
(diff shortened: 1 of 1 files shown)


ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -229,10 +237,25 @@ where
...
     let relayer_rewards_key = balance_key(&nam(), relayer);
     // Remove the completed transfers from the bridge pool
-    for event in transfers {
+    for (event, is_valid) in
+        transfers.iter().zip(valid_transfers.iter().copied())
+    {
         let pending_transfer = event.into();
         let key = get_pending_key(&pending_transfer);
         if likely(wl_storage.has_key(&key)?) {
+            if likely(is_valid) {
```
batconjurer, line 246: herm. This works because the rejected event will eventually time out, but it might be nicer is we de-escrow NAM / re-mint wrapped erc20s immediately.

PR #1267, opened 2023-03-30
index only required event attributes

yito88:


bengtlofgren:
pls spawn devnet [anoma-devnet-max4,8,heliaxdev@1be9883,ON]

cwgoes:
@yito88 What's the status of this?

yito88:
I'll close this since Tendermint index is disabled by default now. IBC events can be queried from the event_log without Tendermint index.
If queries for other events with Tendermint index are required, I'll do this change again.

PR #1271, opened 2023-03-31
Handle Ethereum event nonces

sug0:
* Increment the nonce of the Bridge pool when we process `TransfersToEthereum` Ethereum events
* Increment the nonce of `TransfersToNamada` Ethereum events
* Reject Ethereum events with invalid nonces in `CheckTx` and `ProcessProposal`
* Process `TransfersToNamada` Ethereum events in order

**TODO:** Handle nonces derived from updates to the token whitelist

### DIFF
--- a/core/src/types/ethereum_events.rs
+++ b/core/src/types/ethereum_events.rs
@@ -510,7 +510,7 @@ pub mod testing {
     }
 
     pub fn arbitrary_nonce() -> Uint {
-        123.into()
+        0.into()
     }
 
     pub fn arbitrary_keccak_hash() -> KeccakHash {
--- a/core/src/types/ethereum_events.rs
+++ b/core/src/types/ethereum_events.rs
@@ -198,6 +198,45 @@ impl KeySeg for EthAddress {
     }
 }
 
+/// Event transferring batches of ether or Ethereum based ERC20 tokens
+/// from Ethereum to wrapped assets on Namada
+#[derive(
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Hash,
+    Ord,
+    Clone,
+    Debug,
+    BorshSerialize,
+    BorshDeserialize,
+    BorshSchema,
+)]
+pub struct TransfersToNamada {
+    /// Monotonically increasing nonce
+    pub nonce: Uint,
+    /// The batch of transfers
+    pub transfers: Vec<TransferToNamada>,
+    /// The indices of the transfers which succeeded or failed
+    pub valid_transfers_map: Vec<bool>,
+}
+
+impl From<TransfersToNamada> for EthereumEvent {
+    #[inline]
+    fn from(event: TransfersToNamada) -> Self {
+        let TransfersToNamada {
+            nonce,
+            transfers,
+            valid_transfers_map,
+        } = event;
+        Self::TransfersToNamada {
+            nonce,
+            transfers,
+            valid_transfers_map,
+        }
+    }
+}
+
 /// An Ethereum event to be processed by the Namada ledger
 #[derive(
     PartialEq,
--- a/core/src/types/ethereum_events.rs
+++ b/core/src/types/ethereum_events.rs
@@ -198,6 +198,12 @@ impl KeySeg for EthAddress {
     }
 }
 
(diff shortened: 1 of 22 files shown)


core/src/types/ethereum_events.rs
```diff
@@ -444,7 +510,7 @@ pub mod testing {
     }
 
     pub fn arbitrary_nonce() -> Uint {
-        123.into()
+        0.into()
```
batconjurer, line 513: very arbritrary

core/src/types/storage.rs
```diff
@@ -1151,6 +1154,147 @@ pub struct PrefixValue {
...
+    }
+}
+
+/// Draining iterator over a queue of Ethereum events.
+pub struct EthEventsQueueIter<'queue, E> {
+    current_nonce: Uint,
+    queue: &'queue mut InnerEthEventsQueue<E>,
+}
+
+impl<E> Drop for EthEventsQueueIter<'_, E> {
```
batconjurer, line 1197: Are all events in the queue ensured to be valid? Otherwise this may incorrectly commit to the wrong nonce on drop. If that is ensured, can you add a doc comment explaining this invariant must be maintained?

core/src/types/storage.rs
```diff
@@ -1151,6 +1154,147 @@ pub struct PrefixValue {
...
+                |insert_at| {
+                    tracing::debug!(?new_event, "Queueing Ethereum event");
+                    self.inner.insert(insert_at, new_event)
+                },
+                // the event is already present in the queue... this is
+                // certainly a protocol error
+                |_| {
+                    hints::cold();
+                    unreachable!(
+                        "An event with an identical nonce was already present \
```
batconjurer, line 1278: The doc string and the unreachable message imply different things. I think the unreachable message is the correct one.

core/src/types/storage.rs
```diff
@@ -1151,6 +1154,147 @@ pub struct PrefixValue {
...
+    }
+}
+
+impl<E: GetEventNonce> InnerEthEventsQueue<E> {
+    /// Retrieve the next events to be processed, if any.
+    ///
+    /// This decision is based on the nonce of the next event to be
+    /// processed by the ledger, and the nonce of the event that was just
+    /// confirmed (i.e. achieved a quorum of votes behind it).
+    pub fn get_next_events(
```
batconjurer, line 1229: I find this method name confusing. Maybe something like `push_and_iter`?  Or `push_and_next`?

ethereum_bridge/src/protocol/transactions/ethereum_events/mod.rs
```diff
@@ -138,11 +137,16 @@ where
...
+    let exists_in_storage = 'exists: {
+        let Some(seen) = votes::storage::maybe_read_seen(wl_storage, &eth_msg_keys)? else {
+            break 'exists false;
+        };
+        if seen {
+            tracing::debug!(?update, "Ethereum event is already seen");
+            return Ok((ChangedKeys::default(), false));
+        }
+        true
+    };
```
batconjurer, line 149:
```suggestion
    let exists_in_storage =  if let Some(seen) = votes::storage::maybe_read_seen(wl_storage, &eth_msg_keys)? {
        if seen {
            tracing::debug!(?update, "Ethereum event is already seen");
            return Ok((ChangedKeys::default(), false));
        }
        true
    } else {
        false
    };
```

batconjurer: Let's not get overly excited about rust's new goto statements :stuck_out_tongue_closed_eyes:

batconjurer (changes requested):

core/src/types/storage.rs
sug0: it wasn't the case before my last commit, but now I'm pretty certain we maintain this invariant. I'll add the docstring

sug0:
e2e tests need to be fixed. `send_transfer_to_namada_event()` sends transfer to namada events to the ledger with a randomly generated event nonce. the derived state changes will not get applied, since we are now enforcing a total order on the events' nonces.

PR #1275, opened 2023-04-04
docs/dev: add page table-of-contents plugin

tzemanovic:
adds https://github.com/slowsage/mdbook-pagetoc to dev docs

PR #1276, opened 2023-04-04
Add git developer and maintainer docs

tzemanovic:
depends on #1275 (because the pages are long having ToC makes it easier to navigate them)

The CI will fail here until #1275 is merged.


tzemanovic:
> can we merge this?

@juped wanted to re-org the page a bit, but maybe we could merge as is and do that later

Issue #1010, opened 2023-01-11
Gas & Fee


grarco:
Depends on #976 for wrapper txs to be included first in a block and max_gas block compliance

# TODO

- [x] #889 
- [ ] Add gas accounting mechanism (#1242)
    - `mempool_validate` check on the `GasLimit`
    - `process_proposal` check on the `GasLimit`
    -  `prepare_proposal` check on the `GasLimit`
    - In storage table of gas for every whitelisted tx/vp
    - `finalize_block` computation of the actual gas used and validation of the tx/block
    - Update client to produce wrappers with the correct `GasLimit` (account also for compilation costs)
    - Testing
    - Linear accounting of signatures verification (multisig #1102) -> Get the amount of signatures from `SignedTxData` and define a multiplicative coefficient given the cost of a single signature verification
    - No need to set gas as `i64` for tendermint compatibility anymore since we don't care about passing the used gas to tendermint (see [this](https://github.com/tendermint/abci/issues/57) issue and [this](https://github.com/tendermint/tendermint/issues/1065) one)
- [ ] Add optional unshielding in `WrapperTx` (#1281)
    - `mempool_validate`check
    - `process_proposal` check
    - `prepare_proposal` check
    - `finalize_block` unshield execution
    - update client
    - e2e tests
- [ ] Adjust the balance checks added in #962 to be coherent with the proposed specification
    - `mempool_validate` check
    - Correct `FeeAmount` in client, no default in cli and set default to the storage parameter
    - Define the cost for unit of gas in storage (parameter) + whitelisted assets + checks
    - `process_proposal` check
    - Fee payment in `finalize_block`
    - Force order of txs (wrappers first) in `process_proposal` (should already be done in #976)
    - Remove fixed fee from protocol parameters
    - Unit testing (transfer to block proposer, fee with invalid token)
    - Devnet testing (especially more unshielding in a single block)
    - Adjust `dry_run_tx` to handle a `WrapperTx` for a correct gas estimation
    - Define genesis values

PR #1281, opened 2023-04-11
Fee unshielding

grarco:
Based on #1242.

Addresses #1010.

Adds an optional, unencrypted unshielding `Transaction` to the `WrapperTx` to allow unshielding some tokens to pay fees when the transparent balance is insufficient.

- Static and dynamic (execution simulation) checks of the unshielding in `mempool_validate`, `process_proposal` and `prepare_proposal`
- Performs the unshield in `finalize_block` when the wrapper is accepted
- Updates the client to allow the user to pass a spending key to produce the unshield
- Adds an e2e test


grarco:
pls update wasm

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -809,6 +838,100 @@ where
...
+            // The unshielding tx does not charge gas, instantiate a limitless
+            // custom gas meter for this step
+            let mut gas_meter = TxGasMeter::new(u64::MAX);
+
+            // NOTE: given the checks in process_proposal, a failure in the
+            // unshielding tx should not happen. If it does, do not return early
+            // from this function but try to take the funds from the unshielded
+            // balance
+            match apply_tx(
+                TxType::Decrypted(DecryptedTx::Decrypted {
```
grarco, line 860: In here, I fake a decrypted tx to execute the unshielding. It's basically the same trick we use to execute governance proposals. At this point, I don't know whether we should think about allowing `Raw` txs execution (making sure that no other Raw txs can end up in a block and be executed)

apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -128,21 +137,34 @@ where
...
-                if let Ok(tx) = Tx::try_from(tx_bytes.as_slice()) {
-                    // If tx doesn't have an expiration it is valid. If time cannot be
-                    // retrieved from block default to last block datetime which has
-                    // already been checked by mempool_validate, so it's valid
-                    if let (Some(block_time), Some(exp)) = (block_time.as_ref(), &tx.expiration) {
-                        if block_time > exp { return None }
-                    }
-                    if let Ok(TxType::Wrapper(_)) = process_tx(tx) {
-                        return Some(tx_bytes.clone());
+                match self.validate_wrapper_bytes(tx_bytes, &mut temp_block_gas_meter, block_time, &mut temp_wl_storage, &gas_table, &mut vp_wasm_cache, &mut tx_wasm_cache) {
```
grarco, line 158: The gas check inside `validate_wrapper_bytes` could actually be extracted from there and moved into the `BlockSpaceAllocator`, effectively transforming it into a `BlockSpaceGasAllocator`. It would probably make for a more coherent logic but would not improve performance. Actually, given that the allocator stops at the first element of the iterator that cannot be placed in the block, we could end up with blocks containing less transactions

apps/src/lib/client/signing.rs
```diff
@@ -146,31 +146,31 @@ pub async fn tx_signer(
 ///
 /// If it is a dry run, it is not put in a wrapper, but returned as is.
```
tzemanovic, line 147:
```suggestion
/// If it is a dry run, it is not put in a wrapper, but returned as is.
///
/// If the tx fee is to be unshielded, it also returns the unshielding epoch.
```

core/src/types/transaction/wrapper.rs
```diff
@@ -171,6 +173,8 @@ pub mod wrapper_tx {
         pub epoch: Epoch,
         /// Max amount of gas that can be used when executing the inner tx
         pub gas_limit: GasLimit,
+        /// The optional, unencrypted, unshielding tx for fee payment
+        pub unshield: Option<Tx>,
```
tzemanovic, line 177: I think rather than using the `Tx` type here which allows arbitrary data and code, which would require some validation in the protocol (and also bloats the payload), it would be better to only attach masp `Transaction` here.

cwgoes: Agreed

grarco: Yes, this should be easy to implement given that we've already added the support for tx hashes in #1297. We can load the wasm code before running the operation. Still, some validation will be necessary even though we can get rid of the check on the content of the unshielding tx that will be done now with the type system

grarco:
pls update wasm

grarco:
pls update wasm

grarco:
pls update wasm

grarco:
pls update wasm

grarco:
Closed in favor of #1327

PR #1282, opened 2023-04-12
Make the the backing integer type for amounts 256 bit

batconjurer:
In order to be compatible with the Ethereum and IBC ecosystems, we need native support for token amounts with a backing 256 bit unsigned integer. This furthermore needs to be integrated with MASP, we requires breaking these amounts down in 64 bit words.

This PR also includes a new type `Dec` that replaces the dependency on the `rust_decimal` crate and provides a fixed-precision decimal number for computations that require it (mostly in proof-of-stake)


apps/src/lib/client/tx.rs
```diff
@@ -1377,133 +1411,165 @@ async fn gen_shielded_transfer(
...
+            unreachable!("The function `gen_shielded_transfer` is only called by `submit_tx` which validates amounts.")
+        };
         // Transaction fees need to match the amount in the wrapper Transfer
-        // when MASP source is used
+        // when MASP source is used. This amount should be <= `u64::MAX`.
         let (_, fee) = convert_amount(
             epoch,
             &ctx.get(&args.tx.fee_token),
-            args.tx.fee_amount,
+            &args.sub_prefix.as_ref(),
```
batconjurer, line 1438: wrong!

apps/src/lib/client/tx.rs
```diff
@@ -1546,16 +1622,37 @@ pub async fn submit_transfer(mut ctx: Context, args: args::TxTransfer) {
...
+                    &sub_prefix,
+                    balance,
+                )
+                .await;
                 eprintln!(
                     "The balance of the source {} of token {} is lower than \
                      the amount to be transferred. Amount to transfer is {} \
                      and the balance is {}.",
-                    source, token, args.amount, balance
+                    source, token, validated_amount, balance_amount
```
batconjurer, line 1655: add the sub-prefix here

apps/src/lib/client/tx.rs
```diff
@@ -1745,11 +1845,25 @@ pub async fn submit_ibc_transfer(ctx: Context, args: args::TxIbcTransfer) {
...
+                    &sub_prefix,
+                    balance,
+                )
+                .await;
                 eprintln!(
                     "The balance of the source {} of token {} is lower than \
                      the amount to be transferred. Amount to transfer is {} \
                      and the balance is {}.",
-                    source, token, args.amount, balance
+                    source, token, formatted_amount, formatted_balance
```
batconjurer, line 1866: add sub-prefix.

core/src/types/address.rs
```diff
@@ -576,18 +578,35 @@ pub fn masp_tx_key() -> crate::types::key::common::SecretKey {
...
+{
     vec![
-        (nam(), (0, 100)),
-        (btc(), (1, 100)),
-        (eth(), (2, 100)),
-        (dot(), (3, 100)),
-        (schnitzel(), (4, 100)),
-        (apfel(), (5, 100)),
-        (kartoffel(), (6, 100)),
+        ((nam(), None, MaspDenom::Zero), (0, 100)),
```
batconjurer, line 603: The rewards are equal across denominations, so we don't need to return a denomination here.

core/src/types/governance.rs
```diff
@@ -13,7 +13,7 @@ use crate::types::hash::Hash;
 use crate::types::key::common::{self, Signature};
 use crate::types::key::SigScheme;
 use crate::types::storage::Epoch;
-use crate::types::token::{Amount, SCALE};
+use crate::types::token::{Amount, NATIVE_SCALE};
 
 /// Type alias for vote power
 pub type VotePower = u128;
```
grarco, line 19: Given what we said in the meeting, this `u128` is no longer needed. I'd still keep the type alias as
```rust
pub type VotePower = Uint;
```
to use a more appropriate term for this context. We could probably just use an `Amount` but this is something that we can refactor later on with @Fraccaman.

core/src/types/token.rs
```diff
@@ -30,95 +35,383 @@ use crate::types::storage::{DbKeySeg, Key, KeySeg};
...
+        let mut raw = [0u64; 4];
+        raw[denom as usize] = val;
+        Self { raw: Uint(raw) }
+    }
+
+    /// Get a string representation of a native token amount.
+    pub fn to_string_native(&self) -> String {
+        DenominatedAmount {
+            amount: *self,
+            denom: 6.into(),
```
brentstone, line 210: use `NATIVE_MAX_DECIMAL_PLACES` instead of `6`?

core/src/types/uint.rs
```diff
@@ -0,0 +1,424 @@
...
+        assert!((-zero).non_negative());
+        let negative = SignedUint(Uint([1u64, 0, 0, 2u64.pow(63)]));
+        assert!(!negative.non_negative());
+        assert!((-negative).non_negative());
+        let positive = SignedUint(MAX_SIGNED_VALUE);
+        assert!(positive.non_negative());
+        assert!(!(-positive).non_negative());
+    }
+
+    /// Test that the absolute vale is computed correctly
```
brentstone, line 354: value

core/src/types/uint.rs
```diff
@@ -0,0 +1,424 @@
...
+        let ten = SignedUint(Uint::from(10));
+
+        assert_eq!(zero.abs(), Uint::zero());
+        assert_eq!(neg_one.abs(), Uint::from(1));
+        assert_eq!(neg_eight.abs(), Uint::from(8));
+        assert_eq!(two.abs(), Uint::from(2));
+        assert_eq!(ten.abs(), Uint::from(10));
+    }
+
+    /// Test that the absolute vale is computed correctly
```
brentstone, line 370: value

apps/src/lib/client/rpc.rs
```diff
@@ -1031,39 +1231,53 @@ pub async fn query_shielded_balance(
...
-                        .await
-                        .expect("context should contain viewing key")
-                };
-                if balance[&asset_type] != 0 {
-                    let asset_value =
-                        token::Amount::from(balance[&asset_type] as u64);
-                    println!("  {}, owned by {}", asset_value, fvk);
-                    found_any = true;
+                let mut balance = token::Amount::default();
+                for denom in MaspDenom::iter() {
```
murisi, line 1246: This logic seems to repeat what is at line 1049 of this file, maybe it could be factored out into a function? See also the comment at line 1070 of this file.

apps/src/lib/client/rpc.rs
```diff
@@ -1104,45 +1319,77 @@ pub async fn query_shielded_balance(
...
         }
     }
 }
 
-pub fn print_decoded_balance(
+pub async fn print_decoded_balance(
     ctx: &mut Context,
-    decoded_balance: Amount<Address>,
+    client: &HttpClient,
+    decoded_balance: MaspDenominatedAmount,
```
murisi, line 1331: This approach is valid. However this function could instead take an `Amount<Address, i256>`. That would relieve this function from having to aggregate `u64` denominations into a `u256`. This should be straightforward to do if functions like `decode_amount` also returned `Amount<Address, i256>`s.

apps/src/lib/client/rpc.rs
```diff
@@ -1104,45 +1319,77 @@ pub async fn query_shielded_balance(
...
+                    .unwrap_or_default(),
+                format_denominated_amount(client, addr, sub_prefix, amount,)
+                    .await,
+            );
+        }
     }
 }
 
-pub fn print_decoded_balance_with_epoch(
+pub async fn print_decoded_balance_with_epoch(
```
murisi, line 1360: Analogous comments to the ones at `print_decoded_balance` above.

apps/src/lib/client/rpc.rs
```diff
@@ -2609,9 +2886,99 @@ fn lookup_alias(ctx: &Context, addr: &Address) -> String {
...
+
+/// Make asset type corresponding to given address and epoch
+pub fn make_asset_type<T: std::fmt::Display>(
+    epoch: Epoch,
+    token: &Address,
+    sub_prefix: &Option<T>,
+    denom: MaspDenom,
+) -> AssetType {
+    // Typestamp the chosen token with the current epoch
+    let token_bytes = (
```
murisi, line 2930: Good call to create a `make_asset_type` function  as this prevents people from mixing up the tuple ordering. Another idea: maybe we can make a `MaspToken` or `DenominatedToken` struct to represent this quadruple. I say this, because this quadruple is also referred to frequently in the Merkle tree production code. I say a struct instead of a tuple because this structure now has 4 entries instead of 2, and hence field naming may make it easier to understand.

murisi (changes requested):
I've only reviewed these changes as they affect MASP. I think these changes are essentially valid and should achieve the desired goal of `u256` support in relation to MASP. One thing that I believe should be changed is the printing of MASP amounts in separate `u64` denominations to the user in certain places. How amounts are denominated by `u64`s should not be visible to end-users.

Other than that, I think my other comments are more stylistic. My general sentiment is that the high-level MASP API we expose should use `u256`s where logically possible. This would make the MASP functions easier to use in a lot of places by saving API users from constantly worrying about denominations. I.e. by giving them the illusion of working over `u256`s. Of course, behind the scenes we would still be  using `u64` MASP notes.

core/src/ledger/storage/masp_conversions.rs
```diff
@@ -49,54 +59,74 @@ where
...
-            (MaspAmount::from_pair(old_asset, -(reward.1 as i64)).unwrap()
-                + MaspAmount::from_pair(new_asset, reward.1).unwrap()
-                + MaspAmount::from_pair(reward_asset, reward.0).unwrap())
-            .into(),
-        );
-        // Add a conversion from the previous asset type
-        wl_storage.storage.conversion_state.assets.insert(
-            old_asset,
-            (
+        for denom in token::MaspDenom::iter() {
```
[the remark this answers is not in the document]
cwgoes: I think this is fine. Even if we rotate through conversion trees in an identifiable way the privacy loss is very minimal. Thank you for checking!

Makefile
```diff
@@ -13,17 +13,23 @@ wasms_for_tests := wasm_for_tests/wasm_source
 # Paths for all the wasm templates
 wasm_templates := wasm/tx_template wasm/vp_template
 
+ifdef JOBS
```
tzemanovic, line 16: this can be set in your local cargo config https://doc.rust-lang.org/cargo/reference/config.html#configuration-format

mariari:
Currently known issues

1. `sub_prefix`'s of a token do not have their demoniation saved
2. `chain_id`'s hash is different
3. Even if `sub_prefix` is ignored for `denom` lookup, transaction gets rejected
   - cause of this is still unknown

batconjurer:
pls update wasm

mariari:
Note for real history, merge #1630 into main.

Eth integration + amounts can still use this branch as they share a common ancestor, and the existence of this branch should make it easier to reuse existing work.

brentstone:
Closed as completed by #1630.

PR #1283, opened 2023-04-12
Specs revision

sug0:
Update the specs of the Ethereum bridge, removing outdated info.


batconjurer (changes requested):
Need to fix the bootstrapping section

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
```diff
@@ -1,117 +1,115 @@
...
+   it's up to governance validators to decide which addresses to trust.
+2. A governance proposal should be held to agree on an epoch $E_{bridge} > E_0$ at
+   which to launch the Ethereum bridge. The proposal should be executed at some
+   grace epoch $E_{grace}$, with $E_{grace} \le E_{bridge}$. The wasm code of
+   the proposal should modify storage with:
+
+    - The epoch $E_{bridge}$ when the bridge is to be enabled.
+    - The aforementioned governance parameters of the Ethereum bridge.
+
+   Additionally, the knowledge of $E_0$ must be communicated through the governance
```
batconjurer, line 47: I assume a hash of the validator set will be used?

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
```diff
@@ -1,117 +1,115 @@
...
+3. Validators should vote on the proposal if the wasm code correctly updates
+   Namada's storage, the proposal contains the epoch $E_0$ of the first set of
+   validators in the deployed contracts and the Ethereum height at which the
+   contracts were deployed, and the contracts are initialized correctly.
+4. Eventually, the proposal passes at some epoch $E_{end} \le E_{grace}$, if enough
+   validators vote on it. Then, the Ethereum oracle receives an update command, so it
+   can start processing Ethereum blocks to extract confirmed events. Should the proposal
+   not pass, no state updates are applied in Namada, therefore the bridge is not enabled.
+
+Notice that nothing stops users from preemptively updating the smart contracts
```
batconjurer, line 61: I'm confused here. Do you mean that the person deploying the contract lied about which validator set was in the contract?

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -123,11 +125,46 @@ If vote extensions are not available, this signed root may lag behind the
...
+
+## Timeouts
+
+Pending transfers to Ethereum sitting in the Bridge Pool for some duration of
+$T_{dur}$ blocks are subject to timeouts. The exact value of $T_{dur}$, for
+now, is the minimum duration in blocks of an epoch. Transactions that time out
+should revert the state changes in Namada, including refunding the paid fees
+and escrowed assets.
+
+Timeouts only happen after we process transfer to Ethereum events, to guarantee
```
batconjurer, line 147:
```suggestion
Timeouts only happen after we process transfer to Ethereum events to guarantee
```

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -123,11 +125,46 @@ If vote extensions are not available, this signed root may lag behind the
...
+the consistency of the shared state between Namada and Ethereum. Nonces from
+transfer to Ethereum events will place a total order on these transfers and
```
batconjurer, line 149:
```suggestion
`TransferToEthereum` events will place a total order on these transfers and
```

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
sug0: the epoch should be enough, since you can query the validator set of a given epoch

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
sug0: no, the contract was deployed with the right address, but because the governance proposal is published on-chain as public knowledge, anyone can start pushing validator set updates to the contracts before the governance proposal to enable the bridge passes

batconjurer: ah, I see.

batconjurer (approved):

documentation/specs/src/interoperability/ethereum-bridge.md
```diff
@@ -30,363 +36,8 @@ Ethereum side.
  - [Proofs and validator set updates](./ethereum-bridge/proofs.md)
  - [Smart Contracts](./ethereum-bridge/ethereum_smart_contracts.md)
```
cwgoes, line 38: I like this simplified introduction, but with this we should briefly mention what the bridge allows for (transfer of fungible & non-fungible tokens).

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
```diff
@@ -1,117 +1,115 @@
...
+- `eth_bridge_governance_version` - The version of the `Governance` contract,
+  starting from 1.
+- `eth_bridge_wnam_address` - The address of the deployment of the native
+  ERC20 token, representing NAM in Ethereum.
+
+Here is an overview of the steps to follow to enable the Ethereum bridge through a
+governance proposal for a given Namada chain:
+
+1. The [smart contracts](./ethereum_smart_contracts.md) of the Ethereum bridge
+   should be deployed a priori, with an initial validator set and other parameters,
```
cwgoes, line 36:
```suggestion
   should be deployed beforehand, with an initial validator set and other parameters,
```

I get what you mean :smile:, and from Namada's perspective it is _a priori_, but this might be confusing for some readers).

documentation/specs/src/interoperability/ethereum-bridge/bootstrapping.md
```diff
@@ -1,117 +1,115 @@
...
+   such as nonces, from epoch $E_0$. This step can be done by anyone, but ultimately
+   it's up to governance validators to decide which addresses to trust.
```
cwgoes, line 38:
```suggestion
   it's up to Namada stakers & validators to decide which addresses to trust.
```

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -1,101 +1,83 @@
...
 
-The only way this is impossible is if more than 1/3 of the validators by 
-stake from that epoch delete their ethereum keys, which is extremely unlikely.
+Proofs themselves do not alter any state in Ethereum. Rather, they
+authenticate state transitions in Ethereum based on the security
+properties (in the BFT sense) of the Namada network. Carrying out
+state updates in Ethereum is the task of an __Ethereum bridge relayer__.
+
+The job of an Ethereum bridge relayer is to submit proofs to the Ethereum
+bridge smart contracts before the conclusion of each Namada epoch. The default
```
cwgoes, line 66: Do proofs specifically have to be submitted before the end of the epoch? Do you mean submitting batches before the validator set changes?

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -18,15 +21,26 @@ pub enum EthereumEvent {
...
+Once an event has been seen by at least 2/3 of voting power, it is locked into a
 `seen` state, and acted upon.
 
-There is no adjustment across epoch boundaries - e.g. if an event is seen by 1/3
-of voting power in epoch n, then seen by a different 1/3 of voting power in
-epoch m>n, the event will be considered `seen` in total. Validators may never
-vote more than once for a given event.
+If the voting power of Namada changes across epoch boundaries, then events in
+storage which are yet to be achieve a quorum decision behind them (i.e. whose
+`seen` state is still `false`) must have their voting power adjusted. It is
```
cwgoes, line 31: How is this done, exactly? Do you mean that we don't recalculate the voting power until the event is voted again by a new validator, at which point we iterate?

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -18,15 +21,26 @@ pub enum EthereumEvent {
...
+enough to lazily adjust an event's voting power whenever a new vote is made
+for it, to avoid iterating over each Ethereum event in storage.
+
+Validators may never vote more than once for a given event. To ensure that this
+invariant is held, events are timed out if they are not `seen` within the span
```
cwgoes, line 36: We also check that a specific validator hasn't voted for a specific event more than once, right?

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
```diff
@@ -35,40 +49,71 @@ on Namada. This minimum number of confirmations will be changeable via
...
 `TransferToNamada` events may include a custom minimum number of
-confirmations, that must be at least the protocol-specified minimum number of
-confirmations but is initially set to __100__.
+confirmations that must be at least the protocol-specified minimum number of
+confirmations. However, this value is initially set to __100__.
+
+Validators must not vote to include events that have not met the required
+number of confirmations. Votes on unconfirmed events will eventually time
+out in storage, unless the number of confirmations was only off by a few
+block heights in Ethereum. Assuming that an honest majority of validators
```
cwgoes, line 58: What's the honest majority threshold here? Should be <1/3 Byzantine, as usual, right?

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
```diff
@@ -1,101 +1,83 @@
...
+the Ethereum bridge will halt in the Namada to Ethereum direction, since
+it will not be able to authenticate user transfers.
 
-If a significant number of validators become inactive in the next epoch, we 
-need a means to complete validator set update. Until this is done, the 
-bridge will halt. 
+In this case, the validators from $E$ will need to coordinate, offline, the
+crafting of a transaction including their vote on the validator set of $E + 1$.
+The only way this is impossible is if more than $1/3$ (by stake) of the consensus
+validators from $E$ delete their Ethereum keys, which is extremely unlikely.
```
cwgoes, line 56: if they delete both their hot and cold keys, right?

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
```diff
@@ -123,11 +125,46 @@ If vote extensions are not available, this signed root may lag behind the
...
+
+## Timeouts
+
+Pending transfers to Ethereum sitting in the Bridge Pool for some duration of
+$T_{dur}$ blocks are subject to timeouts. The exact value of $T_{dur}$, for
+now, is the minimum duration in blocks of an epoch. Transactions that time out
+should revert the state changes in Namada, including refunding the paid fees
+and escrowed assets.
+
+Timeouts only happen after we process transfer to Ethereum events to guarantee
```
cwgoes, line 147: What if a particular bridge pool update is never relayed to Ethereum, though? Can we ever claim the assets back on Namada?

cwgoes: I guess if we require ordered processing of bridge pool updates anyways, there's no way to recover from this anyways. This does mean that we'll be less able to be responsive to Ethereum gas markets, though :thinking:. Just curious, did we consider a design where the timeouts reference Ethereum-side block heights, and processing a particular transfer after that block height will fail + you can make a proof that a transfer was not processed before that block height (emitting an Ethereum event which can be confirmed on Namada, releasing the assets)?

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_namada.md
```diff
@@ -1,16 +1,18 @@
...
+will be two internal accounts with associated native validity predicates:
 
 - `#EthBridge` - Controls the `/eth_msgs/` [storage](ethereum_events_attestation.md#storage)
-- and ledgers of balances
-  for wrapped Ethereum assets (ERC20 tokens) structured in a
-  ["multitoken"](https://github.com/anoma/anoma/issues/1102) hierarchy
-- `#EthBridgeEscrow` which will hold in escrow wrapped Namada tokens which have
-  been sent to Ethereum.
+  and ledger of balances of minted wrapped Ethereum assets (ERC20 tokens) structured in a
+  ["multitoken"](https://github.com/anoma/anoma/issues/1102) hierarchy.
```
cwgoes, line 8: Ideally the specs shouldn't link to Github issues, which are pretty ephemeral. Do we have multitoken docs somewhere?

cwgoes: What about the multitoken section of the specs, actually?

cwgoes (changes requested):
Really nice specs work! Left several questions & a few nits.

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
sug0: yeah, exactly, so you can still authenticate stuff

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
sug0: So, this is related with #1288

> Do you mean that we don't recalculate the voting power until the event is voted again by a new validator

I'm still not 100% sure how this will be implemented, however that would be the idea, yes. We store a map of a validator's adress to the block height they voted on some event, so we can scale voting power based on the earliest height we find (maybe?). Basically we need a least common multiple of some sort to then add all of the voting powers together, I think.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
sug0: yeah. `seen_by`, in storage, tracks who voted on some event. it's a function from a validator's address to the height in storage at which they voted on some eth event

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
sug0: yep

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
sug0: the keys we use to sign validator set updates are hot keys, so only those

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
sug0: you can pick a subset of whichever transfers from the bridge pool you want to relay. quorum certificates contain signatures over the merkle root of the bridge pool and some nonce. you provide an inclusion proof for the subset you want to relay. so, there should always be relays happening, regardless of whether a particular transfer is included or not. eventually, transfers that are never relayed will time out, and the assets and fees can be reclaimed back on Namada.

we did not consider timing out events based on their Ethereum-side block height, to answer your question

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
cwgoes: fwiw I think recalculating by just re-iterating over all of the current voting powers is fine, the number of validators is a low constant.

documentation/specs/src/interoperability/ethereum-bridge/ethereum_events_attestation.md
cwgoes: We should clarify that, then

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
cwgoes: But we could recover the bridge-escrowed funds with the cold keys, right?

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
cwgoes: aye - we should just clarify that in this sentence then

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
cwgoes: I'm less concerned with Namada not reaching a quorum decision on a particular nonce than I am with the Ethereum gas prices getting high and no one wanting to relay batches, during which a large queue builds up because the Ethereum side won't allow later batches tto be relayed until earlier ones have been - I'm assuming that the Ethereum contracts check the nonce and increment it each time, right?

documentation/specs/src/interoperability/ethereum-bridge/proofs.md
sug0: yes, this is only about validator set update proofs, which are generated by Namada validators automatically, and thus require hot keys

documentation/specs/src/interoperability/ethereum-bridge/transfers_to_ethereum.md
sug0: > I'm assuming that the Ethereum contracts check the nonce and increment it each time, right?

yup

this is definitely an issue that affects throughput. the escrowed funds would not be able to be reclaimed, with the current design, since they either need to time out or be relayed to be released from the bridge pool, and this only happens when we get a transfer to Ethereum event

cwgoes: alright - I think this is OK for now, but let's definitely note it down

PR #1300, opened 2023-04-17
Refactor Bridge pool VP

sug0:
Fix a bug in the Bridge pool's `escrow_check` method(*) and refactor the VP code a bit, for clarity.

(*) The code path of the bug was not being triggered. The logic of this method was faulty nonetheless.

### DIFF
--- a/shared/src/ledger/native_vp/ethereum_bridge/bridge_pool_vp.rs
+++ b/shared/src/ledger/native_vp/ethereum_bridge/bridge_pool_vp.rs
@@ -84,9 +84,9 @@ where
         }
     }
 
-    /// Check that the correct amount of wrapper erc20s were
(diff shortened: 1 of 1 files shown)


shared/src/ledger/native_vp/ethereum_bridge/bridge_pool_vp.rs
```diff
@@ -82,6 +84,51 @@ where
         }
     }
 
+    /// Check that the correct amount of wrapper erc20s were
+    /// sent from the correct account into escrow
+    fn check_werc20s_escrowed(
```
batconjurer, line 89: Such a nit, but `werc` looks funny to me. Can we just call the function `check_erc20s_escrowed`?

batconjurer (approved):

PR #1304, opened 2023-04-18
update proptest state machine 

tzemanovic:
based on v0.15.1

This PR updates to a new version of proptest with state machine support from https://github.com/proptest-rs/proptest/pull/310


tzemanovic:
switched to use a rev hash instead of a branch name as there might be some more changes in the branch coming in https://github.com/anoma/namada/compare/a227c7aefe535fd30f2085c7101806b0674a44df..ad5a6fe0fcc80c6478318eb355b773602345790f

Issue #1289, opened 2023-04-13
Burn wrapped ERC20s after a successful transfer to Ethereum


sug0:
The wrapped assets escrowed under the Bridge pool's internal address should be burned, when a transfer to Ethereum moving value of the same kind succeeds.

https://github.com/anoma/namada/blob/f1c2fe0816b28a657af3fb6d7be5406ec65ea8b2/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs#L304-L317

E.g.
```rs
             if likely(is_valid) { 
                 tracing::debug!( 
                     ?pending_transfer, 
                     "Valid transfer to Ethereum detected, compensating the \ 
                      relayer and burning assets in Namada" 
                 );
                burn_transferred_assets(wl_storage, &pending_transfer)?;
            } else {
// ...
```

The supply keys for such an asset must also be updated.

**NOTE:** Remove this comment from the code:
```rs
                // Until burning is implemented, we disallow changes to any
                // supply keys via wasm transactions
```
At `shared/src/ledger/native_vp/ethereum_bridge/vp.rs:209`

PR #1307, opened 2023-04-18
Burn ERC20s

sug0:
Closes https://github.com/anoma/namada/issues/1289

### DIFF
--- a/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
+++ b/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
@@ -314,8 +314,9 @@ where
                 tracing::debug!(
                     ?pending_transfer,
                     "Valid transfer to Ethereum detected, compensating the \
-                     relayer"
+                     relayer and burning any Ethereum assets in Namada"
                 );
+                burn_transferred_assets(wl_storage, &pending_transfer)?;
             } else {
                 tracing::debug!(
                     ?pending_transfer,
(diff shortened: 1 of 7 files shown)


ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -482,6 +488,48 @@ where
...
+    wl_storage: &mut WlStorage<D, H>,
+    transfer: &PendingTransfer,
+) -> Result<BTreeSet<Key>>
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+{
+    let mut changed_keys = BTreeSet::default();
+
+    let native_erc20_addr = match wl_storage
```
batconjurer, line 501: doesn't storage already have a method that returns an error if deserialization fails?

tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -156,6 +156,181 @@ fn run_ledger_with_ethereum_events_endpoint() -> Result<()> {
     Ok(())
 }
 
+/// Test we can transfer some DAI to an implicit address on Namada,
+/// then back to Ethereum, burning the assets we minted after the
+/// first transfer.
+#[tokio::test]
+async fn test_roundtrip_eth_transfer() -> Result<()> {
```
batconjurer, line 163: It seems this test does not actually confirm that a burn happened

batconjurer (changes requested):
Couple of tiny comments. The big things is that I would like a unit test in the `protocol/transactions/ethereum_events/events.rs` file. It should not just check that the supply key got touched on burn, but that the burn was performed correctly. We should also test that no burns happen for wnam

batconjurer (approved):

PR #1315, opened 2023-04-24
doc/specs: remove katex renderer

tzemanovic:
Follow-up to https://github.com/anoma/namada/pull/1314 - the katex renderer is being removed in v0.5 and it seems to be failing in v0.4.0 (https://github.com/anoma/namada/actions/runs/4783109491/jobs/8503057783).

See https://github.com/lzanini/mdbook-katex/issues/68 for details.

### DIFF
--- a/documentation/specs/book.toml
+++ b/documentation/specs/book.toml
@@ -15,8 +15,6 @@ git-branch = "main"
 [output.html.search]
 expand = true
 
-[output.katex]
-
 [output.linkcheck]
 
 [preprocessor.katex]


SichangHe:
The KaTeX renderer is depricated, but not removed, thus should not cause CI failure.
The failed CI run linked seems like a glibc incompatibility between the mdBook-KaTeX binary built on Ubuntu 22.04 and the CI environment on Ubuntu 20.04.
Your latest CIs are passing, so that is good, although I have no idea why.

tzemanovic:
@SichangHe ah, thank you for clarifying the confusion! It is strange that it's now working :shrug:

PR #1324, opened 2023-04-28
Install protoc to build with the new prost (3/n) 

tzemanovic:
follow-up to #1306 

added GITHUB_TOKEN to protoc installation to avoid GH API rate limit as per https://github.com/arduino/setup-protoc/issues/63


yito88 (approved):
Thank you for following-up

PR #1325, opened 2023-05-02
Fix `update_data` for epoched data

brentstone:
For both `Epoched` and `EpochedDelta` objects, there are improvements that can be made to the function `update_data`. The base is currently #892 because a bug in the existing changes is exposed in the `pos_state_machine_test`.


tzemanovic:
pls update wasm

tzemanovic (approved):

PR #1327, opened 2023-05-02
Gas & Fee

### LINKED ISSUE #1010: Gas & Fee

### LINKED ISSUE #1473: Include gas in `BlockSpaceAllocator`
The `BlockSpaceAllocator` should be refactored to a `BlockSpaceGasAllocator` to include validation on the gas limit of transactions at block construction time, which, at the moment, is done in a separate way.

As @sug0 suggested, we could refactor `TryAlloc` trait to include an associated type to account for the length of the **serialized** transaction (as stored in the mempool) and the `gas_limit` field of the wrapper. The implementation for the `EncryptedTxBatchAllocator` should then be adjusted to validate also the gas.

Validation of the tx expiration and chain id can be kept in `validate_wrapper_bytes` and this function can now return the gas required to be passed to the allocator.

The block gas limit parameter should be inserted into the `TxBin` but just for the encrypted txs: to do so, we can make this field of the space allocator hold a new struct carrying two `TxBin`s (renaming the fields to remove the trailing in bytes), one for the space and one for the gas: by doing this we can reuse the same logic of `try_dump` for both the checks

### LINKED ISSUE #1489: Temporary disable `GAS_LIMIT_RESOLUTION`
The `GAS_LIMIT_RESOLUTION` const in `core/src/types/transaction/wrapper.rs` is used to obfuscate the gas required by the tx to prevent leakage of information. Given that dkg is not yet implemented and that the `GasLimit` is used for fee computation, we should temporarily set this value to 1 to provide an accurate accounting of fees

### LINKED ISSUE #1597: Disposable key pair for fee payment
The client should support an optional flag to allow for the on-the-fly creation of a disposable keypair to sign a wrapper transaction and, therefore, pay fees. This would work in combination with a mandatory unshielding operation (required cli argument) releasing the entire (and exact) amount of tokens to cover for the fees. This way we can provide more privacy for the fee payer as the unshielding transfer is not directly linked to their address.

grarco:
## Describe your changes

Closes #1010.
Closes #1473.
Closes #1489.
Closes #1597.

This PR finalizes the first version of the gas & fee system:

- Tx and block `GasLimit` checks in `mempool_validate`, `prepare_proposal` and `process_proposal`
- Runtime tx gas check against the declared `GasLimit` in `finalize_block`
- Removes MASP fees
- Removes fixed fees added in #962 
- Implements variable fees
- Implements fee payment to the actual block proposer
- Adds a `dry-run-wrapper` client command to estimate gas cost of a tx
- Extend the `BlockSpaceAllocator` to also allocate block gas
- Adds an optional masp unshielding section to the `Tx` for fee payment
- Implements disposable keypairs for wrapper signing and fee payment
- Disable `GAS_LIMIT_RESOLUTION`
- Expects a valid block proposer with every block
- Testing

v0.21.1

## Checklist before merging to draft

- [X] Git history is in acceptable state


grarco:
pls update wasm

grarco:
pls spawn devnet [devnet-0.15-gas-fee,3,heliaxdev@1be9883,ON]

apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
```diff
@@ -353,7 +359,36 @@ mod test_bp_vote_extensions {
...
+                    as i64,
+            }),
+            signed_last_block: true,
+        }];
+        let req = FinalizeBlock {
+            proposer_address: pkh1,
+            votes,
+            ..Default::default()
+        };
+        assert_eq!(shell.start_new_epoch(Some(req)).0, 1);
```
sug0, line 391:
```rs
assert_eq!(shell.start_new_epoch().0, 1);
```
did this stop working? :\

sug0 (changes requested):
some suggestions + nits + changes requested

apps/src/lib/node/ledger/shell/vote_extensions/bridge_pool_vext.rs
grarco: No it's just that we need a block proposer all the times from this pr, so I needed to construct a proper `FinalizeBlock` request

apps/src/lib/node/ledger/shell/block_alloc.rs
```diff
@@ -47,164 +53,239 @@ pub mod states;
...
     D: 'static + storage::DB + for<'iter> storage::DBIter<'iter>,
     H: 'static + storage::StorageHasher,
 {
     #[inline]
     fn from(storage: &WlStorage<D, H>) -> Self {
-        Self::init(storage.pos_queries().get_max_proposal_bytes().get())
+        let max_block_gas = storage
+            .read(&parameters::storage::get_max_block_gas_key())
+            .expect("Error while reading from storage")
+            .expect("Missing max_block_gas parameter in storage");
```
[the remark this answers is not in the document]
grarco: Can I? This is not strictly related to PoS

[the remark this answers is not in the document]
sug0: hm, I guess not. nevermind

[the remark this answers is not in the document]
grarco: Ok, I can maybe place this helper function in `core/sr/ledger/gas.rs`

apps/src/lib/node/ledger/shell/block_alloc/states.rs
```diff
@@ -64,36 +64,41 @@ pub struct BuildingEncryptedTxBatch<Mode> {
...
 pub trait TryAlloc {
-    /// Try to allocate space for a new transaction.
-    fn try_alloc(&mut self, tx: &[u8]) -> Result<(), AllocFailure>;
+    type Resource<'tx>;
+
+    /// Try to allocate resources for a new transaction.
+    fn try_alloc(
+        &mut self,
+        resource_required: Self::Resource<'_>,
+    ) -> Result<(), AllocFailure>;
```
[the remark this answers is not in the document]
grarco: Ah, this can't actually be done because some times we need the `Resource` to be of type `BlockResources` (so both gas and len) which cannot implement the proposed `Resource` trait given that a single `TxBin` can track only a single resource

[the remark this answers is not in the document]
sug0: you can add a `Resource` impl for `BlockResources` I guess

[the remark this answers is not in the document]
grarco: I could but the fact is that `TxBin` is now generic over a `R: Resource`, but the bin itself cannot actually track more than one resource. It's not a problem if we never instantiate a `TxBin<BlockResource>` but I don't know, it still lefts this possibility open

sug0 (approved):
lgtm

grarco:
pls update wasm

grarco:
pls update wasm

bengtlofgren:
pls spawn devnet [devnet-0.21.1,3,heliaxdev@1be9883,ON]

bengtlofgren:
pls spawn devnet [devnet-0.22,3,heliaxdev@1be9883,ON]

grarco:
pls update wasm

bengtlofgren:
pls spawn devnet [devnet-0.22,3,heliaxdev@1be9883,ON]

Makefile
```diff
@@ -64,7 +64,8 @@ package: build-release
 
 check-wasm = $(cargo) check --target wasm32-unknown-unknown --manifest-path $(wasm)/Cargo.toml
 check:
-	$(cargo) check && \
+	$(cargo) check --workspace --exclude namada_benchmarks && \
+	$(cargo) +$(nightly) check --benches && \
```
tzemanovic, line 68: does this need nightly?

apps/src/lib/cli.rs
```diff
@@ -4645,28 +4652,38 @@ pub mod args {
...
                 "The expiration datetime of the transaction, after which the \
                  tx won't be accepted anymore. All of these examples are \
                  equivalent:\n2012-12-12T12:12:12Z\n2012-12-12 \
                  12:12:12Z\n2012-  12-12T12:  12:12Z",
             ))
+            .arg(
+                DISPOSABLE_SIGNING_KEY
+                    .def()
+                    .help(
+                        "Generates an ephimeral, disposable keypair to sign \
```
tzemanovic, line 4681:
```suggestion
                        "Generates an ephemeral, disposable keypair to sign \
```

Makefile
grarco: I tried to remove it but it complains about not being able to compile `namada_tests`

bengtlofgren:
pls spawn devnet [devnet-0.22,3,heliaxdev@1be9883,ON]

apps/src/lib/cli/context.rs
```diff
@@ -96,16 +98,35 @@ impl Context {
...
+                crate::wallet::load_or_new_from_genesis(
+                    &chain_dir,
+                    default_genesis,
+                ),
+                genesis.native_token,
+            ) // If the WASM dir specified, put it in the config
+        } else {
+            #[cfg(not(any(test, feature = "dev")))]
+            panic!("Missing genesis file");
+            #[cfg(any(test, feature = "dev"))]
```
tzemanovic, line 120: let's use this as "dev" will be soon:tm:  removed
```suggestion
            #[cfg(not(any(test, feature = "testing")))]
            panic!("Missing genesis file");
            #[cfg(any(test, feature = "testing"))]
```

Makefile
tzemanovic: we'll be able to use stable after https://github.com/anoma/namada/pull/1771

apps/src/lib/client/signing.rs
tzemanovic, line 1: it looks like this file is not being used - we must have accidentally revived it at some point

apps/src/lib/node/ledger/shell/block_alloc.rs
tzemanovic, line 1: This was renamed from `block_space_alloc` but the original is not deleted (same for the nested mods)

shared/src/vm/mod.rs
```diff
@@ -25,7 +25,7 @@ const UNTRUSTED_WASM_FEATURES: WasmFeatures = WasmFeatures {
     relaxed_simd: false,
     threads: false,
     tail_call: false,
-    floats: false,
+    floats: true,
```
tzemanovic, line 28: I don't think we want to allow floats in wasm as it's allowed to give non deterministic results for NaNs. It's not needed for the gas, is it?

apps/src/lib/client/signing.rs
grarco: Ah yes, could be my fault since the base of this pr is very old, I might have kept it by mistake (should be correctly removed in the draft branch though)

shared/src/vm/mod.rs
grarco: Ah yes, sorry I should have left a note on this one. So basically I reworked the way we call `fetch_or_compile` and `compile_and_fetch` and also the implementation of `fetch_or_compile`. This is because `compile_or_fetch` is eventually called in any case (either the code being the actual code or only the hash of it). So I basically rewrote it to call `fetch_or_compile` in any case (this ease the gas accounting operation that must be done in any case and can therefore be done in this function only) and then let this function call `compile_or_fetch` in turn.

Now, in the process  of doing this, I also moved the call to `validate_untrusted_wasm` because, at the moment, we are validating the code only when we receive an hash but I think we should do the opposite. When we get an hash, if the precompiled module is not in cache, we end up reading the code from storage. But I think there should be no need to validate code placed in storage (validation should happen when we write the code to storage which I think we are not doing at the moment). If instead of an hash we get some code, we need to validate it before running it (because if the tx is not whitelisted it would only be discarded by the vps, so after it's run).

Now, to the point of the floats. In the compilation of the ibc transaction we emit some `f64` operations (I checked it with `wasm2wat` on the code emitted using both `build-wasm-scripts` and `build-wasm-scripts-docker` from different releases, even in the current version of main these instructions are there). The rearrangement of the compilation steps I mentioned before brought up this thing which, usually, doesn't manifest itself because the precompiled wasm module happens to always be in cache and so skips the validation part. To double check this, in the main branch, I removed the load from cache and forced the compilation of the module every time and indeed, the validation step rejected the ibc transaction because of forbidden floating point operations.

I had a look at the ibc code and it looks that these f64 operations where coming from one of ibc dependencies, in particular from a single insert in an HashMap which, for some strange reasons, triggered the generation of f64 operations in wasm, so I simply allowed floats to speed up the development process since I think I cannot solve this issue by myself

apps/src/lib/client/signing.rs
grarco: Oh, actually it's still there...

shared/src/vm/mod.rs
tzemanovic: Ah, I see. The change makes sense, thanks! 

I had a look at tx_ibc (I left the symbols in the wasm [by removing `RUSTFLAGS='-C link-arg=-s'` from makefile]) and found that the `f64` usage comes from `serde_json` and it's compiled into a number parses even if floats are not used :/  It's being called from e.g. `send_transfer_execute` imported from ibc-rs. If we want to remove the usage of floats we have to use a fork (https://github.com/serde-rs/json/issues/567). There's one from Near https://github.com/nearprotocol/json, it's a bit behind so might need to rebase on later version.

cc @yito88

tzemanovic: https://github.com/cosmos/ibc-rs/issues/850

Issue #1292, opened 2023-04-13
Improve UX on relays of partially accepted BP proofs


sug0:
Say the Bridge pool's nonce is at value $N$, in Namada. If $\frac{1}{3}$ of the total stake has voted on a transfers to Ethereum event with nonce $N$, it is very likely the event is not forged, and will be acted upon very shortly. Bridge pool proofs relays with nonce $N$ should emit a CLI message, warning users that this process may fail, given the voting power behind the events received with nonce $N$.

If the voting power is $\ge \frac{2}{3}$ behind such an event, then it has definitely already been acted upon, but the Bridge pool proof of the next nonce may not have been generated yet.

**EDIT:** Maybe a similar UX improvement can be made to validator set updates?

PR #1340, opened 2023-05-09
Emit warnings for fallible Bridge pool relays

sug0:
Partially solves https://github.com/anoma/namada/issues/1292
(Missing validator set update UX improvements)


cwgoes:
I'm a little bit confused. What does it mean for an event to have been relayed, but not have a quorum of validator signatures? I thought the Ethereum-side contracts checked that all transfers from Namada _have_ a quorum of validator signatures - shouldn't it be impossible to relay an event until it does?

cwgoes:
Ah, you mean a quorum of validators agreeing on step (3), that the Ethereum-side contracts have confirmed the transfers. That makes sense but I think we're overloading the word "relay" a bit much, maybe we could say "relayed to Ethereum" or "relayed from Ethereum" so that it's not ambiguous?

PR #1344, opened 2023-05-10
pos: add a query to find validator by Tendermint address

tzemanovic:
adds `namadac find-validator --tm-address <>`


Fraccaman:
LGTM! thanks :)

tzemanovic:
> this needs to be SDKized

updated

hkey0:
> > this needs to be SDKized
> 
> updated

function `validator_by_tm_addr` is not available in rpc.this should be added

PR #1345, opened 2023-05-10
Fix Key from always parsing (do not rename ("Key" is the name of a struct, which is upcased))

mariari:
Previously the key would always parse, even if the string is empty, this causes an issue where if we wanted an Optional Key, then it would always be parsed as Some, causing issues for various parts of the system.


Fraccaman:
what should we do here?

mariari:
@juped is the new commit a fine solution to this matter?

batconjurer:
> Looks like a few unit tests construct keys with empty segments; not sure yet whether it's a bug in the tests or the code

The empty key is to ensure we always match when iterating over keys.

PR #1362, opened 2023-05-12
apps/client/utils/init-network: avoid adding tokens to wallet twice

tzemanovic:
follow-up to https://github.com/anoma/namada/pull/1081 - the token addresses are being added twice into the wallet when running `init-network`. In here it's a no-op but with #1350 changes it triggers a replace prompt which is unwanted


tzemanovic:
> for some reason, did not address issue; still prompts on first token encountered

Ah, there are more dup inserts for the main non-validator wallet and some for validators too. Let's address this differently in #1350

PR #1365, opened 2023-05-12
potentially breaking change

bengtlofgren:
This will require some review before i can merge it. I'm afraid changing a global variable is dangerous


Fraccaman:
im not sure I follow, whats the purpose of this PR?

bengtlofgren:
I have tested this locally, and it seems to work for at least `init-validator`. What is the issue? @juped

bengtlofgren:
Ah right, DEFAULT_BASE_DIR is used when creating the pre-genesis folder @juped

PR #1367, opened 2023-05-13
apps/config: remove DEFAULT_BASE_DIR constant

juped:
Now that Namada defaults to platform-compliant (e.g. XDG) default data
directories, the DEFAULT_BASE_DIR constant (".namada") should no longer
exist. Remove it, using get_default_namada_folder() in places which
referenced it.

The ultimate fallback in case get_default_namada_folder() fails remains
".namada", but this is now just a string inside
get_default_namada_folder(). This will not happen except in
pathological cases where e.g. no $HOME exists; the details are in the
documentation for the directories crate.

### DIFF
--- a/tests/src/e2e/setup.rs
+++ b/tests/src/e2e/setup.rs
@@ -412,6 +412,7 @@ impl Test {
     }
 
     pub fn get_base_dir(&self, who: &Who) -> PathBuf {
+        let default_namada_folder = config::get_default_namada_folder();
         match who {
             Who::NonValidator => self.test_dir.path().to_owned(),
             Who::Validator(index) => self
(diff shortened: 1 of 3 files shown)


tests/src/e2e/setup.rs
```diff
@@ -420,7 +421,7 @@ impl Test {
                 .join(self.net.chain_id.as_str())
                 .join(utils::NET_ACCOUNTS_DIR)
                 .join(format!("validator-{}", index))
-                .join(config::DEFAULT_BASE_DIR),
+                .join(default_namada_folder),
```
tzemanovic, line 424: This won't work as `default_namada_folder` is an absolute path and so with `.`join` it overrides the path in full. We could probably keep using `.namada` for e2e tests, or perhaps better just use the `"validator-{}"` as a base dir

PR #1378, opened 2023-05-15
ci: remove clippy-abcipp check

juped:
SDK is going to break the abci++ build, and the consensus on whether we
care is that we don't and should consider removing abci++ anyway.
Therefore, disable this check so we are ready for SDK.


Fraccaman:
Im not sure we want to do this, whats the matter with clippy and abci++ ?

Fraccaman (approved):

PR #1404, opened 2023-05-19
Add IBC event queries

### LINKED ISSUE #1358: IBC: Query IBC events without Tendermint index
Tendermint index uses lots of storage space. So, it's disabled by default.
IBC relayer (Hermes) requires the index to query IBC events now. If Namada can query them without the index, we don't have to enable the index in a node that interacts with the relayer.

yito88:
closes #1358

To get IBC "update_client", "send_packet" and "write_acknowledgement" events for an IBC relayer without Tendermint index because Tendermint index is disabled by default.


tzemanovic (dismissed):
@yito88 I've added a small refactor to re-use the async runtime in 72270bee7c09660209cce54e5ff20fb36562b8ce

Fraccaman:
@yito88 can I merge this to the draft branch?

PR #1405, opened 2023-05-19
Stateful replay protection

grarco:
Implements stateful replay protection in `prepare_proposal` to avoid adding the same transaction more than once in a block

### DIFF
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -22,6 +22,7 @@ use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
+use namada::core::types::hash::Hash;
 use namada::ledger::events::log::EventLog;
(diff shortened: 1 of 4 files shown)


apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -22,6 +22,7 @@ use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
+use namada::core::types::hash::Hash;
```
bengtlofgren, line 25: Grarco? More like narco

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -648,6 +651,55 @@ where
         response
     }
 
+    /// Checks that neither the wrapper nor the inner transaction have already
```
bengtlofgren, line 654: LGTM

bengtlofgren (dismissed):
All good in my 👀

adrianbrink (approved):

PR #1415, opened 2023-05-20
edited language on Namada page as well as a couple other edits

0xapriori:


0xapriori:
"Note, should have built off the correct repository" 
-Bengt

PR #1444, opened 2023-05-22
Common sub-expression elimination in inflation calculation

cwgoes:
(as title)


cwgoes:
cc @brentstone typically we should give names to repeatedly-used expressions and only calculate their values once

PR #1438, opened 2023-05-21
Makefile: Use `NAMADA_E2E_DEBUG` variable

gijswijs:
When using prebuilt binaries for e2e testing, it defaults to './target/release'.
For the `test-e2e` makefile target we want it to use './target/debug'.

PR #1450, opened 2023-05-22
tomas/fix failing storage test

tzemanovic:
Fixes a test issue found in https://github.com/anoma/namada/pull/1438. The check that's failing is redundant and doesn't need to be satisfied, as is a case with the seed added here in which the epoch duration is 1 sec.


brentstone (approved):
LGTM

PR #1454, opened 2023-05-23
use cometbft binary

adrianbrink:
@Fraccaman could you fix the CI integration. cometbft just works and we should start using it over our fork of tendermint


adrianbrink:
closed in favor of #1476

PR #1455, opened 2023-05-23
storage: persist `update_epoch_blocks_delay`

tzemanovic:
Persists a newly added storage field for epoch update blocks delay to be available after node restart when not `None`.


Fraccaman:
pls spawn devnet [devnet-0.15-faster-epochs,3,heliaxdev@1be9883,ON]

PR #1461, opened 2023-05-24
Fix IBC token VP to reject invalid refund

### LINKED ISSUE #1460: IBC Token VP should check if IBC-related data has been changed
For example, the current IBC token VP cannot reject a transaction to transfer from IbcMint to another account with a dummy IBC ack/timeout message if the tx doesn't touch any IBC-related data (i.e. IBC VP isn't triggered).

yito88:
closes #1460


cwgoes (changes requested):
I don't think this check is sufficiently specific - we need to check that the transfer(s) performed _corresponded_ to the executed IBC message, no?

cwgoes:
Also, if we're just checking that another VP has also been called, can't we check that with the `verifiers` field?

yito88:
This will be fixed in #1526

Issue #1376, opened 2023-05-15
MASP CI False Positives and Negatives


murisi:
The CI reports that the 3 MASP E2E tests are passing when they should be failing. That the tests should be failing can be established from the fact that trying to manually construct shielded-to-shielded transactions with `namadac` (almost?) always fails due to issues with the MASP transaction signer, and also because the MASP E2E tests (almost?) always fail when executed locally.

Interestingly when the aforementioned issues with `namadac` are fixed, the local E2E tests start passing but the 3 MASP E2E tests in the CI begin to fail. For reference, these fixes are at https://github.com/anoma/namada/blob/5504002ff6d3c234c78afe64df2a1417653912e5/apps/src/lib/client/signing.rs#L157 and https://github.com/anoma/namada/blob/5504002ff6d3c234c78afe64df2a1417653912e5/apps/src/lib/client/tx.rs#L1909 .

A solution to this issue would prevent future regressions in the MASP code. cc @Fraccaman @gijswijs

cwgoes:
Closing since this seems to be out-of-date, please reopen if I'm mistaken.

Issue #1374, opened 2023-05-15
Unsafe MASP Transaction Validation


murisi:
Some of the code executed during MASP transaction validation is unsafe and could be exploited to crash a node. More specifically, the following lines may be problematic:
* Spend description zero-knowledge proof reading at: https://github.com/anoma/namada/blob/8e830ef7e482caa539a4824dc1a3ed298c5f2d09/shared/src/ledger/masp.rs#L101 . A client may inadvertently or maliciously place an invalid spend description ZK-proof into a transaction causing a node to crash. More specifically, placing invalid group elements into the ZK-proof could trigger crashing.
* Analogous issue to above, but for output descriptions at: https://github.com/anoma/namada/blob/8e830ef7e482caa539a4824dc1a3ed298c5f2d09/shared/src/ledger/masp.rs#L121
* Analogous issue to above, but for convert descriptions at: https://github.com/anoma/namada/blob/8e830ef7e482caa539a4824dc1a3ed298c5f2d09/shared/src/ledger/masp.rs#L138
* Spend authorization signature extraction at: https://github.com/anoma/namada/blob/8e830ef7e482caa539a4824dc1a3ed298c5f2d09/shared/src/ledger/masp.rs#L108 . A missing spend authorization signature in a transaction spend description may cause a node to crash.

Additionally, while the type conversion of sighashes (at https://github.com/anoma/namada/blob/8e830ef7e482caa539a4824dc1a3ed298c5f2d09/shared/src/ledger/masp.rs#L156 ) is probably valid, it will crash a node if the MASP crate returns a vector with unexpected length. Maybe it would be better to just log an error and reject a transaction triggering this (hopefully impossible) condition?

These are the unsafe operations (acting on untrusted inputs) I've seen so far. cc @gijswijs @juped

grarco:
@murisi I actually think this has already been solved and can be closed

Issue #1255, opened 2023-03-27
Restructure Namada Transactions


murisi:
Namada transactions are similar to executable files in that they contain executable code (the WASM) and initialized data (Transfer, Bond, Unbond, ...). Continuing the analogy, execution of Namada transactions consists of running the WASM code on the Namada VM with the initialized data being accessible. Therefore the techniques used to organize and design executable file formats may be applicable to Namada's transactions.

# Current Approach
Before considering other executable file formats, it makes sense to describe the current approach we take:

* Outer transaction:
    * Signed data:
        * Unsigned data: Wrapper transaction:
            * Gas payment info
            * Fee payer info
            * Inner transaction hash over:
                * Hash(inner tx wasm_code)
                * Hash(inner_tx extra_data)
                * Inner_tx signed data
        * Sig: Wrapper transaction signature over:
            * Unsigned data
    * Inner tx: Encrypted inner transaction:
        * WASM code
        * Extra data
        * Signed data:
            * Unsigned data
            * Signature over:
                * Hash(wasm_code)
                * Hash(extra_data)
                * Unsigned data

## Advantages
This approach is done in a way that allows:
* hardware constrained wallets to sign over only WASM code hashes and extra data hashes
* inner transactions to be fully encrypted after all wallet signing operations are concluded
* all signing operations to be done by a hardware wallet with one communication round
* inner and outer transactions to be signed by potentially different parties

## Drawbacks
This approach as it's currently implemented works well enough, but it seems to have the following drawbacks:
* The nested treatment of transactions by the protocol means that VPs cannot access gas payment info due to wrapper and outer layers having been stripped
* The reuse of the `Tx` structure for both inner and outer transactions means that some of the structure's fields are redundant in either the inner/outer context
* The nested (de)serialization of transactions (into the `data` field) does not map nicely onto existing protocol message formats like protobufs
* The nested (de)serialization of transactions subverts the Rust type system by replacing structures witth byte vectors
* Full decryption and decoding of an entire is rather cumbersome as it necessarily involves deserializations of its byte vector subcomponents
* Conceptual overloading of the transaction concept (inner, outer, wrapper) makes discussions and reasoning about them slightly harder
* Removal of subsections of a transaction for consumption by certain consumers (like hardware constrained wallets) is cumbersome

# Alternative Approach
[...]

PR #1462, opened 2023-05-24
Murisi/restructure tx draft

murisi:
This is an attempt to achieve #1255 with the tip of `main` merged in. Overall the main changes introduced by this branch are:
- Instead of having an outer transaction structure containing an inner transaction, there is now a single transaction structure containing a flat list of sections.
- Instead of wrapping inner transactions in a structure that supplies a signature, the unsigned data is added as a section to the transaction at the same level as it corresponding signature section.
- The WASM code now receives a `Tx` object instead of a data byte vector. This is because the data section's signature is stored in a different section, and because the extra `data` also resides in a separate section.
- Encryption is now done separately after all signing is complete - in particular this means that there are no longer any signatures over encrypted data.
- Upgraded the MASP client and validation code to use the latest MASP crate.
- Added a builder section to `Tx`s where the auxiliary inputs that were used to construct a MASP transaction can be stored. This may be useful for hardware wallets when validating a given Transaction object.

Additional notes:
- Also addressed the issues in https://github.com/anoma/namada/issues/1374 .
- The MASP E2E tests yield false negatives in the CI. See: https://github.com/anoma/namada/issues/1376


murisi:
pls update wasm

core/src/proto/types.rs
```diff
@@ -38,416 +54,1077 @@ pub enum Error {
...
+    BorshDeserialize,
+    BorshSchema,
+    Serialize,
+    Deserialize,
+)]
+pub enum Section {
+    /// Transaction data that needs to be sent to hardware wallets
+    Data(Data),
+    /// Transaction data that does not need to be sent to hardware wallets
+    ExtraData(Code),
```
bengtlofgren, line 536: Is this meant to be `ExtraData(Code)`?
```suggestion
    ExtraData(Data),
```

core/src/proto/types.rs
```diff
@@ -38,416 +54,1077 @@ pub enum Error {
...
+)]
+pub struct Header {
+    /// The chain which this transaction is being submitted to
     pub chain_id: ChainId,
+    /// The time at which this transaction expires
     pub expiration: Option<DateTimeUtc>,
+    /// A transaction timestamp
+    pub timestamp: DateTimeUtc,
+    /// The SHA-256 hash of the transaction's code section
+    pub code_hash: crate::types::hash::Hash,
```
bengtlofgren, line 689: This may be a privacy concern, as mentioned by @Fraccaman

murisi: The `code_hash` field of `Header` is salted. That is, the WASM code hash is appended to a salt, and then hashed in order to not reveal what kind of transaction is being done. See: https://github.com/anoma/namada/blob/11a272b343e1067fe2e0c7e4673f0917496ea88e/core/src/proto/types.rs#L162 . cc @Fraccaman

core/src/proto/types.rs
murisi: Yes, it is meant to be `ExtraData(Code)`. The reason why is that the extra data section shares a greater resemblance to the the code section than it does to the data section. More specifically, extra data sections frequently contain too many bytes, just like the WASM code in code sections are frequently too large. Hence both extra data and code sections are "compressed" by first taking the SHA-256 hash of their payload and putting that into the section instead. This is in contrast to the data section, which is mostly unique to each transaction, and is never compressed because hardware wallets need to render all contained data (which would not be possible with just a hash).

PR #1464, opened 2023-05-24
Murisi/fix dry run header

murisi:
Now make sure that transactions in dry run are being wrapped by a `DecryptedTx` as it was done before.


murisi:
Superseded by #1524 .

PR #1474, opened 2023-05-25
sdk: use hash instead of wasm blob

Fraccaman:
based on #1462


tzemanovic (dismissed):

Fraccaman:
e2e are not passing due to #1376. We have a fix in #1472.

PR #1491, opened 2023-06-01
apps: add namadac utils default-base-dir

juped:
Add a utility command to print the default base directory that would be
used by the Namada command line if a base directory were not provided
on the command line or in the environment.

### DIFF
--- a/apps/src/lib/client/utils.rs
+++ b/apps/src/lib/client/utils.rs
@@ -27,7 +27,7 @@ use crate::config::genesis::genesis_config::{
     self, GenesisConfig, HexString, ValidatorPreGenesisConfig,
 };
 use crate::config::global::GlobalConfig;
-use crate::config::{self, Config, TendermintMode};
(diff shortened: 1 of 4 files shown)


apps/src/lib/client/utils.rs
```diff
@@ -901,6 +901,18 @@ pub fn pk_to_tm_address(
...
 }
 
+pub fn default_base_dir(
+    _global_args: args::Global,
+    _args: args::DefaultBaseDir,
+) {
+    println!(
+        "{}",
+        get_default_namada_folder().to_str().expect(
+            "expected a default namada folder to be possible to determine"
```
gijswijs, line 911: <pedantic>Error message is a bit cryptic</pedantic>

gijswijs (approved):
In theory we should have a unit test even for small features like this.

apps/src/lib/client/utils.rs
bengtlofgren: Yeah +1 I'd like to see something else

PR #1492, opened 2023-06-01
base dir works with 1491

bengtlofgren:
Still need to change `trash` to `rm`


bengtlofgren:
> Still need to change `trash` to `rm`

This was done

PR #1498, opened 2023-06-02
Murisi/separate signing

murisi:
Currently transaction building, signing, and submission are all combined into single functions in the SDK. This setup makes it hard for web clients using the SDK to elect to use a hardware wallet to sign transactions instead. So this PR separates the transaction building operation from the signing and the submission operations in the SDK.


apps/src/lib/client/tx.rs
```diff
@@ -446,36 +446,99 @@ impl masp::ShieldedUtils for CLIShieldedUtils {
...
+            ProcessTxResponse::Applied(resp) if
+            // If a transaction is shielded
+                tx_epoch.is_some() &&
+            // And it is rejected by a VP
+                resp.code == 1.to_string() &&
+            // And the its submission epoch doesn't match construction epoch
+                tx_epoch.unwrap() != submission_epoch =>
+            {
+                // Then we probably straddled an epoch boundary. Let's retry...
+                eprintln!(
```
adrianbrink, line 492: we can merge this but all these `eprintln` statements make this hard to use, similar to `panics` because consumers can't catch the errors to display useful error messages to users.

adrianbrink (dismissed):
I'm okay with these changes

jurevans (approved):
This looks good to me! For reference, here is how we are utilizing these changes:

https://github.com/anoma/namada-interface/blob/23b0dd5499fc676759b80be7dd836df4cc2de3ae/packages/shared/lib/src/sdk/mod.rs#L99-L201

PR #1510, opened 2023-06-06
established addr use full sha256 hash for generation

### LINKED ISSUE #364: Ensure we're utilizing full SHA256 digests in our `Address` types
Currently we only use the first 40 bytes of the 64 byte hex encoding of the 32 byte SHA256 digest. We should instead use all of the digest i.e. the entire 32 bytes, not hex encoded.

e.g. when generating an `EstablishedAddress`

https://github.com/anoma/namada/blob/d1a06a26abcfc8c3c0fd8904c020a79fac1ac52d/shared/src/types/address.rs#L372-L386

Also see https://github.com/anoma/namada/issues/112 for extra context

tzemanovic:
based on v0.17.0

closes #364 

We cannot use the full 32 bytes for the established address type without increasing the bech32m address length, but imho it's not necessary. Instead, this PR replaces the `String` with `[u8; 20]` containing the first 20 bytes of the string.

The `EstablishedAddressGen` now uses the full sha256 digest.


tzemanovic:
pls update wasm

yito88 (approved):

PR #1512, opened 2023-06-06
implicit addr/pkh bytes

tzemanovic:
based on #1510

Switched the PKH to hold first 20 bytes of the PK hash instead of a hex string of the same


tzemanovic:
pls update wasm

yito88 (approved):

PR #1513, opened 2023-06-06
fix: do not add address if it already exists

### LINKED ISSUE #1504: `address add` may overwrite an existing alias
Using `address add --alias <alias> --address <address>` may re-assign address to an alias different from the alias of the respective key.

karbyshev:
Closes #1504

### DIFF
--- a/.changelog/unreleased/bug-fixes/1504-wallet-address-add-fix.md
+++ b/.changelog/unreleased/bug-fixes/1504-wallet-address-add-fix.md
@@ -1,0 +1,2 @@
+- Do not add address if it already exists in the wallet.
+  ([\#1504](https://github.com/anoma/namada/issues/1504))


shared/src/ledger/wallet/store.rs
```diff
@@ -504,6 +504,15 @@ impl Store {
         address: Address,
         force: bool,
     ) -> Option<Alias> {
+        // abort if the address already exists in the wallet
+        if self.addresses.contains_right(&address) {
```
tzemanovic, line 508: I think this should respect the `force` flag

shared/src/ledger/wallet/store.rs
```diff
@@ -504,6 +504,15 @@ impl Store {
...
+            println!(
+                "Address {} already exists in the wallet.",
```
tzemanovic, line 510: let's also print the alias if it exists

shared/src/ledger/wallet/store.rs
karbyshev: Thanks! Fixed that.

shared/src/ledger/wallet/store.rs
karbyshev: Thanks! Added.

tzemanovic (approved):

PR #1516, opened 2023-06-06
fix: do not panic if the user skips the changes

### LINKED ISSUE #1501: Skipping creation of key with the same alias throws error
Assume you have a key in the wallet with alias `aliace`.
Now, if you execute:

`namadaw key gen --alias aliace` 

you will run into the following dialogue: 

```
You're trying to create an alias "aliace" that already exists for a key in your store.
Would you like to replace it? s(k)ip/re(p)lace/re(s)elect: k
```

However, instead of completing this silently (as expected), we receive

```
Message:  Action cancelled, no changes persisted.
Location: /Users/unknowit/Heliax/namada/shared/src/ledger/wallet/store.rs:269
```

karbyshev:
Closes #1501

### DIFF
--- a/shared/src/ledger/wallet/store.rs
+++ b/shared/src/ledger/wallet/store.rs
@@ -235,6 +235,8 @@ impl Store {
     /// Optionally, use a given random seed and a BIP44 derivation path.
     /// Returns the alias of the key and a reference-counting pointer to the
     /// key.
+    /// Returns None if the alias already exists and the user decides to skip
+    /// it. No changes in the wallet store are made.
     pub fn gen_key<U: WalletUtils>(
         &mut self,
         scheme: SchemeType,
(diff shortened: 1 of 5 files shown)


shared/src/ledger/wallet/store.rs
```diff
@@ -257,24 +259,19 @@ impl Store {
         let (keypair_to_store, raw_keypair) = StoredKeypair::new(sk, password);
         let address = Address::Implicit(ImplicitAddress(pkh.clone()));
         let alias: Alias = alias.unwrap_or_else(|| pkh.clone().into()).into();
-        if self
```
karbyshev, line 260: I think there was a bug here: in the case of alias collision, the user may modify the alias for the keypair, while the originally provided alias have been returned.

tzemanovic: ah yes, it was not possible to pick a different alias

tzemanovic (approved):

PR #1520, opened 2023-06-08
Fix slashable stake in slash processing

brentstone:
Based on v0.17.0.

In the procedure for processing slashes, we must compute the slashable stake for an epoch at or ahead of the current epoch in the presence of bonds that began contributing to stake after the infraction epoch. A slight change is needed to fix buggy behavior in the existing implementation of this computation. See the following discussion for more details: https://github.com/informalsystems/partnership-heliax/pull/55#discussion_r1209863459.


brentstone:
Need to also change this implementation in the AbstractState of the state machine test

proof_of_stake/src/tests/state_machine.rs
```diff
@@ -2265,50 +2280,60 @@ impl AbstractPosState {
...
                         tracing::debug!(
                             "Total unbonded (epoch {}) w slashing = {}",
                             epoch,
                             total_unbonded
                         );
                     }
+                    sum_post_bonds += self
+                        .total_bonded
+                        .get(&validator)
+                        .unwrap()
```
tzemanovic, line 2332: There are cases in which this panics when a validator doesn't yet have any stake but it's still in consensus set as it's not yet full with non-zero stake validators. I think we can just change this to `.and_then(|bonded| bonded.get(&Epoch(epoch)))` to get around it and then we default to 0 if it's not set

proof_of_stake/src/tests/state_machine.rs
```diff
@@ -2405,11 +2433,15 @@ impl AbstractPosState {
...
                     tracing::debug!("Updating ABSTRACT voting powers");
-                    let sum_post_bonds = self.get_validator_bond_sums(
-                        &validator,
-                        infraction_epoch.next(),
-                        self.epoch + offset,
-                    );
+                    sum_post_bonds += self
+                        .total_bonded
+                        .get(&validator)
+                        .unwrap()
```
tzemanovic, line 2439: same here

proof_of_stake/src/tests/state_machine.rs
tzemanovic: done in [7b4e333](https://github.com/anoma/namada/pull/1520/commits/7b4e3335cb50994e163a85549e9238c84c7b239b)

proof_of_stake/src/tests/state_machine.rs
tzemanovic: [7b4e333](https://github.com/anoma/namada/pull/1520/commits/7b4e3335cb50994e163a85549e9238c84c7b239b)

tzemanovic:
f-pushed https://github.com/anoma/namada/compare/7b4e3335cb50994e163a85549e9238c84c7b239b..55d63907820c436ac10e3e3b7aa853176d2d7610 to squash fixup

tzemanovic:
pls update wasm

PR #1525, opened 2023-06-08
fix unreleased changelogs for 0.17

tzemanovic:


Fraccaman (approved):

Fraccaman:
should we merge this directly to main or put it in draft? I think main is enough

tzemanovic:
> should we merge this directly to main or put it in draft? I think main is enough

Can go to main I think

PR #1533, opened 2023-06-11
Fix slashing in client queries for bond and unbond info

brentstone:
Based on v0.17.0

The PoS library function `bonds_and_unbonds` is primarily used in client queries of PoS system information and provides bond and unbond information, including known slashes that affect these bonds. The computation of the slashed token amounts corresponding to these bonds and unbonds was outdated and is fixed in this PR.


proof_of_stake/src/lib.rs
```diff
@@ -2708,36 +2710,40 @@ fn make_bond_details(
...
+    let validator_slashes =
+        applied_slashes.entry(validator.clone()).or_default();
+    for slash in slashes {
+        if slash.epoch >= start {
+            let cur_rate = slash_rates_by_epoch.entry(slash.epoch).or_default();
+            *cur_rate = cmp::min(Decimal::ONE, *cur_rate + slash.rate);
+
+            if !prev_applied_slashes
+                .iter()
+                .any(|s| s.clone() == slash.clone())
```
tzemanovic, line 2724: this hasn't changed but we can compare the refs without cloning here

proof_of_stake/src/lib.rs
```diff
@@ -2708,36 +2710,40 @@ fn make_bond_details(
...
+            {
+                validator_slashes.push(slash.clone());
+            }
+        }
+    }
+
+    let amount_after_slashing = token::Amount::from_change(
+        get_slashed_amount(params, amount, &slash_rates_by_epoch).unwrap(),
+    );
+    let slashed_amount = if amount == amount_after_slashing {
```
tzemanovic, line 2734: let's instead call `get_slashed_amount` only when `!slash_rates_by_epoch.is_empty()` as it does some alloc that we can avoid most of the time

proof_of_stake/src/lib.rs
tzemanovic: done in 65e375c5c44873b9df48482bff0316bf3fc3b54a

proof_of_stake/src/lib.rs
tzemanovic: also in 65e375c5c44873b9df48482bff0316bf3fc3b54a

tzemanovic:
pls update wasm

tzemanovic:
f-pushed to squash fixups

tzemanovic (approved):

PR #1534, opened 2023-06-12
fix block query to only return committed blocks

### LINKED ISSUE #1503: e2e::ledger_tests::test_genesis_validators spurious failure
Sometimes the 2nd validator node query fails with not seeing the up-to-date state expected after the tx: https://github.com/anoma/namada/actions/runs/5172848251/jobs/9317577241

### LINKED ISSUE #1532: latest block query at pre-commit
When we issue latest block query via Tendermint after proposal pre-commit but before actual commit, it already returns the new update block height for the pre-commit. In Namada, proposal state is not being cached and so we don't consider the pre-commit height as the latest height.

This is not a problem most of the time, but it has started to happen frequently in 0.17 in a e2e test such as https://github.com/anoma/namada/issues/1503 in which we're checking that the state on 2 validator nodes and 1 non-validator node is updated following a transaction and we're using the latest block query to wait for them to be in sync. However, because TM may return a block height from pre-commit which is not actually yet committed, the assertion for the state fails.

Unfortunately it's not visible from the TM `/block` query (https://docs.tendermint.com/main/rpc/#/Info/block) if it's pre-commit or commit. To fix this, we can instead query the latest block height via a custom Namada handler from `Storage`, where the height is only updated on commit.

tzemanovic:
based on 0.17

fixes #1532, fixes #1503


tzemanovic:
pls update wasm

Fraccaman (approved):

PR #1549, opened 2023-06-12
Update TendermintMode in Namada config for new post-genesis validator

### LINKED ISSUE #1542: `config.toml` inside `$BASE_DIR/$CHAIN_ID/config.toml` does not update
As soon as you send a `namadac init-validator` tx, we should change the `tendermint_mode = "Validator"`. At the moment, it is stuck on `tendermint_mode = "Full"`. 

Note that tendermint is correctly configued as a validator.

brentstone:
Based on v0.17.0. Closes #1542.


tzemanovic (approved):

tzemanovic:
The state machine test failure is most-like a problem with pre-conditions because `get_total_consensus_stake` ends up being zero. It's not likely caused by the changes introduced here

Fraccaman (approved):

PR #1553, opened 2023-06-13
PoS: fix is_validator fn to work for new validators

### LINKED ISSUE #1551: PoS: is_validator check fails for new validators
In #787 (done in #740) we removed the pending validator state and only set the state to the validator set placement at pipeline offset, but the `fn is_validator` is relying on it being set to determine if the given address is a validator. This means that when a new validator is created, this will return false until pipeline offset. This function should be agnostic to epochs. Instead of validator set, we can use non-epoched max commission rate field which is required for validators.

tzemanovic:
based on v0.17.1

fixes #1551

### DIFF
--- a/proof_of_stake/src/lib.rs
+++ b/proof_of_stake/src/lib.rs
@@ -808,14 +808,12 @@ where
 pub fn is_validator<S>(
     storage: &S,
     address: &Address,
-    params: &PosParams,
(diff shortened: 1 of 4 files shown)


tzemanovic:
pls update wasm

brentstone (approved):
LGTM

PR #1555, opened 2023-06-13
improve wasm tx commission rate test

### LINKED ISSUE #1554: PoS: tx_change_commission_rate - improve rate gen to avoid filter
"New rate must not be equal to the previous epoch's rate" filter applied on the generated rate may fail when the prop_filter rejects the input too may time as occurred in https://github.com/anoma/namada/actions/runs/5252337342/jobs/9488434496. We can generate a new rate without using a filter from the prev rate by adding or subtracting it to/from it

tzemanovic:
based on 0.17.1

fixes #1554


brentstone (approved):
LGTM! Thx, this is nicer.

PR #1576, opened 2023-06-15
Add below-threshold validator set

### LINKED ISSUE #1128: PoS: add `below_threshold` validator sub-set
as specified here https://specs.namada.net/economics/proof-of-stake/bonding-mechanism.html#validator-sets

brentstone:
Based on v0.17.4.

Closes #1128.

This PR implements a third validator set in addition to the `consensus` and `below-capacity` sets: the `below-threshold` set. All validators with stake below the `validator_stake_threshold` parameter are placed into this set, which is unordered, unlike the other two validator sets. Smaller amounts of data are needed to be kept in storage for validators in this set.


brentstone:
Several unit tests break because `arb_genesis_validators` generates genesis validators whose staked tokens are all below the `validator_stake_threshold`. Some hackier ways of solving this is to make the `validator_stake_threshold` 0 or 1 namnam or to just lower the threshold to some value where it is less likely to arbitrarily generate a set of validators all of whose stake is below this number.

The more rigorous solution is to modify `arb_genesis_validators` to take a `PosParams` as input. Then if the generated validators all have stake less than `validator_stake_threshold`, we can generate one additional genesis validator with random stake that is assured to be at least as large as this.

@tzemanovic

proof_of_stake/src/lib.rs
```diff
@@ -250,6 +250,12 @@ pub fn below_capacity_validator_set_handle() -> BelowCapacityValidatorSets {
     BelowCapacityValidatorSets::open(key)
 }
 
+/// Get the storage handle to the epoched below-threshold validator set
+pub fn below_threshold_validator_set_handle() -> BelowThresholdValidatorSets {
```
tzemanovic, line 254: I think we can avoid storing the 3rd set altogether as we don't need to keep an order for these validators. When the validator state is `BelowThreshold` it should be in neither consensus or below_capacity set. This will reduce storage cost and we won't need to copy these on a new epoch.

proof_of_stake/src/lib.rs
```diff
@@ -2432,15 +2432,7 @@ where
...
-                // If both previous and current voting powers are 0, skip
-                // update
-                if *prev_tm_voting_power == 0 && *new_tm_voting_power == 0 {
-                    tracing::info!(
-                        "skipping validator update, {address} is in consensus \
-                         set but without voting power"
-                    );
-                    return None;
-                }
+                // TODO: maybe debug_assert that the new stake is >= threshold?
```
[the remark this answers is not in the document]
tzemanovic: same with `if prev_tm_voting_power == 0 ...` for below capacity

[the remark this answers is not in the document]
brentstone: enforce with debug_asserts or by returning `None` values?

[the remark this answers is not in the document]
brentstone: ALSO, do we still need this `Lazy` type wrapped around `prev_tm_voting_power` and `new_tm_voting_power`?

brentstone:
pls update wasm

proof_of_stake/src/lib.rs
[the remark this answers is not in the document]
tzemanovic: I think we could enforce in the paramaters `PosParams::validate` that `validator_stake_threshold` and `tm_votes_per_token * u64::from(validator_stake_threshold)` are non-zero - only thing is that we're currently not using this function but we should probably fix that anyway. We could use it for genesis parameters validation and for on-chain parameters change, which is only allowed via governance.

If we enforce these invariants on the params then we can remove the check in here as we couldn't have a validator with 0 TM voting to enter the consensus set. That being said, what you've done LGTM and I don't think we need to change this immediately.

I think we should keep `Lazy` as the first condition `if matches!(prev_state, Some(ValidatorState::Consensus))` might short-circuit in which case we don't need to compute these.

proof_of_stake/src/lib.rs
```diff
@@ -695,6 +695,31 @@ where
...
+        .iter(storage)?
+        .map(Result::unwrap)
+        .filter(|address| {
+            if let Ok(Some(state)) =
+                validator_state_handle(address).get(storage, epoch, &params)
+            {
+                state == ValidatorState::BelowThreshold
+            } else {
+                false
+            }
```
tzemanovic, line 718:
```suggestion
            matches!(
                validator_state_handle(address).get(storage, epoch, &params),
                Ok(Some(ValidatorState::BelowThreshold))
            )
```

proof_of_stake/src/lib.rs
```diff
@@ -2230,69 +2455,54 @@ where
...
-                tracing::debug!(
-                    "skipping validator update {address}, it's inactive and \
-                     previously had no voting power"
+            // If the validator is still in the Consensus set, we accounted for
+            // it in the `new_consensus_validators` iterator above
+            if matches!(new_state, Some(ValidatorState::Consensus)) {
+                return None;
+            } else if params.validator_stake_threshold
+                == token::Amount::default()
+                && prev_tm_voting_power == 0
```
tzemanovic, line 2492: We can make `prev_tm_voting_power` lazy to avoid the computation if the prev conditions short-circuit

proof_of_stake/src/tests.rs
```diff
@@ -1930,30 +1948,55 @@ fn advance_epoch(s: &mut TestWlStorage, params: &PosParams) -> Epoch {
...
-                GenesisValidator {
-                    address,
-                    tokens,
-                    consensus_key,
-                    commission_rate,
-                    max_commission_rate_change,
+    let tokens = tokens.prop_map(move |mut amounts| {
+        if let Some(thresh) = threshold {
+            if !amounts.iter().any(|&x| x >= thresh) {
+                let mut rng = rand::thread_rng();
```
tzemanovic, line 1959: Having the rng here that's not seeded from proptest makes this non-deterministic and that could be a hindrance for possible issues caught by the tests that are using this. I think we can write this slightly differently - I'll push a fixup

tzemanovic: [3100bfc](https://github.com/anoma/namada/pull/1576/commits/3100bfc41155c19aaf7f996245acba4de92c4281)

proof_of_stake/src/lib.rs
tzemanovic: added in [19f5aa3](https://github.com/anoma/namada/pull/1576/commits/19f5aa34e26f0f82a7c57846143cf3486543e44b)

tzemanovic (dismissed):

tzemanovic:
I'll squash the fixup commits and add this to draft

proof_of_stake/src/tests.rs
brentstone: ah good point, thx, didn't consider this before

brentstone:
Merged in v0.18.0.

PR #1582, opened 2023-06-16
PoS: delay slash validator set update when next epoch is done in TM

tzemanovic:
Based on v0.17.4.

As caught by e.g. https://github.com/anoma/namada/actions/runs/5280174696/jobs/9551903699?pr=1569

When the epoch change conditions are met, we send validator set update to TM, which takes 2 blocks before the change applies. To stay in sync with TM, we delay epoch update by 2 blocks as well. However, when we receive slashable evidence and the validator set update has already been applied but we're not yet on the new epoch, we have to queue up slashed validator removal from validator set by another epoch.


brentstone (approved):
LGTM (pending new CI results based on v0.17.4)

PR #1583, opened 2023-06-16
portable build_network.sh

juped:
- scripts/build_network: remove spurious field separators
  The default value of $IFS (internal field separator) is space, tab,
  and newline. When de-bashifying this script, this line will change
  its meaning to make the literal characters 'n' and 't' field
  separators, causing heinous errors (for example, the word "network"
  is likely to show up, which contains an 'n' and a 't').

  But the line serves no purpose here, so remove it.

- scripts/build_network: port to posix shell
  Bash is a popular interactive shell, but it is not a suitable target
  for shell scripts. Remove bashisms.

  Tested on dash (Debian Almquist Shell) but not extensively across
  systems.

### DIFF
--- a/scripts/build_network.sh
+++ b/scripts/build_network.sh
@@ -1,10 +1,9 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # Script for initializing a Namada chain for local development and joining it.
 # Note that this script trashes any existing local chain directories!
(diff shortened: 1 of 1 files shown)


bengtlofgren (approved):
lgtm though i dunno what's going on lmao

PR #1599, opened 2023-06-20
wasm secp256k1 sig verification

### LINKED ISSUE #1595: wasm secp256k1 sig verification is missing
Trying to use secp256k1 signatures fails verification in WASM VPs with a runtime error. This is because the `fn verify_signature` has conditional compilation for `feature = "secp256k1-sign-verify"`, which controls whether the `hmac` feature is enabled for libsecp256k1 (https://github.com/paritytech/libsecp256k1). However, `hmac` is only needed for creating signatures not for verifying them, so we can enable these. We should make sure to have a test case for tx signed with a secp256k1 key

tzemanovic:
fixes #1595


tzemanovic:
pls update wasm

PR #1605, opened 2023-06-21
apps: use fd-lock instead of file-lock for cross-platform support

tzemanovic:


apps/src/lib/wallet/pre_genesis.rs
```diff
@@ -36,70 +36,72 @@ pub fn gen_and_store(
...
+        )
+    })?);
+    let guard = lock.read().map_err(|err| {
+        ReadError::ReadWallet(
+            wallet_file.to_string_lossy().into_owned(),
+            err.to_string(),
+        )
+    })?;
+    let mut store = Vec::<u8>::new();
+    (&*guard).read_to_end(&mut store).map_err(|err| {
```
sug0, line 66: hm, the compiler isn't able to figure out the proper deref here (`&*guard`)?

sug0 (commented):
lgtm, just a question

apps/src/lib/wallet/pre_genesis.rs
tzemanovic: Unfortunately not because `read_to_end` takes `&mut self`, but the `trait Read` is implemented for `&File`. I think it's probably trying to auto-deref as `&mut self` which it can't

sug0 (approved):

Issue #1567, opened 2023-06-14
Possible malleability attack on tx encrypted sections


grarco:
About the commitment of the header to the sections. There are only the hashes of the code and data section in the header and other commitments are placed in the `Data` section if needed. This looks good but there might be room for an attack on the encrypted sections.

Fees are paid when the entire tx is included in the block: in the following block we decrypt the sections and do validation. If this validation fails we can avoid executing the transaction but the problem is that the wrapper’s signer has already paid the fees, suffering an economic damage.

grarco:
Yes, I think this issue applies to all of the sections that are included in transaction. If I recall correctly what you explained me, we "encrypt" all of the sections except for the signature on the `Header`. Except for this signature, I think all of the other sections need a commitment in the header to prevent tampering. Note that this doesn't have anything to do with the encryption per se, we'd need the commitment even if all of the sections were in plaintext

cwgoes:
MERGED :rocket:

PR #1607, opened 2023-06-21
Murisi/fix tx malleability

murisi:
An attempt to fix the transaction malleability described at #1567 . The approach taken is as follows:
* Enable signature sections to sign over multiple sections as specified in a `Vec<Hash>`
* Now make the header signatures span over all the other transactions sections in addition to the header hash
* The verify signature algorithm now checks that all sections referenced in a signature are still present in the transaction

Additionally the signatures over data and code sections have now been combined to eliminate the possibility of the wrapper signer somehow changing the inner WASM code that is supposed to execute over the signed data in the inner transaction. That is, the inner transaction code and data are now bound together by a signature.


Fraccaman (approved):

Fraccaman:
LGTM! this is more or less how I was thinking this problem could be fixed. We need better API to construct a tx but we can manage in another PR :) thanks!

core/src/proto/types.rs
```diff
@@ -1002,7 +1033,7 @@ impl Tx {
         match &self.header.tx_type {
             // verify signature and extract signed data
             TxType::Wrapper(wrapper) => {
-                self.verify_signature(&wrapper.pk, &self.header_hash())
+                self.verify_signature(&wrapper.pk, &[self.header_hash()])
```
grarco, line 1036: Should we validate the signature over the header and also all of the sections? I see that in `sign_wrapper` we initialize the vector of hashes with the header hash and then push the hashes of all the sections included in the tx. I imagined we would the same here

murisi: Good point, thanks. Currently we're just checking for a valid signature over the header, which indirectly implies a check that all sections targetted in the signature exist (i.e. haven't been removed or tampered with). But we should check that all sections in a transaction are signed over - this would prevent the addition of new sections to a transaction. Let me work on that.

grarco (dismissed):
Looks good to me now, thanks!

core/src/proto/types.rs
```diff
@@ -998,35 +1044,35 @@ impl Tx {
     /// the Tx and verify it is of the appropriate form. This means
     /// 1. The wrapper tx is indeed signed
     /// 2. The signature is valid
-    pub fn validate_header(&self) -> std::result::Result<(), TxError> {
+    pub fn validate_tx(
```
tzemanovic, line 1047: the rustdoc on this is a bit outdated

tzemanovic (dismissed):

PR #1611, opened 2023-06-22
Murisi/move init proposal content

murisi:
Increased the variety of test vectors in order to facilitate the implementation of hardware wallets. Also some minor modifications to certain transaction encodings to enable processing on hardware constrained wallet devices. More specifically:
* Moved the `content` and part of the `type` field of `InitProposalData` out into `ExtraData` sections so that hardware wallets do not have to process these sections.
* Added a CLI subcommand to change the validator commission rate. This was necessary in order to be able to generate test vectors for `CommissionChange` transactions
* Modified the test vector script to generate `ChangeCommission` transactions, and also increased the variety of `InitProposalData` and `VoteProposalData` transactions
* Made the test vector expert output for MASP transactions more intuitive by replacing the transparent MASP pool address with specific full viewing keys and payment addresses


brentstone (commented):
Validator commission rate change things LGTM

PR #1612, opened 2023-06-23
PoS redelegation

### LINKED ISSUE #34: PoS add re-delegation
Add a transaction for re-delegation (using a version with longer delay, which has to wait for unbonding epoch, but simpler to implement than faster re-delegation).

impl depends on https://github.com/anoma/namada/issues/124

### LINKED ISSUE #1556: PoS/test: division by zero
Trace: https://github.com/anoma/namada/actions/runs/5253847650/jobs/9491877974?pr=1549

The state machine test failure is most-like a problem with pre-conditions because `get_total_consensus_stake` ends up being zero.

### LINKED ISSUE #1640: Add counter to check proportional infraction punishment property
Per discussion with @brentstone.

In order to ensure the property that any bond contributing to an infraction is punished proportionally to the stake it committed to that infraction as the time of the infraction (_not_ the time of detection), we do some intricate calculations. Due to delayed detection, it may be possible for the stake amounts to be slashed "below zero", but we do checked subtraction to ensure that this doesn't happen. However, that makes it difficult to check correct adherence to this property. I think we can keep a counter of "overslashed" stake, per delegator, validator, and epoch, and use this to check correct adherence to this property in the state machine tests.

### LINKED ISSUE #1829: PoS: slashing an active validator holding redelegated tokens
State machine testing in #1612 has revealed an issue with the slashing of redelegated tokens.

In Quint (https://github.com/informalsystems/partnership-heliax/pull/64), the function `endOfEpoch` contains a variable `mapValidatorSlash` that the contains the amounts to slash in a given epoch for a given validator. The misbehaving validator as well as any destination validators holding redelegated tokens originally used to commit the infraction are slashed.

The map is constructed such that tokens may be deducted from each validator starting in the following epoch and up through the pipeline epoch relative to the current one. In Namada, we execute this procedure at the beginning of the following epoch rather than at the end of the current one as in Quint, and so to be consistent, we may update the validator deltas starting at the current Namada epoch up through the epoch preceding the pipeline epoch.

This poses an issue in Namada. Now with redelegation, we may need to slash a validator that is not jailed and may be in either the consensus or below-capacity sets. In this case, we can no longer update the validator's stake in the current epoch (which we are in the very beginning of when processing slashes). We can shift the epochs in which we slash the stake to one more in the future, though even this will have considerations for a future DKG, since this would still be the first time in the protocol that we may update a consensus validator's stake before the pipeline offset. However, this can possibly be considered out of scope for Namada, which can function properly by delaying the slashing of such a validator's stake by one full epoch.

### LINKED ISSUE #1830: PoS round up slash rate mult with slashable amount
In PoS when processing slashes we avoid iterating over individual bonds and instead slash total validator stakes. Bonds only get slashes applied when withdrawing, so the decimal precision loss can cause the sum of individual slashes to be lower or greater than the total slashed amount.

E.g. Validator A has bond from delegators X and Y, 1 (micro) token each. If both are slashed at 30% the total slash is rounded to `2*0.3=1`. When slashing individual bonds on withdrawal they're both unslashed (`1*0.3=0`).

Similarly with slash rate at e.g. 60%, the total slashed would be `2*0.6=1` but individual `1*0.6=1` with sum `2`.

I still need to check this exactly against the code.

### LINKED ISSUE #1959: Change all token-related data that won't hold negative values to `token::Amount`
Some data like `bonds` use `token::Change`, which is an alias for an `I256` for the token amounts, despite the fact that it should never hold negative values. All such instances should be changed to `token::Amount`

brentstone:
Based on v0.23.0

Closes #34, closes #1829, closes #1830, closes #1556, closes #1959, closes #1640

Implements redelegation. We allow a delegator address to redelegate bonded tokens from one validator to another such that the bonded tokens will stop contributing to the original (source) validator's voting power and start contributing to the new (destination) validator's voting power at the pipeline epoch. Redelegated tokens remain slashable for unbonding length for any misbehavior committed by the source validator. Validator addresses can only bond to themselves and thus cannot redelegate.

Client cmd e.g.: `namadac redelegate --source-validator {} --destination-validator {} --owner {} --amount {}`

Note that until we resolve https://github.com/informalsystems/partnership-heliax/issues/74, multiple slashes per validator may result in overslashing, but this issue was already present before redelegation.

Notes on things to-do:
- [x] check the lazy storage bounds on `unbonds` and `redelegated_unbonds`, ensure consistency
- [x] estimate gas for new tx_redelegate
- [x] more redelegation unit tests? (i.e. overslashing)

Further testing goals:
- [x] State machine test
- [x] Unit test with complexity, a la `test_multiple_misbehaviors`, with redelegations
- [x] e2e test


tzemanovic:
pls run pos tests

tzemanovic:
pls run pos tests

tzemanovic:
pls run pos tests

tzemanovic:
pls run pos tests

tzemanovic:
pls run pos tests

tzemanovic:
pls run pos tests

brentstone:
pls run pos tests

tzemanovic (approved):

PR #1616, opened 2023-06-23
fix storage rollback

tzemanovic:
depends on #1615

The issue fixed in #1615 was affecting rollback because the iterator without prefix didn't find anything. There's another issue in rollback itself - a key-val that's deleted is not being restored as shown in [5ca18bc](https://github.com/anoma/namada/pull/1616/commits/5ca18bc344e1d1f3dbe220c760b2ee26a0b65cb4).


brentstone:
Merged in v0.18.0

PR #1619, opened 2023-06-26
checked masp params in node

juped:
- masp: deduplicate parameters loading
- masp: check size and hash of params in node

### DIFF
--- a/shared/src/ledger/masp.rs
+++ b/shared/src/ledger/masp.rs
@@ -4,7 +4,6 @@ use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::env;
 use std::fmt::Debug;
-use std::fs::File;
 #[cfg(feature = "masp-tx-gen")]
 use std::ops::Deref;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


shared/src/ledger/masp.rs
```diff
@@ -90,70 +87,37 @@ pub const OUTPUT_NAME: &str = "masp-output.params";
...
+    if !spend_path.exists() || !convert_path.exists() || !output_path.exists() {
+        let paths = masp_proofs::download_masp_parameters(None).expect(
+            "MASP parameters were not present, expected the download to \
+             succeed",
+        );
+        if paths.spend != spend_path
+            || paths.convert != convert_path
+            || paths.output != output_path
+        {
+            panic!(
```
adrianbrink, line 108: This needs to be fixed in the future. We shouldn't panic but rather return an error, because otherwise consumers of the SDK can't really handle these failures with error messages.

adrianbrink:
Did you verify that the hashes [here](https://github.com/anoma/masp/blob/main/masp_proofs/src/lib.rs#L58) are correct?

grarco (approved):

PR #1621, opened 2023-06-27
namadac epoch-sleep command

juped:
read the commit messages

### DIFF
--- a/.changelog/unreleased/improvements/1621-utils-next-epoch.md
+++ b/.changelog/unreleased/improvements/1621-utils-next-epoch.md
@@ -1,0 +1,2 @@
+- Added a command to wait for the next epoch: `client utils epoch-sleep`.
+  ([\#1621](https://github.com/anoma/namada/pull/1621))


tests/src/e2e/ledger_tests.rs
```diff
@@ -4432,11 +4433,15 @@ fn test_epoch_sleep() -> Result<()> {
     // 3. Use epoch-sleep to sleep for an epoch
     let args = ["epoch-sleep", "--node", &validator_one_rpc];
     let mut client = run!(test, Bin::Client, &args, None)?;
+    let reached_epoch = parse_reached_epoch(&mut client)?;
     client.assert_success();
 
     // 4. Confirm the current epoch is larger
+    // possibly badly, we assume we get here within 30 seconds of the last step
+    // should be fine haha (future debuggers: sorry)
```
tzemanovic, line 4441: I think 30 secs should be plenty even with very limited resources, but we don't have to add the check for equality

tzemanovic (dismissed):

tzemanovic:
> what should we do with this one?

let's merge

tzemanovic:
it's doable to make this a utils sub-command which would be preferable - I'll rebase it and make the change

tzemanovic (approved):

tests/src/e2e/ledger_tests.rs
```diff
@@ -4663,6 +4664,52 @@ fn implicit_account_reveal_pk() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_epoch_sleep() -> Result<()> {
```
tzemanovic, line 4668: the new test has to be added to `.github/workflows/scripts/e2e.json` to run in CI

PR #1626, opened 2023-06-28
Fix the hard coded epoch that we need to wait for.

mariari:
Before it waited for a hard coded amount 7, which is not correct. In reality it should be waiting until 13, seen from this string

Amount 412 withdrawable starting from epoch 13

However even if this changes in the future, this regex will correctly get it.

### DIFF
--- a/tests/src/e2e/ledger_tests.rs
+++ b/tests/src/e2e/ledger_tests.rs
@@ -2202,7 +2202,7 @@ fn pos_rewards() -> Result<()> {
 /// 3. Wait for epoch 4
 /// 4. Submit another delegation to the genesis validator
 /// 5. Submit an unbond of the delegation
-/// 6. Wait for epoch 7
+/// 6. Wait for epoch withdraw_epoch
 /// 7. Check the output of the bonds query
 #[test]
 fn test_bond_queries() -> Result<()> {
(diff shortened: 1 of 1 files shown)


brentstone (approved):
LGTM

tests/src/e2e/ledger_tests.rs
```diff
@@ -2317,6 +2317,11 @@ fn test_bond_queries() -> Result<()> {
...
     let mut client = run!(test, Bin::Client, tx_args, Some(40))?;
     client.exp_string("Transaction is valid.")?;
+    let (_, res) = client
+        .exp_regex(r"withdrawable starting from epoch [0-9]+")
+        .unwrap();
+    let withdraw_epoch =
+        Epoch::from_str(res.split(' ').last().unwrap()).unwrap();
     client.assert_success();
 
     // 6. Wait for epoch 7
```
tzemanovic, line 2327:
```suggestion
    // 6. Wait for withdraw epoch
```

brentstone:
Should this be merged? @mariari

mariari:
> Should this be merged? @mariari

Yes

tzemanovic:
looks like this change got already incorporated via the amounts pr

PR #1627, opened 2023-06-28
Remove spin locks from E2E tests

mariari:
Previously the code would spin lock, creating busy CPU waiting, and creating a bunch of log files which are counterproductive when debugging, this topic removes all those issues


tzemanovic:
I think this should be addressed by #1621

tzemanovic:
> > I think this should be addressed by #1621
> 
> @tzemanovic do you know if 40 or 120 is a better timeout here?

I think for 30 sec epochs, 40 should be sufficient

tzemanovic:
closing in favor of #1621

PR #1629, opened 2023-06-28
wasmer alignment fix

tzemanovic:
based on #1604


tzemanovic:
the issue is still present with these changes

tzemanovic:
unsuccessful attempt to fix #1721

PR #1632, opened 2023-06-30
storage: add delete prefix

### LINKED ISSUE #1225: storage: consider adding `delete_prefix`
Akin to `prefix_iter`, for dealing with e.g. lazy collections it would be convenient to be able to delete anything with a matching storage prefix. We'd need to do this via WriteLog, but when the WriteLog is being committed to DB, there's `delete_range` in RocksDB that can potentially be used to then apply it.

This would be helpful for finalize-block's `apply_inflation` rewards accumulator added in #714.

tzemanovic:
based on v0.18.0

Closes #1225 

This depends on #1633 which needs to be fixed first as it's failing the test that's being added here.


brentstone (approved):
LGTM! Thx :)

PR #1656, opened 2023-07-04
query bonded-stake and order

bengtlofgren:


apps/src/lib/client/rpc.rs
```diff
@@ -1581,12 +1581,17 @@ pub async fn query_bonded_stake<C: namada::ledger::queries::Client + Sync>(
                         .await,
                 );
 
+            let sorted_consensus = consensus.into_iter().sorted_by_key(|v| {
```
tzemanovic, line 1584: I think we could probably just return `BTreeSet` from the set queries as the `WeightedValidator` already has order

shared/src/ledger/queries/vp/pos.rs
tzemanovic, line 1: in `consensus_validator_set` and `below_capacity_validator_set` we have some code duplication - we can just call the functions from pos/lib.rs

apps/src/lib/client/rpc.rs
tzemanovic: done in b3b1f261049f03cf1a70da7f378e806c47204500

tzemanovic (dismissed):

brentstone (approved):

PR #1667, opened 2023-07-05
converts faucet_withdrawal_limit to be correct

bengtlofgren:


wasm/wasm_source/src/vp_testnet_faucet.rs
```diff
@@ -9,6 +9,25 @@
 use namada_vp_prelude::*;
 use once_cell::unsync::Lazy;
 
+fn read_denom(
```
tzemanovic, line 12: let's expose this from `core/src/ledger/storage_api/token.rs` instead

wasm/wasm_source/src/vp_testnet_faucet.rs
```diff
@@ -57,7 +76,10 @@ fn validate_tx(
                                 &ctx.pre(),
                                 &addr,
                             )?;
-                        change >= -max_free_debit.change()
+
+                        token::Amount::from_uint(change.abs(), 0).unwrap()
```
tzemanovic, line 80: should this be the same denom? i.e.
```suggestion
                        token::Amount::from_uint(change.abs(), denom).unwrap()
```

wasm/wasm_source/src/vp_testnet_faucet.rs
tzemanovic: via vp_prelude

core/src/types/uint.rs
```diff
@@ -27,6 +24,62 @@ construct_uint! {
...
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error as serdeError;
+        let amount_string: String =
+            serde::Deserialize::deserialize(deserializer)?;
+
+        let digits = amount_string
+            .chars()
+            .filter_map(|c| {
+                if c.is_numeric() {
```
tzemanovic, line 52:
```suggestion
                if c.is_digit(10) {
```
there are non-decimal numberic chars

apps/src/lib/config/genesis.rs
```diff
@@ -116,7 +118,7 @@ pub mod genesis_config {
         pub faucet_pow_difficulty: Option<testnet_pow::Difficulty>,
         #[cfg(not(feature = "mainnet"))]
         /// Testnet faucet withdrawal limit - defaults to 1000 NAM when not set
```
tzemanovic, line 120:
```suggestion
        /// Testnet faucet withdrawal limit - defaults to 1000 tokens when not set
```

core/src/types/uint.rs
```diff
@@ -27,6 +24,62 @@ construct_uint! {
...
+impl<'de> serde::Deserialize<'de> for Uint {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error as serdeError;
+        let amount_string: String =
+            serde::Deserialize::deserialize(deserializer)?;
+
+        let digits = amount_string
```
tzemanovic, line 49: can we not use `FromStr` from the uint crate?

batconjurer: Unfortunately it's buggy. I used to use it for Amounts, but then round trips aren't idempotent. I found it in e2e tests where we load an write genesis files multiple times. The numbers would just wander off to places I couldn't understand.

wasm/wasm_source/src/vp_testnet_faucet.rs
tzemanovic: I see, thanks!

tzemanovic (approved):

tzemanovic (approved):

PR #1670, opened 2023-07-05
Removed unused associated type from ShieldedUtils

batconjurer:
This type was an attempt to reduce the number of generic parameters needed when interacting with MASP but resulted in making code re-use extremely difficult.

### DIFF
--- a/apps/src/lib/client/rpc.rs
+++ b/apps/src/lib/client/rpc.rs
@@ -60,7 +60,7 @@ use crate::wallet::CliWalletUtils;
 ///
 /// If a response is not delivered until `deadline`, we exit the cli with an
 /// error.
-pub async fn query_tx_status<C: namada::ledger::queries::Client + Sync>(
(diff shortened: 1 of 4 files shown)


apps/src/lib/client/rpc.rs
```diff
@@ -60,7 +60,7 @@ use crate::wallet::CliWalletUtils;
 ///
 /// If a response is not delivered until `deadline`, we exit the cli with an
 /// error.
-pub async fn query_tx_status<C: namada::ledger::queries::Client + Sync>(
+pub async fn query_tx_status<C: namada::ledger::queries::Client + Send + Sync>(
```
grarco, line 63: @batconjurer do we need the `Send` bound on the client?

grarco (dismissed):
LGTM, thanks!

murisi:
@mateuszjasiuk Would you be able to review this since it impacts the MASP web interface? Thanks!

mateuszjasiuk (approved):
Works fine in the interface, we just need to remove 2 lines.

PR #1681, opened 2023-07-06
test/PoS: setup quint sync test from state machine

tzemanovic:
Adds a test of PoS impl against the Quint model (https://github.com/informalsystems/partnership-heliax/blob/tomas/rust-2-quint/2023/Q2/artifacts/PoS-quint/namada.qnt) using the state machine test that calls into Quint's REPL. The basic approach works but it still needs a bit more work to parametrize more of the data (params in the other actions and the genesis validator set with non-zero stake) currently hard-coded in the Quint model.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3851,6 +3851,7 @@ dependencies = [
  "borsh",
  "data-encoding",
  "derivative",
+ "expectrl",
  "itertools",
  "namada_core",
  "once_cell",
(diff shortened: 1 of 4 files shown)


brentstone:
do we still want to do this @tzemanovic ?

tzemanovic:
> do we still want to do this @tzemanovic ?

yeah, let's leave this open for now

PR #1686, opened 2023-07-07
PoS: fix init chain validator set to respect max_validator_slots param

### LINKED ISSUE #1684: PoS: init chain doesn't respect max_validator_slots param
there's no limit on the genesis validator consensus set

tzemanovic:
closes #1684


brentstone (dismissed):
LGTM

PR #1709, opened 2023-07-12
Fix inconsistency state before commit

yito88:
In `finalize_block`, some data are written into the storage directly and a new epoch is updated in the Merkle tree when the new epoch starts before the commit.
It could cause an inconsistent state when shutdown in `finalize_block` and then restarting.

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -201,7 +201,6 @@ where
                     let tx_hash_key =
                         replay_protection::get_tx_hash_key(&tx_hash);
                     self.wl_storage
-                        .storage
                         .delete(&tx_hash_key)
                         .expect("Error while deleting tx hash from storage");
                 }
@@ -221,16 +220,14 @@ where
                             processed_tx.header_hash().0,
                         ));
                     self.wl_storage
-                        .storage
-                        .write(&wrapper_tx_hash_key, vec![])
+                        .write_bytes(&wrapper_tx_hash_key, vec![])
                         .expect("Error while writing tx hash to storage");
 
                     let inner_tx_hash_key = replay_protection::get_tx_hash_key(
                         &tx.clone().update_header(TxType::Raw).header_hash(),
                     );
                     self.wl_storage
-                        .storage
-                        .write(&inner_tx_hash_key, vec![])
+                        .write_bytes(&inner_tx_hash_key, vec![])
                         .expect("Error while writing tx hash to storage");
 
                     #[cfg(not(feature = "mainnet"))]
@@ -256,11 +253,7 @@ where
                     match balance.checked_sub(wrapper_fees) {
                         Some(amount) => {
                             self.wl_storage
-                                .storage
-                                .write(
-                                    &balance_key,
-                                    amount.try_to_vec().unwrap(),
-                                )
+                                .write(&balance_key, amount)
                                 .unwrap();
                         }
                         None => {
@@ -271,12 +264,9 @@ where
                             if reject {
                                 // Burn remaining funds
                                 self.wl_storage
-                                    .storage
                                     .write(
                                         &balance_key,
-                                        Amount::native_whole(0)
-                                            .try_to_vec()
-                                            .unwrap(),
+                                        Amount::native_whole(0),
                                     )
                                     .unwrap();
                                 tx_event["info"] =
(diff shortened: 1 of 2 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -481,6 +471,7 @@ where
                     );
                     stats.increment_errored_txs();
 
+                    self.wl_storage.drop_tx();
```
adrianbrink, line 474: Does this only drop the current tx or does it clear the entire wl_storage? From reading the wl_storage code I'm not entirely sure.

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -491,15 +482,16 @@ where
...
                                 replay_protection::get_tx_hash_key(&hash);
                             self.wl_storage
-                                .storage
                                 .delete(&tx_hash_key)
                                 .expect(
                                     "Error while deleting tx hash key from storage",
                                 );
+                            // Apply only to remove its hash,
+                            // since all other changes have already been dropped
+                            self.wl_storage.commit_tx();
```
adrianbrink, line 491: Why do we need to do this?

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: only the current tx

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: it is to commit the removal of the `tx_hash_key` above

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -1834,7 +1826,14 @@ mod test_finalize_block {
...
                 ..Default::default()
             };
+            // merkle tree root before finalize_block
+            let root_pre = shell.shell.wl_storage.storage.block.tree.root();
+
             let _events = shell.finalize_block(req).unwrap();
+
+            // the merkle tree root should not change after finalize_block
+            let root_post = shell.shell.wl_storage.storage.block.tree.root();
+            assert_eq!(root_pre.0, root_post.0);
```
tzemanovic, line 1836: when I remove the fixes and run the tests they do not yet catch the issue - I think we need to add some txs in here

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -2234,6 +2235,10 @@ mod test_finalize_block {
             })
             .expect("Test failed")[0];
 
+        // the merkle tree root should not change after finalize_block
+        let root_post = shell.shell.wl_storage.storage.block.tree.root();
+        assert_eq!(root_pre.0, root_post.0);
```
tzemanovic, line 2240: this one also seems to pass without the fixes

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: added in f30e158585cba829c9055cd218159e2e89d0fafe

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -3543,4 +3488,83 @@ mod test_finalize_block {
             control_receiver.recv().await.expect("Test failed");
         assert_eq!(u64::from(cmd.min_confirmations), 42);
     }
+
+    /// Make a wrapper tx and a processed tx from the wrapped tx that can be
+    /// added to `FinalizeBlock` request.
+    fn mk_wrapper_tx(
```
batconjurer, line 3494: nit: can these helper functions go to the top of the module? Otherwise people have to read the code from the bottom up (as I currently am)

apps/src/lib/node/ledger/shell/finalize_block.rs
yito88: Yes, it's a bit tricky. But we have to `commit_tx()` to apply changes on the write log when the following commit phase.

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: this one was already passing before because `TxType::Decrypted` didn't commit anything directly to storage so this is ok

tzemanovic (approved):
need to auto-squash before merging

cwgoes (approved):
Discussed w/ @yito88

tzemanovic:
auto-squashed fixups (no diff)

tzemanovic (approved):

PR #1714, opened 2023-07-13
Refactor Masp E2E tests

batconjurer:
The MASP E2E tests are notoriously difficult to maintain and debug. This PR introduces new tooling under `integration` tests that allows the ledger and client to be run in a single process. Theses tests are designed to be deterministic and can have debugger attached.

While this tooling can be used for more than MASP e2e tests, those test are the one's I've refactored as PoC as they are highest priority.


apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -368,34 +368,34 @@ where
...
+    pub tx_wasm_cache: TxCache<WasmCacheRwAccess>,
     /// Taken from config `storage_read_past_height_limit`. When set, will
     /// limit the how many block heights in the past can the storage be
     /// queried for reading values.
     storage_read_past_height_limit: Option<u64>,
     /// Proposal execution tracking
     pub proposal_data: HashSet<u64>,
     /// Log of events emitted by `FinalizeBlock` ABCI calls.
-    event_log: EventLog,
+    pub event_log: EventLog,
```
sug0, line 398: there is already a pub method to access the event log, to acquire an immutable and mutable ref (`Shell::event_log()` and `Shell::event_log_mut()`, respectively)

tests/Cargo.toml
```diff
@@ -33,12 +34,16 @@ namada_core = {path = "../core", features = ["testing"]}
 namada_test_utils = {path = "../test_utils"}
 namada_vp_prelude = {path = "../vp_prelude"}
 namada_tx_prelude = {path = "../tx_prelude"}
+async-trait = {version = "0.1.51"}
```
tzemanovic, line 37:
```suggestion
async-trait.workspace = true
```

tests/Cargo.toml
```diff
@@ -33,12 +34,16 @@ namada_core = {path = "../core", features = ["testing"]}
...
 chrono.workspace = true
+clap.workspace = true
 concat-idents.workspace = true
 derivative.workspace = true
 hyper = {version = "0.14.20", features = ["full"]}
 ibc-relayer-types.workspace = true
 ibc-relayer.workspace = true
+lazy_static = "1.4.0"
```
tzemanovic, line 45: let's do this one via workspace too (it's also in encoding_spec crate)

tzemanovic:
One thing to note is that when you don't yet have masp params present and run the new tests in parallel they'll race to fetch them and all but one will fail. It should be easy to fix with a lock file though - I'd say we don't need to block this PR on it though and do it as a follow up (with the new crate from #1605)

tests/src/integration/setup.rs
```diff
@@ -0,0 +1,167 @@
...
+pub fn initialize_genesis(
+    mut update_genesis: impl FnMut(GenesisConfig) -> GenesisConfig,
+) -> Result<MockNode> {
+    let working_dir = std::fs::canonicalize("..").unwrap();
+    let keep_temp = match std::env::var(ENV_VAR_KEEP_TEMP) {
+        Ok(val) => val.to_ascii_lowercase() != "false",
+        _ => false,
+    };
+    let test_dir = {
+        std::env::set_var(ENV_VAR_KEEP_TEMP, "true");
```
tzemanovic, line 40: this re-use of `TestDir` from e2e seems to forcibly add unnecessarily coupling and unsafe code - maybe we do it better without it?

shared/src/ledger/queries/router.rs
tzemanovic, line 1: is this change needed?

apps/src/lib/cli/utils.rs
```diff
@@ -331,3 +333,35 @@ pub fn safe_exit(code: i32) -> ! {
...
+#[cfg(feature = "testing")]
+/// A helper to exit after flushing output, borrowed from `clap::util` module.
+pub fn safe_exit(_: i32) -> ! {
+    let _ = std::io::stdout().lock().flush();
+    let _ = std::io::stderr().lock().flush();
+
+    panic!("Test failed because the client exited unexpectedly.")
+}
+
+pub fn prompt_aux<R, W>(mut reader: R, mut writer: W, question: &str) -> String
```
tzemanovic, line 346:
```suggestion
fn prompt_aux<R, W>(mut reader: R, mut writer: W, question: &str) -> String
```

apps/src/lib/cli/utils.rs
```diff
@@ -331,3 +333,35 @@ pub fn safe_exit(code: i32) -> ! {
...
+    W: Write,
+{
+    write!(&mut writer, "{}", question).expect("Unable to write");
+    std::io::stdout().flush().unwrap();
+    let mut s = String::new();
+    reader.read_to_string(&mut s).expect("Unable to read");
+    s
+}
+
+#[cfg(feature = "dev")]
```
tzemanovic, line 358: let's pls use "testing" feature instead

tzemanovic:
It's a nice achievement to translate the masp tests without much change to them, but tbh I find the changes required to get there a bit invasive and we're adding some maintenance burden with a bit more coupling, mock code and also with using a compiler internal feature. Personally, I'd prefer to refactor the masp code to remove coupling and test it this way which imho would also improve its maintainability, but nothing is stopping us from doing that as well as having this and this is already unarguably a better experience over the e2e tests and for anything that doesn't rely on cometBFT running, this will be a huge improvement!

tests/src/integration/masp.rs
```diff
@@ -0,0 +1,1205 @@
+use std::path::PathBuf;
+
+use color_eyre::eyre::Result;
+use color_eyre::owo_colors::OwoColorize;
+use namada_apps::client::tx::CLIShieldedUtils;
+use namada_core::types::address::{btc, eth, masp_rewards};
+use namada_core::types::token;
+use namada_core::types::token::{DenominatedAmount, NATIVE_MAX_DECIMAL_PLACES};
```
tzemanovic, line 8:
```suggestion
use namada_core::types::token::{DenominatedAmount, NATIVE_MAX_DECIMAL_PLACES};
use test_log::test;
```
to allow capturing tracing logs with `RUST_LOG`

tests/src/integration/client.rs
```diff
@@ -0,0 +1,664 @@
...
+        let global = {
+            let locked = self.shell.lock().unwrap();
+            Global {
+                chain_id: Some(locked.chain_id.clone()),
+                base_dir: locked.base_dir.clone(),
+                wasm_dir: Some(locked.wasm_dir.clone()),
+            }
+        };
+        let mut ctx = Context::new(global)?;
+        match cmd {
```
tzemanovic, line 97: for a future improvement it would be nice if we could re-use the existing definitions without repeating them here as we're not going through `apps/src/bin/namada-client/cli.rs` anymore with this

tests/src/integration/client.rs
```diff
@@ -0,0 +1,664 @@
...
+        self.matches(needle)
+    }
+}
+
+async fn submit_transfer(
+    client: &MockNode,
+    ctx: &mut Context,
+    args: args::TxTransfer,
+) -> Result<(), namada::ledger::tx::Error> {
+    let arg = args.clone();
```
tzemanovic, line 628: also this would be better to run through the original without a mock version

batconjurer: Yeah, I wanted to, but that has problems. It submits txs twice in case masp hits an epoch boundary condition. This greatly complicates checking if the tx succeeded.

batconjurer: But I'm open to suggestions!

shared/src/ledger/queries/router.rs
batconjurer: I thought I'd un -p ubbed that already. WIll fix

tests/src/integration/client.rs
batconjurer: Actually, I think I can use the original if I'm super careful about epoch boundaries in the tests.

tests/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(internal_output_capture)]
```
tzemanovic, line 1:
```suggestion
#![cfg_attr(test, feature(internal_output_capture))]
```

tzemanovic (approved):

tzemanovic:
merged in 0.20.0

PR #1716, opened 2023-07-13
test/e2e/slashing: wait for wasm on original validator

tzemanovic:
as seen in e.g. https://github.com/anoma/namada/actions/runs/5541777593/jobs/10115627826, sometimes the validator-0-copy is created after the original validator-0 has finished compiling its wasm, in which case the copy is created with the wasm cache. With this change, the copy is created only after original is done compiling


karbyshev:
Would it make sense to wrap the three lines
```
validator_0.exp_string("Namada ledger node started")?;
    validator_0.exp_string("This node is a validator")?;
    wait_for_wasm_pre_compile(&mut validator_0)?;
```
into some macro / function?

karbyshev (approved):
Approving my changes :-)

Issue #1290, opened 2023-04-13
Update token whitelists


sug0:
Act on token whitelist Ethereum events.

sug0:
Related to #1234

PR #1718, opened 2023-07-13
Implement the Ethereum token whitelist logic in Namada

sug0:
Related to https://github.com/anoma/namada/issues/1290 and https://github.com/anoma/ethereum-bridge/pull/60

### DIFF
--- a/core/src/types/ethereum_events.rs
+++ b/core/src/types/ethereum_events.rs
@@ -334,16 +334,6 @@ pub enum EthereumEvent {
         #[allow(dead_code)]
         address: EthAddress,
     },
-    /// Event indication a new Ethereum based token has been whitelisted for
-    /// transfer across the bridge
-    UpdateBridgeWhitelist {
-        /// Monotonically increasing nonce
-        #[allow(dead_code)]
-        nonce: Uint,
-        /// Tokens to be allowed to be transferred across the bridge
-        #[allow(dead_code)]
-        whitelist: Vec<TokenWhitelist>,
-    },
 }
 
 impl EthereumEvent {
@@ -460,30 +450,6 @@ pub struct TransferToEthereum {
     pub sender: Address,
     /// The account of fee payer.
     pub gas_payer: Address,
-}
-
-/// struct for whitelisting a token from Ethereum.
-/// Includes the address of issuing contract and
-/// a cap on the max amount of this token allowed to be
-/// held by the bridge.
-#[derive(
-    Clone,
-    Debug,
-    PartialEq,
-    Eq,
-    Hash,
-    PartialOrd,
-    Ord,
-    BorshSerialize,
-    BorshDeserialize,
-    BorshSchema,
-)]
-#[allow(dead_code)]
-pub struct TokenWhitelist {
-    /// Address of Ethereum smart contract issuing token
-    pub token: EthAddress,
-    /// Maximum amount of token allowed on the bridge
-    pub cap: Amount,
 }
 
 #[cfg(test)]
--- a/core/src/types/ethereum_events.rs
+++ b/core/src/types/ethereum_events.rs
@@ -376,6 +376,60 @@ pub struct TransferToNamada {
     pub receiver: Address,
 }
 
+/// Transfer to Ethereum kinds.
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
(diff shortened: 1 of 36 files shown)


core/src/types/ethereum_events.rs
```diff
@@ -391,25 +382,45 @@ pub struct TransferToNamada {
...
-/// held by the bridge.
+impl TryFrom<u8> for TransferToEthereumKind {
+    type Error = eyre::Error;
+
+    fn try_from(kind: u8) -> Result<Self, Self::Error> {
+        match kind {
+            Self::KIND_ERC20 => Ok(Self::Erc20),
+            Self::KIND_NUT => Ok(Self::Nut),
+            _ => Err(eyre!(
+                "Only valid kinds are {} (Erc20) and {} (Nut)",
```
batconjurer, line 415:
```suggestion
                "Only valid kinds are {} (Erc20) and {} (NUT)",
```

ethereum_bridge/src/parameters.rs
```diff
@@ -3,18 +3,41 @@ use std::num::NonZeroU64;
...
+    Copy,
+    Eq,
+    PartialEq,
+    Debug,
+    Deserialize,
+    Serialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+pub struct Erc20WhitelistEntry {
```
batconjurer, line 32: Do we also cap the max amount of wrapped nam that can exist on the bridge?

ethereum_bridge/src/parameters.rs
```diff
@@ -228,6 +281,8 @@ impl EthereumBridgeConfig {
         let eth_start_height = must_read_key(wl_storage, &eth_start_height_key);
 
         Some(Self {
+            // TODO: read this from storage? iterate storage prefix
```
batconjurer, line 284: exactly

ethereum_bridge/src/parameters.rs
sug0: oh uhhh good point. currently we don't. that check can also be added namada side

ethereum_bridge/src/parameters.rs
sug0: I've reworked this to not return the whitelist. the only place where we call `read()` is in the ethereum oracle, which does not need whitelist data

batconjurer (changes requested):
Several NUT test cases missing from the bridge pool vp. We should check the happy flow. We should check that we reject if an asset is incorrectly labelled. The bridge should also accept NUTs and normals for the same asset simultaneously.

ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -220,56 +230,81 @@ where
...
+            tracing::debug!(
+                %supply_key,
+                ?supply,
+                "New value calculated",
+            );
+        })?;
+        _ = changed_keys.insert(supply_key);
 
-    Ok(changed_keys)
+        let minter_key = minter_key(&token);
```
batconjurer, line 301: what is the purpose of this minter key?

sug0: this is IBC specific stuff that yuji added for the multitoken PR. it's not actually needed anymore. it basically stores a VP address to check who minted tokens. the ethereum bridge doesn't mint tokens from wasm though. it's already removed in the upstream multitokens PR, but I haven't updated the base yet

ethereum_bridge/src/parameters.rs
sug0: I feel like that should be a separate PR though. It doesn't really concern the ERC20 token whitelist. The check should be made on the Bridge pool, to cap wNAM that is sent over to Ethereum.

sug0: https://github.com/anoma/namada/issues/1722

sug0:
Took out the patch that removed `ADDR` from native VPs, and opened an issue

sug0:
added a sanity check in [4b43593](https://github.com/anoma/namada/pull/1718/commits/4b43593b6821490d65d4026b8fa91999a1960ac4)
```rs
        if hints::unlikely(cap < supply) {
            panic!(
                "Namada's state is faulty! The Ethereum ERC20 asset {token} \
                 has a higher minted supply than the configured token cap: \
                 cap:{cap:?} < supply:{supply:?}"
            );
        }
```

PR #1729, opened 2023-07-17
PoS: fix rewards boundary

### LINKED ISSUE #1728: PoS: skipped votes after slash processing on epoch boundary
When there's a slash detected on a start of an epoch, it is enqueued and the validator's state is set to `Jailed` before we call `log_block_rewards` from `finalize_block`. Because the validator state is not preserved for any past epochs, when the misbehaving validator's state is updated, the data for past epochs are dropped and `log_block_rewards` is no longer able to look up its state, because when there's a new epoch in the block that's being finalized, we pass `current_epoch.prev()` to this function to calculate the rewards using the epoched data from before we switched to a new epoch.

tzemanovic:
## Describe your changes

fixes #1728 by reordering the calls in finalize block

v0.19.0

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -110,6 +110,14 @@ where
             )?;
         }
 
+        // Invariant: Has to be applied before `record_slashes_from_evidence`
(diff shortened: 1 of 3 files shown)


brentstone (approved):
LGTM

PR #1606, opened 2023-06-21
multisig: remove `--signer` option

Fraccaman:

PR #1741, opened 2023-07-20
Multisignature (rebase)

### LINKED ISSUE #1178: Multisig protocol parameter limit
Currently, the implementation in #1102 allows for an arbitrary amount of signers on an established address. To prevent time and cost escalation we should hard limit the number of public keys associated with an established address.

Fraccaman:
## Describe your changes
Took #1606 and rebased on 0.19.0. 
Close #1178  and #81

Based on 0.19.0

### DIFF
--- a/genesis/e2e-tests-single-node.toml
+++ b/genesis/e2e-tests-single-node.toml
@@ -4,7 +4,7 @@
 
 genesis_time = "2021-09-30T10:00:00Z"
 native_token = "NAM"
-faucet_pow_difficulty = 1
+faucet_pow_difficulty = 0
 faucet_withdrawal_limit = "1000"
 
 [validator.validator-0]
--- a/tests/src/e2e/ledger_tests.rs
+++ b/tests/src/e2e/ledger_tests.rs
@@ -30,6 +30,7 @@ use namada_apps::config::genesis::genesis_config::{
 use namada_apps::config::utils::convert_tm_addr_to_socket_addr;
 use namada_apps::facade::tendermint_config::net::Address as TendermintAddress;
 use namada_test_utils::TestWasms;
+use namada_vp_prelude::testnet_pow;
 use serde_json::json;
 use setup::constants::*;
 use setup::Test;
@@ -408,7 +409,22 @@ fn stop_ledger_at_height() -> Result<()> {
 /// 8. Query the raw bytes of a storage key
 #[test]
 fn ledger_txs_and_queries() -> Result<()> {
-    let test = setup::network(|genesis| genesis, None)?;
+    let test = setup::network(
+        |genesis| {
+            #[cfg(not(feature = "mainnet"))]
+            {
+                GenesisConfig {
+                    faucet_pow_difficulty: testnet_pow::Difficulty::try_new(1),
+                    ..genesis
+                }
+            }
+            #[cfg(feature = "mainnet")]
+            {
+                genesis
+            }
+        },
+        None,
+    )?;
 
     set_ethereum_bridge_mode(
         &test,
@@ -442,6 +458,9 @@ fn ledger_txs_and_queries() -> Result<()> {
     let tx_data_path = tx_data_path.to_string_lossy();
 
     let validator_one_rpc = get_actor_rpc(&test, &Who::Validator(0));
+
+    let multisig_account =
+        format!("{},{},{}", BERTHA_KEY, ALBERT_KEY, CHRISTEL_KEY);
 
     let txs_args = vec![
         // 2. Submit a token transfer tx (from an established account)
@@ -553,10 +572,34 @@ fn ledger_txs_and_queries() -> Result<()> {
             "--public-keys",
             // Value obtained from `namada::types::key::ed25519::tests::gen_keypair`
             "001be519a321e29020fa3cbfbfd01bd5e92db134305609270b71dace25b5a21168",
+            "--threshold",
+            "1",
             "--code-path",
             VP_USER_WASM,
             "--alias",
             "Test-Account",
+            "--gas-amount",
+            "0",
+            "--gas-limit",
+            "0",
+            "--gas-token",
+            NAM,
+            "--signing-keys",
+            BERTHA_KEY,
+            "--node",
+            &validator_one_rpc,
+        ],
+        // 5. Submit a tx to initialize a new multisig account
+        vec![
+            "init-account",
+            "--public-keys",
+            &multisig_account,
+            "--threshold",
+            "2",
+            "--code-path",
+            VP_USER_WASM,
+            "--alias",
+            "Test-Account-2",
             "--gas-amount",
             "0",
             "--gas-limit",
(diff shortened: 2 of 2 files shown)


core/src/types/account.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Retrive the index of a public key
+    pub fn get_index_from_public_key(
+        &self,
+        public_key: &common::PublicKey,
+    ) -> Option<u8> {
+        self.pk_to_idx.get(public_key).cloned()
+    }
+
+    /// Return an empty AccountPublicKeysMap
+    pub fn empty() -> Self {
```
tzemanovic, line 88: I think you can `derive(Default)` for this

core/src/ledger/parameters/storage.rs
```diff
@@ -119,6 +120,14 @@ pub fn is_max_proposal_bytes_key(key: &Key) -> bool {
     is_max_proposal_bytes_key_at_addr(key, &ADDRESS)
 }
 
+/// Returns if the key is the max signature per transacton key
+pub fn is_max_signatures_per_transaction_key(key: &Key) -> bool {
+    matches!(&key.segments[..], [
+        DbKeySeg::AddressSeg(addr),
+        DbKeySeg::StringSeg(max_signatures_per_transaction),
+    ] if addr == &ADDRESS && max_signatures_per_transaction == Keys::VALUES.max_signatures_per_transaction)
```
tzemanovic, line 128:
```suggestion
    is_max_signatures_per_transaction_key_at_addr(key, &ADDRESS)
```
generated from `derive(StorageKeys)`

core/src/ledger/parameters/storage.rs
```diff
@@ -183,3 +192,15 @@ pub fn get_faucet_account_key() -> Key {
...
+/// Storage key used for the max signatures per transaction key
+pub fn get_max_signatures_per_transaction_key() -> Key {
+    Key {
+        segments: vec![
+            DbKeySeg::AddressSeg(ADDRESS),
+            DbKeySeg::StringSeg(
+                Keys::VALUES.max_signatures_per_transaction.to_string(),
+            ),
+        ],
+    }
```
tzemanovic, line 205:
```suggestion
    get_max_signatures_per_transaction_key_at_addr(ADDRESS)
```

core/src/ledger/storage_api/account.rs
```diff
@@ -0,0 +1,101 @@
...
+    Ok(AccountPublicKeysMap::from_iter(public_keys))
+}
+
+/// Check if an account exists in storage
+pub fn exists<S>(storage: &S, owner: &Address) -> Result<bool>
+where
+    S: StorageRead,
+{
+    let vp_key = Key::validity_predicate(owner);
+    storage.has_key(&vp_key)
```
tzemanovic, line 74:
```suggestion
    match owner {
        Address::Established(_) => {
            let vp_key = Key::validity_predicate(owner);
            storage.has_key(&vp_key)
       }
       Address::Implicit(_) | Address::Internal(_) => Ok(true),
```

core/src/proto/types.rs
```diff
@@ -360,6 +371,146 @@ impl Code {
...
+            )
+        } else {
+            Err(VerifySigError::MissingData)
+        }
+    }
+}
+
+impl Ord for SignatureIndex {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.index.cmp(&other.index)
```
tzemanovic, line 423: I think if you put the `index` field first in the struct then you can derive the impl as it will be sorted by the index (you might need to add `Ord` to `common::Signature` which is fine)

genesis/e2e-tests-single-node.toml
```diff
@@ -4,7 +4,7 @@
 
 genesis_time = "2021-09-30T10:00:00Z"
 native_token = "NAM"
-faucet_pow_difficulty = 1
+faucet_pow_difficulty = 0
```
tzemanovic, line 7: do we need to switch it off here?

.changelog/unreleased/features/1606-multisignature-draft.md
```diff
@@ -0,0 +1,2 @@
+- Introduce multisignature account and transaction format.
```
tzemanovic, line 1: it would be nice to write a little bit more about this and/or link to docs for anyone who'll want to try it in the new release

apps/src/lib/config/mod.rs
```diff
@@ -283,7 +283,7 @@ impl Config {
             .and_then(|c| c.merge(config::File::with_name(file_name)))
             .and_then(|c| {
                 c.merge(
-                    config::Environment::with_prefix("namada").separator("__"),
```
tzemanovic, line 286: is this case sensitive?

shared/src/ledger/tx.rs
```diff
@@ -241,13 +253,29 @@ pub async fn prepare_tx<
     wallet: &mut Wallet<U>,
     args: &args::Tx,
     tx: Tx,
+    owner: Option<Address>,
```
tzemanovic, line 256: it looks like this arg is not needed here - it's just returned untouched

shared/src/ledger/signing.rs
```diff
@@ -209,24 +195,83 @@ pub async fn sign_tx<U: WalletUtils>(
...
     Ok(())
 }
 
+/// Return the necessary data regarding an account to be able to generate a
+/// multisignature section
+pub async fn aux_signing_data<C: crate::ledger::queries::Client + Sync>(
+    client: &C,
+    owner: Option<Address>,
+    public_keys: Vec<common::PublicKey>,
+) -> (AccountPublicKeysMap, u8) {
```
tzemanovic, line 260: I think this function is not very clear - the default threshold `0` when there's no owner doesn't seem right, is it? In e.g. `init_account` that doesn't have an owner beforehand it should respect the tx arg.

I suggest that we change this and use it for all txs except for `reveal_pk`, `init_account` and `init_validator` to only take `owner: &Address` without the `public_keys` and fail if the account info cannot be queried.

For the `reveal_pk`, `init_account` and `init_validator`, we can call `AccountPublicKeysMap::from_iter(public_keys)` directly and take the threshold from the tx arg (with the logic where we set it to 1 when the length of the keys == 1 for new accounts and 1 for `reveal_pk`)

core/src/proto/types.rs
tzemanovic: ah, I see. I think we can add it, we already have a manual PartialOrd using the bytes:

```rust
impl Ord for Signature {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        self.0.to_bytes().cmp(&other.0.to_bytes())
    }
}
```

.changelog/unreleased/features/1606-multisignature-draft.md
tzemanovic: yes pls!

apps/src/lib/config/mod.rs
tzemanovic: sounds good

shared/src/ledger/signing.rs
tzemanovic: thank you!

PR #1746, opened 2023-07-21
Generic IO

batconjurer:
This code replaces IO in sdk and client code with a trait that allows custom IO handling.


apps/src/lib/cli/wallet.rs
```diff
@@ -97,31 +98,33 @@ fn address_key_find(
     let alias = alias.to_lowercase();
     if let Ok(viewing_key) = wallet.find_viewing_key(&alias) {
         // Check if alias is a viewing key
-        println!("Viewing key: {}", viewing_key);
+        IO::println(format!("Viewing key: {}", viewing_key));
```
tzemanovic, line 101: let's use the macros in this file

tzemanovic (dismissed):

tzemanovic (dismissed):

mateuszjasiuk (approved):
Sorry for later review!
So I was able to implement the trait in browser.
In general all of this works great, the two things to comes to my mind though:
- Sometimes it is a bit annoying to pass generic IO as last type `namada::ledger::tx::build_transfer::<_, _, _, IO>`, as first three types can be inferred
- it would be nice to make prompt and read async. The reason is that when we use those functions in the extension background(service worker) we do not have access to prompt(or any other DOM related features). So to make it work we would have to do something like this: [stackblitz](https://stackblitz.com/edit/typescript-jdvple?file=index.ts). On the other hand though, at the moment we are not using prompts at all, so this might not be such a big issue.

batconjurer:
So I can change the order of the generic arguments, but it's a choice of `namada::ledger::tx::build_transfer::<_, _, _, IO>` vs `namada::ledger::tx::build_transfer::<IO _, _, _>`. I don't know if it's worth it for the same amount of typing in the end.

batconjurer:
However, I made the `read` and `prompt` methods async. That is easy.

mateuszjasiuk:
> So I can change the order of the generic arguments, but it's a choice of `namada::ledger::tx::build_transfer::<_, _, _, IO>` vs `namada::ledger::tx::build_transfer::<IO, _, _, _>`. I don't know if it's worth it for the same amount of typing in the end.

Ah my bad, I thought they can be skipped the same as in TS 😅 
Nvm then

mateuszjasiuk (dismissed):

shared/src/types/io.rs
```diff
@@ -40,11 +42,11 @@ pub trait Io {
         eprintln!("{}", output.as_ref());
     }
 
-    fn read() -> std::io::Result<String> {
+    async fn read() -> std::io::Result<String> {
         read_aux(std::io::stdin().lock())
```
tzemanovic, line 46: let's also switch these to use non-blocking io with `tokio::io::stdin`

shared/src/types/io.rs
```diff
@@ -40,11 +42,11 @@ pub trait Io {
...
     }
 
-    fn prompt(question: impl AsRef<str>) -> String {
+    async fn prompt(question: impl AsRef<str>) -> String {
         prompt_aux(
             std::io::stdin().lock(),
```
tzemanovic, line 51: and here

tzemanovic (approved):

mateuszjasiuk (dismissed):

sug0 (changes requested):
Hey, I noticed `apps/src/bin/namada-client/cli.rs` hasn't been deleted yet, despite the module not being referenced from any rust source file.

sug0 (dismissed):

tzemanovic (approved):

PR #1753, opened 2023-07-25
test/e2e/ledger_tests: add a happy-path masp txs and queries test

tzemanovic:
re-adding a simplified happy-path masp e2e test to cover transfers:
- shielding
- shielded -> shielded
- unshielding


tzemanovic:
> I think we should also add the test to the `e2e.json` file

thanks! I forgot

Issue #1760, opened 2023-07-25
Broken e2e testing for ARM computers from #1695


brentstone:
This issue was noted in the PR description for #1695, but we merged the PR anyway without fixing it. After this update, on my Mac M1, the ledger reads the available memory for VP and TX wasm compilation to be 0 bytes.

Fraccaman:
this should be fixed

PR #1761, opened 2023-07-25
Fix wasm compilation cache size setting for all computers

brentstone:
## Describe your changes

See #1760, which is broken for ARM computers it seems. Some default cache size is provided if the available memory bytes is read as 0.

v0.20.0


brentstone:
@Fraccaman @tzemanovic should we provide constants set to some number of bytes rather than hard-coded here?

tzemanovic:
let's close this in favor of #1775

PR #1763, opened 2023-07-25
[fix]: Fixed ordering for Dec and I256 types

### LINKED ISSUE #1762: Ordering for `Dec` type is buggy
The ordering for `Dec` types does not work properly for large numbers it seems. An example encountered in the ledger can be summarized with the following lines:
```
let smaller = Dec::from_str("6483947304.195066085701").unwrap();
let larger = Dec::from_str("32418116583.390243854642").unwrap();
assert!(smaller < larger);
```
The above assertion fails.

batconjurer:
Fixes #1762

### DIFF
--- a/core/src/types/dec.rs
+++ b/core/src/types/dec.rs
@@ -613,4 +613,11 @@ mod test_dec {
                 .unwrap()
         );
     }
+
+    #[test]
+    fn test_ordering() {
+        let smaller = Dec::from_str("6483947304.195066085701").unwrap();
+        let larger = Dec::from_str("32418116583.390243854642").unwrap();
+        assert!(smaller < larger);
+    }
 }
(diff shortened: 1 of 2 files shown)


brentstone (dismissed):
Thanks! LGTM

bengtlofgren (commented):
Could you just describe what the difference was here? What did the .0 do?

core/src/types/dec.rs
```diff
@@ -613,4 +613,11 @@ mod test_dec {
                 .unwrap()
         );
     }
+
+    #[test]
+    fn test_ordering() {
+        let smaller = Dec::from_str("6483947304.195066085701").unwrap();
+        let larger = Dec::from_str("32418116583.390243854642").unwrap();
+        assert!(smaller < larger);
+    }
```
sug0, line 622: wouldn't this test have passed before, with the faulty comparisons?

batconjurer: Nope, it needs 2 u64 words to be stored.

PR #1768, opened 2023-07-26
pre-built MASP proofs for integration tests

tzemanovic:
## Describe your changes

To speed-up MASP integration tests that were timing out in CI, this PR adds the ability to save and load proofs to/from files stored in `test_fixtures/masp_proofs`.

```shell
# Run the tests with the saved proofs.
make test-integration

# Delete old proofs, run the tests and save the new proofs.
# Use this is there's a breaking change in either the proof creation or 
# in the MASP integration tests that invalidates the previously saved proofs.
make test-integration-save-proofs

# Run without pre-built proofs
make test-integration-slow
```

The prebuilt proofs are added to `make test-coverage` to allow running integration tests in the CI.

The MASP proof builder in `fn gen_shielded_transfer` is optionally given a seed to produce deterministic output. When not specified, the original behavior is preserved (i.e. using `OsRng`, but now wrapped in [StdRng](https://rust-random.github.io/rand/rand/rngs/struct.StdRng.html) which is cryptographically secure - see https://rust-random.github.io/rand/rand/trait.SeedableRng.html#method.from_rng for more details).

To enable deterministic builders, the `Conversions` type and notes positions in `ShieldedContext` are changed to `BTreeMap` and `BTreeSet` resp to provide deterministic output from `fn collect_unspent_notes`.

0.20.0


tzemanovic:
> this is awesome! Minor thing: can we remove the masp test from the `e2e.json` file?

cheers! It is already removed on main but this branch is based on 0.20 so it still appears here

PR #1771, opened 2023-07-27
Remove unstable feature from integration tests

batconjurer:
## Describe your changes

This refactors the integrations tests using generic I/O. This allows use to remove the unstable feature gate.

Depends on #1746


tzemanovic (dismissed):
nice! (I tried this together with #1768)

tzemanovic:
we could remove nightly from tests builds now - even though it's not 100% necessary it would help with reusing build artifacts with stable build

batconjurer:
> we could remove nightly from tests builds now - even though it's not 100% necessary it would help with reusing build artifacts with stable build

We still have the unstable option `report-time` though right?

tzemanovic:
> > we could remove nightly from tests builds now - even though it's not 100% necessary it would help with reusing build artifacts with stable build
> 
> We still have the unstable option `report-time` though right?

ah, you're right, nvm then

tzemanovic (approved):

PR #849, opened 2022-11-30
refactor and add more bech32m encodings

tzemanovic:
based on v0.14.0

Since we're doing breaking changes for genesis in https://github.com/anoma/namada/pull/685, it's a good opportunity to add bech32m string encoding for `common::PublicKey` and `DkgPublicKey` and use that for genesis instead the hex

This PR:
- refactors bech32m for transparent and MASP addresses to easier re-use
- adds bench32m encoding for `common::{PublicKey, Signature}` and `DkgPublicKey`

PR #1782, opened 2023-08-02
Bat/pre genesis wallet

### LINKED ISSUE #784: interacting with the cli requires a node locally
I tried to generate a key: 

```
namada wallet key gen --alias marbar
Error:
   0: couldn't read genesis config file from .anoma/anoma-internal.000000000000000.toml
   1: No such file or directory (os error 2)

Location:
   apps/src/lib/config/genesis.rs:580

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
``` 

but got this error. There isn't a command to set a node ip here so im unable to generate a key. 

same thing goes for address

```
namada wallet address gen
Error:
   0: couldn't read genesis config file from .anoma/anoma-internal.000000000000000.toml
   1: No such file or directory (os error 2)

Location:
   apps/src/lib/config/genesis.rs:580

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

steps to reproduce: 

```bash
git clone https://github.com/anoma/namada.git
cd namada 
make install
```

```bash
namada wallet address gen
```

then the error is outputted

batconjurer:
based on #849, this PR adds --pre-genesis argument to the wallet commands to allow to generate keys, implicit addresses and MASP keys without having a chain setup - this is a needed for #685. If no chain is setup yet (i.e. there's no .namada or it's empty), the wallet defaults to use the pre-genesis wallet even without the --pre-genesis flag. The pre-genesis wallet is located in .namada/pre-genesis/wallet.toml.

closes #784


tests/src/e2e/ledger_tests.rs
```diff
@@ -678,151 +677,6 @@ fn ledger_txs_and_queries() -> Result<()> {
     Ok(())
 }
```
tzemanovic, line 679: let's not delete this test, integration tests are easier to dbg and faster - this is just a simplified happy path test from https://github.com/anoma/namada/pull/1753 as a last safety net

apps/src/lib/cli/client.rs
```diff
@@ -41,8 +41,10 @@ impl<IO> CliApi<IO> {
                         let dry_run = args.tx.dry_run;
                         tx::submit_custom(&client, &mut ctx, args).await?;
                         if !dry_run {
```
tzemanovic, line 43: let's move this expr into `tx::submit_custom` in `apps/src/lib/client/tx.rs` since we already have the `chain_ctx` there

apps/src/lib/cli/client.rs
```diff
@@ -104,7 +106,8 @@ impl<IO> CliApi<IO> {
                         tx::submit_init_account(&client, &mut ctx, args)
                             .await?;
                         if !dry_run {
```
tzemanovic, line 108: and this one

tzemanovic (approved):

PR #1784, opened 2023-08-03
Fix IBC amount handling

### LINKED ISSUE #1744: Fix amount handling in IBC transfer
In IBC transfer,
- Need to validate the given amount and the fee
- Fix the conversion from `IbcAmount` to `Amount` for non-native denominations

yito88:
## Describe your changes
Closes #1744 

v0.21.1


core/src/ledger/ibc/context/common.rs
```diff
@@ -357,25 +358,96 @@ pub trait IbcCommonContext: IbcStorageContext {
...
+                                "Decoding the token denom failed: Token {}",
+                                token
+                            ),
+                        })
+                    })?;
+                Ok(Some(denom))
+            }
+            None => Ok(None),
+        }
+    }
```
sug0, line 422: maybe something like this instead?
```rs
Ok(bytes.map(|b| try_from_slice(&b)).transpose()?)
```

core/src/ledger/ibc/mod.rs
```diff
@@ -136,27 +134,47 @@ where
...
+                    .ctx
+                    .borrow()
+                    .get_ibc_event("denomination_trace")
+                    .map_err(|_| {
+                        Error::Denom("Reading the IBC event failed".to_string())
+                    })?;
+                if let Some(event) = result {
+                    if let (Some(trace_hash), Some(ibc_denom)) = (
+                        event.attributes.get("trace_hash"),
+                        event.attributes.get("denom"),
```
sug0, line 148: to avoid writing two chained ifs, you can map and flatten the options
https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten

shared/src/ledger/tx.rs
```diff
@@ -1337,72 +1337,79 @@ pub async fn build_pgf_stewards_proposal<
...
+        client,
+        args.tx.gas_amount,
+        &args.tx.gas_token,
+        args.tx.force,
+    )
+    .await
+    .expect("expected to be able to validate fee");
+
+    args.amount = InputAmount::Validated(validated_amount);
+    args.tx.gas_amount = InputAmount::Validated(validate_fee);
```
sug0, line 1370: I feel like this is unnecessary. can't you use the inner amounts directly?

sug0 (commented):
mostly some code style related comments/suggestions

shared/src/ledger/tx.rs
yito88: You're right. For IBC transfers, we don't have to update them.

sug0 (approved):

PR #1788, opened 2023-08-07
refactor: move vote proposal tx to shared

mateuszjasiuk:
## Based on v0.20.1

- moved logic to shared
- `build_vote_proposal` returns result instead of calling `safe_exit`, which is called on client on Err


mateuszjasiuk:
Closed because it's already done in changes to governance!

PR #1793, opened 2023-08-08
Fix bug in unjail-validator that crashed testnet v0.20.1

### LINKED ISSUE #1790: Apparent duplicate votes on validator set update vote extensions
When a majority quorum is split between more than one validator, we can observe "ignoring duplicate vote" log lines during validator set update vote extension state updates. We should:

- [x] Set up a devnet with 2 validators to replicate the issue locally.
- [x] Check the cause of the output.
- [x] Check if a complete proof (>2/3 voting power) on validator set updates can still be acquired, in spite of these "duplicate voting" attempts.

brentstone:
## Describe your changes

Resolves a bug in the error handling of `fn unjail_validator`.

This bug led to a failure in the testnet based on v0.20.1. In the testnet, a validator who had previously been jailed, and was eligible to unjail themselves, seems to have submitted an unjail transaction twice. The second submission mistakenly did not throw an error, leading to a second insertion of the validator back into the consensus set, which then led to a panic from CometBFT for duplicate validators.

closes #1790

v0.20.1

### DIFF
--- a/.changelog/unreleased/bug-fixes/1793-fix-unjail-bug.md
+++ b/.changelog/unreleased/bug-fixes/1793-fix-unjail-bug.md
@@ -1,0 +1,4 @@
+- Fixes buggy error handling in pos unjail_validator. Now properly enforces that
+  if an unjail tx is submitted when the validator state is something other than
+  Jailed in any of the current or future epochs, the tx will error out and fail.
+  ([\#1793](https://github.com/anoma/namada/pull/1793))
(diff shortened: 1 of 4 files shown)


brentstone:
pls update wasm

tzemanovic (approved):

sug0:
unrelated to #1790, @Fraccaman confirmed this in a devnet with this patchset

PR #1807, opened 2023-08-11
masp: use hardcoded verifying keys

juped:


Fraccaman:
shouldn't we keep the keys in-memory (maybe as part of `VpEnv` ?) instead of deserializing them from bytes every time?

PR #1795, opened 2023-08-08
Arbitrary Bridge pool fee token types

sug0:
## Describe your changes

Closes #1704
Doesn't solve #1799 nor #1800

The PR is configured to merge into the base ([tiago/update-ethbridge-rs](https://github.com/anoma/namada/tree/tiago/update-ethbridge-rs)) to make reviewing easier. It should be changed back to `main`, once the base is upstreamed.

## Indicate on which release or other PRs this topic is based on

Based on #1789

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #1811, opened 2023-08-13
Fix Bridge pool recommendations

### LINKED ISSUE #1799: Fix Bridge pool recommendations SDK functionality
Related to #1704

As of #1795, the Bridge pool recommendations algorithm is "broken" (or more accurately, incomplete). It should take into account the fact that pending transfers in the Bridge pool might be paid in more gas token types other than NAM.

sug0:
## Describe your changes

Closes #1799

Based on #1795

### DIFF
--- a/.changelog/unreleased/features/1781-cap-wnam.md
+++ b/.changelog/unreleased/features/1781-cap-wnam.md
@@ -1,0 +1,2 @@
+- Control the flow of NAM over the Ethereum bridge
+  ([\#1781](https://github.com/anoma/namada/pull/1781))
--- a/.changelog/unreleased/features/1789-update-ethbridge-rs.md
+++ b/.changelog/unreleased/features/1789-update-ethbridge-rs.md
@@ -1,0 +1,2 @@
+- Update ethbridge-rs to v0.22.0
+  ([\#1789](https://github.com/anoma/namada/pull/1789))
--- a/.changelog/unreleased/features/1795-bridge-pool-fees.md
+++ b/.changelog/unreleased/features/1795-bridge-pool-fees.md
@@ -1,0 +1,2 @@
+- Allow Bridge pool transfer fees to be paid in arbitrary token types (except
+  NUTs) ([\#1795](https://github.com/anoma/namada/pull/1795))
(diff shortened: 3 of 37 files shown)


batconjurer (changes requested):
The code changes look fine, but needs to new unit tests to cover the new code paths for provided conversion tables.

sug0:
@batconjurer  there's currently a gotcha in the recommendations algorithm, I think. the `transfer_to_ethereum_progress` RPC method skips "seen" Ethereum transfers. a transfer to be relayed is not recommended if it is not found in the array of transfers returned by `transfer_to_ethereum_progress`; in turn, all transfers considered for relaying are obtained from querying the current signed pool.

the current signed pool may be behind the most up to date Bridge pool, because of the vote extension tx lag. so, I believe the algorithm may recommend a transfer that has already been relayed, that still shows up in the signed pool.

a simple solution may be to return with an error if the signed Bridge pool nonce does not match the up to date Bridge pool nonce in storage. this way we never recommend transfers that have already been relayed and acted upon.

core/src/std_ext.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn maybe_collect_result<C>(self) -> Result<C, E>
+    where
+        C: Default + Extend<T>;
+}
+
+impl<I, T, E> MaybeCollectResult<T, E> for I
+where
+    I: Iterator<Item = Result<T, E>>,
+{
+    fn maybe_collect_result<C>(self) -> Result<C, E>
```
batconjurer, line 16: Is this used?

sug0:
**TODO:** test recommending two different token types

core/src/std_ext.rs
tzemanovic, line 1: I think you can already do this via https://doc.rust-lang.org/stable/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA,+E%3E%3E-for-Result%3CV,+E%3E without `Default` and `Extend`. In bridge_pool.rs where it's used you can call `Result::from_iter` or assign the iter over result to `let iter_res: Result =`. But if you prefer the chained calls I don't mind keeping this, but I'd change it to:

```rust
pub trait MaybeCollectResult<T, E> {
    /// Collect all [`Ok`] values in the iterator, or fail.
    fn maybe_collect_result<C>(self) -> Result<C, E>
    where
        C: FromIterator<T>;
}

impl<I, T, E> MaybeCollectResult<T, E> for I
where
    I: Iterator<Item = Result<T, E>>,
{
    fn maybe_collect_result<C>(self) -> Result<C, E>
    where
        C: FromIterator<T>,
    {
        Result::from_iter(self)
    }
}
```

tzemanovic (approved):

PR #1818, opened 2023-08-14
Remove io::Error from DecodeError

mariari:
We replace io::Error with an ErrorKind and String, and String when the error kind is not available.

This should allow future work to make the API for errors in our codebase easier to satisfy various traits we may wish to derive.

### DIFF
--- a/core/src/types/address.rs
+++ b/core/src/types/address.rs
@@ -4,6 +4,7 @@
 use std::collections::HashMap;
 use std::fmt::{Debug, Display};
 use std::hash::Hash;
+use std::io::ErrorKind;
 use std::str::FromStr;
 
 use bech32::{self, FromBase32, ToBase32, Variant};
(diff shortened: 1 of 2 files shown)


core/src/types/address.rs
```diff
@@ -111,8 +112,10 @@ pub enum DecodeError {
     UnexpectedBech32Prefix(String, String),
     #[error("Unexpected Bech32m variant {0:?}, expected {BECH32M_VARIANT:?}")]
     UnexpectedBech32Variant(bech32::Variant),
+    #[error("Invalid address encoding: {0}, {1}")]
+    InvalidInnerEncoding(ErrorKind, String),
     #[error("Invalid address encoding")]
-    InvalidInnerEncoding(std::io::Error),
+    InvalidInnerEncodingStr(String),
```
tzemanovic, line 118: I think you can remove the `std::io::ErrorKind` too as we don't need to differentiate between them (the strings should capture what the issue is already)

tzemanovic (approved):

core/src/types/address.rs
mariari: Probably, but here it's not too bad, and it's like appending more information on, in a way.

PR #1821, opened 2023-08-14
Mariari/shared error type

mariari:
## Describe your changes

We create a shared error type for most kinds of errors that are seen in the shared crate.

This will be used for a followup where we remove all possible panic locations found within shared

This PR relies on
- #1818 
- #1819 
- #1820 


shared/src/types/error.rs
```diff
@@ -0,0 +1,282 @@
...
+    Other(String),
+}
+
+/// Checks if the given error is an invalid viewing key
+pub fn is_pinned_error<T>(err: &Result<T>) -> bool {
+    match err {
+        Ok(_) => false,
+        Err(Pinned(PinnedBalanceError::InvalidViewingKey)) => true,
+        Err(_) => false,
+    }
```
tzemanovic, line 281:
```suggestion
    matches!(err, Err(Pinned(PinnedBalanceError::InvalidViewingKey)))
```

tzemanovic (dismissed):

shared/src/types/error.rs
mariari: Fixed in the latest patch series

tzemanovic (approved):

PR #1833, opened 2023-08-18
Properly Error on imporper Jailing

mariari:
## Describe your changes

We make the erroing a bit more explicit in the crash here


mariari:
#1834 is an explicit integration branch to see how this topic interacts with the changes made in #1825 

It should help inform what integration should be done when we merge it into draft

PR #1836, opened 2023-08-19
Chain id check in `prepare_proposal`

grarco:


apps/src/lib/node/ledger/shell/prepare_proposal.rs
```diff
@@ -227,6 +227,10 @@ where
     {
         let tx = Tx::try_from(tx_bytes).map_err(|_| ())?;
 
+        if tx.header.chain_id != self.chain_id {
+            return Err(());
+        }
```
sug0, line 232: this should be validated by `mempool_validate` IMO

grarco: We already do that. I was wondering if it was better to double check it in `prepare_proposal` too, just in case someone removed it from `mempool_validate`. But I guess we can assume that if someone starts touching the code they know what they're doing (and I think this is why this check is not done in `prepare_proposal`). I'm gonna close this, thanks 🙏🏻

PR #1846, opened 2023-08-28
Fix submit_reveal_aux

yito88:
## Describe your changes
To call `submit_reveal_aux` without the CLI context from a client application, e.g. IBC relayer.
It is enough to give only the wallet since `submit_reveal_aux` requires only the wallet of the CLI context.

v0.21.1

### DIFF
--- a/.changelog/unreleased/improvements/1846-simplify-reveal.md
+++ b/.changelog/unreleased/improvements/1846-simplify-reveal.md
@@ -1,0 +1,2 @@
+- Call submit_reveal_tx() without CLI context
+  ([\#1846](https://github.com/anoma/namada/issues/1846))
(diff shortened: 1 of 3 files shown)


mariari (commented):
Overall seems fine, the API change is inline with other functions in the file, so it doesn't seem out of place.

murisi:
Hi @yito88 ! Could I please see the context in which you need to call `submit_reveal_aux` without a `Context` available? I have a rough idea of how to integrate your changes, but I just want to make sure I understand this SDK usage pattern.

yito88:
@murisi Sorry for the late reply.
Hermes (IBC relayer) submits a Namada transaction. It requires to call `submit_reveal_aux` [here](https://github.com/heliaxdev/hermes/blob/526851f92b9cb08da062270b72a24fb4623608d2/crates/relayer/src/chain/namada/tx.rs#L122). Welcome a better way to submit a tx.

yito88:
Sorry for the mess. I've checked the latest SDK changes.
I'd like to close this PR and fix Hermes for it.

PR #1868, opened 2023-09-06
Mariari/sdk folder

mariari:


sug0 (commented):
this is a step in the right direction, but it should be a separate crate altogether. `shared` should import the `sdk` crate, so it's actually decoupled at the dependency level. the `sdk` should be a leaner crate, and therefore easier+faster to build.

another thing is, I think you should wait for `v0.22.0`, which is a huge release (it has like a 17k loc diff)

mariari:
> this is a step in the right direction, but it should be a separate crate altogether. `shared` should import the `sdk` crate, so it's actually decoupled at the dependency level. the `sdk` should be a leaner crate, and therefore easier+faster to build.
> 
> another thing is, I think you should wait for `v0.22.0`, which is a huge release (it has like a 17k loc diff)

No that is too much effort. `vm` relies upon some of the functions that belong in the `sdk` which is apart of shared. After battling it for a day, I decided to just do it in the current crate, where anything apart of `sdk` is apart of the SDK itself.

If I were to go down that route, then we'd end up splitting shared into 3 crates

1. the crate that doesn't call into SDK
2. SDK
3. the crate with modules like vm that call into the SDK

mariari:
I've rebased it on v0.22.0 as I wanted to bring in other changes

adrianbrink (approved):

PR #1873, opened 2023-09-07
ledger: remove PoW/faucet

Fraccaman:
## Describe your changes
- Remove `pow-for-fee` feature
- Remove faucet vp/account


apps/src/lib/node/ledger/shell/testing/node.rs
```diff
@@ -404,6 +405,7 @@ impl<'a> Client for &'a MockNode {
         let tx_bytes: Vec<u8> = tx.into();
         self.submit_tx(tx_bytes);
         if !self.success() {
+            resp.code = tendermint::abci::Code::Err(1337); // TODO: submit_tx should return the correct error code + message
```
Fraccaman, line 408: @batconjurer I had to add this otherwise it would keep try to send the tx (cc @grarco )

apps/src/lib/client/signing.rs
grarco, line 1: I think this file keeps coming back but we don't actually need it

core/src/types/storage.rs
```diff
@@ -540,6 +540,18 @@ impl Key {
         self.segments.first()
     }
 
+    /// Return the owner of this subspace. None if empty or not an Address.
+    pub fn owner(&self) -> Option<Address> {
```
grarco, line 544: Do we need this?

grarco (dismissed):

PR #1878, opened 2023-09-08
Removes fee panics from sdk

grarco:
## Describe your changes

Removes some wrong `panic!` that were mistakenly brought into the sdk by the gas&fee pr.

v0.22.0

### DIFF
--- a/shared/src/ledger/signing.rs
+++ b/shared/src/ledger/signing.rs
@@ -277,10 +277,11 @@ pub async fn aux_signing_data<
     };
 
     if fee_payer == masp_tx_key().to_public() {
-        panic!(
+        other_err(
             "The gas payer cannot be the MASP, please provide a different gas \
              payer."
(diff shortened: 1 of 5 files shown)


shared/src/ledger/signing.rs
```diff
@@ -277,10 +277,11 @@ pub async fn aux_signing_data<
     };
 
     if fee_payer == masp_tx_key().to_public() {
-        panic!(
+        return Err(Error::Tx(TxError::Other(
```
mariari, line 280: For this one it may be god to just return a Error::Other. Though it doesn't matter too much.

mariari (approved):
Overall looks reasonable.

Only suggestion I would have is maybe split the `error.rs` additions into their own commit so that there is a more clear point in where they are added

adrianbrink (approved):

PR #1884, opened 2023-09-11
Nondegenerate multisignature hardware wallets

murisi:
## Describe your changes
* Combined Signature and MultiSignature to reduce logic duplication. Specifically:
    * Decompression logic no longer needs to be separately implemented for both types of signatures
    * Hardware wallet would no longer need to implement separate hashers, storage logic, and serializers for both types of signatures
* Allow signer to be specified in signature section, otherwise indexing data would need to be sent to hardware wallet for nondegenerate multisig transactions
* Added decompression logic in order to eliminate the decompression hacks we are using in `namada-interface` (we should explore in future whether this can be used to reduce size of signature sections on protocol)
* Using map instead of set to store indexed signatures in order to remedy the signature double counting issues (multiple valid signatures can exist for given hash and public key).
* Eliminated signature double counting across different sections by storing successfully verified indices in a set.

Namada v0.22.0


core/src/proto/types.rs
```diff
@@ -1346,68 +1375,73 @@ impl Tx {
...
+                            Error::InvalidSectionSignature(
+                                "found invalid signature.".to_string(),
+                            )
+                        });
+                    // Compute the cost of the signature verifications
+                    if let Some(x) = gas_meter.as_mut() {
+                        let amt_verified = usize::from(amt_verifieds.is_err())
+                            + verified_pks.len()
+                            - prev_verifieds;
+                        x.consume(VERIFY_TX_SIG_GAS_COST * amt_verified as u64)
```
grarco, line 1429: I think that given this rework it would be slightly better to move the gas accounting part directly into the `verify_signature` method of `Signature`, to have a highest possible resolution. But we can do this later

grarco (approved):
Looks good to me, thanks!

PR #1894, opened 2023-09-14
Makes async traits in the SDK be `Send`

grarco:
## Describe your changes

Requires the `Send` bound on the `async_traits`s in the SDK. This allow the usage of the SDK in an environment which is both async and multithread (see https://rust-lang.github.io/async-book/07_workarounds/03_send_approximation.html).

v0.22.0

### DIFF
--- a/.changelog/unreleased/SDK/1894-sdk-async-traits-send.md
+++ b/.changelog/unreleased/SDK/1894-sdk-async-traits-send.md
@@ -1,0 +1,3 @@
+- Added the `Send` bound to the `Client` and `ShieldedUtils` `async_trait`s'.
+  This allows the SDK to be used in environments which are both asynchronous and
+  multithread. ([\#1894](https://github.com/anoma/namada/pull/1894))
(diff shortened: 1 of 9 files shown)


mariari (dismissed):
Looks good, Thanks for the changelog.

sug0 (commented):
I think this will be a problem for the namada interface; see https://github.com/rustwasm/wasm-bindgen/issues/2833

and also https://github.com/anoma/namada-interface/blob/2ccc40b1a5092373a1223c55a74d3dcaa2258d6c/packages/shared/lib/src/rpc_client.rs#L16C3-L20

We can gate the ability to send futures across threads with something like:
```rust
#[cfg_attr(not(feature = "send-sdk-client"), async_trait(?Send))]
#[cfg_attr(feature = "send-sdk-client", async_trait)]
trait Client { ... }
```

Fraccaman:
@sug0  yeah thats what we were also thinking! We asked @jurevans to test this branch, lets see how it goes

jurevans (approved):
Tested, and works well for me!

PR #1898, opened 2023-09-15
Fix the trimming of the record of the first block height of each epoch

### LINKED ISSUE #1897: Ledger panics (index out of bounds) on 2nd new epoch after block height > consensus_params.evidence.max_age_num_blocks
(using Namada version `v0.22.0` and Cometbft version `0.37.2`)

All validators and full nodes on our local testnet would panic with the following error shortly after block height `100000`, which corresponds to the default value of `consensus_params.evidence.max_age_num_blocks`:

```
I[2023-09-01|10:07:45.614] finalizing commit of block                   module=consensus height=100421 hash=F87B325287D0864A6B1D5F964B173137D801EBB9BE39C2D26A3C845B4F1396CB root=735053B51C4A798D319906F46C05351CD135F5078602277381606FE089161B0B num_txs=0
2023-09-01T10:07:45.618931Z  INFO namada_core::ledger::storage::wl_storage: Began a new epoch 457
2023-09-01T10:07:45.618952Z  INFO namada_apps::node::ledger::shell::finalize_block: Block height: 100421, epoch: 457, is new epoch: true.
Message:  index out of bounds: the len is 456 but the index is 456
Location: apps/src/lib/node/ledger/shell/finalize_block.rs:695

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
2023-09-01T10:07:47.281671Z  INFO namada_apps::node::ledger::shims::abcipp_shim: ABCI response channel didn't respond
E[2023-09-01|10:07:47.282] Stopping abci.socketClient for error: read message: EOF module=abci-client connection=consensus
Message:  called `Result::unwrap()` on an `Err` value: RecvError(())
Location: /usr/local/cargo/git/checkouts/tower-abci-0d01b039e0b7a0c9/cf9573d/src/server.rs:163

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
I[2023-09-01|10:07:47.283] service stop                                 module=abci-client connection=consensus msg="Stopping socketClient service" impl=socketClient
E[2023-09-01|10:07:47.283] error in proxyAppConn.EndBlock               module=state err="read message: EOF"
E[2023-09-01|10:07:47.284] consensus connection terminated. Did the application crash? Please restart CometBFT module=proxy err="read message: EOF"
E[2023-09-01|10:07:47.284] CONSENSUS FAILURE!!!                         module=consensus err="failed to apply block; error read message: EOF" stack="goroutine 262 [running]:\nruntime/debug.Stack()\n\truntime/debug/stack.go:24 +0x65\ngithub.com/cometbft/cometbft/consensus.(*State).receiveRoutine.func2()\n\tgithub.com/cometbft/cometbft/consensus/state.go:732 +0x4c\npanic({0xe92440, 0xc001a434a0})\n\truntime/panic.go:838 +0x207\ngithub.com/cometbft/cometbft/consensus.(*State).finalizeCommit(0xc0000ae000, 0x18845)\n\tgithub.com/cometbft/cometbft/consensus/state.go:1709 +0xf05\ngithub.com/cometbft/cometbft/consensus.(*State).tryFinalizeCommit(0xc0000ae000, 0x18845)\n\tgithub.com/cometbft/cometbft/consensus/state.go:1609 +0x2ff\ngithub.com/cometbft/cometbft/consensus.(*State).enterCommit.func1()\n\tgithub.com/cometbft/cometbft/consensus/state.go:1544 +0xa5\ngithub.com/cometbft/cometbft/consensus.(*State).enterCommit(0xc0000ae000, 0x18845, 0x0)\n\tgithub.com/cometbft/cometbft/consensus/state.go:1582 +0xcb7\ngithub.com/cometbft/cometbft/consensus.(*State).addVote(0xc0000ae000, 0xc0017345a0, {0xc000108810, 0x28})\n\tgithub.com/cometbft/cometbft/consensus/state.go:2212 +0xcbf\ngithub.com/cometbft/cometbft/consensus.(*State).tryAddVote(0xc0000ae000, 0xc0017345a0, {0xc000108810?, 0xc00027ff00?})\n\tgithub.com/cometbft/cometbft/consensus/state.go:2001 +0x2c\ngithub.com/cometbft/cometbft/consensus.(*State).handleMsg(0xc0000ae000, {{0x127bfa0?, 0xc0012b6820?}, {0xc000108810?, 0x0?}})\n\tgithub.com/cometbft/cometbft/consensus/state.go:861 +0x44b\ngithub.com/cometbft/cometbft/consensus.(*State).receiveRoutine(0xc0000ae000, 0x0)\n\tgithub.com/cometbft/cometbft/consensus/state.go:768 +0x419\ncreated by github.com/cometbft/cometbft/consensus.(*State).OnStart\n\tgithub.com/cometbft/cometbft/consensus/state.go:379 +0x12d\n"
I[2023-09-01|10:07:47.284] service stop                                 module=consensus wal=/root/.local/share/namada/luminara.79474f00ace3ef7ca2712/cometbft/data/cs.wal/wal msg="Stopping baseWAL service" impl=baseWAL
I[2023-09-01|10:07:47.284] signal trapped                               module=main msg="captured terminated, exiting..."
I[2023-09-01|10:07:47.285] service stop                                 module=consensus wal=/root/.local/share/namada/luminara.79474f00ace3ef7ca2712/cometbft/data/cs.wal/wal msg="Stopping Group service" impl=Group
I[2023-09-01|10:07:47.285] service stop                                 module=main msg="Stopping Node service" impl=Node
I[2023-09-01|10:07:47.285] Stopping Node                                module=main
```

At the start of a new epoch, PoS will recalculate inflation wrt the previous epoch by checking `pred_epochs.first_block_heights` at the index equal to the previous epoch:

```
let first_block_of_last_epoch = self
      .wl_storage
      .storage
      .block
      .pred_epochs
      .first_block_heights[last_epoch.0 as usize]
      .0;
```

but this doesn't take into account that `pred_epochs` is trimmed to only keep epochs from less than `consensus_params.evidence.max_age_num_blocks` ago and after reaching that height, indices will no longer directly correspond to epoch number. The first time through after reaching this height will incorrectly reference the current epoch instead of the previous and the second time through will result in an out of bounds error.

Steps:
1. Network nodes consistently panic at block height ~100200 on new epoch start
2. Set epoch duration to half; network will produce ~double the number of epochs but still panic at the same block height.
3. Change the constant `evidence_max_age_num_blocks` in `core/src/ledger/storage/wl_storage.rs` from `100000` to `1000`; nodes now panic on second new epoch after height 1000 instead.
4. After making the following change and relaunching network; nodes no longer panic at heights > `consensus_params.evidence.max_age_num_blocks`. Our localnet is currently at block 111000 and counting
```
        // Get the number of blocks in the last epoch
        let last_epoch_index: usize = last_epoch.0 as usize - self.wl_storage.storage.block.pred_epochs.first_known_epoch.0 as usize;
        let first_block_of_last_epoch = self
            .wl_storage
            .storage
            .block
            .pred_epochs
            .first_block_heights[last_epoch_index]
            .0;
```

(I'm not sure if this is the 'right' way to modify the code but at least it seems to confirm the cause/solution)

brentstone:
## Describe your changes
Closes #1897. To resolve the issue, we no longer trim this data and now keep the first block height of every past epoch in storage.

Based on v0.22.0

### DIFF
--- a/core/src/ledger/storage/wl_storage.rs
+++ b/core/src/ledger/storage/wl_storage.rs
@@ -215,13 +215,8 @@ where
             self.storage.next_epoch_min_start_height =
                 height + min_num_of_blocks;
             self.storage.next_epoch_min_start_time = time + min_duration;
-            // TODO put this into PoS parameters and pass it to tendermint
-            // `consensus_params` on `InitChain` and `EndBlock`
-            let evidence_max_age_num_blocks: u64 = 100000;
-            self.storage
-                .block
-                .pred_epochs
-                .new_epoch(height, evidence_max_age_num_blocks);
+
+            self.storage.block.pred_epochs.new_epoch(height);
             tracing::info!("Began a new epoch {}", self.storage.block.epoch);
         }
         Ok(new_epoch)
--- a/core/src/types/storage.rs
+++ b/core/src/types/storage.rs
@@ -1168,29 +1168,8 @@ impl Default for Epochs {
 }
 
 impl Epochs {
-    /// Record start of a new epoch at the given block height and trim any
-    /// epochs that ended more than `max_age_num_blocks` ago.
-    pub fn new_epoch(
-        &mut self,
-        block_height: BlockHeight,
-        max_age_num_blocks: u64,
-    ) {
-        let min_block_height_to_keep = (block_height.0 + 1)
-            .checked_sub(max_age_num_blocks)
-            .unwrap_or_default();
-        // trim off any epochs whose last block is before the limit
-        while let Some((_first_known_epoch_height, rest)) =
-            self.first_block_heights.split_first()
-        {
-            if let Some(second_known_epoch_height) = rest.first() {
-                if second_known_epoch_height.0 < min_block_height_to_keep {
-                    self.first_known_epoch = self.first_known_epoch.next();
-                    self.first_block_heights = rest.to_vec();
-                    continue;
-                }
-            }
-            break;
-        }
+    /// Record start of a new epoch at the given block height
+    pub fn new_epoch(&mut self, block_height: BlockHeight) {
         self.first_block_heights.push(block_height);
     }
 
--- a/core/src/types/storage.rs
+++ b/core/src/types/storage.rs
@@ -1638,10 +1638,9 @@ mod tests {
         println!("epochs {:#?}", epochs);
         assert_eq!(epochs.get_height(Epoch(0)), Some(BlockHeight(0)));
         assert_eq!(epochs.get_epoch(BlockHeight(0)), Some(Epoch(0)));
-        let mut max_age_num_blocks = 100;
 
         // epoch 1
-        epochs.new_epoch(BlockHeight(10), max_age_num_blocks);
+        epochs.new_epoch(BlockHeight(10));
         println!("epochs {:#?}", epochs);
         assert_eq!(epochs.get_height(Epoch(1)), Some(BlockHeight(10)));
         assert_eq!(epochs.get_epoch(BlockHeight(0)), Some(Epoch(0)));
@@ -1671,7 +1670,7 @@ mod tests {
         );
 
         // epoch 2
-        epochs.new_epoch(BlockHeight(20), max_age_num_blocks);
+        epochs.new_epoch(BlockHeight(20));
         println!("epochs {:#?}", epochs);
         assert_eq!(epochs.get_height(Epoch(2)), Some(BlockHeight(20)));
         assert_eq!(epochs.get_epoch(BlockHeight(0)), Some(Epoch(0)));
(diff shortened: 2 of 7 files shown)


core/src/ledger/storage/wl_storage.rs
```diff
@@ -215,13 +216,13 @@ where
             self.storage.next_epoch_min_start_height =
                 height + min_num_of_blocks;
             self.storage.next_epoch_min_start_time = time + min_duration;
+
             // TODO put this into PoS parameters and pass it to tendermint
             // `consensus_params` on `InitChain` and `EndBlock`
```
brentstone, line 221: @tzemanovic how should we pass this to `FinalizeBlock`, assuming we only want to do that if the value of the param changes? In this case, it would happen thru a gov proposal right? Would we need to read the value in `finalize_block` before an `execute_governance_proposals`, then read it again to see if it changed, and then if so include it in `FinalizeBlock::consensus_param_updates`?

tzemanovic: Yeah, it can be changed via gov, but we could probably just always read it and pass it to Comet and if it's the same it wouldn't have any effect

proof_of_stake/src/parameters.rs
```diff
@@ -46,6 +46,9 @@ pub struct PosParams {
     /// The minimum amount of bonded tokens that a validator needs to be in
     /// either the `consensus` or `below_capacity` validator sets
     pub validator_stake_threshold: token::Amount,
+    /// The number of blocks in the past for which the first block height of an
+    /// epoch is recorded
+    pub max_age_num_blocks: u64,
```
tzemanovic, line 51: If we were to use this param and pass it to CometBFT, it would invalidate any evidence that's more than this number of blocks in the past, so we'd need to ensure that it's greater than our slashable number of epochs * `min_num_of_blocks`, but that's complicated because the number of blocks in epoch may vary, so I think we're better off not using this param and filter out outdated evidence in the app (as we already do)

brentstone:
pls update wasm

Fraccaman (approved):

core/src/types/storage.rs
```diff
@@ -1161,48 +1159,25 @@ impl Default for Epochs {
...
-        }
+    /// Record start of a new epoch at the given block height
+    pub fn new_epoch(&mut self, block_height: BlockHeight) {
         self.first_block_heights.push(block_height);
     }
 
-    /// Look-up the epoch of a given block height.
+    /// Look up the epoch of a given block height.
+    /// TODO: handle passing of a future block height? (only valid for current
+    /// or past)
```
tzemanovic, line 1175: We can't determine future epochs because epoch progression depends on 2 params - min. num of blocks and min duration. I think if the given `block_height > current_height` we can return `None`

tzemanovic: ah, nvm, we don't have enough context here to check current height. The current behavior is that we return the current epoch, which is probably fine

tzemanovic, line 1175:
```suggestion
    /// Look up the epoch of a given block height. If the given height is
    /// greater than the current height, the current epoch will be returned even
    /// though an epoch for a future block cannot be determined.
```

tzemanovic (dismissed):

core/src/types/storage.rs
brentstone: Yeah, I put this here because it doesn't rly make sense to pass a future block height into this, but figured if this function were to eventually also take in the current block height, we could return `None` perhaps

brentstone:
Merged into v0.23.0

PR #1865, opened 2023-09-05
Rework voting on Ethereum tallies across epoch boundaries

sug0:
## Describe your changes

Closes #1288

This PR changes the way voting across epochs on Ethereum tallies works. Instead of accounting for the average voting power available across all epochs $E$ some tally $t$ takes place in, $t$ is considered `seen` if its accumulated voting power is greater than $2/3 \max(\sigma(e) : e \in E)$, where $\sigma$ is the function that returns the voting power available in the input epoch.

With this strategy, tallies won't stall if they cross epoch boundaries, while still respecting the Tendermint BFT protocol's safety property (i.e. linearizability is preserved).

## Indicate on which release or other PRs this topic is based on

`v0.22.0`

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #1899, opened 2023-09-18
Retransmit some timed out Ethereum events

sug0:
## Describe your changes

In case an Ethereum event times out with $>\frac{1}{3}$ voting power accumulated behind it, it may be proposed once more during the **COMMIT** phase of the protocol, such that its state updates may be reconsidered. BFT protocols often have this notion of weak certificates on proposed data (e.g. see Sousa et al. in Mod-SMaRt).

Based on #1865 

### DIFF
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -28,6 +28,7 @@ use std::rc::Rc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use masp_primitives::transaction::Transaction;
+use namada::core::hints;
 use namada::core::ledger::eth_bridge;
 use namada::ledger::eth_bridge::{EthBridgeQueries, EthereumOracleConfig};
 use namada::ledger::events::log::EventLog;
@@ -902,6 +903,11 @@ where
             events.sort();
             events
         };
+        if hints::likely(eth_events.is_empty()) {
+            // more often than not, there won't by any expired
+            // Ethereum events to retransmit
+            return;
+        }
         if let Some(vote_extension) = self.sign_ethereum_events(eth_events) {
             let protocol_key = self
                 .mode
--- a/apps/src/lib/node/ledger/shell/mod.rs
+++ b/apps/src/lib/node/ledger/shell/mod.rs
@@ -2203,12 +2203,60 @@ mod abciplus_mempool_tests {
     use namada::types::ethereum_events::EthereumEvent;
     use namada::types::key::RefTo;
     use namada::types::storage::BlockHeight;
-    use namada::types::transaction::protocol::{ProtocolTx, ProtocolTxType};
+    use namada::types::transaction::protocol::{
+        ethereum_tx_data_variants, ProtocolTx, ProtocolTxType,
+    };
     use namada::types::vote_extensions::{bridge_pool_roots, ethereum_events};
 
     use super::*;
     use crate::node::ledger::shell::test_utils;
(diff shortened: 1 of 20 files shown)


tzemanovic (approved):

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -849,32 +849,71 @@ where
...
+            let mut events: Vec<_> = self
+                .wl_storage
+                .storage
+                .expired_txs_queue
+                .drain()
+                .map(|expired_tx| match expired_tx {
+                    ExpiredTx::EthereumEvent(event) => event,
+                })
+                .collect();
+            events.sort();
```
batconjurer, line 902: what's the point of sorting them?

apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -124,7 +135,7 @@ where
                 .get_current_decision_height(),
             #[cfg(not(feature = "abcipp"))]
             block_height: self.wl_storage.storage.get_last_block_height(),
-            ethereum_events: self.new_ethereum_events(),
+            ethereum_events,
```
batconjurer, line 138: It seems possible to send a vote extension here when the `ethereum_events` are empty.

ethereum_bridge/src/protocol/transactions/votes/storage.rs
```diff
@@ -36,21 +38,39 @@ where
     Ok(())
 }
 
+/// Delete a tally from storage, and return the associated payload of
+/// type `T` being voted on, in case it has accumulated more than 1/3
+/// of fractional voting power behind it.
+#[must_use = "The optional payload returned by this function must be used"]
```
batconjurer, line 44: nit: the word "payload" seems less natural than something like "value" since this is coming from storage.

apps/src/lib/node/ledger/shell/vote_extensions.rs
sug0: this is the intended behavior. we always sign sets of ethereum events, even if we find none. this can be changed, but it's out of the scope of this PR.

apps/src/lib/node/ledger/shell/mod.rs
sug0: I think this was about being absolutely deterministic in the payload you signed. can't remember the exact details of this tbh. maybe it made more sense when we had digests of events

ethereum_bridge/src/protocol/transactions/votes/storage.rs
sug0: sure, can change this

apps/src/lib/node/ledger/shell/vote_extensions.rs
batconjurer: But I think with the retransmission, it will retransmit empty vexts no?

sug0: oh, I think so yeah. I'll add a check

apps/src/lib/node/ledger/shell/mod.rs
sug0: so the digests were used to reconstruct the vote extensions from every validator, signature pair stored within. I think this was the reason we required events to be sorted, so we had signatures over the same data. it's probably not required with individual vote extensions, for them to be signed

batconjurer (approved):

PR #1905, opened 2023-09-19
Replay protection improvements

### LINKED ISSUE #1568: Remove invalid commitment tx hash from storage
Currently, for replay protection, we remove the hash of an inner transaction from storage (i.e. allow replay) only when the tx goes out of gas. We should do the same in one more case: if the inner tx is not the expected one, i.e. the commitment in the wrapper is wrong

grarco:
## Describe your changes

Closes #1568.

Better handles the case of an invalid wrapper tx (ABCI only) regarding fee payment and replay protection: the hash of the wrapper is committed to storage to avoid replays given that the wrapper signer, in this case, is forced to pay whatever amount of fees they can.

Remove the hash of the inner tx (for replay protection purposes) in more cases:
- if the inner tx is actually `Undecryptable`
- if the signature on the inner tx is invalid
- if the transaction fails because of an invalid commitment to one of its sections (this is because we only use the hash of the `Header` for replay protection)

This makes for a more correct replay protection mechanism. To do this, I used some sentinel variables to be able to detect these errors in wasm. Also fixed a bug regarding the `OutOfGas` error which, for the same reason, could not be detected and was not triggering the hash removal (again, fixed using the same sentinel variable). Theoretically, the wasm error can be downcasted to the original error but the method to do so is fallible (and indeed, in some tests I ran it failed because of multiple strong references on the inner `Arc`): so, to guarantee a consisting behavior, I had to rely on the sentinel.

v0.25.0


Fraccaman:
is this mergeable in 0.24.0?

grarco:
Converting this back to draft since it needs a few touches because of #1867

grarco:
@tzemanovic ready again for review

shared/src/vm/wasm/run.rs
```diff
@@ -243,7 +251,22 @@ where
         keys_changed,
         verifiers,
         gas_meter,
-    )
+    ) {
+        Ok(accept) => {
+            if !accept && sentinel.is_invalid_signature() {
```
cwgoes, line 256: it should never be true that `accept && sentinel.is_invalid_signature()`, right? maybe we should explicitly check

cwgoes (changes requested):
See comments, and @tzemanovic should definitely review this as I'm not familiar with all of the modified code.

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -310,13 +314,17 @@ where
...
+                                self.allow_tx_replay(tx_in_queue.tx);
+
                                 event["info"] =
                                     "Transaction is invalid.".into();
                                 event["log"] = "Transaction could not be \
                                                 decrypted."
                                     .into();
                                 event["code"] =
                                     ErrorCodes::Undecryptable.into();
+                                response.events.push(event);
```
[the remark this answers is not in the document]
grarco: Yes correct

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -521,6 +540,9 @@ where
                             .expect("Error while writing tx hash to storage");
                     }
 
+                    stats.increment_errored_txs();
+                    self.wl_storage.drop_tx();
```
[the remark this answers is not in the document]
grarco: No just moved these two instructions from above, the order is not important

shared/src/vm/host_env.rs
```diff
@@ -1854,32 +1897,40 @@ where
...
+        max_signatures,
+        &mut Some(gas_meter),
+    ) {
+        Ok(_) => Ok(HostEnvResult::Success.to_i64()),
+        Err(err) => match err {
+            namada_core::proto::Error::OutOfGas(inner) => {
+                sentinel.set_out_of_gas();
+                Err(vp_host_fns::RuntimeError::OutOfGas(inner))
+            }
+            _ => {
```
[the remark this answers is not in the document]
grarco: The `verify_signatures` method only returns two error variants atm, either `OutOfGas` or `InvalidSectionSignature` so, for now, this is safe. But even if we modified that method to return, say, a third error variant, my idea was that every error which is not due to gas should be considered a signature error but this assumption might be a little too relaxed?

shared/src/vm/wasm/run.rs
grarco: Yes correct, if the vp does verify the signature and the check on the signature failed than the vp will return `false`

shared/src/vm/host_env.rs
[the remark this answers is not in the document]
cwgoes: I would not make that assumption, seems like it could be violated in the future

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -2386,72 +2442,155 @@ mod test_finalize_block {
...
-        // tx is still there and corresponding wrapper hash has been removed
-        // since useless
-        assert_eq!(event[0].event_type.to_string(), String::from("accepted"));
+        // Check inner tx hash has been removed from storage
+        assert_eq!(event[0].event_type.to_string(), String::from("applied"));
         let code = event[0]
             .attributes
             .get("code")
-            .expect("Test failed")
+            .expect("Testfailed")
```
tzemanovic, line 2450: tiny nit, for cases like this where there's no meaningful message you can just use `unwrap`

shared/src/vm/wasm/run.rs
tzemanovic: With the way we're using `verify_signatures` that would set the sentinel in our VPs the result must be `false` if the sig is invalid (there's no alternative authorization method), so I agree that we should explicitly check this

shared/src/vm/wasm/run.rs
```diff
@@ -690,6 +721,7 @@ mod tests {
         let mut gas_meter = VpGasMeter::new_from_tx_meter(
             &TxGasMeter::new_from_sub_limit(TX_GAS_LIMIT.into()),
         );
+        let _invalid_sig = false;
```
tzemanovic, line 724: are these unused declarations needed?

shared/src/vm/host_env.rs
```diff
@@ -1997,6 +2053,18 @@ where
     Ok(())
 }
 
+/// Set the sentinel for an invalid tx section commitment
+pub fn tx_set_commitment_sentinel<MEM, DB, H, CA>(env: &TxVmEnv<MEM, DB, H, CA>)
```
tzemanovic, line 2057: let's have this return `TxResult<()>` (even though it's infallible, the consistency it will avoid having a special macro case for it in tests)

tzemanovic: actually, I take it back - it would make the usage awkward when we `.map_err` which is more important

tzemanovic (dismissed):

shared/src/vm/wasm/run.rs
grarco: Ah no they are typos from the very first version of the sentinel system, I'll remove them 👍🏻

tzemanovic (approved):

tzemanovic (approved):

PR #1917, opened 2023-09-21
Receive a token to a payment address over IBC

yito88:
## Describe your changes
The sender needs to do two operations to send a token to the payment address on the destination Namada
1. Request to generate MASP transaction with `ibc-gen-shielded` command on the destination Namada
    - Need not only the receiver(payment address), the token and the amount, but also port/channel IDs that are used in the following IBC transfer
    - The command makes a file including the serialized memo for the IBC transfer
2. Set the MASP transfer to the memo field of MsgTransfer
    - with the option `--memo-path` to specify the memo file

Based on #1951


shared/src/ledger/vp_host_fns.rs
```diff
@@ -333,6 +334,20 @@ where
...
+    _gas_meter: &mut VpGasMeter,
+    write_log: &WriteLog,
+    event_type: String,
+) -> EnvResult<Option<IbcEvent>> {
+    for event in write_log.get_ibc_events() {
+        if event.event_type == event_type {
+            return Ok(Some(event.clone()));
+        }
+    }
+    Ok(None)
```
sug0, line 348: instead of doing a linear look-up, we can eventually get the events from a map

core/src/types/ibc.rs
```diff
@@ -84,6 +102,53 @@ mod ibc_rs_conversion {
...
+            Self::try_from_slice(&bytes)
+                .map_err(Error::DecodingShieldedTransfer)
+        }
+    }
+
+    /// Get the shielded transfer from the memo
+    pub fn get_shielded_transfer(
+        event: &IbcEvent,
+    ) -> Result<Option<IbcShieldedTransfer>> {
+        if event.event_type != "fungible_token_packet" {
```
sug0, line 130: is the string `"fungible_token_packet"` part of any ICS (maybe ICS 20)? either way, it would be best to assign a constant binding to it, IMO

core/src/ledger/vp_env.rs
```diff
@@ -97,6 +106,45 @@ where
...
+                    .ok_or_err_msg("unable to find shielded hash")?;
+                let masp_tx = signed
+                    .get_section(&shielded_hash)
+                    .and_then(|x| x.as_ref().masp_tx())
+                    .ok_or_err_msg("unable to find shielded section")?;
+                Ok((transfer, masp_tx))
+            }
+            Err(_) => {
+                if let Some(event) =
+                    self.get_ibc_event("fungible_token_packet".to_string())?
```
sug0, line 128: I see the same event type here

core/src/ledger/vp_env.rs
```diff
@@ -97,6 +106,45 @@ where
...
+                            "No shielded transfer in the IBC event",
+                        ))
+                    }
+                } else {
+                    Err(storage_api::Error::new_const(
+                        "No IBC event for the shielded action",
+                    ))
+                }
+            }
+        }
```
sug0, line 145: I think this block of code could be improved. IMO it doesn't read very well. if possible, I'd like it to remain somewhat flat (at at most two indentation levels)

core/src/ledger/ibc/mod.rs
```diff
@@ -204,6 +209,38 @@ where
...
     }
+
+    /// Handle the MASP transaction if needed
+    fn handle_masp_tx(&mut self, envelope: &MsgEnvelope) -> Result<(), Error> {
+        let shielded_transfer = match envelope {
+            MsgEnvelope::Packet(PacketMsg::Recv(_)) => {
+                let event = self
+                    .ctx
+                    .borrow()
+                    .get_ibc_event("fungible_token_packet")
```
sug0, line 225: same event type here

apps/src/lib/client/tx.rs
```diff
@@ -1467,3 +1468,32 @@ pub async fn submit_tx(
...
+        };
+        let out = File::create(&output_path)
+            .expect("Should be able to create the out file.");
+        serde_json::to_writer_pretty(out, &Memo::from(shielded_transfer))
+            .expect("IBC memo should be deserializable.");
+    } else {
+        eprintln!("No shielded transfer for this IBC transfer.")
+    }
+    Ok(())
+}
```
sug0: it should be possible to hold the intermediary memo in memory, and write it to the ibc transfer msg in a single CLI command, no? it's better from a UX perspective IMO

generating the shielded transfer memo as a separate command can still be useful though.

benches/lib.rs
```diff
@@ -733,44 +730,10 @@ impl BenchShieldedCtx {
...
-            token: address::nam(),
-            amount: InputAmount::Validated(DenominatedAmount {
-                amount,
-                denom: 0.into(),
-            }),
-            native_token: self.shell.wl_storage.storage.native_token.clone(),
-            tx_code_path: TX_TRANSFER_WASM.into(),
+        let denominated_amount = DenominatedAmount {
+            amount,
+            denom: 0.into(),
```
sug0, line 735: we default to the minimum denomination for all ibc amounts? this could be improved, since it steers away from the default behavior of the namada cli in all other instances (i.e. we always validate denominations with rpc calls)

if possible, we would query the IBC base denomination, then, using the obtained token address, query Namada's denomination in storage, eventually defaulting to 0, like we do for non-usable erc20 tokens

core/src/ledger/ibc/mod.rs
```diff
@@ -255,3 +260,37 @@ pub struct ValidationParams {
...
+                    Error::Denom(format!("Trace path is invalid: error {e}"))
+                })?;
+        if is_receiver_chain_source(
+            src_port_id.clone(),
+            src_channel_id.clone(),
+            &ibc_denom,
+        ) {
+            let prefix =
+                TracePrefix::new(src_port_id.clone(), src_channel_id.clone());
+            ibc_denom.remove_trace_prefix(&prefix);
```
sug0: ah so this is how we "pop off" traces from the path, to eventually redeem native tokens!

core/src/types/address.rs
```diff
@@ -254,6 +259,9 @@ impl Address {
                 debug_assert_eq!(string.len(), FIXED_LEN_STRING_BYTES);
                 string
             }
+            Address::Foreign(addr) => {
+                format!("{}::{}", PREFIX_FOREIGN, addr)
+            }
```
sug0, line 264: I think this is buggy, as we require these strings to be of a fixed length (45 bytes iirc?). bech32 decoding will fail for foreign addresses, unless we change the decoding logic

core/src/types/address.rs
```diff
@@ -134,6 +136,8 @@ pub enum Address {
     Implicit(ImplicitAddress),
     /// An internal address represents a module with a native VP
     Internal(InternalAddress),
+    /// An foreign address is provided from other chains
+    Foreign(String),
```
sug0, line 140: btw, since we're storing foreign addresses on-chain, it might also be possible to store denominations for them, if they are a token address. this would allow you to specify a denominated amount when calling `ibc-transfer`

shared/src/ledger/protocol/mod.rs
```diff
@@ -1038,6 +1038,7 @@ where
 
                     accepted
                 }
+                Address::Foreign(_) => Ok(true),
```
sug0, line 1041: wasm can make any changes to foreign addresses?

shared/src/ledger/tx.rs
```diff
@@ -1916,6 +1927,102 @@ pub async fn build_custom<
...
+                     channel {dest_channel_id}"
+                )))
+            }
+        }
+        _ => Err(Error::Other(format!(
+            "Reading channel end failed: port {dest_port_id}, channel \
+             {dest_channel_id}"
+        ))),
+    }
+}
```
sug0, line 2024: doesn't read very well. combining `Result` and `Option` methods functions better for error handling IMO

sug0 (changes requested):

apps/src/lib/client/tx.rs
yito88: This is requested on the destination Namada. The memo is used by the source Namada.
Though a single CLI could request another Namada chain, for a non-Namada source chain, the separate command could be useful, I think.

benches/lib.rs
yito88: I guess that this is just for a bench.

yito88:
`Address::Foreign` wraps the sender/receiver address on the source/destination chain to set them in `token::Transfer` or MASP tx generation. Basically, Namada doesn't know the addresses.
Anyway, as you mentioned, I have to refine the handling.

shared/src/ledger/vp_host_fns.rs
yito88: I tried but I found that multiple "message" events could be issued.

yito88:
I rebased this branch based on #1951.
And, the sender address for a shielded transfer isn't required. So, I remove `Address::Foreign` and use `InternalAddress::Ibc`.

shared/src/ledger/vp_host_fns.rs
sug0: hm, but currently we're short-circuiting on the first `"message"` event we find, anyway. perhaps we should return a vector of events instead?

shared/src/sdk/tx.rs
```diff
@@ -2002,6 +2008,122 @@ pub async fn build_custom<
...
+        None => return Ok(None),
+    };
+    let source = Address::Internal(InternalAddress::Ibc);
+    let (src_port_id, src_channel_id) =
+        get_ibc_src_port_channel(client, &args.port_id, &args.channel_id)
+            .await?;
+    let ibc_denom =
+        rpc::query_ibc_denom::<_, IO>(client, &args.token, Some(&source)).await;
+    let token = namada_core::ledger::ibc::received_ibc_token(
+        &ibc_denom.parse().expect("Invalid IBC denom"),
```
sug0, line 2033: we should get rid of this `expect`

shared/src/sdk/tx.rs
```diff
@@ -2002,6 +2008,122 @@ pub async fn build_custom<
...
+        &args.port_id,
+        &args.channel_id,
+    )
+    .map_err(|e| {
+        Error::Other(format!("Getting IBC Token failed: error {e}"))
+    })?;
+    let validated_amount =
+        validate_amount::<_, IO>(client, args.amount, &token, false)
+            .await
+            .expect("expected to validate amount");
```
sug0, line 2045: same deal

sug0 (commented):
lgtm. some tinies

sug0 (approved):

tzemanovic (approved):

PR #1943, opened 2023-09-26
Refactor `Epoched` storage objects

brentstone:
## Describe your changes
Allow for more dynamic bookkeeping of the number of past epochs kept in an `Epoched` and `EpochedDelta` storage object. Previously, this number was determined by a const `u64`, but now it can be dependent on proof-of-stake parameters.

v0.23.0


tzemanovic:
We affected the tests

```
     protocol::transactions::votes::update::tests::test_apply_duplicate_votes
    protocol::transactions::votes::update::tests::test_calculate_already_seen
```

that now need to have params initialized for epoched data access

brentstone:
pls update wasms

tzemanovic (approved):

proof_of_stake/src/types.rs
```diff
@@ -25,82 +24,43 @@ pub use rev_order::ReverseOrdTokenAmount;
...
     common::PublicKey,
     crate::epoched::OffsetPipelineLen,
+    crate::epoched::OffsetDefaultNumPastEpochs,
 >;
 
 /// Epoched validator's eth hot key.
 pub type ValidatorEthHotKeys = crate::epoched::Epoched<
     common::PublicKey,
     crate::epoched::OffsetPipelineLen,
+    crate::epoched::OffsetDefaultNumPastEpochs,
```
tzemanovic, line 48: @sug0 is the default (2 epochs in the past) sufficient for the eth keys?

sug0: no, we need them around for longer. same as the other consensus data whose lifetime we're extending

brentstone: Ah ok, then @sug0 can you specifically note all the data that you need longer similar to this? And pls include if the max proposal period is the proper length or if smth else is.

brentstone: Also, after we sort what further changes need to be made, let's make the changes in #1944

sug0: here is a diagram of the PoS data required by the ethereum bridge:
image

with some validator's address, we are able to query their ethereum keys. the voting power associated with that validator should also be queriable, as well as the total (consensus) voting power of some epoch. the values should be stored for at least "max proposal period". however, IMO, we should add at least another 2 epochs to the lifetime of the data to guarantee that it can be queried and relayed to ethereum without any problems.

the reasoning behind this is that a proposal to enable the ethereum bridge may last up to "max proposal period". immediately following this period, the bridge must be brought up to date with validator sets from past epochs in Namada. technically, this can happen concurrently, while the proposal is still taking place, but it might never reach a full decision, which effectively means that funds spent relaying validator sets were wasted

sug0: we also require some validator's protocol key, but I think that is not within the domain of PoS (but maybe it should be?)

tzemanovic: Yeah, the protocol keys are outside of PoS - they're written to `key::protocol_pk_key` which isn't epoched and so doesn't preserve a previous value when it changes. Is that something that we need for the bridge or is it sufficient to only keep the latest?

sug0: protocol keys are required to validate votes on protocol transactions. here it's mostly a question of how much delay (expressed in epochs) a tx is expected to experience, from the moment it is crafted and signed, to the moment it is broadcasted to the network, gossiped, added to some validator's mempool and eventually proposed and decided. during `ProcessProposal`, protocol keys need to be available in storage to validate the tx in question

tzemanovic: I see - I think in the case a protocol key is changed by a validator, the tx to change it will be applied before the protocol txs that are in the queue (`prepare_proposal` adds protocol txs after decrypted txs) so it would invalidate any protocol txs signed with the old key. Depending on the timing this could be 1 (the tx queue) or 2 (mempool + queue) blocks of transactions (provided the queue is not too full). We could switch it to the PoS epoched type if we want to avoid this

sug0: yeah, it should definitely account for this

brentstone: Thx for this discussion, the eth key changes have been made in #1944. It looks like Tomas opened #1961 to address the protocol keys. Let's move further discussion to #1944, as that is the relevant topic (this PR just refactored bookkeeping mechanism for the epoched data)

PR #1951, opened 2023-09-27
Query for IBC tokens

### LINKED ISSUE #1946: Query balances for IBC tokens
To find a received token over IBC, we have to query balances with `balance` command without `owner`.
Also, they show the raw addresses.
It isn't easy to find the IBC token if you have lots IBC tokens.

yito88:
## Describe your changes
closes #1946 
- Transfer (`transfer` and `ibc-transfer`) with a token prefixed with IBC `trace-path` for IBC token instead of the raw address.
    ```
    $ namadac transfer \
        --amount 100 \
        --source albert \
        --target bertha \
        --token transfer/channel-0/nam \
        --signing-keys albert-key
    ```
- Balance query checks also IBC tokens and shows the alias instead of the raw addresses
    ```
    $ namadac balance --token nam --owner albert
    nam: 1000000
    transfer/channel-0/nam: 50000

    $ namadac balance --owner albert
    apfel: 1000000
    btc: 1000000
    dot: 1000000
    eth: 1000000
    kartoffel: 1000000
    nam: 1000000
    schnitzel: 1000000
    transfer/channel-0/nam: 50000
    ```

v0.23.0

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -2516,6 +2516,7 @@ pub mod args {
     use std::path::PathBuf;
     use std::str::FromStr;
 
+    use namada::ibc::applications::transfer::TracePath;
     use namada::ibc::core::ics24_host::identifier::{ChannelId, PortId};
     pub use namada::sdk::args::*;
     use namada::types::address::Address;
@@ -2725,6 +2726,7 @@ pub mod args {
     pub const TM_ADDRESS: Arg<String> = arg("tm-address");
     pub const TOKEN_OPT: ArgOpt<WalletAddress> = TOKEN.opt();
     pub const TOKEN: Arg<WalletAddress> = arg("token");
+    pub const TRACE_PATH: ArgOpt<TracePath> = arg_opt("trace-path");
     pub const TRANSFER_SOURCE: Arg<WalletTransferSource> = arg("source");
     pub const TRANSFER_TARGET: Arg<WalletTransferTarget> = arg("target");
     pub const TX_HASH: Arg<String> = arg("tx-hash");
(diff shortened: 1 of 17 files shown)


apps/src/lib/cli.rs
```diff
@@ -3517,13 +3520,15 @@ pub mod args {
             let source = TRANSFER_SOURCE.parse(matches);
             let target = TRANSFER_TARGET.parse(matches);
             let token = TOKEN.parse(matches);
+            let trace_path = TRACE_PATH.parse(matches);
```
sug0, line 3523: adding a `trace_path` argument to `TxTransfer` kind of leaks ibc logic into something that should be generic. is it possible to implement this purely CLI-side, like we have for the Ethereum bridge, when parsing Ethereum addresses? this doesn't bode too well with the fact that you can specify a `token` with an IBC trace path already, then add another trace path on top of it, ending up with something like: `port/chan/token/port/chan/token`.

if you feel like that would be too disruptive to change into, we can leave it as is

apps/src/lib/client/rpc.rs
```diff
@@ -348,35 +359,47 @@ pub async fn query_transparent_balance<
...
+                {
+                    Ok(balance) => {
+                        let balance = format_denominated_amount::<_, IO>(
+                            client, &token, balance,
+                        )
+                        .await;
+                        display_line!(IO, "{}: {}", token_alias, balance);
+                    }
+                    Err(e) => {
+                        display_line!(IO, "Eror in querying: {e}");
```
sug0, line 388: nit
```suggestion
                        display_line!(IO, "Querying error: {e}");
```

apps/src/lib/cli.rs
yito88: Thank you! That also works for non-Namada tokens.

sug0 (approved):

PR #1954, opened 2023-09-28
Moves signatures verification gas in `Signature`

grarco:
## Describe your changes

Moves the gas accounting for the signature verification in the `verify_signature` method of `Signature` for higher resolution (i.e. being able to short circuit early in case of an out-of-gas event).

v0.23.0


murisi (approved):
Looks good to me. Thanks!

PR #1956, opened 2023-09-28
Add `ZeroizeOnDrop` to ed25519 keys

sug0:
## Describe your changes

- Tag `ed25519` keys with `ZeroizeOnDrop`, to indicate their bytes are cleared uppon the key being dropped.
- ~~Switch to `k256` for `secp256k1` keys. It natively implements `Zeroize`~~ - done in base #1958

#1958 (first commit here 609e70d62ce56842633b85aab5958d0324122416)

### DIFF
--- a/.changelog/unreleased/miscellaneous/1958-k256.md
+++ b/.changelog/unreleased/miscellaneous/1958-k256.md
@@ -1,0 +1,2 @@
+- Switched from using `libsecp256k1` to `k256` crate.
+  ([\#1958](https://github.com/anoma/namada/pull/1958))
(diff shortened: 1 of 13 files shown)


brentstone:
DId these keys not already have this behavior? See the tests `types::key::more_tests::zeroize_keypair_ed25519` and `types::key::more_tests_zeroize_keypair_secp256k1`?

sug0:
oh @brentstone you're right. I didn't realize we were using a fork of libsecp that cleared these keys on drop. to be honest though, we should switch back to the upstream version of the crate. we can clear the sk in a way similar to what has been implemented in this pr (albeit it involves unsafe code). lmk what you think

brentstone:
@sug0 yeah you're prob right that we should avoid using the fork if we can accomplish the same thing with your method. I defer to your judgment.

tzemanovic:
looking back at this I think we should switch to use https://crates.io/crates/k256

Issue #50, opened 2022-01-17
atomic wallet writes


tzemanovic:

cwgoes:
@karbyshev can you take this on as part of the wallet work?

cwgoes:
Until a real user has a real problem, we are not doing this.

opsecx:
> Until a real user has a real problem, we are not doing this.

Hi :)

https://github.com/anoma/namada/issues/4633

(I don't blame y'all if this isn't a prio but for the sake of completeness)

PR #1963, opened 2023-10-03
Murisi/sdk refactor rebased

murisi:
## Describe your changes
Created an SDK interface in order to streamline common operations like building, signing, and submitting constructions. Also modified the codebase to use this new interface as a means of testing its usability. More specifically, the following changes have been made:
* Provided standard implementations of "WalletUtils" (`WalletIo + WalletStorage`) and `ShieldedUtils` so that users do not have to implement them themselves
* Alleviated the user from having to specify a client, shielded context, wallet context, and input/output stream for each transaction by bundling these into one object
* Now supply default field values in the builders for each of the transaction types in order to reduce the verbosity of making transactions
* Integrated the saving and loading of wallets into the "WalletUtils" trait so that these operations can be done without knowing the underlying wallet storage
* Combined the construction of a transaction and its corresponding singing data in order to reduce the number of calls an SDK user to make from transaction construction through to submission
  * Doing this also reduced the number of parameters SDK users have to supply to transaction building functions because they now internally compute the signing data
* Made the SDK interface work for parallel use cases by using locks for the wallet and the shielded context. Formerly, the user had to inefficiently construct new wallets and shielded contexts from file storage for each thread.
* Separated SDK functionality (everything required to construct, sign, and send transactions; perform queries) out of the `shared` crate and into a new crate named `sdk`. Hence the dependency graph is now more like `shared -> sdk -> core`.

See an example of this PR's usage here: https://github.com/anoma/namada-sdk-starter/pull/1 .

v0.23.0


sug0 (dismissed):
these changes seem generally good to me. the only thing(s) I'm worried about are the conflicts with the chain of PRs whose tip is #1957 😅

apps/src/bin/namada-relayer/main.rs
```diff
@@ -14,5 +14,6 @@ async fn main() -> Result<()> {
 
     let cmd = cli::namada_relayer_cli()?;
     // run the CLI
-    CliApi::<CliIo>::handle_relayer_command::<HttpClient>(None, cmd).await
+    CliApi::<CliIo>::handle_relayer_command::<HttpClient>(None, cmd, &CliIo)
```
batconjurer, line 17: It may be way more work than is worth it right now, but does this mean we can get rid of the explicit generic param?

murisi: Yes, that should be the case. Let me try removing the explicit generic parameter now...

shared/src/ledger/mod.rs
```diff
@@ -14,6 +14,521 @@ pub mod queries;
...
+use crate::types::io::Io;
+use crate::types::key::*;
+use crate::types::masp::{TransferSource, TransferTarget};
+use crate::types::token;
+use crate::types::token::NATIVE_MAX_DECIMAL_PLACES;
+use crate::types::transaction::GasLimit;
+
+#[async_trait::async_trait(?Send)]
+/// An interface for high-level interaction with the Namada SDK
+pub trait Namada<'a>: Sized {
```
batconjurer, line 51: nit: I'm not entirely happy with the name because of how it appears in the apps crate. It's amost always assigned to something called `context`.  I feel like it should be `NamadaCtx` or `NamadaApi` or `NamadaUtils` or something.

shared/src/ledger/mod.rs
```diff
@@ -14,6 +14,521 @@ pub mod queries;
...
+        &self,
+        tx: Tx,
+        args: &args::Tx,
+    ) -> crate::sdk::error::Result<ProcessTxResponse> {
+        tx::process_tx(self, args, tx).await
+    }
+}
+
+/// Provides convenience methods for common Namada interactions
+pub struct NamadaImpl<'a, C, U, V, I>
```
batconjurer, line 393: And I also would change this name too.

shared/src/ledger/mod.rs
```diff
@@ -14,6 +14,521 @@ pub mod queries;
...
+    /// Stores the current state of the shielded pool
+    pub shielded: RwLock<&'a mut ShieldedContext<V>>,
+    /// Captures the input/output streams used by this object
+    pub io: &'a I,
+    /// The default builder for a Tx
+    prototype: args::Tx,
+}
+
+/// The Namada token
+pub const NAM: &str = "atest1v4ehgw36x3prswzxggunzv6pxqmnvdj9xvcyzvpsggeyvs3cg9qnywf589qnwvfsg5erg3fkl09rg5";
```
batconjurer, line 413: Hard coded addresses give me nightmares since I've been working on the new genesis flow :sweat_smile: . Is there any way this can be looked up instead?

batconjurer: I mean eventually it will have to be looked up, it will just fall on me.

shared/src/sdk/signing.rs
```diff
@@ -364,14 +365,10 @@ pub async fn wrap_tx<
...
         Some(amount) => {
-            let validated_fee_amount = validate_amount::<_, IO>(
-                client,
-                amount,
-                &args.fee_token,
-                args.force,
-            )
-            .await
-            .expect("Expected to be able to validate fee");
+            let validated_fee_amount =
```
batconjurer, line 368: I just remember, are we allowed to have panics in the sdk still?

shared/src/ledger/mod.rs
murisi: Nightmares are not good. Formerly I used to look it up in the wallet - but the SDK is frequently used in contexts where the wallet is empty (see https://github.com/anoma/namada-sdk-starter/pull/1 and also the benchmarks). Would it be valid to have an RPC endpoint that returns the native token address?

shared/src/sdk/signing.rs
murisi: Thanks. You're correct: panics are not allowed in the SDK. Since SDK integration is urgent - perhaps we can make a future PR where we identify the remaining occurrences of panics and fix them?

shared/src/ledger/mod.rs
batconjurer: That works for me.

shared/src/ledger/mod.rs
murisi: The name `context` is a bit of a hangover from how we were naming instances of `Context` - this too is changeable. This trait is intended to be the primary interface for SDK users (who won't be aware of its `apps` crate usage and our naming distinctions), so I was thinking more about the best look/outcome in this context: https://github.com/anoma/namada-sdk-starter/blob/murisi/updating-sdk-usage/src/main.rs. Also, SDK users (and we) will probably be typing `&impl Namada<'a>` a lot, so I'm keen to minimize the name's length. That being said, I'm not too attached to any particular name, and we can always export/import `as`.

shared/src/ledger/mod.rs
```diff
@@ -14,6 +14,531 @@ pub mod queries;
...
+where
+    C: crate::ledger::queries::Client + Sync,
+    U: WalletIo,
+    V: ShieldedUtils,
+    I: Io,
+{
+    /// Used to send and receive messages from the ledger
+    pub client: &'a C,
+    /// Stores the addresses and keys required for ledger interactions
+    pub wallet: RwLock<&'a mut Wallet<U>>,
```
tzemanovic, line 395: Unfortunately, I don't think the lock on the wallet makes it parallelizable as one task might overwrite the changes of another task because in-memory changes are not synced to fs without an explicit `save()` call. We have an issue to fix this (#50 - the wallet should lock internally during any mutable operations on it and only release the lock after implicitly saving to fs with as small scope as possible to avoid contention), but I'd like to avoid the illusion of this being thread-safe before that's done as it can lead to loss of keys. I'm not quite sure what to do with this until then

murisi: Agreed, this PR buys us asynchrony within a Namada application at the cost of safety for multiple running Namada applications. (We might use it in this parallel application for instance: https://github.com/anoma/namada-sdk-starter/blob/murisi/updating-sdk-usage/src/main.rs .) Since this PR does fix SDK and API issues and the locking issue is not so trivial, could we perhaps merge this PR and then follow up with a fix for wallet safety?

One quick way to address the key loss issue may be to leave the API as-is and modify the `save` method of the wallet and shielded context to: 1) obtain lock on file 2) load the file into memory 3) merge the new changes into that file 4) save the merged file 5) release the lock. See https://github.com/anoma/namada/blob/3f979bf3b2707db4d065a4fae6e8c2340ae1d54b/shared/src/sdk/masp.rs#L633 . Then we will have more time to implement https://github.com/anoma/namada/issues/50 while knowing that keys are not being lost.

tzemanovic: That's a great idea! I think we should go with that

sug0 (dismissed):
I had some PRs open that touched the sdk. I need your help fixing them later @murisi haha

grarco (approved):
Only reviewed from 17c3e6cdb767ed907e8519b75ab6eff536e86288 onward but looks good to me

PR #1984, opened 2023-10-13
core/lazy_map+set: avoid calling delete when given key is not present

### LINKED ISSUE #1981: PoS bond tx touchs more validator delta keys than expected
A transaction such as tx_bond touches more storage keys than expected.

An example of this from testnet https://pastebin.com/Cch73t7d touched all the validator deltas since epoch 0 to the pipeline offset, which also made it exceed the default gas limit.

tzemanovic:
## Describe your changes

closes #1981 

0.23.0


brentstone (approved):
Awesome, I checked the behavior locally in a dummy e2e test env and it works. Thx!

PR #1916, opened 2023-09-21
masp inflation 0.23

juped:
## Describe your changes
0.23-draft masp inflation for review, not merging. changes not properly separated. tests broken

## Checklist before merging to `draft`
- [nope] I have added a changelog
- [nope] Git history is in acceptable state

PR #1985, opened 2023-10-13
Implemented compounding inflation for native tokens.

murisi:
## Describe your changes
Implemented compounding inflation for native tokens and now use a PD-controller to determine inflation values. More specifically, the changes are as follows:
* Implemented compounding inflation for native tokens so that users can now get correspondingly denominated rewards for holding NAM tokens.
* Generalized the PD-controller to allow the token used for computing the shielded pool inflation cap to be different from the one used to compute the locked token amount. And then implemented inflation based on this PD-controller's outputs.
* Updated the `masp_incentives` integration test so that it no longer assumes fixed reward rates, but instead has PD-controller outputs hardcoded into it.
* Increased the precision of inflation computations for each asset so that high denomination tokens like ETH will be able to receive non-zero NAM rewards. This involved representing reward components as `i128`s.

This PR is based on:
* https://specs.namada.net/economics/inflation-system
* https://specs.namada.net/economics/shielded-pool-incentives
* https://github.com/anoma/namada/pull/1916
* https://github.com/anoma/namada/pull/1025

v0.23.0


core/src/ledger/storage/masp_conversions.rs
```diff
@@ -27,6 +39,143 @@ pub struct ConversionState {
...
+    // Since we must put the notes in a compatible format with the
+    // note format, we must make the inflation amount discrete.
+    let noterized_inflation = if total_token_in_masp.is_zero() {
+        0u32
+    } else {
+        crate::types::uint::Uint::try_into(
+            (inflation.raw_amount() * PRECISION)
+                / total_token_in_masp.raw_amount(),
+        )
+        .unwrap()
```
mariari, line 129: Is this any different? Otherwise I don't see much about this which is different

mariari (commented):
I can do a more thorough review later, but looking now, nothing looks too out of the ordinary,  I would like to see more what was changed, I should diff this against the mess of a current branch @juped and I worked on.

Was the solution just running some of the calculations only once instead of 4 times?

core/src/types/token.rs
```diff
@@ -814,6 +829,31 @@ pub const TX_KEY_PREFIX: &str = "tx-";
 pub const CONVERSION_KEY_PREFIX: &str = "conv";
 /// Key segment prefix for pinned shielded transactions
 pub const PIN_KEY_PREFIX: &str = "pin-";
+/// Last calculated inflation value handed out
+pub const MASP_LAST_INFLATION: &str = "last_inflation";
+/// The last locked ratio
+pub const MASP_LAST_LOCKED_RATIO: &str = "last_locked_ratio";
+/// The key for the nominal proportional gain of a shielded pool for a given
+/// asset
+pub const MASP_KP_GAIN_KEY: &str = "proptional_gain";
```
brentstone, line 838: `"proportional_gain"`

core/src/types/token.rs
```diff
@@ -847,6 +887,98 @@ pub fn minted_balance_key(token_addr: &Address) -> Key {
         .expect("Cannot obtain a storage key")
 }
 
+/// Obtain the nominal proportional key for the given token
+pub fn masp_kp_gain(token_addr: &Address) -> Key {
```
brentstone, line 891: Just a nit, but can this and the related functions below that yield a `Key` have "key" appended to their name, i.e. `masp_kp_gain_key`?

core/src/types/token.rs
```diff
@@ -847,6 +887,98 @@ pub fn minted_balance_key(token_addr: &Address) -> Key {
...
+)]
+pub struct Parameters {
+    /// Maximum reward rate
+    pub max_reward_rate: Dec,
+    /// Shielded Pool nominal derivative gain
+    pub kd_gain_nom: Dec,
+    /// Shielded Pool nominal proportional gain for the given token
+    pub kp_gain_nom: Dec,
+    /// Locked ratio for the given token
+    pub locked_ratio_target_key: Dec,
```
brentstone, line 937: remove "_key" from `locked_ratio_target_key`

core/src/types/token.rs
```diff
@@ -914,6 +1046,24 @@ pub fn is_masp_key(key: &Key) -> bool {
                     || key.starts_with(PIN_KEY_PREFIX)))
 }
 
+/// The last locked ratio of a token
+pub fn masp_last_locked_ratio(token_address: &Address) -> Key {
```
brentstone, line 1050: Same as above regarding adding "_key" to the name. Also maybe change the docstring too

brentstone:
In `Shell::apply_inflation` of `finalize_block.rs`, there is some dummy MASP inflation code that should be removed, since this now effectively occurs in `update_allowed_conversions`

batconjurer (changes requested):

murisi:
> I can do a more thorough review later, but looking now, nothing looks too out of the ordinary, I would like to see more what was changed, I should diff this against the mess of a current branch @juped and I worked on.
> 
> Was the solution just running some of the calculations only once instead of 4 times?

Hi! I assume you want this PR to be differentiated from `ray/0.23-rewards`? I took most of what you and @juped had implemented and mostly just altered the arithmetic in `update_allowed_conversions` and upgraded the `masp_incentives` integration test. The main changes in this PR:
* The reward asset for a token of `MaspDenom` i (=0,1,2,3) also has a reward of `MaspDenom` i. Doing this should allow rewards to be proportional to owned shielded tokens regardless of their denomination.
* The reference inflation is set to a larger value in order to allow a greater variety/set of fractions of it to be distributed as rewards. More concretely, if `ref_inflation == 1`, then initially `normed_inflation = 1`, which means that a `new_normed_inflation` equal to 1, 2, or 3 represent a limited choice between a 0%, 100%, or 200% inflation. Whereas if `ref_inflation` were to equal 100, then initially `ref_inflation == 100`, which means that a `new_normed_inflation` of say 163 represents a 63% inflation. A higher `ref_inflation` allows for more accurate inflation distribution.
* In the case of computing rewards for Namada tokens:
    * Since `reward.0 / reward.1` conceptually represents a percentage increase in money supply, this PR gives `x*(1+(reward.0 / reward.1))` current-epoch-NAM tokens for every `x` previous-epoch-NAM tokens. If we only dished out `normed_inflation * ((reward.0 / normed_inflation) / reward.1)`, or (almost) equivalently `reward.0 / reward.1` tokens, then it seems that users would lose their principal shielded amount in the process of receiving their rewards for it.
    * The `total_reward` is now increased by `(addr_bal / normed_inflation) * new_normed_inflation` in order to support fractional inflation. Were `normed_inflation` to be assumed to bee `I256::from(1)`, we would lose accuracy in the transparent inflation being able to choose only multiples of 100%.
* In `calculate_masp_rewards`, the PD-controller now computes the shielded pool inflation cap using the current supply of NAM instead of the current supply of `addr`. This seems to be closer to what's in https://specs.namada.net/economics/inflation-system, unless if that document is now outdated.

core/src/ledger/storage/masp_conversions.rs
```diff
@@ -90,15 +104,80 @@ where
...
+                            reward_assets[denom as usize],
+                            real_reward as i32,
+                        )
                         .unwrap())
-                .into(),
-            );
+                    .into(),
+                );
+                // Operations that happen exactly once for each token
+                if denom == MaspDenom::Three {
```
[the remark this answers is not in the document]
murisi: I don't think that the body of this conditional was called 4 times before. This `denom == MaspDenom::Three` check became necessary to stop repetitions once I had moved the logic occurring before the `for denom in token::MaspDenom::iter() {` loop into the `for denom in token::MaspDenom::iter() {` loop. I made this movement as a matter of convenience, but I'm sure that there are other ways this logic could be structured.

I chose `denom::Three` because the conditional's body modifies variables (namely `total_reward` and `normed_inflation`) that would affect the next iteration of `for denom in token::MaspDenom::iter() {`. In the current context, the `normed_inflation` can only be modified once we have finished working with the native token.

core/src/ledger/storage/masp_conversions.rs
murisi: Other than use of unsigned amounts, this code is probably the same as what's in the other branches/PRs.

apps/src/lib/config/genesis.rs
```diff
@@ -1001,8 +1012,8 @@ pub fn genesis(num_validators: u64) -> Genesis {
         implicit_vp_code_path: vp_implicit_path.into(),
         implicit_vp_sha256: Default::default(),
         max_signatures_per_transaction: 15,
-        epochs_per_year: 525_600, /* seconds in yr (60*60*24*365) div seconds
-                                   * per epoch (60 = min_duration) */
+        epochs_per_year: 365, /* seconds in yr (60*60*24*365) div seconds
```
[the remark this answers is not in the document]
murisi: Ah, thanks for pointing this out. I need `epochs_per_year` to equal a small number for the integration tests in order to be able to observe non-zero inflation. This value is directly queried by `calculate_masp_rewards` and used as a quotient in some computations.

For the integration tests, this doesn't seem to be an issue since we manually drive the epochs forward using `MockNode::next_epoch`. And for the E2E tests, this doesn't seem to be an issue since they override the default genesis with `e2e-tests-single-node.toml`. Is there an area where this becomes an issue? Or do you know a better way to achieve the above goal of having observable inflation?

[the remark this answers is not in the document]
batconjurer: I suppose not. I guess in basically all cases, we can override if we need to.

core/src/ledger/inflation.rs
```diff
@@ -67,10 +69,12 @@ impl RewardsController {
             .expect("Should not fail to convert token Amount to Dec");
         let total = Dec::try_from(total_tokens.raw_amount())
             .expect("Should not fail to convert token Amount to Dec");
+        let total_native = Dec::try_from(total_native_tokens.raw_amount())
```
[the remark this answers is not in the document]
murisi: Agreed, I hadn't really thought about the correctness of the PD-controller code but instead followed its style and format. That being said, the `RewardsController::run` method and specs seem to be linear in the `locked_tokens`, `total_tokens`, `total_native_tokens`, and `last_inflation_amount` variables. So the uses of incorrect scales seem to cancel out in the result of the computations. This seems to be borne out when trying the formula with specific numbers (like in the `masp_incentives` integration tests). But yeah, we should fix this as it still represents an arithmetic overflow risk.

core/src/ledger/storage/masp_conversions.rs
```diff
@@ -163,24 +162,21 @@ where
...
+                0,
+            )
+            .unwrap(),
         )
         .expect("unable to encode new inflation rate (Decimal)");
 
     wl_storage
-        .write(&token::masp_last_locked_ratio(addr), locked_ratio)
+        .write(&token::masp_last_locked_ratio_key(addr), locked_ratio)
         .expect("unable to encode new locked ratio (Decimal)");
```
Fraccaman, line 177: is this safe to unwrap?

core/src/ledger/storage/masp_conversions.rs
```diff
@@ -163,24 +162,21 @@ where
...
         .write(
-            &token::masp_last_inflation(addr),
-            (total_token_in_masp / PRECISION) * u64::from(noterized_inflation),
+            &token::masp_last_inflation_key(addr),
+            token::Amount::from_uint(
+                (total_token_in_masp.raw_amount() / precision)
+                    * crate::types::uint::Uint::from(noterized_inflation),
+                0,
+            )
+            .unwrap(),
```
Fraccaman, line 171: is this safe to unwrap? same for the one above

PR #1991, opened 2023-10-16
Prevent crash on failed pgf payment

grarco:
## Describe your changes

Fixes a bug for which the ledger crashed in case of a failed PGF payment.

Commit 3f979bf3b2707db4d065a4fae6e8c2340ae1d54b from `base` branch


Fraccaman:
should we also do the same for cpgf? or does it already have the same behavior?

grarco:
I think that the cPGF payment done in `finalize_block` is fine as it is. We are only propagating the error when crediting the tokens for the PGF account itself, which I think is fine since we can't really recover from that. For cpgf and stewards crediting we are instead already managing any possible errors

https://github.com/anoma/namada/blob/3f979bf3b2707db4d065a4fae6e8c2340ae1d54b/apps/src/lib/node/ledger/shell/finalize_block.rs#L839-L917

Fraccaman (approved):

PR #1993, opened 2023-10-17
fix the start block height of the first epoch

### LINKED ISSUE #648: Mismatch in initial block heights at genesis
The initial block height of the chain according to `request::InitChain` from tower ABCI is 1 ([see code here](https://github.com/anoma/namada/blob/546a043f023e329b815d7cb6dfa49d33425a02a9/apps/src/lib/node/ledger/shell/init_chain.rs#L52-L55)), which does not match the initial block height of the chain in the namada protocol, which is 0.

The initial block height of 1 from tower ABCI is used to determine when the next epoch begins [right here](https://github.com/anoma/namada/blob/546a043f023e329b815d7cb6dfa49d33425a02a9/apps/src/lib/node/ledger/shell/init_chain.rs#L64-L69), so possibly epoch 1 (the second epoch) starts a block later than intended.

tzemanovic:
## Describe your changes

closes #648

0.23


core/src/types/storage.rs
```diff
@@ -267,6 +273,17 @@ impl TryFrom<i64> for BlockHeight {
...
 }
 impl BlockHeight {
+    /// The first block height 1.
+    pub fn first() -> Self {
+        Self(1)
+    }
+
+    /// A sentinel value block height 0 may be used before any block is
+    /// committed or in queries to read from the latest committed block.
+    pub fn sentinel() -> Self {
```
sug0, line 283:
```suggestion
    pub const fn sentinel() -> Self {
```
nit

core/src/types/storage.rs
```diff
@@ -1161,13 +1178,12 @@ pub struct Epochs {
...
-
 impl Default for Epochs {
-    /// Initialize predecessor epochs, assuming starting on the epoch 0 and
-    /// block height 0.
+    /// Initialize predecessor epochs, assuming start of the first epoch at
+    /// block height 1.
     fn default() -> Self {
         Self {
-            first_block_heights: vec![BlockHeight::default()],
+            first_block_heights: vec![BlockHeight(1)],
```
sug0, line 1186: this is still somewhat incorrect, though, no? because it assumes height 1 has already been committed even before we commit the first block

tzemanovic: yeah, good point! I think we can just initialize it empty as it's later set from init chain

sug0: @tzemanovic if it doesn't break the protocol, sure, but I think it's still semantically wrong. when calling finalize-block for the first time, no block has been committed yet, so `Epochs` in theory would be empty

tzemanovic: I think it's ok - the first epoch is initialized on `InitChain` instead of `FinalizeBlock` as for many sub-systems it needs different logic than the epoch change in `FinalizeBlock`

core/src/types/storage.rs
tzemanovic: I also did this for `fn first`

core/src/types/storage.rs
tzemanovic: done in [6fb1984](https://github.com/anoma/namada/pull/1993/commits/6fb1984cbdc0522031c5e1c46e4c83b5b8f794d2)

yito88 (approved):

tzemanovic:
I squashed the fixup commits in the last f-push (no diff)

PR #2045, opened 2023-10-24
Validator metadata

### LINKED ISSUE #1609: PoS validator metadata
allow validators to optionally set metadata such as an alias, website, description

brentstone:
## Describe your changes
Closes #1609.

This PR implements the storage of validator metadata, such as an email, description, website, and discord_handle. An email is the only required piece of this data. A TX is implemented to change this metadata that also allows one to change a validator's commission rate at the same time.

v0.25.0


s0uk4:
Adding an email should be required, with a tx implemented to change this field if need be (it should not be immutable)

Ideally, this should allow us to select for everyone who is either active and without VP and those who are active and have VP if there is ever a need to coordinate.

apps/src/lib/config/genesis.rs
```diff
@@ -372,6 +372,7 @@ pub mod genesis_config {
                         "Max commission rate change must be between 0.0 and \
                          1.0",
                     ),
+                metadata: Default::default(),
```
tzemanovic, line 375: we should add the metadata to `ValidatorConfig` and pass them through here to allow to set these at genesis

brentstone: done in d356f824843c2a50089651d773240c0d2b7aa077

sdk/src/tx.rs
```diff
@@ -605,6 +607,138 @@ pub async fn build_validator_commission_change<'a>(
...
+
+    // If there is a new email, it cannot be an empty string that indicates to
+    // remove the data (email data cannot be removed)
+    if let Some(email) = email.as_ref() {
+        if email.is_empty() {
+            edisplay_line!(
+                context.io(),
+                "Cannot remove a validator's email, which was implied by the \
+                 string {}",
+                email
```
tzemanovic, line 651:
```suggestion
                "Cannot remove a validator's email, which was implied by the \
                 empty string"
```

proof_of_stake/src/types.rs
```diff
@@ -328,6 +328,45 @@ pub struct GenesisValidator {
...
+    /// Validator description
+    pub description: Option<String>,
+    /// Validator website
+    pub website: Option<String>,
+    /// Validator alias
+    pub alias: Option<String>,
+    /// Validator's discord handle
+    pub discord_handle: Option<String>,
+}
+
```
tzemanovic, line 359:
```suggestion

#[cfg(any(test, feature = "testing"))]
```
(to avoid using the default email somewhere in non-test code)

tzemanovic (approved):

apps/src/lib/config/genesis.rs
```diff
@@ -200,6 +201,16 @@ pub mod genesis_config {
         /// Tendermint node key is used to derive Tendermint node ID for node
         /// authentication
         pub tendermint_node_key: Option<HexString>,
+        /// Validator email (required)
+        pub email: String,
```
tzemanovic, line 205: @s0uk4 @Fraccaman because this is required, this will be a breaking change for testnet setup - you'll need to ask all validators to add this

tzemanovic (approved):

apps/src/lib/config/genesis.rs
adrianbrink: It can just be an empty string. We can't validate the email anyway during the submission.

adrianbrink: So based on that I think we can also make this an option too. Ie, someone can make their email `potatoe@potatoe@potatoe.com`, which is equivalent to None really.

s0uk4: even if this is true, that's the validator's problem, not ours. they need to put an email in for the case where the network needs them to respond to potential security issues. asking for an email field helps us, but it's also a favor to the valds themselves.

if they put in an empty string rather than the proper email and there is a breaking change or risk requiring validators to respond, and they don't see discord /or/ the email, they will go down/put the network at risk and be penalized, which means they will lose money. there is incentive for them not to defect here.

brentstone: It's unclear to me if @s0uk4 is convincing others here, so can we get some clear consensus on how the email should be done? Atm, it is required, and providing an empty string for this arg will return an error, since empty strings are used to indicate that one wants to remove metadata from storage (i.e. if I want to delete my validator website, I pass `--website ""` at the CLI).

tzemanovic (approved):

PR #2050, opened 2023-10-25
Reduce merkle tree snapshots

### LINKED ISSUE #2043: Store only essential Merkle tree snapshots
Currently, we store a lot of Merkle tree snapshots. In the current testnet, the size of a snapshot was over 2GB and the 60 snapshots took over 100GB.

The Merkle tree has 4 subtrees under the `Base` tree; `Account`, `PoS`, `Ibc`, and `BridgePool`. `Base` tree has roots of subtrees. Each subtree has hashed key-value pairs. We store `Base` tree snapshot at every height(block) for the root hash. 4 subtree snapshots are stored every epoch for the readable period (60 epochs in the testnet).
https://github.com/anoma/namada/blob/2138c968f07d22714a25a20b30ba8526b2afffd2/apps/src/lib/node/ledger/storage/rocksdb.rs#L912-L940

However, only IBC and EthBridge require the Merkle proof though the Merkle tree snapshot is used for proof generation. The `Account` and `PoS` snapshots are never used (except for node restart).

So, we need to store only `Ibc` and `BridgePool` tree snapshots for the period and store `Account` and `PoS` tree snapshots only for the last epoch for restarting.
The changes would be about:
- Check if the snapshots should be stored or not when each block commit
- Restore a Merkle tree
    - When all snapshots are loaded
    - When not all snapshots are loaded
        - Load the last snapshots and restore the tree by diffs

It would drastically reduce the storage size. The size would be from 100GB  to 4GB.

yito88:
## Describe your changes
closes #2043

`v0.25.0`


apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -775,6 +779,35 @@ impl DB for RocksDB {
...
+                    .map_err(|e| Error::DBError(e.into_string()))?
+                {
+                    merkle_tree_stores.set_root(
+                        st,
+                        types::decode(bytes).map_err(Error::CodingError)?,
+                    );
+                }
+                let store_key = key_prefix
+                    .push(&"store".to_owned())
+                    .map_err(Error::KeyError)?;
```
sug0, line 801:
```suggestion
                let store_key = key_prefix
                    .with_segment("store".to_owned());
```
this method doesn't need to error out

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -1033,35 +1065,22 @@ impl DB for RocksDB {
...
-        let tree_key = Key::from(stored_height.to_db_key())
-            .push(&"tree".to_owned())
-            .map_err(Error::KeyError)?;
         let mut merkle_tree_stores = MerkleTreeStoresRead::default();
-        for st in StoreType::iter() {
-            let prefix_key =
-                tree_key.push(&st.to_string()).map_err(Error::KeyError)?;
+        let store_types = store_type
+            .map(|st| vec![st])
+            .unwrap_or(StoreType::iter().copied().collect());
```
sug0, line 1077: [`Either`](https://docs.rs/itertools/latest/itertools/enum.Either.html#impl-Iterator-for-Either%3CL,+R%3E) from itertools implements `Iterator`, so you don't need to allocate here. not really a big deal I guess, more like a FYI

core/src/ledger/storage/merkle_tree.rs
```diff
@@ -259,6 +298,26 @@ impl fmt::Display for StoreType {
...
 }
 
+/// Get the key prefix with which the base root and store are stored in the
+/// storage
+pub fn base_tree_key_prefix(height: BlockHeight) -> Key {
+    Key::from(height.to_db_key())
+        .push(&"tree".to_owned())
+        .expect("Creating a key shouldn't fail")
+        .push(&StoreType::Base.to_string())
+        .expect("Creating a key shouldn't fail")
```
sug0, line 308: `with_segment`, instead of `push` here?

core/src/ledger/storage/merkle_tree.rs
```diff
@@ -259,6 +298,26 @@ impl fmt::Display for StoreType {
...
+
+/// Get the key prefix with which the subtree root and store are stored in the
+/// storage
+pub fn subtree_key_prefix(st: &StoreType, epoch: Epoch) -> Key {
+    Key::from(epoch.to_db_key())
+        .push(&"tree".to_owned())
+        .expect("Creating a key shouldn't fail")
+        .push(&st.to_string())
+        .expect("Creating a key shouldn't fail")
+}
```
sug0, line 319: same

core/src/ledger/storage/mod.rs
```diff
@@ -1086,19 +1142,23 @@ where
...
+                }
+                // Remove stores at the previous epoch because the Merkle tree
+                // stores at the starting height of the epoch would be used to
+                // restore stores at a height (> min_height) in the epoch
+                for st in StoreType::iter_provable() {
+                    self.db.prune_merkle_tree_store(batch, st, epoch.prev())?;
                 }
             }
         }
+        // remove non-provable stores at the previous epoch
```
sug0, line 1154: we can have an extra condition in `prune_merkle_tree_stores` to remove bridge pool stores with nonces lower than the latest bridge pool nonce signed by a quorum of validators. for more context, see the comment at: https://github.com/anoma/namada/issues/2043#issuecomment-1780567749

something like (pseudo code):
```rs
// remove bridge pool stores with outdated nonces
let signed_nonce = get_bridge_pool_signed_nonce();
let trees = get_stored_bridge_pool_trees();

for t in trees.filter(|t| t.nonce < signed_nonce) {
    /* remove t from storage */
}
```

ethereum_bridge/src/test_utils.rs
```diff
@@ -191,8 +189,9 @@ pub fn init_storage_with_validators(
     // set last height to a reasonable value;
     // it should allow vote extensions to be cast
     if let Some(b) = wl_storage.storage.last_block.as_mut() {
-        b.height = 3.into();
+        b.height = 0.into();
```
sug0, line 192: vote extension validation fails when `last_height = 0`. I think semantically this is wrong, given the validation code we have for vote extensions

sug0 (changes requested):

core/src/ledger/storage/mod.rs
yito88: https://github.com/anoma/namada/issues/2043#issuecomment-1783562436

grarco (approved):

sug0 (approved):

PR #2051, opened 2023-10-26
masp: move to native VP

### LINKED ISSUE #1924: Native VPs should account for cpu cycles spent during validation
>Just as a note, I think this pattern should be shared across all native vps given the impossibility of tracking their execution at runtime. If a native vp is computationally expensive (access to storage are already correctly tracked though host functions, so no need to care about them) we should extract the complex part out to a host function to which we can assign a gas cost

_Originally posted by @grarco in https://github.com/anoma/namada/issues/1923#issuecomment-1731092440_

juped:
## Describe your changes

Closes #1924.

Moves the masp vp to native. Removes the now useless host env for masp validation and updates gas accounting and cost. Updated the masp vp benchmarks.

commit 3a0ecd89e8bfea1ca1eda7887d562666674a41d8 (v0.25.0)


benches/vps.rs
```diff
@@ -466,94 +463,5 @@ fn vp_validator(c: &mut Criterion) {
     group.finish();
 }
 
-fn vp_masp(c: &mut Criterion) {
```
grarco, line 469: We should move this benchmark to `benches/native_vps`

shared/src/ledger/native_vp/masp.rs
```diff
@@ -0,0 +1,296 @@
...
+            Some(Ordering::Greater) => {
+                tracing::debug!(
+                    "Transaction fees cannot be paid inside MASP transaction."
+                );
+                return Ok(false);
+            }
+            _ => {}
+        }
+        // Verify the proofs.
+        Ok(verify_shielded_tx(&shielded_tx))
```
grarco, line 288: It would be nice to benchmark this function because I think this is the most computationally expensive one and charge some gas before calling it in here

grarco:
https://github.com/anoma/namada/blob/3a0ecd89e8bfea1ca1eda7887d562666674a41d8/shared/src/vm/wasm/host_env.rs#L132

I believe we don't need this host fn anymore

sdk/src/signing.rs
```diff
@@ -277,8 +277,16 @@ pub async fn aux_signing_data<'a>(
             Some(AccountPublicKeysMap::from_iter(public_keys.clone())),
             1u8,
         ),
-        Some(owner @ Address::Internal(_)) => {
-            return Err(Error::from(TxError::InvalidAccount(owner.encode())));
+        Some(owner @ Address::Internal(internal)) => {
+            match internal {
+                // TODO use sentinel?
```
grarco, line 282: I don't think we need a signature on the transaction at all if the source is the MASP cause no VP would check it anyway

PR #2062, opened 2023-10-26
Remove IBC token denomination and refactoring

### LINKED ISSUE #1445: IBC: refactor for typed storage
We should create helper functions to ensure that we never write values of the wrong type to particular storage keys, and we should avoid mingling code which handles logic and code which handles serialisation.

For example (from `shared/src/ledger/ibc/mod.rs`)

```rust
    // the client counter
    let key = client_counter_key();
    let value = 0_u64.to_be_bytes().to_vec();
    storage
        .write_bytes(&key, value)
        .expect("Unable to write the initial client counter");
```

This should be refactored into (in some appropriate file):

```rust
pub fn write_client_counter(value: u64) {
    let key = client_counter_key();
    let value = value.to_be_bytes().to_vec();
    storage
        .write_bytes(&key, value.to_be_bytes())
        .expect("Unable to write the client counter");
}
```

such that the original code needs to be only

```rust
ibc_storage::write_client_counter(0_u64);
```

Does this make sense? We should apply this pattern generally, starting with the IBC code.

### LINKED ISSUE #1947: IBC token denomination doesn't have to be stored
The amount of ICS-20 transfer is `uint256`. So, the IBC token's denomination should be always zero. (I don't mean about ICS-20 token denomination here.)
When receiving the token over IBC, the denomination is stored as zero. It's redundant because we can always use zero denomination when they are queried.

yito88:
## Describe your changes
closes #1947 and closes #1445 by refactoring for IBC, 

`v0.25.0`


core/src/ledger/ibc/context/common.rs
```diff
@@ -31,149 +33,426 @@ use crate::ibc::mock::consensus_state::MockConsensusState;
...
+                }
+                .into()
+            }),
+            None => Err(ClientError::ClientSpecific {
+                description: format!(
+                    "The client update height doesn't exist: ID {client_id}",
+                ),
+            }
+            .into()),
+        }
```
sug0, line 226: here are my two cents on this kind of error handling: using matches to handle optionals, if not strictly necessary (e.g. because you need to `return` in one of the branches), introduces unnecessary strain when reading the code. IMO, this is because we are forced to evaluate two possible scenarios mentally (what happens when the code fails? what happens when the code succeeds?). handling errors with methods like `ok_or`, `map_err`, etc, improves readability I think, because errors are handled linearly, if that makes sense, so the logic flows better (e.g. you are not interrupted with error handling).

anyway, this is just a nit

sug0 (dismissed):
lgtm, managed to shave off quite a bit of lines!

core/src/ledger/ibc/context/common.rs
```diff
@@ -203,160 +482,139 @@ pub trait IbcCommonContext: IbcStorageContext {
...
+    /// Store the packet receipt
+    fn store_packet_receipt(
+        &mut self,
+        port_id: &PortId,
+        channel_id: &ChannelId,
+        sequence: Sequence,
+    ) -> Result<()> {
+        let key = storage::receipt_key(port_id, channel_id, sequence);
+        // the value is the same as ibc-go
+        let bytes = [1_u8].to_vec();
```
batconjurer, line 548: that's a shame we have to allocated a vec for a single byte

core/src/ledger/ibc/context/common.rs
```diff
@@ -203,160 +482,139 @@ pub trait IbcCommonContext: IbcStorageContext {
...
-    ) -> Result<(), ContextError> {
-        self.write(key, u64::from(sequence).to_be_bytes().to_vec())
-            .map_err(|_| {
-                ContextError::PacketError(PacketError::Channel(
-                    ChannelError::Other {
+    /// Read a counter
+    fn read_counter(&self, key: &Key) -> Result<u64> {
+        match self.read_bytes(key)? {
+            Some(value) => {
+                let value: [u8; 8] =
```
batconjurer, line 597: this is odd. Why can't we just store the u64 directly?

batconjurer (dismissed):
Generally looks good.

core/src/ledger/ibc/context/common.rs
yito88: Thanks! I modified them with `ok_or`

core/src/ledger/ibc/context/common.rs
yito88: You're right. We can use u64 directly (I used it for sequences before. but it isn't used anymore.)

batconjurer (approved):

sug0 (approved):

Issue #70, opened 2022-02-04
ledger (HW wallet) keys integration


tzemanovic:
Allow to import keys from the ledger app (to be built, hence externally blocked). This should allow for interactive signing via the hardware wallet.

cwgoes:
@karbyshev can you take this as part of the wallet work?

PR #2068, opened 2023-10-27
Added Ledger support to the wallet.

murisi:
## Describe your changes
Added Ledger support to the `namadaw` command, expanded the `Store` to also hold derivation paths and public keys, and attempted to simplify the wallet logic. More specifically:
* Derivation paths are now stored in the wallet to ease future signing operations
* Public keys are now stored in the wallet so that the `namadaw key list` command remains as informative even in the absence of the secret key (which may be being stored externally on a hardware wallet)
* Modified the key generation command to always generate and print out a seed phrase
* Encryption is now done in `store.rs` to ensure that the key derivations are always maximized and are done correctly:
    * I.e. a public key and public key hash of the same alias should always be derived from inserting secret key to wallet
* The SDK now takes derivation paths using the type `DerivationPath` instead of `String`
    * This enabled the removal of string parsing logic (like checking for "default" keyword) from the SDK
* Broke up the SDK key derivation function because it was doing many things and its parameters were like flags
* Implemented key derivation given a derivation path using the hardware wallet
    * This is dependent upon a `ledger-namada` branch enabling SLIP10 and returns larger key byte vectors that contain the key discriminant byte: https://github.com/heliaxdev/ledger-namada/tree/murisi/fix-rs-0.24.0
* Enabled transactions to be signed using the hardware wallet when it is connected
    * This is dependent upon a `ledger-namada` branch fixing the hashing of signature sections involved in multisig transactions: https://github.com/heliaxdev/ledger-namada/tree/murisi/fix-rs-0.24.0
* Parameterized SDK functions depending on randomness with a randomness parameter
* Removed code that allowed multiple wallet entries sharing the same alias to be inserted separately
* Implemented Serde (de)serialization of public keys and derivation paths to enable `Store` (de)serialization
* Augmented the `namadaw key list` command to also print out public keys (labelled as `external`) even when no corresponding secret key exists

closes https://github.com/anoma/namada/issues/70


apps/src/lib/client/tx.rs
```diff
@@ -57,6 +65,129 @@ pub async fn aux_signing_data<'a>(
...
 }
 
+// Sign the given transaction using a hardware wallet as a backup
+pub async fn sign<'a>(
+    context: &impl Namada<'a>,
+    tx: &mut Tx,
+    args: &args::Tx,
+    signing_data: SigningTxData,
+) -> Result<(), error::Error> {
+    // Setup a reusable context for signing transactions using the Ledger
```
cwgoes, line 75: I prefer that we use the Ledger app only explicitly, when the user has instructed us to do so, by either:
- passing some flag (such as `--use-device`), or
- using a key which we know (from the saved data) to have been generated with the Ledger

sdk/src/signing.rs
```diff
@@ -208,42 +228,96 @@ pub async fn tx_signers<'a>(
...
+                sigidx
+            })
             .collect();
         tx.add_signatures(signatures);
-    } else if let Some(account_public_keys_map) =
-        signing_data.account_public_keys_map
+    }
+
+    // Then try to sign the raw header with private keys in the software wallet
+    if let Some(account_public_keys_map) = signing_data.account_public_keys_map
```
cwgoes, line 255: Are we possibly signing with multiple keys? I'm a little bit confused by this flow. It seems to me that we should first lookup the key pair that the user instructed us to sign with, then use it (or return an error if not found)

sdk/src/wallet/mod.rs
```diff
@@ -225,6 +226,31 @@ pub mod fs {
...
+pub fn gen_secret_key(
+    scheme: SchemeType,
+    csprng: &mut (impl CryptoRng + RngCore),
+) -> common::SecretKey {
+    match scheme {
+        SchemeType::Ed25519 => ed25519::SigScheme::generate(csprng).try_to_sk(),
+        SchemeType::Secp256k1 => {
+            secp256k1::SigScheme::generate(csprng).try_to_sk()
+        }
+        SchemeType::Common => common::SigScheme::generate(csprng).try_to_sk(),
```
cwgoes, line 239: What is this?

sdk/src/wallet/mod.rs
```diff
@@ -481,52 +493,59 @@ impl<U: WalletIo> Wallet<U> {
...
-        ))
+            None,
+            Some(derivation_path),
+        )
+        .map(|alias| (alias, sk))
     }
 
-    /// Generate a new keypair and derive an implicit address from its public
+    /// Generate a spending key similarly to how it's done for keypairs
+    #[cfg(feature = "masp-tx-gen")]
```
cwgoes, line 530: Why is this under a feature flag?

cwgoes (commented):
Most of the changes look fine, but I have some minor concerns about the flow.

apps/src/lib/client/tx.rs
murisi: I've implemented option 1, `--use-device`, to trigger hardware wallet signing.

sdk/src/signing.rs
murisi: Agreed, this flow is a little confusing. Yes, we are potentially signing with multiple keys. I guess either way is possible: first try to apply the signatures the user has supplied via the CLI (used in multisig contexts), and then try signing with the keys the user has specified. Or the other way round. If wanted to immediately error on a key pair not being found, we would have to do some refactoring in order to keep the multisig implementation/flow working.

sdk/src/wallet/mod.rs
murisi: This function used to be called `gen_sk_rng` and lived in`sdk/src/wallet/store.rs`. Each implementation of signing scheme has to be identified with a `SchemeType` (`SchemeType::Secp256k1`, `SchemeType::Ed25519`, ...) (because of `SigScheme` trait requirements). Since we needed a signing scheme that is not fixed at compile-time, we ended up making `common::PublicKey`, `common::SecretKey`, `common::Signature` and identifying them with `SchemeType::Common`.

Since `common::SigScheme::generate(csprng)` panics, the entire branch `SchemeType::Common => common::SigScheme::generate(csprng).try_to_sk(),` panics. We could probably make that more clearer.

sdk/src/signing.rs
cwgoes: Ok, gotcha. Let's think about this later/separately.

core/src/types/key/common.rs
```diff
@@ -42,6 +40,48 @@ pub enum PublicKey {
...
+    fn serialize<S>(
+        &self,
+        serializer: S,
+    ) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // String encoded, because toml doesn't support enums
+        let prefix = match self {
+            PublicKey::Ed25519(_) => "ED25519_PK_PREFIX",
```
karbyshev, line 53: Could we define constants for "ED25519_PK_PREFIX" and "SECP256K1_PK_PREFIX", as they are used in both serialize/deserialize?

sdk/src/wallet/store.rs
```diff
@@ -491,25 +490,13 @@ impl Store {
...
-    }
-
     /// Insert a new address with the given alias. If the alias is already used,
     /// will prompt for overwrite/reselection confirmation, which when declined,
     /// the address won't be added. Return the selected alias if the address has
     /// been added.
     pub fn insert_address<U: WalletIo>(
         &mut self,
-        alias: Alias,
+        mut alias: Alias,
```
karbyshev, line 499: Now alias is both mutable and returned. That seems redundant to me.

sdk/src/wallet/derivation_path.rs
```diff
@@ -105,6 +106,29 @@ impl fmt::Display for DerivationPath {
     }
 }
 
+impl FromStr for DerivationPath {
+    type Err = DerivationPathError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
karbyshev, line 112: Why was it introduced? Can `DerivationPath::from_path_str` be used instead, which also takes the scheme into account?

karbyshev: For `SchemeType::Ed25519` all path segments must be hardened.

murisi: `FromStr` was introduced to support the implementation of `Serde::Deserialize` for `DerivationPath`. And this is required in order to be able to store `DerivationPath`s in the wallet's `Store`, which itself is serialized to file storage using Serde.

I see your point. Currently if a user supplies a non-hardened path like `m/44'/877'/0'/0/0`, we implicitly harden it to `m/44'/877'/0'/0'/0'` before commencing processing. Have we considered just rejecting non-hardened Ed25519 paths as is done in one of the implementations linked to from the SLIP10 specification? See https://github.com/elucidsoft/dotnetstandard-bip32/blob/master/dotnetstandard-bip32/BIP32.cs#L65 where they use the regex `^m(\/[0-9]+')+$`.

sdk/src/wallet/store.rs
```diff
@@ -443,23 +409,56 @@ impl Store {
...
-        self.payment_addrs.remove(alias);
-        self.view_keys.remove(alias);
-        self.spend_keys.remove(alias);
-        self.keys.remove(alias);
-        self.addresses.remove_by_left(alias);
-        self.pkhs.retain(|_key, val| val != alias);
+    /// Insert public keys
+    pub fn insert_public_key<U: WalletIo>(
+        &mut self,
+        mut alias: Alias,
```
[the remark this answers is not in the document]
murisi: Only a local copy of the parameter is modified (so the caller doesn't see the mutation). You might be thinking of `alias: &mut Alias`...

sdk/src/wallet/store.rs
murisi: Only a local copy of the parameter is modified (so the caller doesn't see the mutation). You might be thinking of `alias: &mut Alias`...

core/src/types/key/common.rs
murisi: Good point. Done, thanks!

sdk/src/wallet/mod.rs
murisi: You're right. This is a holdover from back when the WASM builds depended on the shared crate. Back then we were trying to prevent the WASM builds from indirectly depending on the `rand` and `rand_core` crates whilst not breaking MASP transaction generation that uses randomness.

But now, logically, if the SDK can generate secret keys without breaking MASP, then it should also be able to generate spending keys without breaking anything. Experimentally I've removed the flags and rebuilt the WASM binaries and their hashes haven't changed. I'm running the tests, but this removal is unlikely to have broken anything.

sdk/src/wallet/derivation_path.rs
karbyshev: > I guess requiring scheme information to deserialize derivation paths would mean having to tag derivation paths with their scheme types.
I would prefer this solution. I think that explicit storage of the scheme tag might be also useful from the user perspective.

PR #2089, opened 2023-11-03
Fix querying error while checking token balance

CroutonDigital:
- [x] Fix querying error while checking token balance


.gitignore
```diff
@@ -28,3 +28,6 @@ wasm/*.wasm
 
 # app version string file
 /apps/version.rs
+
+# mac
+.DS_Store
```
tzemanovic, line 33: pls don't add OS specific settings here - you can add it to your global git config

apps/src/lib/client/rpc.rs
```diff
@@ -340,8 +340,7 @@ pub async fn query_transparent_balance<'a>(
                             balance
                         );
                     }
-                    Err(e) => {
-                        display_line!(context.io(), "Querying error: {e}");
+                    Err(_) => {
```
tzemanovic, line 343: I think this error should be suppressed only when it's `QueryError::NoSuchKey` - any other errors should be shown instead of "No balance found" print

.gitignore
CroutonDigital: Removed. 

Thank you!

apps/src/lib/client/rpc.rs
CroutonDigital: All build tests passed

CroutonDigital: @tzemanovic can you my changes? 

Thank you!

PR #2091, opened 2023-11-04
Gas&fee updates

### LINKED ISSUE #1923: IBC transaction gas costs
Especially after #1904 is merged, IBC transactions will not pay many (if any) gas fees. Updating Namada's state machine with an IBC transaction is somewhat expensive, as validation and storage modifications do the exact same amount of work (i.e. validation is not cheaper, the same code is executed for wasm as is for the native VP).

We should assign a constant gas price to invoking the IBC tx actions host function, and some other cost to executing the IBC native VP. These values should be determined based on the computational effort of running the associated IBC code, with benchmarks.

### LINKED ISSUE #2054: Validator minimum gas fee parameter
We need to add a local configuration parameter to allow a validator to set the minimum gas fee that they want to accept. This should be done for every token they are willing to accept gas payment in. The validation of transactions in `prepare_proposal` should be based on this parameter instead of the protocol one.

### LINKED ISSUE #2056: Miscellaneous gas&fee updates
The gas & fee system needs the following miscellaneous updates:

- Updated the benchmarks
- Charge gas for the validation of the wrapper tx
- Charge gas for wasm validation
- Remove checked arithmetic ops if unneeded
- Set a definitive value for the `SCALE` constant
- Excessive fee redistribution among the delegators (nice to have, not mandatory)

grarco:
## Describe your changes

Addresses #1924.
Closes #1923.
Closes #2054.
Closes #2056.

This PR addressed the following:

- Updates the benchmarks
- Updates the gas costs and multipliers
- Charges gas for wrapper tx validation
- Tracks gas for ibc native vp expensive sections
- Adds a local configuration parameter for validators to customize their gas-fee tokens preferences

v0.25.0


apps/src/bin/namada-node/cli.rs
```diff
@@ -57,6 +58,24 @@ pub fn main() -> Result<()> {
                     );
                 }
             }
+            cmds::Config::UpdateLocalConfig(cmds::LocalConfig(args)) => {
+                // Validate the new config
+                let updated_config = std::fs::read(args.config_path).unwrap();
+                let _validator_local_config: ValidatorLocalConfig =
+                    toml::from_slice(&updated_config).unwrap();
+
+                // Update validator configuration file with the new accepted
```
cwgoes, line 67: presumably other updates would also be applied, if present in the file read?

apps/src/lib/bench_utils.rs
```diff
@@ -111,6 +111,10 @@ pub const TX_WITHDRAW_WASM: &str = "tx_withdraw.wasm";
 pub const TX_INIT_ACCOUNT_WASM: &str = "tx_init_account.wasm";
 pub const TX_INIT_VALIDATOR_WASM: &str = "tx_init_validator.wasm";
 
+pub const TX_RESIGN_STEWARD: &str = "tx_resign_steward.wasm";
```
cwgoes, line 114: how do these changes relate to this PR

cwgoes (commented):
Only some minor comments.

apps/src/lib/bench_utils.rs
grarco: Ah, we were just missing the benchmark for this specific transaction, so I added it in this PR, nothing more than this

apps/src/bin/namada-node/cli.rs
cwgoes: this is fine, but we should make sure the comments and cli usage info match what the code actually does

tzemanovic (approved):

PR #2103, opened 2023-11-06
deps: switch to tower-abci branch with overloaded service handling

tzemanovic:
## Describe your changes

https://github.com/heliaxdev/tower-abci/pull/12:
- added exponential backoff when a service overloaded
- peek requests in case a service call fails to be able to retry them - this avoids crashing CometBFT (it *should* be possible to shed query requests when the query service is overloaded, but at the moment the implementation doesn't distinguish this and shedding mempool requests does crash Comet)

0.25.0


tzemanovic:
this is to be rebased on #2141 with a rebased tower-abci error handling fix

tzemanovic:
done with #2141 instead

PR #2104, opened 2023-11-06
Replay protection delay write

### LINKED ISSUE #2036: Delay replay protection key write
Currently we write the hash of the inner tx when including the wrapper in the block. In some failure cases we remove this hash to allow for a replay. It would be slightly better though, to write the inner tx hash after it has been executed. This would allow to include in a single block different wrappers carrying the same inner transaction (slightly undesirable), but it would prevent a valid transaction from being excluded from a block because of another invalid tx with the same hash, which represents a more serious issue.

To do so we need to:

- in `prepare_proposal` and `process_proposal`, validate the inner tx hash only against the storage (no more write log check)
- in `finalize_block`, write the wrapper hash when accepting it, check the inner hash agains the write log and then execute the inner tx and
    - if tx successful or failed but no replay allowed, write the inner tx hash and remove the corresponding wrapper tx hash
    - otherwise do nothing

grarco:
## Describe your changes

Closes #2036.

Moves the write of the decrypted tx hash after its execution in `finalize_block` and introduces a new check to prevent applying the same decrypted tx more than once.

Updates unit tests.

Based on #1905 


cwgoes:
Note: reviewed in https://github.com/anoma/namada/pull/2108#pullrequestreview-1717094944 for the correct diff

tzemanovic (approved):

tzemanovic (approved):

PR #2108, opened 2023-11-07
Grarco/replay protection delay write REVIEW ONLY

cwgoes:
REVIEW ONLY

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -525,8 +525,22 @@ where
                                 | Error::TxApply(
                                     protocol::Error::MissingSection(_)
                                 )
+                                | Error::TxApply(
+                                    protocol::Error::ReplayAttempt(_)
+                                )
                         ) {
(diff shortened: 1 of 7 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -518,26 +515,33 @@ where
...
+                                    protocol::Error::ReplayAttempt(_)
+                                )
                         ) {
-                            self.allow_tx_replay(wrapper);
+                            self.commit_inner_tx_hash(wrapper);
+                        } else if let Error::TxApply(
+                            protocol::Error::ReplayAttempt(_),
+                        ) = msg
+                        {
+                            // Remove the wrapper hash but keep the inner tx
```
cwgoes, line 537: Can we further explicate this comment to explain why this is safe? (as I understand it, because the inner transaction hash will still be in storage, and block any future replay)

cwgoes (commented):
Only a minor comment, but @tzemanovic should also review this.

PR #2128, opened 2023-11-08
Governance `Abstain` vote

### LINKED ISSUE #2090: Implement an `Abstain` vote option in governance
See spec at https://github.com/anoma/namada-docs/pull/196

grarco:
## Describe your changes

Closes #2090.

Implements the `Abstain` governance vote. Also reviews the tally process for the different proposal types.

Commit 3a0ecd89e8bfea1ca1eda7887d562666674a41d8 on `base` (v0.25.0)

### DIFF
--- a/core/src/ledger/governance/cli/onchain.rs
+++ b/core/src/ledger/governance/cli/onchain.rs
@@ -356,6 +356,11 @@ impl ProposalVote {
         matches!(self, ProposalVote::Nay)
     }
 
+    /// Check if the vote type is abstain
+    pub fn is_abstain(&self) -> bool {
+        matches!(self, ProposalVote::Abstain)
+    }
+
     /// Check if two votes are equal
     pub fn is_same_side(&self, other: &Self) -> bool {
         std::mem::discriminant(self) == std::mem::discriminant(other)
--- a/core/src/ledger/governance/cli/onchain.rs
+++ b/core/src/ledger/governance/cli/onchain.rs
@@ -330,8 +330,6 @@ pub enum ProposalVote {
     Nay,
     /// Represent an abstain proposal vote
     Abstain,
-    /// Represent an invalid proposal vote
-    Invalid,
 }
 
 impl TryFrom<String> for ProposalVote {
--- a/core/src/ledger/governance/cli/onchain.rs
+++ b/core/src/ledger/governance/cli/onchain.rs
@@ -313,7 +313,7 @@ pub struct PgfFundingTarget {
     pub address: Address,
 }
 
-/// Rappresent an proposal vote
+/// Represent an proposal vote
 #[derive(
     Debug,
     Clone,
@@ -324,11 +324,13 @@ pub struct PgfFundingTarget {
     PartialEq,
 )]
 pub enum ProposalVote {
-    /// Rappresent an yay proposal vote
+    /// Represent an yay proposal vote
     Yay,
-    /// Rappresent an nay proposal vote
+    /// Represent an nay proposal vote
     Nay,
-    /// Rappresent an invalid proposal vote
+    /// Represent an abstain proposal vote
+    Abstain,
+    /// Represent an invalid proposal vote
     Invalid,
 }
 
@@ -339,14 +341,25 @@ impl TryFrom<String> for ProposalVote {
         match value.trim().to_lowercase().as_str() {
             "yay" => Ok(ProposalVote::Yay),
             "nay" => Ok(ProposalVote::Nay),
+            "abstain" => Ok(ProposalVote::Abstain),
             _ => Err("invalid vote".to_string()),
         }
     }
 }
 
 impl ProposalVote {
-    /// Check if the proposal type is yay
+    /// Check if the vote type is yay
     pub fn is_yay(&self) -> bool {
         matches!(self, ProposalVote::Yay)
+    }
+
+    /// Check if the vote type is nay
+    pub fn is_nay(&self) -> bool {
+        matches!(self, ProposalVote::Nay)
+    }
+
+    /// Check if two votes are equal
+    pub fn is_same_side(&self, other: &Self) -> bool {
+        std::mem::discriminant(self) == std::mem::discriminant(other)
     }
 }
--- a/core/src/ledger/governance/utils.rs
+++ b/core/src/ledger/governance/utils.rs
@@ -109,7 +109,7 @@ impl TallyResult {
     ) -> Self {
         let passed = match tally_type {
             TallyType::TwoThirds => {
-                yay_voting_power >= total_voting_power / 3 * 2
+                yay_voting_power >= total_voting_power * 2 / 3
             }
             TallyType::OneHalfOverOneThird => {
                 let at_least_one_third_voted =
@@ -180,10 +180,10 @@ impl ProposalResult {
         let at_least_two_thirds_voted = self.total_yay_power
             + self.total_nay_power
             + self.total_abstain_power
-            >= self.total_voting_power / 3 * 2;
+            >= self.total_voting_power * 2 / 3;
 
         let at_least_two_thirds_nay = self.total_nay_power
-            >= (self.total_nay_power + self.total_yay_power) / 3 * 2;
+            >= (self.total_nay_power + self.total_yay_power) * 2 / 3;
 
         at_least_two_thirds_voted && at_least_two_thirds_nay
     }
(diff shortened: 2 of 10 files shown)


core/src/ledger/governance/utils.rs
```diff
@@ -161,9 +174,18 @@ impl Display for ProposalResult {
...
-    /// Return true if two third of total voting power voted nay
-    pub fn two_third_nay(&self) -> bool {
-        self.total_nay_power >= (self.total_voting_power / 3) * 2
+    /// Return true if at least 1/3 of the total voting power voted and at least
+    /// two third of the non-abstained voting power voted nay
+    pub fn two_thirds_nay_over_two_thirds_total(&self) -> bool {
+        let at_least_two_thirds_voted = self.total_yay_power
+            + self.total_nay_power
+            + self.total_abstain_power
+            >= self.total_voting_power / 3 * 2;
```
cwgoes, line 183:
```suggestion
            >= self.total_voting_power * 2 / 3;
```

in general, we should practice multiply-then-divide

core/src/ledger/governance/utils.rs
```diff
@@ -161,9 +174,18 @@ impl Display for ProposalResult {
...
+
+        let at_least_two_thirds_nay = self.total_nay_power
+            >= (self.total_nay_power + self.total_yay_power) / 3 * 2;
```
cwgoes, line 186:
```suggestion
            >= (self.total_nay_power + self.total_yay_power) * 2 / 3;
```

core/src/ledger/governance/utils.rs
```diff
@@ -196,12 +218,25 @@ impl TallyVote {
...
-
-        both_yay || !both_nay
+        match (self, other) {
+            (TallyVote::OnChain(vote), TallyVote::OnChain(other_vote)) => {
+                vote.is_same_side(other_vote)
+            }
+            (TallyVote::Offline(vote), TallyVote::Offline(other_vote)) => {
+                vote.is_same_side(other_vote)
+            }
+            _ => false,
```
cwgoes, line 238: when would this case happen? shouldn't this be impossible?

core/src/ledger/governance/utils.rs
```diff
@@ -226,14 +261,17 @@ pub fn compute_proposal_result(
...
 
     for (address, vote_power) in votes.validator_voting_power {
         let vote_type = votes.validators_vote.get(&address);
         if let Some(vote) = vote_type {
             if vote.is_yay() {
                 yay_voting_power += vote_power;
-            } else {
+            } else if vote.is_nay() {
                 nay_voting_power += vote_power;
+            } else {
```
cwgoes, line 273: `vote.is_abstain()` would be more defensive, in case we add another voting option in the future

core/src/ledger/governance/utils.rs
```diff
@@ -249,16 +287,33 @@ pub fn compute_proposal_result(
                 if !validator_vote.is_same_side(delegator_vote) {
                     if delegator_vote.is_yay() {
                         yay_voting_power += voting_power;
-                        nay_voting_power -= voting_power;
-                    } else {
+                        if validator_vote.is_nay() {
+                            nay_voting_power -= voting_power;
+                        } else {
```
cwgoes, line 292: ditto

core/src/ledger/governance/utils.rs
```diff
@@ -249,16 +287,33 @@ pub fn compute_proposal_result(
...
+                            abstain_voting_power -= voting_power;
+                        }
+                    } else if delegator_vote.is_nay() {
                         nay_voting_power += voting_power;
-                        yay_voting_power -= voting_power;
+                        if validator_vote.is_yay() {
+                            yay_voting_power -= voting_power;
+                        } else {
```
cwgoes, line 299: ditto

core/src/ledger/governance/utils.rs
```diff
@@ -249,16 +287,33 @@ pub fn compute_proposal_result(
...
+                            abstain_voting_power -= voting_power;
+                        }
+                    } else {
+                        abstain_voting_power += voting_power;
+                        if validator_vote.is_yay() {
+                            yay_voting_power -= voting_power;
+                        } else {
```
cwgoes, line 306: ditto

core/src/ledger/governance/utils.rs
```diff
@@ -249,16 +287,33 @@ pub fn compute_proposal_result(
...
+                            nay_voting_power -= voting_power;
+                        }
                     }
                 }
             } else if delegator_vote.is_yay() {
                 yay_voting_power += voting_power;
-            } else {
+            } else if delegator_vote.is_nay() {
                 nay_voting_power += voting_power;
+            } else {
```
cwgoes, line 315: ditto

tests/src/e2e/ledger_tests.rs
```diff
@@ -2095,7 +2095,7 @@ fn proposal_submission() -> Result<()> {
     let mut client = run!(test, Bin::Client, query_proposal, Some(15))?;
     client.exp_string("Proposal Id: 0")?;
     client.exp_string(
-        "passed with 120900.000000 yay votes and 0.000000 nay votes (0.%)",
+        "passed with 120000.000000 yay votes and 900.000000 nay votes (0.%)",
```
cwgoes, line 2098: why did this change? :thinking:

cwgoes (changes requested):

core/src/ledger/governance/utils.rs
grarco: Never, it's just the the method is implemented on the `TallyVote` enum which contains two variants, one for the on-chain votes and the other one for off-chain votes and we still need the pattern matching to be exhaustive, so we need to cover also the case in which the two instances are actually two different variants. We can return an error instead of `false` to make it more clear

core/src/ledger/governance/cli/onchain.rs
```diff
@@ -324,11 +324,13 @@ pub struct PgfFundingTarget {
...
+    /// Represent an yay proposal vote
     Yay,
-    /// Rappresent an nay proposal vote
+    /// Represent an nay proposal vote
     Nay,
-    /// Rappresent an invalid proposal vote
+    /// Represent an abstain proposal vote
+    Abstain,
+    /// Represent an invalid proposal vote
     Invalid,
```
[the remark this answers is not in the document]
grarco: Actually nowhere, I'm removing it thanks!

core/src/ledger/governance/utils.rs
grarco: I think the operators precedence in Rust already evaluates multiplications before divisions, but I can change it. Do you think at this point it's better to split the fraction among the two sides of the inequality an be `(yay + nay + abstain) * 3 >= total * 2`?

core/src/ledger/governance/utils.rs
```diff
@@ -196,12 +218,25 @@ impl TallyVote {
...
+            TallyVote::Offline(vote) => vote.is_nay(),
+        }
+    }
+
     /// Check if two votes are equal
     pub fn is_same_side(&self, other: &TallyVote) -> bool {
-        let both_yay = self.is_yay() && other.is_yay();
-        let both_nay = !self.is_yay() && !other.is_yay();
-
-        both_yay || !both_nay
```
grarco, line 204: @cwgoes here's the bug for the wrong e2e tally, it should have been `both_yay || both_nay`

tests/src/e2e/ledger_tests.rs
grarco: In test we submit two votes, a `yay` with the validator (120000 voting power) and a `nay` with a delegator (900 voting power) and previously we were not seeing the nay votes before because of a bung in the `is_same_side` function. I've tagged you at the wrong line

core/src/ledger/governance/utils.rs
cwgoes: wherever a case isn't expected to happen, we should return an error

core/src/ledger/governance/utils.rs
cwgoes: how you have it is fine, I just prefer `* 3 / 2`. admittedly this is somewhat of an aesthetic choice

tests/src/e2e/ledger_tests.rs
cwgoes: ah

cwgoes (approved):
lgtm

PR #2130, opened 2023-11-08
Add client command to dry run genesis files

### LINKED ISSUE #761: Validators should check that they can compile wasm at startup.
On certain machines, the wasm compiler that is used isn't supported.  Currently, the failure to compile wasm is only detected the first time a tx with wasm is sent to the ledger. Furthermore, the compilation error leads the validator to reject the state changes and thus its state diverges from other validators. 

To mitigate this issue, on startup, validators should attempt to compile wasm to ensure it is supported  on their machine.

batconjurer:
A new client command has been added that takes a set of pre-genesis template files, validates them, and runs them through `init_chain`. All errors are collected into a report. 

This also closes #761


sug0:
This should be a descendant of #2186, please don't merge it until that PR is ready

PR #2133, opened 2023-11-08
Fix governance bugs

### LINKED ISSUE #1107: `is_delegator` return `true` for a validator in Gov VP
Inside the governance VP , `fn is_delegator()` will return `true` for a validator with self-bonds, which could unintentionally cause a `true` return value for `is_valid_vote_key`.

`shared/src/ledger/native_vp/governance/mod.rs`

grarco:
## Describe your changes

Closes #1107.

Fixes the `is_delegator` function in the governance vp which could return true also for validators with self bonds.
Corrects the inflation mechanism for pgf.

Commit 3a0ecd89e8bfea1ca1eda7887d562666674a41d8 on `base` (v0.25.0)

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -800,17 +800,7 @@ where
         let pgf_pd_rate =
             pgf_parameters.pgf_inflation_rate / Dec::from(epochs_per_year);
         let pgf_inflation = Dec::from(total_tokens) * pgf_pd_rate;
-
-        let stewards = pgf::get_stewards(&self.wl_storage)?;
-        let pgf_stewards_pd_rate =
-            pgf_parameters.stewards_inflation_rate / Dec::from(epochs_per_year);
-        let pgf_steward_inflation =
-            Dec::from(total_tokens) * pgf_stewards_pd_rate;
-        let total_pgf_stewards_inflation =
-            pgf_steward_inflation * Dec::from(stewards.len());
-
-        let pgf_inflation_amount =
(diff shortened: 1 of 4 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -830,22 +830,23 @@ where
         pgf_fundings.sort_by(|a, b| a.id.cmp(&b.id));
 
         for funding in pgf_fundings {
-            if credit_tokens(
+            if storage_api::token::transfer(
```
cwgoes, line 833: I don't think this is complete, we also need to update/change https://github.com/anoma/namada/blob/main/apps/src/lib/node/ledger/shell/finalize_block.rs#L858 - I think we can actually use `credit_tokens` there, but then we shouldn't send the steward inflation to the PGF account in the first place (https://github.com/anoma/namada/blob/main/apps/src/lib/node/ledger/shell/finalize_block.rs#L812)

shared/src/ledger/governance/mod.rs
```diff
@@ -669,21 +669,13 @@ where
         epoch: Epoch,
         verifiers: &BTreeSet<Address>,
         address: &Address,
-        delegation_address: &Address,
+        _delegation_address: &Address,
     ) -> Result<bool> {
-        let bond_handle = pos::namada_proof_of_stake::bond_handle(
+        Ok(pos::namada_proof_of_stake::is_delegator(
```
Fraccaman, line 674: I think we could also add a check if `address != delegation_address`

cwgoes (approved):

PR #2137, opened 2023-11-09
Change validator's consensus key

### LINKED ISSUE #36: PoS validator consensus key change
Add a tx and update PoS validity predicate to allow validators to change consensus key 

NOTE: how do we update this in tendermint? We currently using the key file on disk. The key change on-chain will happen in a future epoch, so we'd need to update the file at the right time.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763047146961) by [Unito](https://www.unito.io)

brentstone:
## Describe your changes
Closes #36.

TODO:
- e2e test? (RIP)

v0.26.0


brentstone:
@tzemanovic @cwgoes Should the validator be required to explicitly provide a new consensus key or should we allow a new consensus key to be generated for them as well?

cwgoes:
> @tzemanovic @cwgoes Should the validator be required to explicitly provide a new consensus key or should we allow a new consensus key to be generated for them as well?

They should be required to explicitly provide one, only experienced validators should be doing this.

brentstone:
> They should be required to explicitly provide one, only experienced validators should be doing this.

@adrianbrink

PR #2163, opened 2023-11-12
init validator tx - verify ownership of used keys

### LINKED ISSUE #106: PoS - verify validator keys ownership with signatures
For the on-chain validator account creation and for transactions to change consensus key (pending on anoma/namada#36), we must also require that a signature with the consensus key is attached to the transactions to verify the key's ownership. Depends on anoma/namada#85 to be able to attach multiple signatures to a tx.

For this, we should update the client commands and the PoS VP to enforce this rule.

tzemanovic:
## Describe your changes

closes #106 

0.26.0

### DIFF
--- a/sdk/src/signing.rs
+++ b/sdk/src/signing.rs
@@ -384,6 +384,51 @@ pub async fn aux_signing_data<'a>(
     })
 }
 
+pub async fn init_validator_signing_data<'a>(
+    context: &impl Namada<'a>,
+    args: &args::Tx<SdkTypes>,
+    validator_keys: Vec<common::PublicKey>,
+) -> Result<SigningTxData, Error> {
+    let mut public_keys = if args.wrapper_fee_payer.is_none() {
+        tx_signers(context, args, None).await?
+    } else {
+        vec![]
+    };
+    public_keys.extend(validator_keys.clone());
+
+    let account_public_keys_map =
+        Some(AccountPublicKeysMap::from_iter(validator_keys));
+
+    let fee_payer = if args.disposable_signing_key {
+        context
+            .wallet_mut()
+            .await
+            .gen_disposable_signing_key(&mut OsRng)
+            .to_public()
+    } else {
+        match &args.wrapper_fee_payer {
+            Some(keypair) => keypair.to_public(),
+            None => public_keys.get(0).ok_or(TxError::InvalidFeePayer)?.clone(),
+        }
+    };
+
+    if fee_payer == masp_tx_key().to_public() {
+        other_err(
+            "The gas payer cannot be the MASP, please provide a different gas \
+             payer."
+                .to_string(),
+        )?;
+    }
+
+    Ok(SigningTxData {
+        owner: None,
+        public_keys,
+        threshold: 0,
+        account_public_keys_map,
+        fee_payer,
+    })
+}
+
 /// Informations about the post-tx balance of the tx's source. Used to correctly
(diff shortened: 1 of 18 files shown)


tzemanovic:
The bench for init-validator is affected, it’s most likely missing the newly required signatures

tx_prelude/src/lib.rs
```diff
@@ -361,3 +363,36 @@ impl TxEnv for Ctx {
...
     unsafe { namada_tx_ibc_execute() }
 }
+
+/// Verify section signatures against the given list of keys
+pub fn verify_signatures_of_pks(
+    ctx: &Ctx,
+    tx: &Tx,
+    pks: Vec<common::PublicKey>,
+) -> EnvResult<bool> {
+    let max_signatures_per_transaction =
```
brentstone, line 373: What if the `max_signatures_per_transaction` is lower than the number of validator keys at initialization?

sdk/src/signing.rs
```diff
@@ -384,6 +384,51 @@ pub async fn aux_signing_data<'a>(
...
+            "The gas payer cannot be the MASP, please provide a different gas \
+             payer."
+                .to_string(),
+        )?;
+    }
+
+    Ok(SigningTxData {
+        owner: None,
+        public_keys,
+        threshold: 0,
```
brentstone, line 426: what does a threshold of 0 mean here?

brentstone (dismissed):
LGTM

shared/src/vm/host_env.rs
```diff
@@ -2076,6 +2076,85 @@ where
...
+    )?;
+    match tx.verify_signatures(
+        &hashes,
+        public_keys_map,
+        &None,
+        threshold,
+        max_signatures,
+        // This uses VpGasMeter, so we're charging the gas here manually
+        // instead
+        &mut None,
```
grarco, line 2142: Minor thing here, we can change the `verify_signatures` method of `Tx` to accept an `&mut dyn GasMetering` trait object just like we do in `fetch_or_compile` so that we can still charge gas for every single signature without the need to do it manually, if you prefer. Alternatively, since this function is only called by the `init_validator` tx, and we always require 4 valid signatures, we could also just skip charging gas in here since its cost is already included in the benchmark of the tx itself

tx_prelude/src/lib.rs
tzemanovic: As we don't have support for validator multisig, it would only be an issue if the source of the tx is a mutisig with threshold > `max_signatures_per_transaction - 5`. We use `15` as the default which should be sufficient for either case

sdk/src/signing.rs
tzemanovic: it's unused for signing - this field is only used for offline proposal votes

shared/src/vm/host_env.rs
tzemanovic: I've refactored it in [88578bf](https://github.com/anoma/namada/pull/2163/commits/88578bfa1b48d6587bc18c112f31611c829dea38) but without the dynamic dispatch using a lambda instead. It also moves the gas metering logic back into the relevant places in `vm/host_env.rs` instead of being done in `proto/types.rs`

tx_prelude/src/lib.rs
tzemanovic: ah, I was wrong on this - we do have multisig validator support, but with the default this would still allow up to 10 multisig participants which should be sufficient

PR #2178, opened 2023-11-14
Misc PoS tasks

### LINKED ISSUE #17: PoS - use checked arithmetics
Use checked arithmetics in the PoS library code (has a TODO left in code)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762910654658) by [Unito](https://www.unito.io)

### LINKED ISSUE #1270: PoS: panic on slashing failure
Depends on #892 

In `apps/src/lib/node/ledger/shell/mod.rs` calls to PoS `slash` and `process_slashes` should be switched from error handling with error logging to panics as slashing is a critical system so it's better to shut down in case of an issue. For now, for devnets and testnets stability it's better to just log any errors

brentstone:
## Describe your changes
Closes #17, closes #1270. Some other things addressed in this PR:
- rename `store_total_consensus_stake` to `compute_and_store_total_consensus_stake`
- Improves docstrings, comments, and logging
- Removes outdated TODOs in PoS
- `is_delegator` issue

v0.26.0

### DIFF
--- a/proof_of_stake/src/lib.rs
+++ b/proof_of_stake/src/lib.rs
@@ -753,9 +753,18 @@ where
             }
             Ok(false)
         }
-        None => Ok(storage_api::iter_prefix_bytes(storage, &prefix)?
-            .next()
-            .is_some()),
+        None => {
+            let iter = storage_api::iter_prefix_bytes(storage, &prefix)?;
+            for res in iter {
+                let (key, _) = res?;
+                if let Some((bond_id, _epoch)) = is_bond_key(&key) {
+                    if bond_id.source != bond_id.validator {
+                        return Ok(true);
+                    }
+                }
+            }
+            Ok(false)
+        }
     }
 }
 
--- a/proof_of_stake/src/lib.rs
+++ b/proof_of_stake/src/lib.rs
@@ -518,7 +518,13 @@ where
     let val = handle
         .get_delta_val(storage, current_epoch + offset)?
         .unwrap_or_default();
-    handle.set(storage, val + delta, current_epoch, offset)
+    handle.set(
+        storage,
+        val.checked_add(&delta)
+            .expect("Validator deltas updated amount should not overflow"),
+        current_epoch,
+        offset,
+    )
 }
 
 /// Read PoS total stake (sum of deltas).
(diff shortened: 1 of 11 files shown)


proof_of_stake/src/lib.rs
```diff
@@ -327,7 +322,7 @@ where
             current_epoch,
             epoch,
         )?;
-        store_total_consensus_stake(storage, epoch)?;
+        // compute_and_store_total_consensus_stake(storage, epoch)?;
```
cwgoes, line 325: Why is this now commented?

proof_of_stake/src/lib.rs
```diff
@@ -308,7 +302,8 @@ where
     Ok(())
 }
 
-/// new init genesis
+/// Copies the validator sets into all epochs up through the pipeline epoch at
+/// genesis. Also computes the total
```
cwgoes, line 306: Sentence fragment

proof_of_stake/src/lib.rs
```diff
@@ -716,6 +721,8 @@ pub fn is_validator<S>(
 where
     S: StorageRead,
 {
+    // TODO: should this check be made different? I suppose it does work but
```
cwgoes, line 724: What do you mean?

proof_of_stake/src/rewards.rs
```diff
@@ -91,6 +90,11 @@ impl PosRewardsCalculator {
 
     /// Implement as ceiling of (2/3) * validator set stake
     fn get_min_required_votes(&self) -> Amount {
-        ((self.total_stake * 2u64) + (3u64 - 1u64)) / 3u64
+        (self
+            .total_stake
+            .checked_mul(2.into())
+            .expect("Amount overflow while computing minimum required votes")
+            + (3u64 - 1u64))
```
cwgoes, line 97: Couldn't it technically overflow here?

cwgoes (changes requested):
Minor nits.

proof_of_stake/src/lib.rs
brentstone: It was a redundant operation. We already call this function in `init_chain` and in `finalize_block` upon a new epoch, both times before calling `copy_validator_sets`. It is only necessary to call the function at the beginning of a new epoch and compute the total consensus stake for that current epoch only.

proof_of_stake/src/lib.rs
brentstone: Thx, will fix up.

proof_of_stake/src/lib.rs
brentstone: Technically this works because all validators need to have a `max_commission_rate_change` set in storage, but it just seems odd from a readability perspective. May change this in a small PR before the next release

proof_of_stake/src/rewards.rs
brentstone: Yeah, you are right, this should have an extra `checked_add`

PR #2183, opened 2023-11-15
Restore non-hd keys generation functionality

karbyshev:
This restores option for generation of random non-hd keys which was removed in #2158


sug0:
I think this should be handled separately, with a cmdline flag that would conflict with `--hd-path` (e.g. `--no-mnemonic`). dispatching on the presence of `--hd-path` results in a bit of a strange UX, IMO

karbyshev:
Integrated in #2260

PR #2167, opened 2023-11-13
fix to get proofs even if no height is specified

yito88:
## Describe your changes
closes #2166 

## Indicate on which release or other PRs this topic is based on
`0.26.0`

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #2190, opened 2023-11-17
Save conversion state into the state storage

### LINKED ISSUE #2189: Save conversion state in the block state storage
Currently, we save the MASP conversion state in the subspace every epoch. However, we don't need the old conversion state.
So, we can save the conversion state into the block state (RocksDB's `state` column family) not to record the diffs.
It would reduce the diffs size in the storage.

yito88:
## Describe your changes
closes #2189 

`0.26.0`

### DIFF
--- a/.changelog/unreleased/improvements/2189-save-conversion-state.md
+++ b/.changelog/unreleased/improvements/2189-save-conversion-state.md
@@ -1,0 +1,2 @@
+- Save MASP conversion state to the state storage instead of the diffs
+  ([\#2189](https://github.com/anoma/namada/issues/2189))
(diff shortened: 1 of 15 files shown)


grarco (approved):
LGTM, thanks!

murisi (approved):
Looks good to me, thanks. A question: does the `invalid_transactions` test usually fail?

grarco:
> Looks good to me, thanks. A question: does the `invalid_transactions` test usually fail?

I think it's failing because this PR doesn't include #2167, which is currently in draft

yito88:
Thanks, `invalid_transactions` failed after this change. `run_ledger_ibc` failed due to #2167. Let me check `invalid_transactions`.

PR #2193, opened 2023-11-20
Cwgoes/for review only

cwgoes:
ONLY FOR REVIEW DO NOT MERGE


apps/src/lib/config/genesis.rs
```diff
@@ -336,14 +366,54 @@ pub fn make_dev_genesis(
...
     if let Some(bonds) = genesis.transactions.bond.as_mut() {
-        bonds.retain(|bond| {
-            bond.source
-                != transactions::AliasOrPk::Alias(
-                    Alias::from_str("albert").unwrap(),
-                )
-        })
+        let default_addresses: HashMap<Alias, Address> =
+            defaults::addresses().into_iter().collect();
+        let fat_alberts_address =
```
cwgoes, line 381: what is this, lol. shouldn't we just change the original config?

apps/src/lib/config/genesis.rs
```diff
@@ -336,14 +366,54 @@ pub fn make_dev_genesis(
...
+                    .unwrap(),
+            ));
+        bonds.retain(|bond| bond.source != fat_alberts_address);
+    };
+    // fetch validator's balances
+    let (first_val_balance, first_val_bonded) = {
+        let nam_balances = genesis
+            .balances
+            .token
+            .get_mut(&Alias::from_str("nam").unwrap())
```
cwgoes, line 394: "nam" should be a constant

apps/src/lib/config/genesis/transactions.rs
```diff
@@ -579,6 +452,10 @@ pub struct ValidatorAccountTx<PK> {
     pub eth_cold_key: PK,
 }
 
+impl DeriveEstablishedAddress for UnsignedValidatorAccountTx {
+    const SALT: &'static str = "validator-account-tx";
```
cwgoes, line 456: Why do we need different constant salts at all? Are we trying to avoid some kind of collision? I don't really follow.

cwgoes (commented):
See questions.

Issue #2110, opened 2023-11-07
Prune old Bridge pool merkle trees


sug0:
we might be able to withgo storing multiple bridge pool merkle trees, as well. we only need to keep trees from the latest root that has been signed with a quorum signature onward. so, suppose the latest merkle tree nonce is $5$, but the latest signed root's nonce is $3$; then, we can delete all the trees signed with a nonce of $0$ through $2$, whilst upholding the liveness of the bridge.

_Originally posted by @sug0 in https://github.com/anoma/namada/issues/2043#issuecomment-1780567749_

PR #2206, opened 2023-11-21
Yuji/prune bp merkle tree

yito88:
## Describe your changes
#2110 

`0.27.0`

### DIFF
--- a/apps/src/lib/node/ledger/storage/mod.rs
+++ b/apps/src/lib/node/ledger/storage/mod.rs
@@ -55,6 +55,7 @@ mod tests {
 
     use itertools::Itertools;
     use namada::core::ledger::masp_conversions::update_allowed_conversions;
+    use namada::ledger::eth_bridge::storage::bridge_pool;
     use namada::ledger::gas::STORAGE_ACCESS_GAS_PER_BYTE;
     use namada::ledger::ibc::storage::ibc_key;
     use namada::ledger::parameters::{EpochDuration, Parameters};
@@ -62,6 +63,7 @@ mod tests {
     use namada::ledger::storage::{types, StoreType, WlStorage};
     use namada::ledger::storage_api::{self, StorageWrite};
     use namada::types::chain::ChainId;
+    use namada::types::ethereum_events::Uint;
     use namada::types::hash::Hash;
     use namada::types::storage::{BlockHash, BlockHeight, Key};
     use namada::types::time::DurationSecs;
@@ -487,6 +489,9 @@ mod tests {
             let value_bytes = types::encode(&storage.block.height);
             storage.write(&key, value_bytes)?;
         }
+        let key = bridge_pool::get_nonce_key();
+        let bytes = types::encode(&Uint::default());
+        storage.write(&key, bytes)?;
 
         // Update and commit
         let hash = BlockHash::default();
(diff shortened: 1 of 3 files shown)


core/src/ledger/storage/mod.rs
```diff
@@ -1182,6 +1193,52 @@ where
...
+                match self.block.pred_epochs.get_start_height_of_epoch(epoch) {
+                    Some(h) => h,
+                    None => continue,
+                };
+            let (bytes, _) = self.read_with_height(&nonce_key, height)?;
+            let bytes = bytes.expect("Bridge pool nonce should exits");
+            let nonce = Uint::try_from_slice(&bytes)
+                .map_err(Error::BorshCodingError)?;
+            if nonce < current_nonce {
+                break;
```
grarco, line 1236: Isn't this returning at the first epoch with an invalid nonce instead of the last one matching the current nonce?

apps/src/lib/node/ledger/storage/mod.rs
```diff
@@ -606,6 +616,13 @@ mod tests {
             .write(&key, types::encode(&value))
             .expect("write failed");
 
+        let nonce = nonce + 1;
+        let bytes = types::encode(&nonce);
+        storage.write(&bp_nonce_key, bytes).unwrap();
+
+        let bytes = types::encode(&Uint::default());
+        storage.write(&bp_nonce_key, bytes).unwrap();
```
grarco, line 624: Do we need this second write since we've just written the same key?

core/src/ledger/storage/mod.rs
yito88: The nonce was incremented to the current one during the epoch though the epoch has the invalid nonce at the starting height. The Merkle tree at the epoch starting would be required to rebuild one at a height (after the nonce was incremented) during the epoch. So, the Merkle tree in the epoch should have a valid nonce.
There is a corner case; the nonce was incremented at the same height when the epoch was updated. However, the Merkle tree snapshot would be pruned eventually.

grarco (dismissed):
lgtm, thanks!

apps/src/lib/node/ledger/storage/mod.rs
yito88: thanks, it's not needed

grarco (approved):

core/src/ledger/storage/mod.rs
```diff
@@ -1182,6 +1193,52 @@ where
...
+        };
+        self.block
+            .pred_epochs
+            .get_epoch(oldest_height)
+            .unwrap_or_default()
+    }
+
+    /// Get oldest epoch which has the valid signed nonce of the bridge pool
+    pub fn get_oldest_epoch_with_valid_nonce(&self) -> Result<Epoch> {
+        let nonce_key = bridge_pool::get_nonce_key();
```
sug0, line 1212: this is incorrect. `get_nonce_key` doesn't return a signed nonce; rather, it returns the latest Bridge pool nonce, which may lag behind. the key you want is `get_signed_bridge_pool_root`, which returns a tuple containing a `BridgePoolRootProof` as one of the elements. inside of `BridgePoolRootProof` there will be a `Uint` corresponding to the signed nonce.

yito88: Can we move `BridgePoolRootProof` and other types from `ethereum_bridge` to `core` to decode `BridgePoolRootProof` in `core` storage?

PR #2202, opened 2023-11-21
combine user vp and validator vp

tzemanovic:
## Indicate on which release or other PRs this topic is based on

v0.27.0

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

Issue #408, opened 2022-08-31
vp_user: consider being less permissive on unknown keys


tzemanovic:
in VPs implementation, we decided to be permissive on changed storage keys that are unknown or irrelevant to a validity predicate that is being implemented. 

In case of e.g. vp_token I think this makes a lot of sense, it shouldn't need to know or check what happens in PoS (a bond is just a transfer from source to PoS from its perspetive). I think in vp_user this is could be dangerous though, as this VP is used for signature authorization for certain actions (for e.g. receiving transparent tokens, it doesn't require signature from the receiver, but naturally it does for spending tokens), so the vp_user is coded to understand certain storage keys of other accounts.

Perhaps if the vp_user cannot make a sense of the storage change from another account, the safer default for it would be to still require a valid signature. Otherwise, one might have a transaction mistakenly thinking that's been authorized by a signature of a key associated with the account by vp_user, while instead the storage change(s) might just be unknown to it.

In vp_user, it's this case:

```rust
            KeyType::Unknown => {
                if key.segments.get(0) == Some(&addr.to_db_key()) {
                    // Unknown changes to this address space require a valid
                    // signature
                    *valid_sig
                } else {
                    // Unknown changes anywhere else are permitted
                    true
                }
            }
```

we'd change to:
```rust
            KeyType::Unknown => {
                // Unknown changes require a valid signature
                *valid_sig
            }
```

cwgoes:
Agree, we should be very defensive here.

tzemanovic:
I'll add this a milestone as it's important

PR #2208, opened 2023-11-22
replace tx_init_validator with tx_become_validator that doesn't create an account

tzemanovic:
## Describe your changes

A new `tx_become_validator` replaces `tx_init_validator`. This tx doesn't initialize a new account and instead it modifies an existing established address to become a validator. This currently requires that there are no delegations on the source account before it can become a validator (if there are some, they have to be unbonded, but they don't have to be withdrawn).

The client command `init-validator` is kept for convenience and updated to send `tx_init_account` tx before `tx_become_validator`.

#2202

Diff against base PR for review: https://github.com/anoma/namada/pull/2208/files/7de64272819719887bdc1e2087e25f7d25884886..2881c44643d4f30bbb2dbaefa3150c61a74145bd

### DIFF
--- a/apps/src/lib/client/rpc.rs
+++ b/apps/src/lib/client/rpc.rs
@@ -2288,7 +2288,7 @@ pub async fn get_public_key<C: namada::ledger::queries::Client + Sync>(
     rpc::get_public_key_at(client, address, index).await
 }
 
-/// Check if the given address is a known validator.
+/// Check if the given address has any bonds.
 pub async fn is_validator<C: namada::ledger::queries::Client + Sync>(
     client: &C,
     address: &Address,
@@ -2318,7 +2318,7 @@ pub async fn is_delegator_at<C: namada::ledger::queries::Client + Sync>(
         .unwrap()
 }
 
-/// Check if the given address is a known validator.
+/// Check if the given address has any bonds.
 pub async fn has_bonds<C: namada::ledger::queries::Client + Sync>(
     client: &C,
     address: &Address,


apps/src/lib/client/tx.rs
```diff
@@ -512,6 +513,79 @@ pub async fn submit_init_validator<'a>(
...
+        );
+        if !tx_args.force {
+            safe_exit(1)
+        }
+    };
+
+    // If the address is not yet a validator, it cannot have self-bonds, but it
+    // may have delegations. It has to unbond those before it can become a
+    // validator.
+    if rpc::has_bonds(namada.client(), &address).await {
```
brentstone, line 543: If the established address then goes and unbonds all of its delegations, with this logic it will be able to immediately then become a validator successfully before waiting the unbonding period, since `unbond` will decrement the bonds to 0 immediately. Just want to check that this is the desired behavior or if we want to wait the unbonding period (which would require `has_bonds` to check existing `unbonds` as well)

apps/src/lib/client/tx.rs
```diff
@@ -721,19 +739,22 @@ pub async fn submit_init_validator<'a>(
...
         website,
         discord_handle,
-        validator_vp_code_hash: extra_section_hash,
     };
 
     // Put together all the PKs that we have to sign with to verify ownership
-    let mut all_pks = data.account_keys.clone();
+    let account = namada_sdk::rpc::get_account_info(namada.client(), &address)
+        .await?
+        .expect("TODO");
```
brentstone, line 747: fill in this TODO

brentstone (dismissed):
left some comments / questions, and suggested to take care of one "TODO", but otherwise LGTM

apps/src/lib/client/tx.rs
cwgoes: but this check is only in the transaction creation code, no? we'd need this check to be in the tx verification logic later (maybe it is)

cwgoes: I found the other check in the code later. I think not checking unbonds is fine in principle, we should have a quick test for this though to make sure nothing breaks.

tzemanovic: Yeah, I think this is safe. Only thing that can be done with `unbonds` is to withdraw them once they're ready. If the delegation gets slashed, then only the withdrawal action will affected. Rewards are not auto-bonded so they wouldn't affect this either (they have to be claimed back into the account first)

wasm/wasm_source/src/vp_user.rs
```diff
@@ -122,12 +126,85 @@ fn validate_tx(
...
+                                // Becoming a validator must be authorized
+                                *valid_sig
+                            }
+                            _ => true,
+                        }
+                    }
+                    None => true,
+                };
+
+                valid_bond_or_unbond_change
```
cwgoes, line 204: if the change is unrecognized, shouldn't we reject it by default? looks like we accept it by default here

tzemanovic: ah, yes, we decided to do this in https://github.com/anoma/namada/issues/408. Let's deal with this separately though as it affects other sub-systems too

tzemanovic: > I think not checking unbonds is fine in principle, we should have a quick test for this though to make sure nothing breaks.

I think if anything breaks it will most likely be interactions with some other thing like slashing or rewards (though to my understanding it is safe - see https://github.com/anoma/namada/pull/2208#discussion_r1403114572). I'd also like to test this and imho the best way to check this would be to add this to state machine, but we don't yet cover rewards and couple other things

apps/src/lib/client/tx.rs
tzemanovic: done in [9122528](https://github.com/anoma/namada/pull/2208/commits/9122528e0aab864fa440b0262e049c473ca5a292)

tzemanovic:
f-pushed squashed fixup commit (no diff: https://github.com/anoma/namada/compare/9122528e0aab864fa440b0262e049c473ca5a292..8159327a0296aa766bf009aad6e317fca7f3772d)

PR #2213, opened 2023-11-23
less permissive VPs

### LINKED ISSUE #408: vp_user: consider being less permissive on unknown keys

tzemanovic:
## Describe your changes

fixes #408

Unknown changes in vp_user and vp_implicit now require valid signature(s). 

For PoS changes, we have to check against all possible storage changes that can be applied from a tx and require signature(s) for those that have to be authorized by the source (for vp_user it can be both delegator and validator actions, for vp_implicit only delegator actions). If the change within PoS account subspace is not recognized we default to require sig(s).

To ease debugging in vp_user and vp_implicit, I changed the logging of rejected keys to also be be emitted in release build.

v0.28.0

### DIFF
--- a/wasm/wasm_source/src/vp_user.rs
+++ b/wasm/wasm_source/src/vp_user.rs
@@ -11,7 +11,6 @@
 //!
 //! Any other storage key changes are allowed only with a valid signature.
 
-use namada_vp_prelude::storage::KeySeg;
 use namada_vp_prelude::*;
 use once_cell::unsync::Lazy;
 use proof_of_stake::types::ValidatorState;
@@ -115,15 +114,15 @@ fn validate_tx(
                         proof_of_stake::storage::is_unbond_key(key)
                             .map(|(bond_id, _, _)| bond_id)
                     });
-                let valid_bond_or_unbond_change = match bond_id {
+                let is_valid_bond_or_unbond_change = match bond_id {
                     Some(bond_id) => {
                         // Bonds and unbonds changes for this address
                         // must be signed
                         bond_id.source != addr || *valid_sig
                     }
                     None => {
-                        // Any other PoS changes are allowed without signature
-                        true
+                        // Unknown changes are not allowed
+                        false
                     }
                 };
                 // Commission rate changes must be signed by the validator
@@ -131,26 +130,26 @@ fn validate_tx(
                     proof_of_stake::storage::is_validator_commission_rate_key(
                         key,
                     );
-                let valid_commission_rate_change = match comm {
+                let is_valid_commission_rate_change = match comm {
                     Some((validator, _epoch)) => {
                         *validator == addr && *valid_sig
                     }
-                    None => true,
+                    None => false,
                 };
                 // Metadata changes must be signed by the validator whose
                 // metadata is manipulated
                 let metadata =
                     proof_of_stake::storage::is_validator_metadata_key(key);
-                let valid_metadata_change = match metadata {
+                let is_valid_metadata_change = match metadata {
                     Some(address) => *address == addr && *valid_sig,
-                    None => true,
+                    None => false,
                 };
 
                 // Changes due to unjailing, deactivating, and reactivating are
                 // marked by changes in validator state
                 let state_change =
                     proof_of_stake::storage::is_validator_state_key(key);
-                let valid_state_change = match state_change {
+                let is_valid_state_change = match state_change {
                     Some((address, epoch)) => {
                         let params_pre =
                             proof_of_stake::read_pos_params(&ctx.pre())?;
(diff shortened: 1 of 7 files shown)


tzemanovic:
looks like lots of e2e tests are affected, looking into it

tzemanovic:
there are some more changes needed for IBC

wasm/wasm_source/src/vp_user.rs
```diff
@@ -158,35 +136,200 @@ fn validate_tx(
...
+    let is_valid_bond_or_unbond_change = || {
+        let bond_id = storage::is_bond_key(key)
+            .map(|(bond_id, _)| bond_id)
+            .or_else(|| storage::is_bond_epoched_meta_key(key))
+            .or_else(|| {
+                storage::is_unbond_key(key).map(|(bond_id, _, _)| bond_id)
+            });
+        if let Some(bond_id) = bond_id {
+            // Bonds and unbonds changes for this address must be signed
+            return &bond_id.source != owner || **valid_sig;
```
brentstone, line 172: Should the `&bond_id.source != owner` still be here now? How do we enforce that a delegation is signed / truly submitted by the delegator?

wasm/wasm_source/src/vp_user.rs
```diff
@@ -158,35 +136,200 @@ fn validate_tx(
...
+                        ) {
+                            true
+                        } else {
+                            // Unknown state changes are not allowed
+                            false
+                        }
+                    }
+                    (None, Some(_post)) => {
+                        // Becoming a validator must be authorized
+                        **valid_sig
```
brentstone, line 251: maybe unnecessary, but should we have a `address == owner && *valid_sig` here too?

wasm/wasm_source/src/vp_user.rs
```diff
@@ -158,35 +136,200 @@ fn validate_tx(
...
+            || storage::is_validator_eth_hot_key_key(key).is_some()
+            || storage::is_validator_max_commission_rate_change_key(key)
+                .is_some()
+            || storage::is_validator_address_raw_hash_key(key).is_some()
+        {
+            // A signature is required to become validator
+            return **valid_sig;
+        }
+        false
+    };
```
brentstone, line 308: should this look at the data in these keys from `pre` and `post` and be valid if the `pre`s are `None` and `post`s are `Some`?

wasm/wasm_source/src/vp_user.rs
```diff
@@ -478,10 +621,184 @@ mod tests {
...
+            let max_commission_rate_change = Dec::new(1, 2).unwrap();
+            let args = transaction::pos::BecomeValidator {
+                address: address.clone(),
+                consensus_key,
+                eth_cold_key,
+                eth_hot_key,
+                protocol_key,
+                commission_rate,
+                max_commission_rate_change,
+                email: "cucumber@tastes.good".to_string(),
```
brentstone, line 687: agreed

wasm/wasm_source/src/vp_user.rs
tzemanovic: I think we don’t need to go that deep in here. The detailed PoS validation should happen in PoS VP and this VP should only be concerned with authorization

wasm/wasm_source/src/vp_user.rs
```diff
@@ -158,35 +136,200 @@ fn validate_tx(
...
+    };
+
+    let is_valid_reward_claim = || {
+        if let Some(bond_id) = storage::is_last_pos_reward_claim_epoch_key(key)
+        {
+            // Claims for this address must be signed
+            return &bond_id.source != owner || **valid_sig;
+        }
+        false
+    };
```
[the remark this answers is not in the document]
tzemanovic: Ah yes! Otherwise a claim might get rejected

wasm/wasm_source/src/vp_user.rs
tzemanovic: I believe that’s what this does - if the source is eq to the owner then the sig is required

brentstone: ah yes ok, forgot that this is now the VP for normal users and validators. So in the delegator's VP, the sig will be required, but in the validator's VP it won't be, as expected

PR #2218, opened 2023-11-28
Update consensus key change

brentstone:
## Describe your changes
Add an e2e test and require to sign the tx with the new consensus key.

v0.28.1

### DIFF
--- a/apps/src/lib/client/tx.rs
+++ b/apps/src/lib/client/tx.rs
@@ -374,6 +374,7 @@ pub async fn submit_change_consensus_key(
         consensus_key_alias =
             format!("{base_consensus_key_alias}-{key_counter}");
     }
+    drop(wallet);
 
     let mut wallet = namada.wallet_mut().await;
     let consensus_key = consensus_key
--- a/apps/src/lib/client/tx.rs
+++ b/apps/src/lib/client/tx.rs
@@ -343,6 +343,7 @@ pub async fn submit_change_consensus_key(
         tx: tx_args,
         validator,
         consensus_key,
+        unsafe_dont_encrypt,
         tx_code_path: _,
     }: args::ConsensusKeyChange,
 ) -> Result<(), error::Error> {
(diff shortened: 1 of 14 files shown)


tzemanovic (approved):

apps/src/lib/client/tx.rs
```diff
@@ -459,17 +459,14 @@ pub async fn submit_change_consensus_key(
...
             display_line!(
                 namada.io(),
-                "  Consensus key \"{}\"",
-                consensus_key_alias
+                "New consensus key stored with alias \
+                 \"{consensus_key_alias}\". It will become active \
+                 {EPOCH_SWITCH_BLOCKS_DELAY} blocks before pipeline offset \
+                 from the current epoch, at which point you'll need to give \
+                 the new key to CometBFT in order to be able to sign with it \
+                 in consensus.",
```
brentstone, line 469: can we somehow make this message dependent on if the transaction is valid or not? Playing around and testing to make sure if validator-1 submits the consensus key change for validator-0, then the tx is invalid. However, we still currently print this message when it is invalid.

tzemanovic: ah good catch! Looks like we need to do this for couple other txs too, so I'll do it in a follow up PR - the client output could be overall improved

tzemanovic: also noticed that for some reason, the storage keys for this inner tx show up in the client result as empty (they don't show in `Transaction applied with result:` info when they're empty) - I can see them populated in the node, so I'll open another issue to find why

brentstone: Opened #2267 for this

tzemanovic (approved):

tzemanovic:
fixed the issue in benches [1c4f085](https://github.com/anoma/namada/pull/2218/commits/1c4f085590c283fe9679662067ae6797a3ebfdf2) - the tx needed the extra sig with the new consensus key

PR #2220, opened 2023-11-28
Light SDK

adrianbrink:
## Describe your changes

Implements a first version of a simpler, more user-friendly SDK.

Commit 74d412e14 of `base` (v0.28.1)


light_sdk/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    "tx_update_steward_commission.wasm";
+/// Redelegate transaction WASM path
+pub const TX_REDELEGATE_WASM: &str = "tx_redelegate.wasm";
+/// Target chain ID
+pub const CHAIN_ID: &str = "localnet.3e837878d84b54a40f-0";
+/// Reveal public key transaction code hash
+pub const TX_REVEAL_PK_HASH: [u8; 32] = [0x8d, 0x4e, 0x20, 0x08, 0x30, 0xa9, 0xf6, 0x03, 0xd6, 0x47, 0x79, 0x9f, 0xaf, 0xdb, 0xf1, 0x94, 0x6e, 0x07, 0xa0, 0x38, 0xa6, 0x7c, 0xfa, 0x09, 0xd4, 0xee, 0xb3, 0x1c, 0x80, 0xc7, 0x74, 0xce];
+
+/// Takes any kind of inner tx and gives me back my sign bytes
+pub fn get_inner_sign_bytes(mut tx: Tx) -> Hash {
```
adrianbrink, line 155: These should be immutable functions on a specific RevealPk struct. Ie. this signature should be repeated for every type of transaction.

Also this function should be immutable.

light_sdk/src/tx_builders/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+use namada_core::types::token;
+use namada_core::types::token::{Amount, DenominatedAmount, MaspDenom};
+use namada_core::types::transaction::Fee;
+use namada_core::types::transaction::GasLimit;
+use std::collections::BTreeMap;
+use std::str::FromStr;
+
+pub mod pos;
+
+pub(in crate::tx_builders) fn build_tx(
```
adrianbrink, line 25: no generics - everything should be passed in as strings

light_sdk/src/tx_builders/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        secret_keys,
+        public_keys_index_map,
+        signer,
+    );
+
+    (tx, signatures)
+}
+
+pub(in crate::tx_builders) fn attach_raw_signatures(
+    mut tx: Tx,
```
adrianbrink, line 60: no mutability - we should consume the tx object and return a new one

light_sdk/src/tx_builders/pos.rs
```diff
@@ -0,0 +1,653 @@
...
+            code_hash,
+            TX_UNBOND_WASM.to_string(),
+            chain_id,
+        ))
+    }
+
+    /// Generate the signature(s) for the given transaction and signers
+    pub fn generate_signatures(
+        mut self,
+        secret_keys: &[common::SecretKey],
```
adrianbrink, line 139: the library doesn't take secret keys. 

it should expose `sign_bytes()` and then the caller figures out how to sign it

light_sdk/src/tx_builders/pos.rs
```diff
@@ -0,0 +1,653 @@
...
+            timestamp,
+            expiration,
+            code_hash,
+            TX_INIT_VALIDATOR_WASM.to_string(),
+            chain_id,
+        ))
+    }
+
+    /// Generate the signature(s) for the given transaction and signers
+    pub fn generate_signatures(
```
adrianbrink, line 213: likewise - this kind of function shouldn't exist in the public api

light_sdk/src/tx_builders/mod.rs
grarco: That's what I'm doing here but I need the object to be mutable in the scope of the function

light_sdk/src/tx_builders/mod.rs
grarco: This function is not exposed to the SDK user

light_sdk/src/reading/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+            })?
+            .data;
+        println!("Dry-run result: {}", result);
+        Ok(result)
+    }
+
+    /// Lookup the full response accompanying the specified transaction event
+    pub fn query_tx_response(
+        tendermint_addr: &str,
+        tx_query: TxEventQuery<'_>,
```
adrianbrink, line 99: This is essentially a hash, right? Can this type be instantiated by the caller?

light_sdk/src/reading/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+            TendermintAddress::from_str(tendermint_addr)
+                .map_err(|e| Error::Other(e.to_string()))?,
+        )
+        .map_err(|e| Error::Other(e.to_string()))?;
+        let rt = Runtime::new().unwrap();
+        rt.block_on(rpc::query_bond(&client, source, validator, epoch))
+    }
+
+    /// Query a validator's unbonds for a given epoch
+    pub fn query_and_print_unbonds(
```
adrianbrink, line 446: let's not have print options in the api. if a caller wants to print something, they can do it in their own code.

light_sdk/src/reading/mod.rs
```diff
@@ -0,0 +1,679 @@
+use std::str::FromStr;
```
adrianbrink, line 1: Overall I like the api.

It's probably better to split the individual submodules into their own files.

light_sdk/src/writing/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod blocking {}
```
adrianbrink, line 1: This should be just 1 function - submit_tx().

light_sdk/src/reading/mod.rs
batconjurer: yeah, I changed this to take a `&str`. It means you can only query applied txs and not accepted, but I think that is totally fine.

batconjurer (dismissed):

Issue #2094, opened 2023-11-05
MASP finalization plan


cwgoes:
Proposed order after reading through the issue backlog:

1. [ ] Investigate the epoch transition proof issue and agree on a plan
2. [ ] Convert MASP to a native validity predicate (rebase and merge https://github.com/anoma/namada/pull/2051)
3. [ ] Check that our imported versions of the upstream MASP crates are the latest
4. [ ] Implementing missing validation (https://github.com/anoma/namada/issues/1373 and https://github.com/anoma/namada/issues/1962)
5. [ ] Check/fix unsafe validation (https://github.com/anoma/namada/issues/1374) 
6. [ ] Audit MASP & MASP rewards
7. [ ] Other stuff

adrianbrink:
Ensure that wallets have an easy library to import to generate proofs that we can independently optimize.

adrianbrink:
also check that proofing times don't scale with epoch sizes.

cwgoes:
Two points to check on the epoch transition issue:
1. We should allow proofs with reference to past commitment tree anchors.
2. We should allow unshielding from previous epochs (the user will receive fewer rewards, we can show a warning in extreme cases).

Together per discussion we think that these two changes should resolve the issue.

cwgoes:
Notes:
- Rework to avoid unnecessary conversions on AssetType, which should be used all the way through to the edge of CLI client input.
- Discrete conversions with threshold: amounts below conversion are currently left behind - we should just send them anyways, and later allow unshielding (with no conversions)
- Check on proof generation time increasing (maybe more notes to convert?)
- Check on duplicate SpendDescriptions, OutputDescriptions, etc. - seems to be the case in recent iterations of the client.

cwgoes:
- Fix MASP verifying key / parameters loading (should only load once into memory when the ledger starts)

PR #2222, opened 2023-11-28
Masp epoch transition proof fix

### LINKED ISSUE #307: Handle MASP Transactions Across Epoch Boundaries

### LINKED ISSUE #1880: Retry transactions on fee unshielding epoch error
The optional fee unshielding operation that can be attached to a transaction for fee payment is a masp unshielding operations. As such, it can fail because of an epoch boundary. At the moment though, we are not attempting to resubmit the transaction in case of failure like we are doing for the `transfer` tx. We should implement a similar mechanism to resubmit it if the failure was likely due to an epoch change

grarco:
## Describe your changes

Addresses #2094.
Closes #307 (@murisi not completely sure about this).
Closes #1880.

Allows to unshield assets from the MASP that belong to older epochs.

v0.27.0


grarco:
Note: the integration tests are actually passing for me locally, they fail here because of a time limit on the CI step

tests/src/integration/masp.rs
```diff
@@ -1248,3 +1245,90 @@ fn wrapper_fee_unshielding() -> Result<()> {
     assert!(tx_run);
     Ok(())
 }
+
+// Test that a masp unshield transaction can be succesfully executed even across
+// an epoch boundary.
+#[test]
+fn cross_epoch_tx() -> Result<()> {
```
murisi, line 1252: Can I assume that assume that CI integration test failure is due solely to the additional time taken by this test?

murisi (approved):
Looks good to me, thanks! Can I assume that the integration tests don't take significantly more time to execute on your machine than before your changes?

PR #2225, opened 2023-11-28
Removed uses of lifetimes in the SDKs Namada trait and implementation

murisi:
## Describe your changes
Modified the `Namada` trait and implementation to own its their clients, streams, wallet, and shielded utilities. Then removed all the redundant lifetimes. This was relatively straightforward because it was rare that after a `NamadaImpl` has been instantiated that any other object would want to own the client, streams, wallet, or shielded utilities.

Namada 0.27.0


sug0 (dismissed):
Nice, this is a big improvement. It should solve many lifetime related annoyances in the SDK.

sug0:
don't forget to add an sdk changelog

tzemanovic (approved):

Issue #1373, opened 2023-05-15
Missing MASP Transaction Validation


murisi:
MASP transaction validation seems to be incomplete, and hence may accept MASP transactions that should be rejected. In particular it seems to be missing the following:
* Spend description anchor validation. The anchor of each spend description must be confirmed to refer to an earlier treestate. If this is not done malicious clients will be able to spend notes from a fabricated note commitment tree. This requirement is inherited from the consensus rules subsection of https://zips.z.cash/protocol/protocol.pdf#spendsandoutputs .
* Nullifier uniqueness enforcement. The nullifier revealed by a given spend description must not already be in the nullifier set. If this is not done malicious clients will be able to do double-spends. This requirement is inherited from the consensus rules subsection of https://zips.z.cash/protocol/protocol.pdf#nullifierset .
* Convert description anchor validation. The anchor of a convert description must be confirmed to refer to the current conversion tree. If this is not done malicious clients will be able to apply conversions from a fabricated allowed conversion tree. This requirement is implicit in the specs, but perhaps ought to be stated more explicitly.
* Note commitment tree capacity check. A transaction must be rejected if it would cause the note commitment tree to exceed its capacity. If this were allowed to happen the MASP shielded pool would enter an undefined state. See the consensus rules subsection of https://zips.z.cash/protocol/protocol.pdf#merkletree .

These seem to be all relevant consensus rules inherited from the Zcash spec that cannot/are not already enforced directly by the MASP crate. cc @juped @gijswijs @cwgoes @joebebel .

cwgoes:
Ah! I didn't realize we weren't doing these already. Is anything here likely to be particularly difficult to implement?

PR #2240, opened 2023-11-30
MASP nullifier uniqueness

grarco:
## Describe your changes

Addresses #1373.

Implements a nullifier set to prevent double spending in masp transactions. Updates the masp VP to check that the transactions writes the correct nullifiers and that these haven't already been revealed.

**NOTE**: testing this with an integration or e2e test is impossible cause we cannot force the submission of the tx if the builder fails to construct the shielded `Transaction`. We'd need to test this with a unit test (which I think would be the correct way regardless of the `force` issue) but at the moment we don't have the tools to write unit tests for the masp vp (at least to my knowledge)

v0.27.0

### DIFF
--- a/shared/src/ledger/native_vp/ibc/context.rs
+++ b/shared/src/ledger/native_vp/ibc/context.rs
@@ -5,8 +5,7 @@ use std::collections::{BTreeSet, HashMap, HashSet};
 use borsh_ext::BorshSerializeExt;
 use masp_primitives::transaction::Transaction;
 use namada_core::ledger::ibc::{IbcCommonContext, IbcStorageContext};
-use namada_core::types::hash::Hash;
-use namada_core::types::token::MASP_NULLIFIERS_KEY_PREFIX;
+use namada_core::ledger::masp_utils;
 
 use crate::ledger::ibc::storage::is_ibc_key;
 use crate::ledger::native_vp::CtxPreStorageRead;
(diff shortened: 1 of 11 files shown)


shared/src/ledger/native_vp/masp.rs
```diff
@@ -161,6 +166,43 @@ where
...
+                    || revealed_nullifiers.contains(&nullifier_key)
+                {
+                    tracing::debug!(
+                        "MASP double spending attempt, the nullifier {:#?} \
+                         has already been revealed previously",
+                        description.nullifier.0
+                    );
+                    return Ok(false);
+                }
+                revealed_nullifiers.insert(nullifier_key);
```
murisi, line 192: Right now in the VP we check that the nullifier keys changed are a subset of those in the transaction. But where do we check that there is a key change for each of the nullifiers in the transaction? I would expect this check to be somewhere in this loop...

shared/src/ledger/native_vp/ibc/context.rs
```diff
@@ -237,6 +239,19 @@ where
         );
         self.write(&current_tx_key, record.serialize_to_vec())?;
         self.write(&head_tx_key, (current_tx_idx + 1).serialize_to_vec())?;
+        for description in shielded
```
murisi, line 242: I see a bit of duplication in how we update the nullifier set in the transaction codes. But maybe getting rid of it is easier said than done and can be deferred to a later PR?

murisi (changes requested):
Do you know why the unit and integration tests are failing in the CI? They seem to be timing out for some reason... Do they pass on your machine? Have you tried regenerating the MASP test fixtures maybe?

shared/src/ledger/native_vp/ibc/context.rs
grarco: I've tried to improve the thing with a shared function now

shared/src/ledger/native_vp/masp.rs
grarco: Nice catch thanks! I've added the missing check now

grarco:
> Do you know why the unit and integration tests are failing in the CI? They seem to be timing out for some reason... Do they pass on your machine? Have you tried regenerating the MASP test fixtures maybe?

Both unit and integration tests pass on my machine without the need to rebuild the proofs

grarco:
> Nice. Could you please just check out the `serialize_to_vec` issue in the comments to see if there is anything to it? And if you get the chance (but it's not strictly necessary), could you please try seeing if moving `handle_masp_tx` from `tx_prelude/src/token.rs` into `core/src/ledger/masp_utils.rs` is trivial or not?

Ok' I've moved the entire `handle_masp_tx` to the `masp_utils` file. This has also solved the double serialization issue (which yes, I believe like you do it was a bug). To be able to share this function also with the IBC context I had to remove the call to `insert_verifier(masp_addr)` from the function (others I would have needed to implement `TxEnv` on the IBC context), which shouldn't be a problem because the tx writes some masp keys anyway so those should already trigger the vp

PR #2241, opened 2023-12-01
Update README.md

kylemasterd1:
Hi,

"An example setup with a single validator used to run a localnet can be found in localnet directory." - "used to run a localnet" may be clearer as "used for running a localnet". So I made that change. 

Thanks.


brentstone:
Unncessary

PR #2237, opened 2023-11-30
Build Hermes in CI

yito88:
## Describe your changes
Build Hermes with the current Namada commit instead of downloading

## Indicate on which release or other PRs this topic is based on
`v0.27.0`

## Checklist before merging to `draft`
- [ ] I have added a changelog
- [ ] Git history is in acceptable state

PR #1797, opened 2023-08-09
Add IBC E2E test with Hermes

yito88:
## Describe your changes
closes #773
depends on #2237
Add a new IBC E2E test with Hermes relaying

## Indicate on which release or other PRs this topic is based on
~~#1784~~ ~~`0.28.0`~~ `0.29.0`

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #2243, opened 2023-12-04
PGF over IBC

### LINKED ISSUE #1395: CPGF & RPGF over IBC
Namada should support CPGF (recurring) and RPGF (one-time) allocations which are sent to an address over IBC (maybe also over the Ethereum bridge, but we can start with only IBC). When these allocations are processed, instead of sending tokens to a Namada address (as we do for CPGF right now), we should generate an ICS20 transfer packet on the specified port, connection, and channel, to the specified destination address on the other chain. We don't need to worry about handling timeouts for now (but we should set a long timeout on the packet, maybe slightly under an epoch length).

@Fraccaman Does this seem straightforward to implement, or is anything potentially tricky?

yito88:
## Describe your changes
closes #1395
CI for this PR requires Hermes: #2237

#1797

### DIFF
--- a/apps/src/lib/node/ledger/shell/governance.rs
+++ b/apps/src/lib/node/ledger/shell/governance.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 
 use namada::core::ledger::governance::storage::keys as gov_storage;
 use namada::core::ledger::governance::storage::proposal::{
-    AddRemove, PGFAction, ProposalType, StoragePgfFunding,
+    AddRemove, PGFAction, PGFTarget, ProposalType, StoragePgfFunding,
 };
 use namada::core::ledger::governance::utils::{
     compute_proposal_result, ProposalVotes, TallyResult, TallyType, TallyVote,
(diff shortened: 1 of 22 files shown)


Fraccaman:
can you add the new `e2e` test to the `e2e.json` file pls

apps/src/lib/node/ledger/shell/governance.rs
```diff
@@ -340,64 +347,71 @@ where
...
-                ) {
+                let result = match &target {
+                    PGFTarget::Internal(target) => token::transfer(
+                        storage,
+                        token,
+                        &ADDRESS,
+                        &target.target,
+                        target.amount,
+                    ),
+                    PGFTarget::Ibc(target) => {
```
grarco, line 398: This actually adds support for rPGF payments over ibc which doesn't show up as a requirement in #1395. It looks good to me I just want to make sure this is intended

cwgoes: ah I think having both is good

grarco (dismissed):
Looks good to me, thank you!

Fraccaman (approved):

yito88:
Rebased. I moved `ibc.rs` from `app` to `core`, which implements `transfer_over_ibc` for a protocol transaction because of conflict.
@Fraccaman @grarco Could you review it again?

apps/src/lib/node/ledger/shell/governance.rs
```diff
@@ -136,6 +136,20 @@ where
                             id
                         );
 
+                        for ibc_event in
```
grarco, line 139: It looks to me like we are doing this loop to set the height for ibc events three times: once here, once in the `apply_inflation` function and once when we handle the `TxResult` coming from `dispatch_tx`. I wonder if we could do this just once at the end of `finalize_block`

grarco: Nevermind, this doesn't make sense, we need to extract the events from the write log and push them to the abci response

grarco (approved):

PR #2244, opened 2023-12-04
Spend descriptions validation

grarco:
## Describe your changes

Addresses #1373.

Creates a note commitment tree in storage and modifies the masp tx to update this tree.
Changes `finalize_block` to update the anchor of this tree at the end of a block.
Updates the masp VP to verify the changes made to the tree (also checks that the tree capacity is not exceeded) and the validity of the anchor for each spend description.

Commit 39e00c5c16fdb68f88626c6ff175ff6948cf2dda of #2240 

### DIFF
--- a/core/src/ledger/masp_utils.rs
+++ b/core/src/ledger/masp_utils.rs
@@ -1,15 +1,22 @@
 //! MASP utilities
 
+use masp_primitives::merkle_tree::CommitmentTree;
+use masp_primitives::sapling::Node;
 use masp_primitives::transaction::Transaction;
+use masp_proofs::bls12_381;
 
 use super::storage_api::{StorageRead, StorageWrite};
-use crate::ledger::storage_api::Result;
+use crate::ledger::storage_api::{Error, Result};
 use crate::types::address::MASP;
 use crate::types::hash::Hash;
 use crate::types::storage::{BlockHeight, Epoch, Key, KeySeg, TxIndex};
 use crate::types::token::{
     Transfer, HEAD_TX_KEY, MASP_NULLIFIERS_KEY_PREFIX, PIN_KEY_PREFIX,
     TX_KEY_PREFIX,
+};
+use crate::types::token::{
+    MASP_NOTE_COMMITMENT_ANCHOR_PREFIX, MASP_NOTE_COMMITMENT_TREE,
+    MASP_NULLIFIERS_KEY_PREFIX,
 };
 
 // Writes the nullifiers of the provided masp transaction to storage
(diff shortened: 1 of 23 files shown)


core/src/ledger/masp_utils.rs
```diff
@@ -0,0 +1,113 @@
...
+            .expect("Cannot obtain a storage key");
+        ctx.write(&nullifier_key, ())?;
+    }
+
+    Ok(())
+}
+
+/// Appends the note commitments of the provided transaction to the merkle tree
+/// and updates the anchor
+/// NOTE: this function is public as a temporary workaround because of an issue
```
grarco, line 39: So initially this was compiled and called directly from WASM. Unfortunately I found the the wasm context was crashing constantly when calling the `append` method of the tree. At the same time, though, the MASP native VP had no problem running the same function. I even tried ignoring the error coming out of the `append` method but nothing changed (expected because that error is only triggered when the tree is full which is not our case). I did some tests and found that the issue manifests when the `append` function calls the `combine` method of `Node`, so everything works fine as long as one of the left or right field of the tree are empty but crashes otherwise. 

So as a workaround, this function is now called separately and as a host function if we are in the wasm `transfer`,

shared/src/ledger/native_vp/masp.rs
```diff
@@ -161,6 +322,18 @@ where
                     return Ok(false);
                 }
             }
+
+            if !(self.valid_spend_descriptions_anchor(&shielded_tx)?
+                && self.valid_nullifiers_reveal(keys_changed, &shielded_tx)?)
```
grarco, line 327: In case `source != Address::Internal(Masp)`, should we add a check to verify that no `shielded_spends` are attached to the `Transaction`?

core/src/ledger/masp_utils.rs
murisi: This is fine. We can investigate the causes of this failure in the future for curiosity's sake.

shared/src/ledger/native_vp/masp.rs
murisi: Yes, this sounds good to me, thanks! (Maybe you can also add this to the `Boundary Conditions` section in https://specs.namada.net/masp/ledger-integration ?)

murisi (approved):
This looks good to me, thanks!

shared/src/ledger/native_vp/masp.rs
grarco: For simplicity's sake I'm addressing this in #2248

PR #2245, opened 2023-12-04
Now denominate the fee amount in wrapper headers.

murisi:
## Describe your changes
Made changes necessary to enable hardware wallets to first correctly display amounts in different tokens and second sign over them. Specifically, the following changes were made:
* Converted `WrapperTx::amount_per_gas_unit` from `Amount` to `DenominatedAmount` to enable the hardware wallet to display amount with correct precision regardless of the token.
* Modified the protocol to always align the `DenominatedAmount::denom` field before extracting the `DenominatedAmount::admount` field  so that hardware wallet users cannot be tricked into signing unrelated amounts.
* Added checked high level operations (like add, subtract, and multiply) to `DenominatedAmount` to reduce the potential for arithmetic mistakes when manipulating the significand and denomination.
* Fixed the redacting of `Transfer::amount` and `Transfer::token` for fully shielded transactions. Also fixed the bug where `Transfer::amount` and its denomination did not correspond to `Transfer::token`.

The following are known omissions from this PR:
* Ethereum bridge pool transactions: `EthereumBridgePool::amount` and `EthereumBridgePool::fee_amount` should probably be converted from `Amount` to `DenominatedAmount` to allow the hardware wallet to display quantities properly, and the protocol should safely convert them back to `Amount`s.

Namada v0.27.0


sdk/src/signing.rs
```diff
@@ -544,15 +544,15 @@ pub async fn wrap_tx<'a, N: Namada<'a>>(
                 let target = namada_core::types::masp::TransferTarget::Address(
                     fee_payer_address.clone(),
                 );
-                let fee_amount = DenominatedAmount {
+                let fee_amount = DenominatedAmount::new(
```
grarco, line 547: This variable is only used to feed the next call to `gen_shielded_transfer`. This second function though, takes a `DenominatedAmount` but only ever uses the `amount` field of it. I was wondering if we could maybe change the signature of `gen_shielded_transfer` to take an `Amount` instead

grarco (dismissed):
Left a minor comment but looks good from my side!

batconjurer (approved):

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+        storage: &impl StorageRead,
+    ) -> storage_api::Result<Amount> {
+        let denom = read_denom(storage, token)?.ok_or_else(|| {
+            storage_api::Error::SimpleMessage(
+                "No denomination found in storage for the given token",
+            )
+        })?;
+        self.increase_precision(denom)
+            .map_err(storage_api::Error::new)
+            .map(|x| x.amount)
```
sug0, line 423: `self.scale(...)` since you have this method available

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+        self.increase_precision(Denomination(denom.into()))
+            .map(|x| x.amount)
+    }
+
+    /// Checked multiplication. Returns `None` on overflow.
+    pub fn checked_mul(&self, rhs: DenominatedAmount) -> Option<Self> {
+        let amount = self.amount.checked_mul(rhs.amount)?;
+        let denom = self.denom.0.checked_add(rhs.denom.0)?.into();
+        Some(Self { amount, denom })
+    }
```
sug0, line 441: add unit tests for this

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+            lhs = lhs.increase_precision(rhs.denom).ok()?;
+        } else {
+            rhs = rhs.increase_precision(lhs.denom).ok()?;
+        }
+        let amount = lhs.amount.checked_sub(rhs.amount)?;
+        Some(Self {
+            amount,
+            denom: lhs.denom,
+        })
+    }
```
sug0, line 456: add unit tests for this

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+            lhs = lhs.increase_precision(rhs.denom).ok()?;
+        } else {
+            rhs = rhs.increase_precision(lhs.denom).ok()?;
+        }
+        let amount = lhs.amount.checked_add(rhs.amount)?;
+        Some(Self {
+            amount,
+            denom: lhs.denom,
+        })
+    }
```
sug0, line 471: add unit tests for this

core/src/types/token.rs
```diff
@@ -320,12 +318,17 @@ impl From<Denomination> for u8 {
...
     /// The mantissa
-    pub amount: Amount,
+    amount: Amount,
     /// The number of decimal places in base ten.
-    pub denom: Denomination,
+    denom: Denomination,
 }
 
 impl DenominatedAmount {
+    /// Make a new denominated amount
```
sug0, line 327: I would rework the docstr here to make it a bit more clear that what we're returning is `amount * 10^denom`

core/src/types/token.rs
```diff
@@ -320,12 +318,17 @@ impl From<Denomination> for u8 {
...
+    pub fn new(amount: Amount, denom: Denomination) -> Self {
```
sug0, line 328:
```suggestion
    pub const fn new(amount: Amount, denom: Denomination) -> Self {
```

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+        Some(Self {
+            amount,
+            denom: lhs.denom,
+        })
+    }
+
+    /// Returns the significand of this number
+    pub fn amount(&self) -> Amount {
+        self.amount
+    }
```
sug0, line 476: this method might be a bit confusing, since it doesn't specify that the returned amount is not multiplied by the denomination. maybe mention this in the docstr.

also, I think this can become a `const fn`

core/src/types/token.rs
```diff
@@ -401,6 +404,81 @@ impl DenominatedAmount {
...
+
+    /// Returns the denomination of this number
+    pub fn denom(&self) -> Denomination {
```
sug0, line 479:
```suggestion
    pub const fn denom(&self) -> Denomination {
```

tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -137,11 +137,8 @@ async fn test_roundtrip_eth_transfer() -> Result<()> {
     assert_eq!(dai_supply, Some(transfer_amount));
 
     // let's transfer them back to Ethereum
-    let amount = token::DenominatedAmount {
-        amount: transfer_amount,
-        denom: 0u8.into(),
-    }
-    .to_string();
+    let amount =
+        token::DenominatedAmount::new(transfer_amount, 0u8.into()).to_string();
```
sug0, line 141: I wouldn't bother touching the eth bridge e2e tests, honestly. they're still disabled

sug0 (changes requested):

PR #2248, opened 2023-12-06
Convert descriptions anchor validation

### LINKED ISSUE #1373: Missing MASP Transaction Validation

grarco:
## Describe your changes

Closes #1373.

Modifies `update_allowed_conversions` to update the anchor in storage at every epoch.
Updates the masp vp to check the validity of the convert descriptions' anchors and improves how the masp vp validates the changed keys.
Also updates the VP to run stricter checks on the sapling bundle composition (addressing https://github.com/anoma/namada/pull/2244#discussion_r1417456423)

#2244 

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -1,9 +1,13 @@
 //! Implementation of the `FinalizeBlock` ABCI++ method for the Shell
 
 use data_encoding::HEXUPPER;
+use masp_primitives::merkle_tree::CommitmentTree;
+use masp_primitives::sapling::Node;
+use masp_proofs::bls12_381;
 use namada::core::ledger::inflation;
 use namada::core::ledger::masp_conversions::update_allowed_conversions;
 use namada::core::ledger::pgf::ADDRESS as pgf_address;
+use namada::core::types::storage::KeySeg;
 use namada::ledger::events::EventType;
 use namada::ledger::gas::{GasMetering, TxGasMeter};
 use namada::ledger::parameters::storage as params_storage;
@@ -17,9 +21,13 @@ use namada::proof_of_stake::{
     find_validator_by_raw_hash, read_last_block_proposer_address,
     read_pos_params, read_total_stake, write_last_block_proposer_address,
 };
+use namada::types::address::MASP;
 use namada::types::dec::Dec;
 use namada::types::key::tm_raw_hash_to_string;
 use namada::types::storage::{BlockHash, BlockResults, Epoch, Header};
+use namada::types::token::{
+    MASP_NOTE_COMMITMENT_ANCHOR_PREFIX, MASP_NOTE_COMMITMENT_TREE,
+};
 use namada::types::transaction::protocol::{
     ethereum_tx_data_variants, ProtocolTxType,
 };
@@ -556,6 +564,23 @@ where
 
         tracing::info!("{}", stats);
         tracing::info!("{}", stats.format_tx_executed());
+
+        // Update the MASP commitment tree anchor
+        let tree_key = Key::from(MASP.to_db_key())
+            .push(&MASP_NOTE_COMMITMENT_TREE.to_owned())
+            .expect("Cannot obtain a storage key");
+        let updated_tree: CommitmentTree<Node> = self
+            .wl_storage
+            .read(&tree_key)?
+            .expect("Missing note commitment tree in storage");
+        let anchor_key = Key::from(MASP.to_db_key())
+            .push(&MASP_NOTE_COMMITMENT_ANCHOR_PREFIX.to_owned())
+            .expect("Cannot obtain a storage key")
+            .push(&namada::core::types::hash::Hash(
+                bls12_381::Scalar::from(updated_tree.root()).to_bytes(),
+            ))
+            .expect("Cannot obtain a storage key");
+        self.wl_storage.write(&anchor_key, ())?;
 
         if update_for_tendermint {
             self.update_epoch(&mut response);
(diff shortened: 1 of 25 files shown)


batconjurer (approved):

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -557,6 +566,25 @@ where
         tracing::info!("{}", stats);
         tracing::info!("{}", stats.format_tx_executed());
 
+        // Update the MASP commitment tree anchor if the tree was updated
```
cwgoes, line 569: What is the purpose of writing to one key in the write log then writing the anchor to another key? Why not just write to the anchor key directly?

core/src/ledger/masp_conversions.rs
```diff
@@ -454,6 +459,19 @@ where
     // obtained
     wl_storage.storage.conversion_state.tree =
         FrozenCommitmentTree::merge(&tree_parts);
+    // Update the anchor in storage
+    let anchor_key = Key::from(MASP.to_db_key())
+        .push(&MASP_CONVERT_ANCHOR_KEY.to_owned())
+        .expect("Cannot obtain a storage key");
+    wl_storage.write(
+        &anchor_key,
+        crate::types::hash::Hash(
```
cwgoes, line 468: We can abstract all of this out into a `hash` function on the tree type, perhaps?

core/src/ledger/masp_utils.rs
```diff
@@ -0,0 +1,113 @@
...
+    let masp_addr = MASP;
+    let head_tx_key = Key::from(masp_addr.to_db_key())
+        .push(&HEAD_TX_KEY.to_owned())
+        .expect("Cannot obtain a storage key");
+    let current_tx_idx: u64 =
+        ctx.read(&head_tx_key).unwrap_or(None).unwrap_or(0);
+    let current_tx_key = Key::from(masp_addr.to_db_key())
+        .push(&(TX_KEY_PREFIX.to_owned() + &current_tx_idx.to_string()))
+        .expect("Cannot obtain a storage key");
+    // Save the Transfer object and its location within the blockchain
```
cwgoes, line 86: Do we really need to store the entire transaction data? Seems wasteful. Also, can't this be queried from Tendermint?

core/src/ledger/masp_utils.rs
```diff
@@ -0,0 +1,113 @@
...
+pub fn handle_masp_tx(
+    ctx: &mut (impl StorageRead + StorageWrite),
+    transfer: &Transfer,
+    shielded: &Transaction,
+) -> Result<()> {
+    let masp_addr = MASP;
+    let head_tx_key = Key::from(masp_addr.to_db_key())
+        .push(&HEAD_TX_KEY.to_owned())
+        .expect("Cannot obtain a storage key");
+    let current_tx_idx: u64 =
```
cwgoes, line 81: If we are going to write transactions to storage like this, all of this incrementing current tx index logic should be abstracted into a common function.

cwgoes (changes requested):

apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: I'm not sure I'm getting your point. The masp tx appends to the merkle tree in storage and here we just update the anchor at the end of the block. We could make the tx update the anchor itself and update the vp to validate it but to my understanding it would be of little benefit because no one could use those intermediate anchors. This way we can do the update only once and without the need for validation.

core/src/ledger/masp_conversions.rs
grarco: Ah the `FrozenCommitmentTree` is imported from the masp crate but we can probably refactor this part of the code in other ways

core/src/ledger/masp_utils.rs
grarco: We don't need this data in protocol but we need it for the clients to scan transactions and generate a valid internal state to produce future transactions. I agree with you that this could bloat the storage pretty quick (together with the nullifier set), but I think there are a few concerns with querying this data from comet:

- I'm not sure we can assume that the blocks will be always available (in case of a chain restart for example if we decided to prune history)
- Requesting blocks from full nodes could take a long time/bandwidth
- The time to process the blocks filtering out only masp transactions could be long

Still we might prefer this solution over writing to storage. Wdyt? Alternatively we could think about a way to prune transactions from storage when they are not useful anymore (e.g. all the output descriptions have been spent)

apps/src/lib/node/ledger/shell/finalize_block.rs
cwgoes: They could technically use it in several transactions sent in sequence, but this is fine for now I think.

core/src/ledger/masp_utils.rs
cwgoes: We should be able to use the indexer to pre-filter for MASP transaction data so that clients need only download and scan that. It's true that we need to keep around data after chain upgrades, but this doesn't seem too difficult.

cwgoes (approved):

PR #2255, opened 2023-12-08
 Add client command to dry run genesis files

### LINKED ISSUE #761: Validators should check that they can compile wasm at startup.

batconjurer:
A new client command has been added that takes a set of pre-genesis template files, validates them, and runs them through `init_chain`. All errors are collected into a report. 

This also closes #761


tzemanovic (approved):

apps/src/lib/cli.rs
```diff
@@ -6936,6 +6963,32 @@ pub mod args {
...
+            Self { path, wasm_dir }
+        }
+
+        fn def(app: App) -> App {
+            app.arg(
+                PATH.def()
+                    .help("Path to the directory with the template files."),
+            )
+            .arg(WASM_DIR.def().help(
+                "Can optionally provide a wasm directory as part of verifying \
```
sug0, line 6986: nit: the wording here should be more consistent with the rest of the CLI subcmd help messages. something along the lines of "Optional wasm directory to blah blah ..."

apps/src/lib/cli/client.rs
```diff
@@ -677,6 +678,39 @@ impl CliApi {
...
+                    genesis
+                        .write_toml_files(
+                            &test_dir.path().join(chain_id.to_string()),
+                        )
+                        .unwrap();
+                    ledger::test_genesis_files(
+                        config.ledger,
+                        genesis,
+                        args.wasm_dir,
+                    );
```
sug0, line 712: perhaps move this to `client/utils.rs`?

apps/src/lib/node/ledger/mod.rs
```diff
@@ -739,6 +739,77 @@ async fn maybe_start_ethereum_oracle(
...
+            validator: ValidatorParams {
+                pub_key_types: vec![],
+            },
+            version: None,
+            abci: Default::default(),
+        },
+        validators: vec![],
+        app_state_bytes: Default::default(),
+        initial_height: 0u32.into(),
+    };
```
sug0, line 791: this init chain request could be moved inside  of `shell::InitChainValidation::new`, IMO, since the caller isn't really interested in building a request comprised mostly of values that will be ignored anyway. the only parameter required seems to be the chain id, which can replace the init chain request fn arg. the genesis time can be acquired from inside the fn.

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -169,71 +242,114 @@ where
...
+                            .map_err(|e| {
+                                Panic::ReadingWasm(vp_filename, e.to_string())
+                            }),
+                    )
+                    .or_placeholder(Some(vec![]))?
+                    .unwrap();
+                v.insert(code).clone()
+            }
+        };
+        self.proceed_with(code)
```
sug0, line 284: I'm getting flashbacks :D

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -483,44 +621,251 @@ where
...
+        if self.is_ok() {
+            return Ok(());
+        }
+        if !self.panics.is_empty() {
+            panic!(
+                "Namada ledger failed to initialize due to: {}",
+                self.panics.remove(0)
+            );
+        } else {
+            Err(self.errors.remove(0))
```
sug0, line 802: don't we want a summary of all possible errors? so, something that cancatenates them into a single string, or a list (newline separated entries)

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -483,44 +621,251 @@ where
...
+            warnings: vec![],
+            dry_run,
+        }
+    }
+
+    /// Print out a report of errors encountered while dry-running
+    /// genesis files
+    pub fn report(&self) {
+        use color_eyre::owo_colors::{OwoColorize, Style};
+        let separator: String = ["="; 60].into_iter().collect();
```
sug0, line 707: wtf. lol nice, learned something new today

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -553,4 +898,218 @@ mod test {
             storage_state.iter(),
         );
     }
+
```
sug0, line 901: can we also test tx signature validation, on bonds and validator account txs?

sug0 (commented):
couple of comments

apps/src/lib/node/ledger/shell/init_chain.rs
batconjurer: The current design prints out all errors when dry-running. In "production", it errors out with the first error which was the original behavior.

apps/src/lib/node/ledger/shell/init_chain.rs
batconjurer: `init_chain` takes finalized genesis files. Those files have already had their signatures checked by the finalization procedure and can be checked by the existing validate genesis command. No signature checks are performed at all during initialization.

sug0 (dismissed):

tzemanovic:
I noticed on merging there were artifacts from `make test-unit` - I fixed it in [27bd6fb](https://github.com/anoma/namada/pull/2255/commits/27bd6fbf0e5142e1a9187fc1e4470a1d34681d28)

Issue #2135, opened 2023-11-09
Wallet cleanup brainstorming


cwgoes:
Clean up the Namada CLI wallet, both internally and externally.

Just using the wallet, I observe some UX issues:
- Many commands seem to require genesis files which I wouldn't expect to need them, e.g.

```
namada wallet key list
```

=>

```
The application panicked (crashed).
Message:  Missing genesis files:
   0: Couldn't read Validity predicates config file from /home/cwgoes/.local/share/namada/e2e-test.a23671b0c1d2dc60d75d8/validity-predicates.toml
   1: No such file or directory (os error 2)

Location:
   apps/src/lib/config/genesis/toml_utils.rs:11
```

Why do we need genesis files for the wallet at all? The wallet should only be dealing with its own database and configuration, in general. If we're using genesis files to look up aliases or such, this is fine, but we should gracefully degrade (perhaps with a warning) if they're not found instead of crashing. 

- The relationship between key and address generation is pretty unclear to me. I understand addresses as simply being derived from public keys - if we create a new key, we should also know the address. Why do we have a whole separate subcommand for address management (with many of the same sub-sub-commands as for keys)? Can we get rid of this, and simply make the key commands compute & store the address as well?
- Is there a strong reason to separate MASP and regular key generation and management? I think it might be easier for users if we simply have a `--masp` option or such when generating a new key (and generating payment addresses only works when used with a spending key). This may require us to implement [ZIP32](https://zips.z.cash/zip-0032) (but there is almost certainly a library which we can import).
- The wallet should support an arbitrary-message `sign` operation which does not depend on any transaction-related code (just signs binary data with a key in the wallet).

cwgoes:
Discussed quickly with @tzemanovic who noted that the `address` subcommand also deals with established addresses and allows for alias creation. Agreed to split out all of this ledger-state-dependent logic into `namada client`, and keep `namada wallet` for key logic only (which should not be dependent on genesis state or any particular network). That should provide a cleaner separation of concerns and allow us to simplify the wallet UX.

tzemanovic:
just to add, in 0.25 we added `--pre-genesis` flag to wallet cmds. If the wallet is independent from chain then we probably won't need it anymore

karbyshev:
After the discussion with @murisi, we suggest the following structure for namadaw CLI commands:
- `gen [--masp] [--alias] [--hd-path]` generates new key
  + default, _secret_ key for transparent pool, optionally HD
[...]

PR #2260, opened 2023-12-10
CLI revamp - 0.28.0

karbyshev:
This PR introduces a revamped structure of the wallet CLI suggested in #2135. The changes are based on v0.28.0.

The following changes were additionally introduced.

- For consistency, `--alias` argument is made obligatory for `gen` / `derive` commands.
- Raw (non-HD) key generation functionality is restored, hidden under `--raw` flag.
  Example: `namadaw gen --raw --alias key1`
- Fix alias string normalization
- Implemented `export` for MASP spending keys
- Implemented `import` for secret keys / MASP spending keys
- Implement `add` for secret keys
- Simplified interface of `add` command. Example: `namadaw add --alias my_entry --value val`. The supported values: 
  + secret key
  + public key
  + namada address
  + MASP spending key / viewing key / payment address
- Introduced bi-map for the alias-payment address book
- Implemented alias lookup by MASP payment key

Future work:

- implement ZIP32 functionality for spending key generation
- use `Alias` struct everywhere in order to force alias normalization?

### DIFF
--- a/.changelog/unreleased/features/2260-wallet-cli-revamping-main-rebased.md
+++ b/.changelog/unreleased/features/2260-wallet-cli-revamping-main-rebased.md
@@ -1,0 +1,7 @@
+- The wallet CLI structure has been significantly reworked and simplified.
+  Alias argument is now obligatory for key generation / derivation
+  commands. Feature of raw (non-HD) key generation has been restored,
+  which was removed in the previous release. Key export / import
+  functionality for both transparent and shielded key kinds has been
+  implemented. Additionally, several other improvements have been made.
+  ([\#2260](https://github.com/anoma/namada/pull/2260))
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -6111,57 +6111,54 @@ pub mod args {
 
     impl Args for KeyAddressList {
         fn parse(matches: &ArgMatches) -> Self {
-            let decrypt = DECRYPT.parse(matches);
             let transparent_only = TRANSPARENT.parse(matches);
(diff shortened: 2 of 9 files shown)


cwgoes (changes requested):
ACK modulo notes from testing:

- we should prevent generating two keys/addresses (including shielded & transparent) with the same alias
- `list` should show both shielded and transparent by default, unless filters are provided
- `list` should show both shielded keys and addresses, unless filters are provided
- `export` should support exporting shielded keys without passing `--shielded` (should only need the alias)
- `export` followed by `import` doesn't work - needs to be tested
- add function to remove keys
- test it more, recruit @bengtlofgren to help test

karbyshev:
> ACK modulo notes from testing:
> 
> * we should prevent generating two keys/addresses (including shielded & transparent) with the same alias
> * `list` should show both shielded and transparent by default, unless filters are provided
> * `list` should show both shielded keys and addresses, unless filters are provided
> * `export` should support exporting shielded keys without passing `--shielded` (should only need the alias)
> * `export` followed by `import` doesn't work - needs to be tested
> * test `find` and make sure it isn't case-dependent, also make sure it doesn't require the password
> * add function to remove keys
> * test it more, recruit @bengtlofgren to help test

All are fixed.

Fraccaman:
great work @karbyshev! Do you think this PR is ready?

karbyshev:
@Fraccaman Thanks! I think it can be merged, but additional testing could be helpful.
ZIP32 keys are not implemented yet, and should be implemented in a separate PR, imho.

karbyshev:
ZIP32 key derivation is implemented in #2417

PR #2310, opened 2023-12-19
Wait for genesis time

batconjurer:
## Describe your changes
Make the ledger wait for genesis before starting up any processes

v0.28

### DIFF
--- a/apps/src/lib/node/ledger/mod.rs
+++ b/apps/src/lib/node/ledger/mod.rs
@@ -775,7 +775,7 @@ async fn sleep_until(time: DateTimeUtc) {
     };
     tokio::select! {
         _ = shutdown_signal => {
-            tracing::info!("Shutdown signal receiving, shutting down");
+            tracing::info!("Shutdown signal received, shutting down");
         }
         _ = sleep => {
             tracing::info!("Genesis time reached, starting ledger");
--- a/apps/src/lib/node/ledger/mod.rs
+++ b/apps/src/lib/node/ledger/mod.rs
@@ -249,12 +249,14 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
     // from Tendermint
     let mut spawner = AbortableSpawner::new();
 
-    // Start Tendermint node
-    let tendermint_node = start_tendermint(&mut spawner, &config);
     // wait for genesis time
     let genesis_time = DateTimeUtc::try_from(config.genesis_time.clone())
         .expect("Should be able to parse genesis time");
-    sleep_until(genesis_time).await;
+    if let std::ops::ControlFlow::Break(_) = sleep_until(genesis_time).await {
+        return;
+    }
+    // Start Tendermint node
+    let tendermint_node = start_tendermint(&mut spawner, &config);
 
     // Start oracle if necessary
     let (eth_oracle_channels, eth_oracle) =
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/mod.rs
```diff
@@ -748,3 +754,31 @@ async fn maybe_start_ethereum_oracle(
 fn spawn_dummy_task<T: Send + 'static>(ready: T) -> task::JoinHandle<T> {
     tokio::spawn(async { std::future::ready(ready).await })
 }
+
+/// Sleep until the genesis time if necessary.
+async fn sleep_until(time: DateTimeUtc) {
+    let shutdown_signal = install_shutdown_signal();
```
sug0, line 760: this overrides another shutdown signal no? not sure if ctrl-c won't do anything afterwards

apps/src/lib/node/ledger/mod.rs
```diff
@@ -249,6 +251,10 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
 
     // Start Tendermint node
     let tendermint_node = start_tendermint(&mut spawner, &config);
+    // wait for genesis time
+    let genesis_time = DateTimeUtc::try_from(config.genesis_time.clone())
+        .expect("Should be able to parse genesis time");
+    sleep_until(genesis_time).await;
```
sug0, line 257: won't this cause issues, since cometbft will try to connect to the abci app and crash? should we sleep up to genesis time before starting cometbft instead?

sug0 (commented):
a couple of potential issues

apps/src/lib/node/ledger/mod.rs
batconjurer: When I tested locally, this worked fine. And reading the code, creates another shutdown handler after this one that the rest of the process use. However, to shutdown, I think you have to hit ctrl-c twice right now. I'll fix that.

apps/src/lib/node/ledger/mod.rs
batconjurer: When I tested locally, it seemed fine, but I can easily just move this above the part where we start cometbft

apps/src/lib/node/ledger/mod.rs
sug0: hm this still seems incorrect, because every time we call `install_shutdown_signal`, we spawn a new async task listening for interrupt signals. would it be better to expose the `ShutdownSignal` handler from the `AbortableSpawner`, and poll it from `tokio::select!()` instead?

batconjurer: What's the problem with spawning an async task? No other shutdowns are install at the point and the task gets dropped at the completion of this call.

sug0: `install_shutdown_signal` internally calls `tokio::spawn`, to listen to interrupt signals from the OS. by the end of `sleep_until`, if we return with `ControlFlow::Continue`, the process will have two async tasks running in the background listening for OS signals. I think it's unnecessary, bc/ we can re-use the handler inside of the `AbortableSpawner` to fetch incoming OS signals

sug0: thx!

sug0 (approved):

PR #2311, opened 2023-12-19
Fix ibc-gen-shielded to receive a non-Namada token

### LINKED ISSUE #2308: ibc-gen-shielded transactions aren't working as expected
Every time I try to send funds from Namada's MASP to Cosmos testnet, I got this:

```
namada client ibc-gen-shielded --target cosmos1ay642tntyt36kpgu7x2ynkxptq0gar4qenf7kt --token NAM --amount 10 --channel-id channel-0 --port-id transfer
Message:  called `Result::unwrap()` on an `Err` value: "Unknown payment address cosmos1ay642tntyt36kpgu7x2ynkxptq0gar4qenf7kt"
```

### LINKED ISSUE #2309: sending funds from Cosmos testnet to Namada's MASP isn't working as expected
Every time I try to do this, I got this at Hermes logs:

```
2023-12-18T21:21:06.972836Z ERROR ThreadId(10) worker.batch{chain=public-testnet-15.0dacadb8d663}:supervisor.handle_batch{chain=public-testnet-15.0dacadb8d663}:supervisor.process_batch{chain=public-testnet-15.0dacadb8d663}:worker.packet.cmd{src_chain=public-testnet-15.0dacadb8d663 src_port=transfer src_channel=channel-0 dst_chain=theta-testnet-001}:relay{odata=0a880e3f ->Destination @0-1899; len=1}:send_messages_and_wait_check_tx{chain=theta-testnet-001 tracking_id=0a880e3f}:send_tx_with_account_sequence_retry{chain=theta-testnet-001 account.sequence=2140}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 1: acknowledge packet verification failed: failed packet acknowledgement verification for client (07-tendermint-2649): chained membership proof contains nonexistence proof at index 0. If this is unexpected, please ensure that proof was queried from a height that contained the value in store and was queried with the correct key. The key used: /ibc/acks%2Fports%2Ftransfer%2Fchannels%2Fchannel-0%2Fsequences%2F6: invalid proof [cosmos/ibc-go/v4@v4.4.2/modules/core/23-commitment/types/merkle.go:255] With gas wanted: '0' and gas used: '427211' ", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "19396293"} 
```

While trying to dig into incoming transactions to Namada, I found that only transactions involving client updates are able to be fetched:
```
namadac tx-result --tx-hash ACD19136EB018D53287DC14FF4AA7CA2FDC49FA88DD117F6D03AEED24BF3E59B
Transaction was applied with result: {
  "info": "Transaction is valid. Gas used: 2270; Changed keys: #tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4/clients/07-tendermint-24/clientState, #tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4/clients/07-tendermint-24/consensusStates/0-19396280, #tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4/clients/07-tendermint-24/update_height, #tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4/clients/07-tendermint-24/update_timestamp; VPs result:  Accepted: tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4;",
  "log": "",
  "height": "1892",
  "hash": "ACD19136EB018D53287DC14FF4AA7CA2FDC49FA88DD117F6D03AEED24BF3E59B",
  "code": "0",
  "gas_used": "2270",
  "initialized_accounts": []
}
```
And others:
```
namadac tx-result --tx-hash 4A3C98BF786823A1F63FE5752BE646A95828BC0FF2A802209A48E53B0477280F
Message:  no entry found for key
Location: /.../namada/sdk/src/rpc.rs:657
```

yito88:
## Describe your changes
closes #2308
closes #2309
`ibc-gen-shielded` should get not only a Namada token but also a non-Namada token to transfer it from non-Namada chains

`v0.28.1`

### DIFF
--- a/.changelog/unreleased/bug-fixes/2308-fix-ibc-gen-shielded.md
+++ b/.changelog/unreleased/bug-fixes/2308-fix-ibc-gen-shielded.md
@@ -1,0 +1,2 @@
+- Non-Namada token can be given to ibc-gen-shielded
+  ([\#2308](https://github.com/anoma/namada/issues/2308))
--- a/sdk/src/rpc.rs
+++ b/sdk/src/rpc.rs
@@ -1163,12 +1163,33 @@ pub async fn format_denominated_amount(
 /// Look up the IBC denomination from a IbcToken.
 pub async fn query_ibc_denom<N: Namada>(
     context: &N,
-    token: &Address,
+    token: impl AsRef<str>,
     owner: Option<&Address>,
 ) -> String {
-    let hash = match token {
-        Address::Internal(InternalAddress::IbcToken(hash)) => hash.to_string(),
-        _ => return token.to_string(),
+    let hash = match Address::decode(token.as_ref()) {
+        Ok(Address::Internal(InternalAddress::IbcToken(hash))) => {
+            hash.to_string()
+        }
+        Ok(_) => return token.as_ref().to_string(),
+        Err(_) => match namada_core::types::ibc::is_ibc_denom(token.as_ref()) {
+            Some((trace_path, base_denom)) => {
+                let base_token = context
+                    .wallet()
+                    .await
+                    .find_address(&base_denom)
+                    .map(|addr| addr.to_string())
+                    .unwrap_or(base_denom);
+                return format!("{trace_path}/{base_token}");
+            }
+            None => {
+                return context
+                    .wallet()
+                    .await
+                    .find_address(token.as_ref())
+                    .map(|addr| addr.to_string())
+                    .unwrap_or(token.as_ref().to_string());
+            }
+        },
     };
 
     if let Some(owner) = owner {
@@ -1195,5 +1216,5 @@ pub async fn query_ibc_denom<N: Namada>(
         }
     }
 
-    token.to_string()
+    token.as_ref().to_string()
 }
--- a/sdk/src/rpc.rs
+++ b/sdk/src/rpc.rs
@@ -1170,26 +1170,7 @@ pub async fn query_ibc_denom<N: Namada>(
         Ok(Address::Internal(InternalAddress::IbcToken(hash))) => {
             hash.to_string()
         }
-        Ok(_) => return token.as_ref().to_string(),
-        Err(_) => match namada_core::types::ibc::is_ibc_denom(token.as_ref()) {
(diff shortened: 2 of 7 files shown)


sdk/src/rpc.rs
```diff
@@ -1163,12 +1163,33 @@ pub async fn format_denominated_amount(
...
+                return format!("{trace_path}/{base_token}");
+            }
+            None => {
+                return context
+                    .wallet()
+                    .await
+                    .find_address(token.as_ref())
+                    .map(|addr| addr.to_string())
+                    .unwrap_or(token.as_ref().to_string());
+            }
```
sug0, line 1191: is this the correct behavior? users could have populated their wallet with an alias that might coincide with some other chain's address format. it's odd, but certainly possible. perhaps we should document in the cli the order of parsing/look-ups

.changelog/unreleased/bug-fixes/2308-fix-ibc-gen-shielded.md
```diff
@@ -0,0 +1,2 @@
+- Non-Namada token can be given to ibc-gen-shielded
```
grarco, line 1: I think we also need another changelog for the sdk updates

sdk/src/rpc.rs
yito88: Good point. We shouldn't look up the token address with the wallet of the destination chain.

When `ibc-gen-shielded` is requested, a user is trying to send the token of another chain. A user should be able to get the raw address of the token on the source chain.

And, if the token address is invalid, though the proof will be generated, the following ibc-transfer will fail.

(When `query_ibc_denom` is called from `ibc-transfer`, the token should be parsed and this `Err` branch shouldn't be reached.)

grarco (dismissed):
Looks good to me, thanks!

PR #2312, opened 2023-12-19
Modularization: core refactors

tzemanovic:
## Describe your changes

We're dividing namada_core crate into more granular low-level crates. The updated dependency graph looks like this:

test

with dev-deps (blue):

test-dev

This entails:
- [x] Add new crates from namada_core modules
- [x] Make namada_core compile
- [x] [Make each new crate compile]()
- [x] Update all the existing crate to depend on the new crates
- [x] `cargo check --tests` on each crate until passing

### Low-level
- [x] `account`
- [x] `gas`
- [x] `macros`
- [x] `parameters`
- [x] `storage`
- [x] `vp_env`
- [x] `merkle_tree`
- [x] `state`
- [x] `tx_env`
- [x] `tx_prelude`
- [x] `tx`
- [x] `vm_env`
- [x] `vote_ext`
- [x] `vp_prelude`

### Sub-systems
- [x] `governance`
- [x] `trans_token`
- [x] `ethereum_bridge`
- [x] `ibc`
- [x] `proof_of_stake`
- [x] `shielded_token`
- [x] `token`

### Other
- [x] `apps`
- [x] `encoding_spec`
- [x] `examples`
- [x] `light_sdk`
- [x] `sdk`
- [x] `shared`
- [x] `test_utils`
- [x] `tests`
- [x] `wasm`

v0.29.0 - Diff range for review: https://github.com/anoma/namada/pull/2312/files


core/src/types/sign.rs
```diff
@@ -0,0 +1,95 @@
+//! Types for signing
+
+use std::cmp::Ordering;
+
+use data_encoding::HEXUPPER;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use super::address::Address;
+use super::key::common;
```
batconjurer, line 10: It might make sense to `pub use` here. If I were a new dev and saw this file, I would assume I could fine keys and signature types in here.

core/src/types/token.rs
```diff
@@ -1,6 +1,7 @@
 //! A basic fungible token
 
 use std::cmp::Ordering;
+use std::collections::BTreeMap;
```
batconjurer, line 4: How do we decide what gets its own module and what stays under core? For example, why isn't this its own module or account not under core?

tzemanovic: we should probably move this into the new trans_token crate - there are still few files left in core that depend on it, but they can also be moved out

account/src/lib.rs
```diff
@@ -0,0 +1,47 @@
+//! Support for signature based authorization of actions on a user account
+//! using public key(s) and signature threshold (minimum number of signatures
+//! needed to authorize an action) stored on-chain.
+
+mod storage;
+mod storage_key;
+mod types;
+
+use borsh::{BorshDeserialize, BorshSerialize};
+pub use namada_core::types::account::AccountPublicKeysMap;
```
batconjurer, line 10: What's the reason for keeping this module in core as opposed to pulliing it all into here?

account/src/types.rs
```diff
@@ -1,10 +1,9 @@
-use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use namada_core::borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
```
batconjurer, line 1: I find the the name `types.rs` confusing here. This seems to contain tx data.

account/src/lib.rs
tzemanovic: Good q - there's an unfortunate cyclic dependency. `AccountPublicKeysMap` is also used in the `namada_tx` crate for tx signing and sig verification, but because `namada_account` depends on `namada_storage` and `namada_storage` depends on `namada_tx` (for `TxQueue`), we cannot import `namada_account` in `namada_tx`

account/src/types.rs
tzemanovic: Yeah, this was previously in `core/src/types/transaction/account` (and there are still some data types left over in `core/src/types/transaction`), but I went with a more generic name because the crate itself doesn't have a concept of tx - it just provides some storage operations and tx integration happens in `sdk/src/tx.rs` (construction) and `tx_init_account.rs` (consumption). It would be nice if we could provide all of these from one crate, but I think we'll need more base abstractions for those

proof_of_stake/src/pos_queries.rs
```diff
@@ -208,22 +206,11 @@ where
...
-        //
-        // we can remove this check once that's fixed
-        if self.wl_storage.storage.get_current_epoch().0 == Epoch(0) {
-            let height_offset_within_epoch = BlockHeight(1 + height_offset);
-            return current_decision_height == height_offset_within_epoch;
-        }
-
-        let fst_heights_of_each_epoch = self
-            .wl_storage
+        let pred_epochs = self
```
tzemanovic, line 209: @sug0 I rewrote this module to use `StorageRead` instead of the full `WlStorage`. I removed this hack as this was fixed in https://github.com/anoma/namada/pull/1993 - I had to update the test in https://github.com/anoma/namada/pull/2312/commits/0354bbec89d1947c1e80bf7c4255031b8bc7d766, pls check if that's right

core/src/types/account.rs
```diff
@@ -49,7 +17,7 @@ impl Account {
     Deserialize,
     Default,
 )]
-/// Holds the public key map data as a bimap for efficient querying
+/// Holds the public key map data as a bimap for efficient quering
```
batconjurer, line 20: This was actuallly correct before.

core/src/types/account.rs
```diff
@@ -80,15 +48,15 @@ impl FromIterator<common::PublicKey> for AccountPublicKeysMap {
 }
 
 impl AccountPublicKeysMap {
-    /// Retrieve a public key from the index
+    /// Retrive a public key from the index
```
batconjurer, line 51: same here

core/src/types/account.rs
```diff
@@ -80,15 +48,15 @@ impl FromIterator<common::PublicKey> for AccountPublicKeysMap {
...
     pub fn get_public_key_from_index(
         &self,
         index: u8,
     ) -> Option<common::PublicKey> {
         self.idx_to_pk.get(&index).cloned()
     }
 
-    /// Retrieve the index of a public key
+    /// Retrive the index of a public key
```
batconjurer, line 59: same here

core/src/types/eth_bridge_pool.rs
```diff
@@ -173,16 +218,26 @@ pub struct PendingTransfer {
     pub gas_fee: GasFee,
 }
 
+/// Construct a token address from an ERC20 address.
+pub fn erc20_token_address(address: &EthAddress) -> Address {
```
batconjurer, line 222: This isn't really just a bridge pool thing, it's a bit more general than that. Maybe it could go live with tokens instead

core/src/types/eth_bridge_pool.rs
```diff
@@ -173,16 +218,26 @@ pub struct PendingTransfer {
...
+    Address::Internal(InternalAddress::Erc20(*address))
+}
+
+/// Construct a NUT token address from an ERC20 address.
+pub fn erc20_nut_address(address: &EthAddress) -> Address {
```
batconjurer, line 227: same with this

core/src/types/eth_bridge_pool.rs
tzemanovic: there's unfortunately an issue with this - the `PendingTransfer::token_address` is using it, which itself is used from other crates

core/src/types/account.rs
tzemanovic: sry, must have accidentally revived it at some point, fixed in [7c0364c](https://github.com/anoma/namada/pull/2312/commits/7c0364c44a3a59c1b8a02c221bf64eac7f5beb42)

proof_of_stake/src/pos_queries.rs
tzemanovic: Ah, the `init_chain` still initializes `pred_epochs` with `BlockHeight(0)`. I'll put the hack back in for the refactor and try to fix it in a follow-up

sug0 (changes requested):
So far I've looked at:
1. The `pos_queries` and `eth_bridge_queries` changes, and I like them.
2. The general layout of the repository. The root of the project looks too cluttered now, so I suggest we move non-wasm crates to a separate directory. (`native`?? `ledger`?? idk)
3. (Follow-up of point 2). This is a somewhat counterintuitive point of criticism to the objective at hand, but perhaps we have too many crates now with very little benefit (other than smaller wasm sizes, potentially). Our main goal should be unifying the APIs into an abstraction that will prevent code changes from bubbling up the dependency graph, wrecking havoc during merges, refactors, etc. As a first step to realize that goal, this is still an excellent effort!
4. The dependency graph. I noticed `storage` depended on `tx`, which I found odd, so I looked through the code to see where it got used, and left some suggestions to clean up that dependency from the graph.

I might revisit this PR later today.

storage/src/tx_queue.rs
```diff
@@ -0,0 +1,76 @@
...
+/// add `has_valid_pow` flag for only used in testnets.
+#[derive(Debug, Clone, BorshDeserialize, BorshSerialize)]
+pub struct TxInQueue {
+    /// Wrapper tx
+    pub tx: Tx,
+    /// The available gas remaining for the inner tx (for gas accounting).
+    /// This allows for a more detailed logging about the gas used by the
+    /// wrapper and that used by the inner
+    pub gas: Gas,
+}
```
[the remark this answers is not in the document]
tzemanovic: Yeah, I actually had some issue with having these dependencies elsewhere. Moving it to `state` sounds like a good call

vp_prelude/src/lib.rs
```diff
@@ -355,6 +361,10 @@ impl<'view> VpEnv<'view> for Ctx {
         unsafe { namada_vp_charge_gas(used_gas) };
         Ok(())
     }
+
+    fn get_pred_epochs(&self) -> namada_storage::Result<storage::Epochs> {
+        todo!()
```
tzemanovic, line 366: ah, missed couple of these

storage/src/tx_queue.rs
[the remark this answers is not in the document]
tzemanovic: ah, unfortunately it's not so straight-forward as the `BlockStateRead` in `storage/src/db.rs` depends on it :(

PR #2355, opened 2024-01-03
New storage write API

brentstone:
## Describe your changes
StorageWrite API changes.

Some notes:
- Options to merklize data and add data to diffs are provided to various `write` and `delete` functions
- Specific keys that are changed (all are no longer merklized or persisted in diffs):
  - `masp/note_commitment_anchor/<hash>`
  - `masp/commitment_tree`
  - `masp/convert_anchor`
  - `masp/nullifiers/<hash>`
  - `masp/pin-<id>`
  - `ibc/clients/counter`
  - `ibc/connections/counter`
  - `ibc/channelEnds/counter`
- Data that is not written to diffs still maintain diffs for one block prior to the current height in order to maintain rollback functionality
- Various uses of `write_bytes` are abstracted abstracted to a different write API call
- Tests

v0.30.1


core/src/ledger/storage_api/mod.rs
```diff
@@ -20,6 +20,18 @@ use crate::types::storage::{
     self, BlockHash, BlockHeight, Epoch, Header, TxIndex,
 };
 
+/// Write actions to help commit_block determine what parts of the storage to
+/// update
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum WriteActions {
```
tzemanovic, line 26: nit, perhaps better `WriteOpts`

core/src/ledger/storage_api/mod.rs
```diff
@@ -106,24 +118,65 @@ pub trait StorageRead {
...
     fn write<T: BorshSerialize>(
         &mut self,
         key: &storage::Key,
         val: T,
     ) -> Result<()> {
         let bytes = val.serialize_to_vec();
-        self.write_bytes(key, bytes)
+        self.write_bytes(key, bytes, WriteActions::All)
+    }
+
```
tzemanovic, line 130: how about we keep `write_bytes` as it (defaulting to `WriteActions::All` and instead only add `write_bytes_with_opts` and `write_with_opts`?

core/src/ledger/storage_api/mod.rs
tzemanovic: also, maybe this would be better as
```rust
 struct WriteOpts {
    pub modify_merkle_tree: bool,
    pub write_diffs: bool,
}

impl Default for WriteOpts {
    fn default() -> Self {
        Self { modify_merkle_tree: true, write_diffs: true }
    }
}
```

tzemanovic: even better we could use e.g. https://crates.io/crates/bitflags

tx_prelude/src/lib.rs
```diff
@@ -245,6 +248,15 @@ impl StorageWrite for Ctx {
         unsafe { namada_tx_delete(key.as_ptr() as _, key.len() as _) };
         Ok(())
     }
+
+    // TODO: is this the correct thing to do here?
```
tzemanovic, line 252: I'm not too sure, but we might need to be able to choose this from wasm txs too, then we might need to pass `WriteActions` back to host

core/src/ledger/storage_api/mod.rs
brentstone: done in dd9d661621a428b60434aabba4c365fc9aa7597f

core/src/ledger/storage_api/mod.rs
brentstone: done in bd9265b375722fdb25486849c639690c81c64a3a

shared/src/ledger/native_vp/masp.rs
```diff
@@ -116,6 +116,7 @@ where
         keys_changed: &[&Key],
         transaction: &Transaction,
     ) -> Result<bool> {
+        println!("\nCHECKING VALID NULLIFIERS REVEAL\n");
```
grarco, line 119: Just make sure to remove these prints before merging

benches/txs.rs
```diff
@@ -648,6 +648,8 @@ fn become_validator(c: &mut Criterion) {
             || {
                 let mut shell = BenchShell::default();
                 // Initialize the account to be able to use it
+                // TODO: what write actions are desired here?? can this be turn
+                // into `write` from `write_bytes` anyway?
```
grarco, line 652: Ah yes I think we can

shared/src/ledger/native_vp/masp.rs
brentstone: done in a1143ab7f88fa58cce305c6981dfdb6924887476

tx_prelude/src/lib.rs
brentstone: done in b7b1921cf0a8102a73bc75104dfbb29735a38a8c

shared/src/ledger/ibc/mod.rs
```diff
@@ -18,6 +18,8 @@ where
     // In ibc-go, u64 like a counter is encoded with big-endian:
     // https://github.com/cosmos/ibc-go/blob/89ffaafb5956a5ea606e1f1bf249c880bea802ed/modules/core/04-channel/keeper/keeper.go#L115
 
+    // TODO: what kinds of write actions are desired for data below?
```
yito88: These counters don't need to be merklized or diffs

shared/src/ledger/native_vp/masp.rs
```diff
@@ -178,6 +179,7 @@ where
         &self,
         transaction: &Transaction,
     ) -> Result<bool> {
+        println!("\nCHECKING VALID NOTE COMMITMENT\n");
```
yito88: Are debug codes left in this file?

yito88 (commented):
Overall, it looks good to me. I left some comments

shared/src/ledger/native_vp/masp.rs
brentstone: removed in a1143ab7f88fa58cce305c6981dfdb6924887476

shared/src/ledger/ibc/mod.rs
brentstone: Is it ok to encode them with Borsh? It looks like they are encoded a different way when initialized, but when incremented encoded with Borsh

yito88: oh, yes, it's OK to encode them with Borsh. thanks

core/src/ledger/storage_api/mod.rs
```diff
@@ -21,6 +22,34 @@ use crate::types::storage::{
...
+// /// update
+// #[derive(Clone, Debug, PartialEq, Eq)]
+// pub enum WriteOpts {
+//     /// All
+//     All,
+//     /// Don't add to merkle tree
+//     NoMerkl,
+//     /// Don't add to diffs or merkle tree
+//     NoDiffsOrMerkl,
+// }
```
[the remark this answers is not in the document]
brentstone: done in 8ce00d9d49876785de0c8e35a82b902e49f60697

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -1181,6 +1182,24 @@ impl DB for RocksDB {
...
 
+    fn read_diffs_val(
+        &self,
+        key: &Key,
+        height: BlockHeight,
+        is_old: bool,
+    ) -> Result<Option<Vec<u8>>> {
+        let diffs_cf = self.get_column_family(DIFFS_CF)?;
+        let old_new_seg = if is_old { "old" } else { "new" };
+        let prefix = Key::from(height.to_db_key())
```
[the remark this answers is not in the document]
brentstone: done in 8ce00d9d49876785de0c8e35a82b902e49f60697

yito88 (approved):

crates/namada/src/vm/host_env.rs
```diff
@@ -2537,10 +2557,14 @@ where
     H: StorageHasher,
     CA: WasmCacheAccess,
 {
-    fn write_bytes(
+    // TODO: is it ok to not explicitly use the `_actions` and use the default
+    // behavior????
```
tzemanovic, line 2561: I think we should use `write_log.write_with_opts/delete_with_opts` otherwise the opts won't work in a IBC tx

crates/tx_prelude/src/lib.rs
```diff
@@ -247,14 +248,29 @@ impl StorageWrite for Ctx {
...
+                action.bits() as _,
             )
         };
         Ok(())
     }
 
     fn delete(&mut self, key: &storage::Key) -> namada_storage::Result<()> {
+        self.delete_with_opts(key, WriteOpts::ALL)
+    }
+
```
tzemanovic, line 260:
```suggestion
```
(implemented in the trait)

crates/storage/src/lib.rs
```diff
@@ -400,20 +468,29 @@ pub mod testing {
...
-                .write_subspace_val(self.height, key, val)
+                .write_subspace_val(self.height, key, val, action)
                 .into_storage_result()?;
             Ok(())
         }
 
         fn delete(&mut self, key: &storage::Key) -> Result<()> {
+            self.delete_with_opts(key, WriteOpts::ALL)
+        }
+
```
tzemanovic, line 486:
```suggestion
```

crates/storage/src/collections/lazy_map.rs
```diff
@@ -408,6 +408,25 @@ where
         Ok(previous)
     }
 
+    /// Same as `LazyMap::insert` above but does not add the key-value to the
```
tzemanovic, line 411: this is not needed anywhere atm right? I'd say let's skip it as I think we'd also need another version of `fn remove`. For lazy collections though I think we can do better, we could parametrize the handlers to allow to specify the `WriteOpts` at type level and apply it to all its sub-keys

crates/state/src/wl_storage.rs
```diff
@@ -720,12 +730,24 @@ mod tests {
                 | Level::BlockWriteLog(WlMod::Delete | WlMod::DeletePrefix) => {
                 }
                 Level::TxWriteLog(WlMod::Write(val)) => {
+                    // NOTE: dummy WriteActions used here because this is
+                    // only for testing
+                    // TODO: may want to use test these new features though
+                    // in here.
```
tzemanovic, line 736:
```suggestion
                    // NOTE: Will write to merkle tree and DB diff
```
I think this is fine, it's a private fn and we're only using it for `fn test_prefix_iters`. I don't think we need to test this in `wl_storage` specifically

crates/state/src/lib.rs
```diff
@@ -48,6 +48,11 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// it has 2 blocks delay on validator set update.
 pub const EPOCH_SWITCH_BLOCKS_DELAY: u32 = 2;
 
+/// Returns true if the given key is one whose data is not merklized
+// pub fn is_key_not_merklized(key: &Key) -> bool {
+//     is_masp_key(key) || is_ibc_counter_key(key)
+// }
+
```
tzemanovic, line 55: just a note here to not forget to remove this and the test below if we don't use it

crates/storage/src/collections/lazy_map.rs
tzemanovic: https://github.com/anoma/namada/pull/2355/commits/eea0a55f079fbb9844446e009553214cdc6f2c98

tzemanovic (dismissed):

crates/apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -231,14 +235,38 @@ impl RocksDB {
...
                 .join(key)
                 .to_string();
             self.0
                 .put_cf(cf, new_val_key, new_value)
                 .map_err(|e| Error::DBError(e.into_string()))?;
         }
+
+        // If not persisting the diffs, remove the existing diffs from two block
+        // heights earlier than `height`
+        if !persist_diffs {
```
grarco, line 249: I don't think this works. This deletes a key with height `height - 2`. But say that we write a non-diff key at height 5 and than the next write happens at height 7, than we'll try to delete the keys `6/old/` and `6/new` which don't exist. Instead we should be deleting the keys `5/old` and `5/new`. This can build up and we can end up with quite a few stale keys in storage

crates/apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -266,12 +295,32 @@ impl RocksDB {
...
+                .push(&NEW_DIFF_PREFIX.to_owned())
                 .map_err(Error::KeyError)?
                 .join(key)
                 .to_string();
             batch.0.put_cf(cf, new_val_key, new_value);
         }
+
+        // If not persisting the diffs, remove the existing diffs from two block
+        // heights earlier than `height`
+        if !persist_diffs {
```
grarco, line 307: Same here

cwgoes (changes requested):
The new API is definitely an improvement. A few minor nits and some questions on negative gas.

crates/state/src/write_log.rs
```diff
@@ -198,27 +205,30 @@ impl WriteLog {
...
         {
             Some(prev) => match prev {
-                StorageModification::Write { ref value } => {
-                    len as i64 - value.len() as i64
-                }
-                StorageModification::Delete => len as i64,
+                StorageModification::Write {
+                    ref value,
+                    action: _,
+                } => len as i64 - value.len() as i64,
```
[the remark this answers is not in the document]
tzemanovic: This is only used for `size_diff` and we don't do anything with it on the caller side. We don't subtract from gas cost

crates/state/src/lib.rs
```diff
@@ -427,32 +424,80 @@ where
...
+    pub fn write(
+        &mut self,
+        key: &Key,
+        value: impl AsRef<[u8]>,
+    ) -> Result<(u64, i64)> {
+        self.write_with_opts(key, value, WriteOpts::ALL)
+    }
+
+    /// Write with diffs but no merklization
+    pub fn write_without_merkl(
```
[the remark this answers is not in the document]
tzemanovic: fixed in [9ca40ed](https://github.com/anoma/namada/commit/9ca40edd6c91e5f745beeaf0d31f78de3c2d9436)

tzemanovic:
closing in favor of #2438

PR #2358, opened 2024-01-03
Extend transactions with memo sections

### LINKED ISSUE #2354: Add optional size-limited memo field to transactions
We should add an optional size-limited memo field to transactions (say, limited to 1 KB), which can be filled with arbitrary data (which may be encrypted, likely in the case of a shielded transfer). This can be used to communicate e.g. an identifier corresponding to a payment to the recipient.

/ref https://forum.namada.net/t/messages-in-transactions/375

sug0:
## Describe your changes

Closes #2354

`v0.29.0`


tzemanovic (approved):

tzemanovic:
should we also provide a way to set this from cli?

batconjurer (dismissed):

core/src/proto/types.rs
```diff
@@ -1169,6 +1224,18 @@ impl Tx {
         None
     }
 
+    /// Look-up the first memo field found in the transaction's
+    /// sections.
+    pub fn get_memo(&self) -> Option<&Memo> {
+        self.sections.iter().find_map(|sec| {
```
murisi, line 1230: It seems that transactions are slightly malleable: the transaction sections can be reordered without invalidating the signature sections (which only seem to test for the presence of certain section hashes). No? (cc @grarco ). If so, then we should either fix this malleability, or we should avoid operations like "get the first memo section in the transaction" (since the memo sections may have been reordered).

core/src/proto/types.rs
murisi, line 1: Who is supposed to sign over the memo section? It seems to me that the person who signs over the inner transaction (which is, say, a `Transfer`) should be the one that signs over the memo which describes it. No? If the inner transaction signer (in addition to the wrapper signer) is supposed to sign over the memo, then maybe the hash of the memo section should be included in `Header` (beside `code_hash` and `data_hash`)? (Doing this would also allow us to avoid making assumptions about memo section orderings since we would use this header value to directly look up the right memo section.) cc @grarco .

murisi (changes requested):
Thanks for the changes. I'm just worried about a few things: keeping the `Tx` size down so it can fit on the hardware wallet RAM, keeping the size of hardware wallet `Tx` parsers down so that they can fit in memory, assumptions about the ordering of `Memo` sections, and the necessary signers for a memo section to be accepted.

core/src/proto/types.rs
grarco: Yes, this is the way we've handled the consistency and correctness of the `Tx` so far. The inner signer signs only the `Header` but this contains commitments to the critical sections `Code` and `Data` (and data, in turn, can hold commitments to other sections if needed), so that a malleability attack on the tx cannot be successfully achieved (the tx fails either at signature verification or when loading the required sections). I'm in favor of adding another commitment for the memo section

core/src/proto/types.rs
```diff
@@ -805,6 +807,56 @@ impl borsh::BorshSchema for MaspBuilder {
...
+    Clone,
+    Debug,
+    BorshSerialize,
+    BorshDeserialize,
+    BorshSchema,
+    Serialize,
+    Deserialize,
+)]
+pub struct Memo {
+    inner: Vec<u8>,
```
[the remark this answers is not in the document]
sug0: currently, we're always submitting the full data to the hardware wallet. submitting only the commitment can be done in a separate (future) pr

.changelog/unreleased/improvements/2358-tx-memo-field.md
```diff
@@ -0,0 +1,2 @@
+- Introduce a memo field, to allow including arbitrary data (up to 1 KiB) inside
```
murisi, line 1: Are you still enforcing this 1KiB limit somewhere? Or is this no longer a requirement?

murisi (approved):
The CI seems to be failing in some parts. Do you know the cause by any chance? Otherwise this looks good to me, thanks!

PR #2359, opened 2024-01-03
Include validator avatar in their medatada

### LINKED ISSUE #2357: Add avatar url field to validator metadata
> Can we add a field to the validator metadata which corresponds to an url to an image? This is the same for validators in cosmos, which is neat as explorers or our leaderboard can later display the image.
We’ll need it for the shielded expedition, and I didn’t realize that we were missing this field until today. I assumed we had it already.

sug0:
## Describe your changes

Closes #2357

`v0.29.0`


sug0:
TODO: update regenerate signatures

adrianbrink (dismissed):

sdk/src/lib.rs
```diff
@@ -346,6 +346,7 @@ pub trait Namada: Sized + MaybeSync + MaybeSend {
         description: Option<String>,
         website: Option<String>,
         discord_handle: Option<String>,
+        avatar: Option<String>,
```
murisi, line 349: Not strictly necessary, but we could remove the optional `email`, `description`, `website`, `discord_handle`, `avatar`, and `commission_rate` arguments from here so that `new_change_metadata` is less cumbersome to call. Then we could augment the `args::MetaDataChange` builder to allow setting these properties individually. See the builder for `TxIbcTransfer` which has lots of optional parameters for instance.

murisi (approved):
Looks good to me. Thanks!

tzemanovic (approved):

sdk/src/args.rs
```diff
@@ -1487,6 +1493,54 @@ impl<C: NamadaTypes> MetaDataChange<C> {
...
+    }
+
+    /// New validator commission rate
+    pub fn commission_rate(self, commission_rate: Dec) -> Self {
+        Self {
+            commission_rate: Some(commission_rate),
+            ..self
+        }
+    }
 }
```
brentstone, line 1544: missing a `pub fn max_commission_rate_change` for a validator I think

brentstone: ah nvmd forgot this is immutable

brentstone (approved):

Issue #2343, opened 2023-12-28
MASP storage optimization


grarco:
Currently the MASP storage is managed in a suboptimal way. More specifically, the following can be observed:

1. All the keys are written under the `subspace`, meaning they are subject to diffs and merklization which is unneeded because we only need the most updated value of the masp keys and we already manage the masp merkle trees in a separate way
2. We are currently writing the `Transfer` and `Transaction` objects (plus some other metadata) into storage to allow clients to scan the history of masp transactions and construct their internal states to correctly produce future transactions
3. The `conversion_state` https://github.com/anoma/namada/blob/6c6b3bf56b689e3a47e62d2596152b87536a9fd8/core/src/ledger/masp_conversions.rs#L25-L38 is kept in memory under the `Storage` object

Issue 1 should be solved by #2265.

Issue 2 could be solved by not writing this data to storage since we already have it in the block storage. Clients could query the blocks instead (possibly with the help of an indexer), to get the same results. We should also make sure to keep the blocks around in case of a chain upgrade (see discussion at https://github.com/anoma/namada/pull/2248#discussion_r1430888110).

Issue 3 is probably not as major as the previous two. The `ConversionState` is effectively only queried by clients and updated in protocol once per epoch, so we could probably move it to storage to reduce the memory consumption (hopefully without increasing the query time too much because of the need to retrieve data from storage first)

grarco:
@cwgoes for reference

cwgoes:
Thanks for the writeup. I think we should definitely fix 1 & 2. Regarding 3, I'm not too worried about memory consumption, as long as we have a reasonable understanding of the bounds and total Namada memory usage does not exceed single-digit GBs. Do we have a sense of how large (concretely) this struct would be?

grarco:
I've tried to run some quick, approximate calculations. So the `ConversionState` struct is composed of four fields:

- `normed_inflation`, which is an `Option<u128>`, is constant at ~ 17 bytes
- `tree`, which is a `FrozenCommtimentTree` that we update once per epoch
- `tokens`, which is a map from a token alias to its address should remain low in size (`BTreeMap<String, Address>`)
- `assets`, which is a map between `AssetType` and some data about it, will actually grow once per epoch, more specifically we'll need two inserts per epoch per every unique `(token, denom)` couple, one for the old asset and one for the new asset, but the one for the old asset is actually an overwrite of the previous value, so we can count just one insert

[...]

PR #2363, opened 2024-01-04
Removes MASP record from storage

grarco:
## Describe your changes

Addresses point 2 of #2343.

Removes the masp record from storage except for the pin key. Queries are now done via the cometBFT rpc endpoints `\block` and `block_results`. Also updates key validation in the masp vp.

Adds a new `is_valid_masp_tx` attribute to appropriate tx events in `finalize_block` to allow for a fast lookup on the client side.

Updates the mock implementations of `Client` for integrations tests and benchmarks to support the new queries.

#2345  - diff for review: https://github.com/anoma/namada/pull/2363/files/ddd6368725255f0c6bd468d7e2d849f8b0a407a6..454e138f012e6c522277d3ac69b50f2a5763c982


shared/src/ledger/native_vp/masp.rs
```diff
@@ -289,77 +286,28 @@ where
...
-                && &storage_transaction == transaction) => {}
-            _ => return Ok(false),
-        }
-
         // Validate pin key
-        if let Some(key) = &transfer.key {
+        if let Some(key) = pin_key {
             let pin_key = Key::from(MASP.to_db_key())
                 .push(&(PIN_KEY_PREFIX.to_owned() + key))
                 .expect("Cannot obtain a storage key");
```
yito88, line 306: Can we replace this with `masp_pin_tx_key(key)`?

core/src/ledger/masp_utils.rs
```diff
@@ -71,42 +62,27 @@ pub fn update_note_commitment_tree(
...
     // update_note_commitment_tree(ctx, shielded)?;
     reveal_nullifiers(ctx, shielded)?;
 
     // If storage key has been supplied, then pin this transaction to it
-    if let Some(key) = &transfer.key {
-        let pin_key = Key::from(masp_addr.to_db_key())
+    if let Some(key) = pin_key {
+        let pin_key = Key::from(MASP.to_db_key())
             .push(&(PIN_KEY_PREFIX.to_owned() + key))
             .expect("Cannot obtain a storage key");
```
yito88, line 78: ditto

apps/src/lib/bench_utils.rs
```diff
@@ -725,9 +744,126 @@ impl Client for BenchShell {
...
+
+    async fn block_results<H>(
+        &self,
+        height: H,
+    ) -> Result<
+        tendermint_rpc::endpoint::block_results::Response,
+        tendermint_rpc::Error,
+    >
+    where
+        H: TryInto<namada::tendermint::block::Height> + Send,
```
batconjurer, line 819: Why the different trait bound here vs the block method above?

batconjurer, line 819: why are the different trait bound on H as opposed to how it's done in the block method? In fact, the logic for checking the latest block heeight seems like it should be the same but it's implemented two different ways

core/src/ledger/masp_utils.rs
```diff
@@ -71,42 +62,27 @@ pub fn update_note_commitment_tree(
...
     // TODO: temporarily disabled because of the node aggregation issue in WASM.
     // Using the host env tx_update_masp_note_commitment_tree or directly the
     // update_note_commitment_tree function as a  workaround instead
     // update_note_commitment_tree(ctx, shielded)?;
     reveal_nullifiers(ctx, shielded)?;
 
     // If storage key has been supplied, then pin this transaction to it
-    if let Some(key) = &transfer.key {
-        let pin_key = Key::from(masp_addr.to_db_key())
+    if let Some(key) = pin_key {
```
batconjurer, line 75: you wrote all these nice factory functions for the other keys, why not this one?

sdk/src/masp.rs
```diff
@@ -1370,36 +1513,109 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+            .block(indexed_tx.height.0 as u32)
             .await
-            .map_err(|_| {
-                Error::Other("Ill-formed epoch, transaction pair".to_string())
-            })?;
+            .map_err(|e| Error::from(QueryError::General(e.to_string())))?
+            .block
+            .data;
+
+        let tx = Tx::try_from(block[indexed_tx.index.0 as usize].as_ref())
```
batconjurer, line 1542: It seems like it might be possible that this and the code from `fetch_shielded_transfers` can be re-used here. I'm not quite sure as there is no wrapper check here, but it would be awesome if this stair-step code could be factored out into a separate function and re-used.

batconjurer (approved):

yito88 (approved):

PR #2371, opened 2024-01-09
MASP Bug Fixes

murisi:
## Describe your changes
* Refactored the validation of shielded transaction transparent outputs:
  * Now avoid assuming that transparent outputs are ordered.
  * No longer assuming that there must be transparent outputs.
  * Now ensure that total of transparent output values match the containing transaction.
* Increased the validation of transparent inputs to be similar to that of transparent outputs.
* Redefined `MaspAmount` to be an instantiation of `ValueSum`
* Reduced usage of `HashMap`s in representing numbers
* Reduced the back-and-forth encoding and decoding of `AssetType`s
* Nearly halved average MASP transaction sizes by eliminating redundant notes from two sources:
  * Zero-valued output notes emanating from parts of `u256` integer being zero
  * The note inclusion algorithm including spending notes that are not needed for the transaction
* Allow the spending of assets from previous epochs in addition to the current one. Involved:
  * Adjusting the output description creation algorithm to select assets from more than one epoch
  * Modifying the balance printing algorithms to not ignore assets from older epochs
  * Adjusted the growth of the MASP pool's balance in order to be able to fully satisfy shielded withdrawals involving trace amounts
* Fixed the printing of a shielded balance for a specific token across multiple users
  * Before the balances for all tokens across all users was being printed out (which is more than requested)
* Unified the `encode_asset_type` and `make_asset_type` functions to make sure their encodings never diverge
  * Additionally, `serialize_to_vec` is now called on only a single token representation (inside `encode_asset_type`) to ensure that conceptually different tokens never map to the same `AssetType`
* Enabled shielded transaction validation when the asset type is not in the conversion tree
Namada v0.29.0


tzemanovic (dismissed):

core/src/types/token.rs
```diff
@@ -212,6 +212,26 @@ impl Amount {
...
+        let lo = val as u64;
+        let hi = (val >> 64) as u64;
+        let lo_pos = denom as usize;
+        let hi_pos = lo_pos + 1;
+        let mut raw = [0u64; 4];
+        raw[lo_pos] = lo;
+        if hi != 0 && hi_pos >= 4 {
+            return None;
+        } else if hi != 0 {
+            raw[denom as usize + 1] = hi;
```
grarco, line 230: MInor, could we use `hi_pos` here as the index?

sdk/src/masp.rs
```diff
@@ -1690,61 +1627,124 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-            let hash = ripemd::Ripemd160::digest(sha2::Sha256::digest(
+            Some(ripemd::Ripemd160::digest(sha2::Sha256::digest(
                 target_enc.as_ref(),
-            ));
-            for (denom, asset_type) in MaspDenom::iter().zip(asset_types.iter())
+            )))
+        } else {
+            None
+        };
+        // This indicates how much more assets need to be sent to the receiver
```
batconjurer, line 1673:
```suggestion
        // This indicates how many more assets need to be sent to the receiver
```

sdk/src/masp.rs
```diff
@@ -1690,61 +1627,124 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+        // in order to satisfy the requested transfer amount.
+        let mut rem_amount = amount.amount().raw_amount().0;
+        // If we are sending to a shielded address, we may need the outgoing
+        // viewing key in the following computations.
+        let ovk_opt = spending_key.map(|x| x.expsk.ovk);
+
+        // Now handle the outputs of this transaction
+        // Loop through the value balance components and see how they which
```
batconjurer, line 1681:
```suggestion
        // Loop through the value balance components and see which
```

shared/src/ledger/native_vp/masp.rs
```diff
@@ -396,29 +367,91 @@ where
...
+                        total_in_values += token::Amount::from_masp_denominated(
+                            vin.value, *denom,
+                        );
+                    }
+                    // Maybe the asset type has no attached epoch
+                    None if unepoched_tokens.contains_key(&vin.asset_type) => {
+                        let (token, denom) = &unepoched_tokens[&vin.asset_type];
+                        // Determine what the asset type would be if it were
+                        // epoched
+                        let epoched_asset_type =
```
batconjurer, line 426: Is there no danger that this note is in the conversion with an epoch other than the checked against here? Or is that considered fine, since I guess we upconvert in the else branch?

batconjurer (commented):
What is reasoning behind allowing users to spend assets not in the current epoch? Also, shouldn't we have such cases in the integration tests?

batconjurer:
> > What is reasoning behind allowing users to spend assets not in the current epoch?
> 
> Hi! The reasoning is as follows: MASP conversions are discrete in the sense that for every `x` tokens you hold of some asset, you get `y` tokens of some other asset. Unfortunately, this means that if your shielded balance is not a multiple of `x`, then `balance % x` cannot be upgraded to the latest epoch. In the past this meant that small parts of a user's balance could disappear after epoch changes because assets from older epochs were ignored - I believe that you observed this during your genesis flow work. This PR now handles older epoch assets so that parts of a balance do not disappear.

Ah, I see. That's how we're fixing that issue.

murisi:
> Also, shouldn't we have such cases in the integration tests?

I believe the integration tests are already implicitly using notes from older epochs. But you're right, I will try to write a test to explicitly test for this behaviour. Thanks!

shared/src/ledger/native_vp/masp.rs
grarco: I believe once an asset is in the `ConversionState`, the function `update_allowed_conversions` will update the conversions at every new epoch, so probably we are guaranteed that either the asset is in the conversion state and we have the conversion with the latest epoch (the one we pass here to `encode_asset_type`) or we don't have it at all?

murisi: Yes, this is exactly how I was reasoning. Though this code could be a little bit more robust...

grarco (approved):

PR #2376, opened 2024-01-10
Removes the hardcoded masp sentinel key

grarco:
## Describe your changes

Removes the hardcoded masp key from the codebase. Updates `tx_signers` to return an empty vector of keys if the source of the tx is MASP since there's no need to sign the tx in that case.

v0.29.0


murisi (approved):
Nice one! Looks good to me.

PR #2380, opened 2024-01-11
fix fst epoch start height in apps tests

tzemanovic:
## Describe your changes

On a live chain, we typically receive `InitChain` request with `initial_height = 1`, but the apps `TestShell` was using 0.

#2312 

diff for review: https://github.com/anoma/namada/pull/2380/files/72c1fff90bf35dfc5e2f54fa7fcd5dfa708a4a86..99ac08e70c883020f30bc2b67a7c9f52917bd297

- ~~[ ] I have added a changelog~~ - test changes only

### DIFF
--- a/crates/state/src/wl_storage.rs
+++ b/crates/state/src/wl_storage.rs
@@ -270,16 +270,6 @@ where
     pub fn is_deciding_offset_within_epoch(&self, height_offset: u64) -> bool {
         let current_decision_height = self.get_current_decision_height();
 
-        // NOTE: the first stored height in `fst_block_heights_of_each_epoch`
-        // is 0, because of a bug (should be 1), so this code needs to
-        // handle that case
-        //
-        // we can remove this check once that's fixed
-        if self.storage.get_current_epoch().0 == storage::Epoch(0) {
-            let height_offset_within_epoch = BlockHeight(1 + height_offset);
-            return current_decision_height == height_offset_within_epoch;
-        }
-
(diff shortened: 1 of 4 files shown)


tzemanovic:
there's some issue with the unit tests - when I run `make test-unit` locally I get stuck on `test_oracle_reactivation`, but when I run it on its own it passes (even tried to repeat it many times)

crates/state/src/wl_storage.rs
```diff
@@ -270,16 +270,6 @@ where
...
-        // NOTE: the first stored height in `fst_block_heights_of_each_epoch`
-        // is 0, because of a bug (should be 1), so this code needs to
-        // handle that case
-        //
-        // we can remove this check once that's fixed
-        if self.storage.get_current_epoch().0 == storage::Epoch(0) {
-            let height_offset_within_epoch = BlockHeight(1 + height_offset);
-            return current_decision_height == height_offset_within_epoch;
-        }
-
```
sug0, line 282: unrelated to this PR, but I just realized this method might have an edge case. if `fst_heights_of_each_epoch` is empty, and `height_offset` is 0, the predicate should return true. this is because we're at genesis, and the first block is being decided.

sug0 (approved):
lgtm

crates/state/src/wl_storage.rs
sug0: it isn't being used anywhere under these conditions, but it's just something to keep in mind

PR #2407, opened 2024-01-16
NFT transfer over IBC

yito88:
## Describe your changes
~~Need to wait for ibc-rs upstream for ICS-721 implementation~~
- Updated `ibc-rs` to `v0.50.0` including ICS-721 impl.
- Integrated `ibc-rs`' NFT transfer module/context
- Add Namada's `MsgTransfer` and `MsgNftTransfer`
  - Each message has `ibc-rs`' `MsgTransfer`/`MsgNftTransfer` and `Option<IbcShieldedTransfer>`
- Fixed to retrieve the minted token info. and the memo(MASP transaction) from the IBC message instead of ABCI events
- Replaced `DenominatedAmount` with `Amount` in IBC handling (Also fixed `tx_prelude::token::transfer`)

`v0.31.9`


sug0 (commented):
just one comment to consider, the rest are nits

crates/ibc/src/context/nft_transfer.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn create_or_update_class_validate(
+        &self,
+        class_id: &PrefixedClassId,
+        _class_uri: Option<&ClassUri>,
+        _class_data: Option<&ClassData>,
+    ) -> Result<(), NftTransferError> {
+        match self.get_nft_class(class_id) {
+            Err(NftTransferError::NftClassNotFound) => Ok(()),
+            Err(e) => Err(e),
+            Ok(class) if class.class_id != *class_id => {
```
grarco, line 72: Given that we query for that specific `class_id` can it happen that we retrieve an `NftClass` with a different class id?

crates/sdk/src/masp.rs
```diff
@@ -735,15 +737,11 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
                 .block
                 .data;
 
-            for (idx, tx_event) in txs_results {
+            for idx in txs_results {
```
grarco, line 740: This looks good to me, unfortunately we modified this file in the last release to require tx events for other things but ibc (basically to use the changed keys instead of `Transfer`) so we'll probably need to revert some of your changes here when merging

grarco (approved):
Minor comments but looks good to me, thanks!

crates/ibc/src/context/nft_transfer.rs
```diff
@@ -0,0 +1,318 @@
...
+        &self,
+        from_account: &Self::AccountId,
+        _port_id: &PortId,
+        _channel_id: &ChannelId,
+        class_id: &PrefixedClassId,
+        token_id: &TokenId,
+        _memo: &Memo,
+    ) -> Result<(), NftTransferError> {
+        // Assumes that the class ID is prefixed with "port-id/channel-id" or
+        // has no prefix
```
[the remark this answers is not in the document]
yito88: Thank you. This comment is stale for the latest `ibc-rs`. I copied this function from `token_transfer.rs`.
Removed these comments.

crates/ibc/src/context/nft_transfer.rs
yito88: Thanks, it's redundant because `get_nft_class` always returns the same `class_id`.

PR #2416, opened 2024-01-17
Localnet script with multiple nodes

bengtlofgren:
How to use:

`python3 scripts/gen_localnet.py --num-nodes 6 --num-vals 3`


brentstone:
Nit: I think `--num-node` and `--num-vals` would be better

sug0:
Superseded by #3196

PR #2419, opened 2024-01-18
protocol tx and VP code allowlist

### LINKED ISSUE #2418: add tx whitelist check in the protocol
Currently, the tx whitelist check is done in VPs which means that a tx has to be executed first. This allows to execute arbitrary txs and even if they're going to be rejected, they can be used maliciously to e.g. trigger a panic in host env (#2401)

tzemanovic:
## Describe your changes

Added tx code allowlist at protocol level.

closes #2418

v0.30.1


crates/parameters/src/wasm_allowlist.rs
```diff
@@ -0,0 +1,42 @@
+use namada_core::types::hash::Hash;
```
tzemanovic, line 1: the main change is here

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -577,6 +581,11 @@ where
         vp_wasm_cache,
         tx_wasm_cache,
     } = shell_params;
+    if !crate::parameters::is_tx_allowed(wl_storage, tx.code_sechash())
```
tzemanovic, line 584: and here (with a test below)

crates/apps/src/lib/config/genesis/templates.rs
```diff
@@ -266,12 +266,14 @@ pub struct ChainParams<T: TemplateValidation> {
...
-    /// Hashes of whitelisted txs array. `None` value or an empty array
-    /// disables whitelisting.
-    pub tx_whitelist: Option<Vec<String>>,
+    /// Hashes of allowed vps array. `None` value or an empty array
+    /// disables allowlisting.
+    #[serde(alias = "vp_whitelist")] // old name for compat
+    pub vp_allowlist: Option<Vec<String>>,
+    /// Hashes of allowed txs array. `None` value or an empty array
+    /// disables allowlisting.
+    #[serde(alias = "tx_whitelist")] // old name for compat
```
tzemanovic, line 275: the renaming is non-breaking

crates/vp_prelude/src/lib.rs
```diff
@@ -119,11 +111,10 @@ pub fn verify_signatures(ctx: &Ctx, tx: &Tx, owner: &Address) -> VpResult {
...
-/// - tx is executed by an approved governance proposal (no need to be
-///   whitelisted)
+/// - tx is in allowlist, or
+/// - tx is executed by an approved governance proposal (no need to be allowed)
 pub fn is_valid_tx(ctx: &Ctx, tx_data: &Tx) -> VpResult {
-    if is_tx_whitelisted(ctx)? {
+    if is_tx_allowed(ctx)? {
         accept()
     } else {
         let proposal_id = tx_data
```
Fraccaman, line 120: I guess we dont need the `if` condition anymore as we are checking it in the protocol?

Fraccaman:
looks great! I just have a question about checking for allowed vps inside the vps. Can't we do that in protocol too?

crates/vp_prelude/src/lib.rs
grarco: Do we still need the entire function at all? If the tx comes from a transaction than we do check on the allowlist before vp execution and if it comes from governance it's allowed anyway. I think we need this only if there's another way to submit transactions to wasm which doesn't involve a normal transaction nor passing through governance?

crates/namada/src/ledger/protocol/mod.rs
grarco: I believe this is the same path of execution we follow for governance proposals, we call `dispatch_tx` which in turn calls `apply_wasm_tx` but than we do this check here so I wonder if we could prevent governance proposals txs from being applied here, because it looks like this function only checks the allowlist, not if we are about to execute something coming from a proposal

Fraccaman:
oh actually all e2e test seems to fail

wasm/wasm_source/src/vp_testnet_faucet.rs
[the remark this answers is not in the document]
tzemanovic: ah, looks like it got revived at some point

tzemanovic:
> looks great! I just have a question about checking for allowed vps inside the vps. Can't we do that in protocol too?

yeah, we could potentially check those in `fn tx_validate_vp_code_hash` host env fn and remove it from wasm

crates/namada/src/ledger/protocol/mod.rs
tzemanovic: I changed this in https://github.com/anoma/namada/commit/f288518a09937aadc86c78d8adc475188e07a027

wasm/wasm_source/src/vp_testnet_faucet.rs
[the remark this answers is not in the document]
tzemanovic: done in [018c60e](https://github.com/anoma/namada/commit/018c60e287df5226913b749f94289510a9b4c3d3)

crates/vp_prelude/src/lib.rs
tzemanovic: I removed it in https://github.com/anoma/namada/commit/dc50e580a19e3eb0180457899935237a0e1b2d6c

tzemanovic:
@Fraccaman I've also moved the VP check in [8bd2deb](https://github.com/anoma/namada/commit/8bd2debc2939a41afa13f93d3c9ca57f71d84415) (and tests in [a4e73dc](https://github.com/anoma/namada/commit/a4e73dc5c847a169c545a5f59cbc8a8bf6bb8497))

tzemanovic:
There was a failing unit test in finalize block that was failing on the newly added tx whitelist check because a tx used in the test didn't have a valid code section. Fixed in [159da62](https://github.com/anoma/namada/commit/159da621b54dcc378cd0c1d718e168dcf30c2558) - squashed now

grarco (approved):

PR #2421, opened 2024-01-22
Murisi/masp test vectors

murisi:
## Describe your changes
Implemented the random generation of MASP test vectors and slightly altered MASP-related transaction sections to facilitate the printing of shielded amounts with correct denominations. More specifically, the changes are as follows:
* Used proptest to generate random shielding, shielded, and deshielding MASP transactions
* Implemented a MASP `TxProver`  to aid constructing these transactions with the following properties:
  * It is faster than `LocalTxProver` since it does not generate any zero-knowledge proofs
  * It is more correct than `MockTxProver` since it generates binding signatures that the hardware wallet will check
* Improved the formatting of MASP transactions by using the new denomination information
* Modified the `AssetType` derivation to be parameterized by a token's `token::Denomination`
  * Needed for hardware wallets to safely display and sign shielded denominated amounts
  * Modified the `MaspTxBuilder` supplied to hardware wallets to include `token::Denomination` in the preimages

#2371 


core/src/types/masp.rs
```diff
@@ -15,23 +16,77 @@ use crate::types::string_encoding::{
...
+    Debug,
+    PartialOrd,
+    Ord,
+    PartialEq,
+    Eq,
+    Hash,
+    Serialize,
+    Deserialize,
+)]
+pub struct PreAssetType {
```
batconjurer, line 35: I'm somehow not digging this name. Maybe something like `AssetData`?

batconjurer (commented):
Overall looks fine. I'm somehow hung up on the name `PreAssetType`, but really that's minor.

core/src/types/masp.rs
murisi: This sounds like a better name. Thanks!

PR #2422, opened 2024-01-22
Make shielded syncing a separate command

batconjurer:
## Describe your changes
Previously, the scanning of MASP notes was done as part of various MASP commands. This moves that logic into a separate command so that it can be done out of band.

#2363 

### DIFF
--- a/crates/core/src/types/storage.rs
+++ b/crates/core/src/types/storage.rs
@@ -1,4 +1,5 @@
 //! Storage types
+use std::cmp::Ordering;
 use std::collections::VecDeque;
 use std::convert::{TryFrom, TryInto};
 use std::fmt::Display;
@@ -1461,14 +1462,28 @@ impl<E> GetEventNonce for InnerEthEventsQueue<E> {
     BorshDeserialize,
     Eq,
     PartialEq,
-    Ord,
-    PartialOrd,
 )]
 pub struct IndexedTx {
     /// The block height of the indexed tx
     pub height: BlockHeight,
     /// The index in the block of the tx
     pub index: TxIndex,
+}
+
+impl PartialOrd for IndexedTx {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for IndexedTx {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if self.height == other.height {
+            self.index.cmp(&other.index)
+        } else {
+            self.height.cmp(&other.height)
+        }
+    }
 }
 
 #[cfg(test)]
--- a/crates/core/src/types/storage.rs
+++ b/crates/core/src/types/storage.rs
@@ -1454,14 +1454,7 @@ impl<E> GetEventNonce for InnerEthEventsQueue<E> {
 /// Represents the pointers of an indexed tx, which are the block height and the
 /// index inside that block
 #[derive(
-    Default,
-    Debug,
-    Copy,
-    Clone,
-    BorshSerialize,
-    BorshDeserialize,
-    Eq,
-    PartialEq,
+    Default, Debug, Copy, Clone, BorshSerialize, BorshDeserialize, Eq, PartialEq,
 )]
 pub struct IndexedTx {
     /// The block height of the indexed tx
--- a/crates/sdk/src/masp.rs
+++ b/crates/sdk/src/masp.rs
@@ -4,9 +4,12 @@ use std::cmp::Ordering;
 use std::collections::{btree_map, BTreeMap, BTreeSet, HashMap, HashSet};
 use std::env;
 use std::fmt::Debug;
-use std::ops::Deref;
+use std::future::Future;
+use std::ops::{ControlFlow, Deref};
 use std::path::{Path, PathBuf};
+use std::pin::Pin;
 use std::str::FromStr;
+use std::task::{Context, Poll};
 
 // use async_std::io::prelude::WriteExt;
 // use async_std::io::{self};
@@ -67,6 +70,8 @@ use rand_core::{CryptoRng, OsRng, RngCore};
 use ripemd::Digest as RipemdDigest;
 use sha2::Digest;
 use thiserror::Error;
+use tokio::sync::oneshot::error::TryRecvError;
+use namada_core::ledger::inflation::RewardsController;
 
 #[cfg(feature = "testing")]
 use crate::error::EncodingError;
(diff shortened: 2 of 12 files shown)


murisi (commented):
Thanks for these changes, I think I understand the general flow. But maybe I am missing some of the higher level considerations that motivated this approach to separating MASP synchronization into a separate command. Some questions:
* Have alternatives to interrupts been considered? Like continuously saving/appending new transactions to a file?
* Given interrupt usage and SyncStatus modality, will the MASP functionality be easily usable from the SDK on all platforms (including the web)?
* How do the `ShieldedContext` lock guard in the SDK and `SyncStatus` marker type interact? Is there any overlap or redundancy here?
* How do we handle shielded context file consistency when multiple clients (from separate processes) are running?

batconjurer:
> Thanks for these changes, I think I understand the general flow. But maybe I am missing some of the higher level considerations that motivated this approach to separating MASP synchronization into a separate command. Some questions:
> 
>     * Have alternatives to interrupts been considered? Like continuously saving/appending new transactions to a file?
> 
>     * Given interrupt usage and SyncStatus modality, will the MASP functionality be easily usable from the SDK on all platforms (including the web)?
> 
>     * How do the `ShieldedContext` lock guard in the SDK and `SyncStatus` marker type interact? Is there any overlap or redundancy here?
> 
>     * How do we handle shielded context file consistency when multiple clients (from separate processes) are running?

 * I considered having a daemon process that would do such a thing. I think it's more work and will be more complicated to manage multiple process reading and writing to the same shielded context. It is something I'd like eventually, but Adrian didn't consider it a prioriy.
 * This functionality will only work in the CLI. I will pull the syncing logic up into apps to reflect this. For the web, something bespoke will need to be written
 * The `SyncStatus` is there syncing and non-syncing related functions are actually implemented on separate types. This was especially helpful in making sure I had gotten rid of all the `fetch` calls in client code. This wasn't caught by the lock guards.
  * I hadn't considered this two points because my impression was that the existing Namada trait implementations covered these problems. If that isn't the case, indeed I will need to do more work checking this. In particular, I won't allow more than one process access to a shielded context at a time. This can be relaxed later.

crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -187,7 +187,7 @@ fn run_ledger_ibc() -> Result<()> {
 }
 
 #[test]
-fn run_ledger_ibc_with_hermes() -> Result<()> {
+fn drun_ledger_ibc_with_hermes() -> Result<()> {
```
grarco, line 190: Small typo here

crates/sdk/src/masp.rs
```diff
@@ -2038,181 +1942,493 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+            self.unscanned.txs.extend(fetched);
+            display_line!(io, "");
+        }
+        // Now that we possess the unspent notes corresponding to both old and
+        // new keys up until tx_pos, proceed to scan the new transactions.
+        let mut txs = BTreeMap::new();
+        std::mem::swap(&mut self.unscanned.txs, &mut txs);
+        let txs = ProgressLogging::new(txs, io, ProgressType::Scan);
+        for (indexed_tx, (epoch, tx, stx)) in txs {
+            if self.interrupted() {
```
murisi, line 2190: If an interrupt happens here, would `self.unscanned.txs` be left in a state where it's missing the very first transactions? If this were the case, would this hamper the client's future ability to scan the very first transactions with viewing keys that are not yet known at this point?

crates/sdk/src/masp.rs
```diff
@@ -2038,181 +1942,493 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+            // Load all transactions accepted until this point
+            let fetched = self
+                .fetch_shielded_transfers(
+                    client,
+                    io,
+                    self.latest_unscanned(),
+                    last_query_height,
+                )
+                .await?;
+            self.unscanned.txs.extend(fetched);
```
murisi, line 2136: Before this line, is `self.unscanned.txs` supposed to contain all transactions up till `self.latest_unscanned()`? After this line, is `self.unscanned.txs` supposed to contain all transactions up till `last_query_height`?

murisi:
> * I considered having a daemon process that would do such a thing. I think it's more work and will be more complicated to manage multiple process reading and writing to the same shielded context. It is something I'd like eventually, but Adrian didn't consider it a prioriy.

I was thinking that there could be alternatives to interrupts other than daemon processes. Like for instance, we could do things as before but atomically commit the `ShieldedContext` to file storage (using `ShieldedUtils::save`) whenever we reach consistent states (like a transaction fetched into the context, or a transaction scanned into the context). This way if Ctrl-C is pressed or there's an OS crash, the last saved state is still preserved. I say this because reasoning about interrupts, when they can occur, how to leave/save the context in a consistent state, and how to maintain correctness as more functionality is added to the shielded pool may be more challenging than a model without channels and receivers.

murisi:
> * The `SyncStatus` is there so the syncing and non-syncing related functions are actually implemented on separate types. This was especially helpful in making sure I had gotten rid of all the `fetch` calls in client code. This wasn't caught by the lock guards.

I see, this is a valid approach. That being said,  though it may be undesirable in to do a sync just before displaying balances in our client. I wouldn't say that it is necessarily illogical for users of the SDK to interleave syncing and non-syncing operations for their own applications.

crates/sdk/src/masp.rs
batconjurer: `self.latest_unscanned()` is supposed to tell us the most recent block height in the local cache. However, there is a bug here. We need to fetch from zero up to `self.last_fetched` since we are trying to sync up new keys with existing keys.

crates/sdk/src/masp.rs
batconjurer: Yes, the cache is cleared out. If new keys appear later, we have to repopoulate this cache.

crates/core/src/types/storage.rs
```diff
@@ -1471,6 +1463,22 @@ pub struct IndexedTx {
     pub index: TxIndex,
 }
 
+impl PartialOrd for IndexedTx {
```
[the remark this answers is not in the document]
tzemanovic: let's leave a note about it (or a test)

murisi:
When working with you on the note scanning algorithm, I noticed two things:
1. The `SyncStatus` caused friction in our SDK usage when we were trying to save the `ShieldedContext`. We eventually worked around this issue by cloning the `ShieldedContext` before saving it (which should not be necessary since Borsh only requires a `ShieldedContext` reference to do serialization). We also had to introduce a `PhantomData<SyncStatus>` since the SDK code does not actually depend on the sync status.
2. We have essentially moved the interruption logic out of the SDK and into the `apps` crate. This means that the SDK no longer really deals with partiatially completed synchronizations in its interface.

In light of the above, it may be worth reconsidering if and how we do `SyncStatus`. If we are doing it for aesthetic reasons, then it should be fine. However if we are doing it to enforce correctness, then we should establish (in the absence of interrupts) what sequence of `save`, `fetch`, `load`, and `gen_shielded_transfer` operations will corrupt the internal state of the `ShieldedContext` (and render it incorrect/unusable) and therefore require the overhead of the type system to prevent. My general fear is that we may be ossifying the `apps` crate's latest `ShieldedContext` usage pattern (synchronization subcommand completely separated from other MASP actions) into the SDK unnecessarily, thereby creating friction for other SDK use cases (for instance a non-interactive program that repeatedly fetches and does useful shielded work).

If the `SyncStatus` must remain, then we should also consider moving it into the `apps` crate where such information may be more relevant. Or alternatively, we could move these `SyncStatus` API changes into a separate PR since this partitioning of `ShieldedContext` (rather than the creation of a separate syncing command) seems to form the bulk of this PR's diff. Doing this could make reviewing the `sync` subcommand changes easier. That being said, I'm happy to review whichever approach you decide upon.

brentstone:
Closing since this has already been done.

PR #2434, opened 2024-01-24
Improved localnet script rebased on draft for CI

adrianbrink:
based on #https://github.com/anoma/namada/pull/2416


adrianbrink:
Something is broken in this script. When running `python scripts/gen_localnet.py --num-vals 4 --num-nodes 6 --mode release` it generates 7 directories and 7 validators.

PR #2440, opened 2024-01-25
pgf: added reward distribution limit for stewards

Fraccaman:
## Describe your changes
- limit the number of reward distribution a steward can set
- fix a minor bug in pgf vp


crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2380,6 +2377,114 @@ fn pgf_governance_proposal() -> Result<()> {
     Ok(())
 }
 
+/// Test submission and vote of a PGF proposal
```
grarco, line 2380: Documentation here is wrong

grarco (approved):
One minor comment but looks good to me!

PR #2458, opened 2024-01-26
Changed the note scanning algorithm to not require additional context.

batconjurer:
## Describe your changes
Simplified the note scanning algorithm so that it does not require a merge operation when new keys are introduced into the wallet. Instead of having two paths in the fetching code: one for when the viewing keys are unchanged but there are new shielded transactions from the network, and the other for when new keys are introduced and transactions must be rescanned from the beginning. Instead of doing that, we now have a single path that starts scanning transactions from the most out of date viewing key going forward to the present. This change increases the set of valid states that a `ShieldedContext` can have and allows its (recoverable) state to be saved more frequently.

Namada v0.30.2


grarco (dismissed):

batconjurer (commented):
The fetch commands in the client are still present, so this can't be merged as is. Also, this is missing the batching and caching. I've got that implemented now in another branch, but still needs a bit of debugging.

crates/sdk/src/masp.rs
```diff
@@ -586,142 +650,147 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-    > {
+        logger: &impl ProgressLogger<IO>,
+        last_indexed_tx: Option<BlockHeight>,
+        last_query_height: Option<BlockHeight>,
+    ) -> Result<IndexedNoteData, Error> {
         // Query for the last produced block height
         let last_block_height = query_block(client)
             .await?
             .map_or_else(BlockHeight::first, |block| block.height);
+        let last_query_height = last_query_height.unwrap_or(last_block_height);
```
grarco, line 783: Should we query for the block height only if the arg is `None`? Just to avoid an extra RPC call since we only use `last_block_height`. Anyway this is very minor, we could change it in another PR in case

grarco (dismissed):

cwgoes (dismissed):
Makes sense to me. In the future, we might want to consider additional algorithm improvements such as (possibly dependent on user configuration/instructions):
- scanning from the most recent height backwards instead of from the oldest height forwards
- trial decrypting with a specific viewing key as opposed to all of them, for example if a user knows where they should have received a payment (this applies mostly if we get to a point where this is compute-bound)

PR #2460, opened 2024-01-26
MASP rewards controller using token amounts instead of ratios

brentstone:
## Describe your changes
Two slightly different rewards PD controllers are used for PoS and MASP now. The PoS one remains the same as before, while the MASP one no longer uses target ratios of locked tokens and instead relies on an absolute number of locked tokens.

v0.30.2


crates/core/src/types/token.rs
```diff
@@ -1009,8 +1009,9 @@ pub struct MaspParams {
     pub kd_gain_nom: Dec,
     /// Shielded Pool nominal proportional gain for the given token
     pub kp_gain_nom: Dec,
-    /// Locked ratio for the given token
-    pub locked_ratio_target: Dec,
+    /// Target amount for the given token that is locked in the shielded pool
+    /// TODO: should this be a Uint or DenominatedAmount???
```
cwgoes, line 1013: It should be a uint, I think. Whatever the max supply of a token is (probably u256?)

cwgoes (approved):
I would prefer that we abstract out the common logic in the rewards controller, since most of it is shared. We can live with this for now though.

crates/core/src/types/token.rs
cwgoes: Ah but we have the weird MASP denom splitting thing, better think carefully about this

crates/core/src/ledger/inflation.rs
```diff
@@ -96,19 +95,101 @@ impl RewardsController {
...
+
+        // Max inflation amount for this epoch
+        let max_inflation = total_native * max_reward_rate / epochs_py;
+
+        // Intermediate values
+        let p_gain = p_gain_nom * max_reward_rate / epochs_py;
+        let d_gain = d_gain_nom * max_reward_rate / epochs_py;
+        let error = locked_amount_target - locked;
+        let delta_error = locked_amount_last - locked;
+        let control_val = p_gain * error - d_gain * delta_error;
```
sug0, line 173: review these computations, to make sure they don't have values of 0

crates/core/src/types/token.rs
murisi: `Uint` (or `Amount` in the worst case) should be fine here. The inflation computations are done at a high level, and the low-level MASP denom splitting is taken care of by the MASP conversion logic.

crates/shielded_token/src/storage.rs
```diff
@@ -20,13 +22,17 @@ where
...
     } = params;
     storage.write(&masp_last_inflation_key(address), Amount::zero())?;
-    storage.write(&masp_last_locked_ratio_key(address), Dec::zero())?;
+    storage.write(&masp_last_locked_amount_key(address), Amount::zero())?;
     storage.write(&masp_max_reward_rate_key(address), max_rate)?;
-    storage.write(&masp_locked_ratio_target_key(address), locked_target)?;
     storage.write(&masp_kp_gain_key(address), kp_gain_nom)?;
     storage.write(&masp_kd_gain_key(address), kd_gain_nom)?;
+
+    let raw_target = Uint::from(*locked_amount_target)
```
murisi, line 33: `locked_amount_target` should probably be a `Uint` or `Amount`. If that is the case, then this multiplication (or any other scaling operation) should not be needed.

murisi (changes requested):
Thanks! Everything looks good to me. Just change `locked_amount_target` to be a `Uint` or an `Amount`.

crates/core/src/types/token.rs
brentstone: Yeah I think `Uint` is best, as it is clearly readable and digestible by humans, whereas `Amount` will not be in this scenario.

PR #2474, opened 2024-01-30
Murisi/masp separate parallel sync

murisi:
## Describe your changes
This PR creates a separate synchronization command and heavily draws code from #2422 . More specifically, this PR's changes are as follows:
* Parallelized the downloading of transactions with the scanning of transactions and updating of the note Merkle tree
* Created a separate command to synchronize the shielded context  called `shielded-sync`
* Adjusted all other commands to no longer implicitly synchronize the shielded pool before executing their actions
* Modified the synchronization function to save state frequently so that the interrupted and resumed without data loss
* Adjusted the integration tests so that they now explicitly do `shielded-sync`s before doing any actions

#2458 


crates/apps/src/lib/cli.rs
```diff
@@ -5601,6 +5630,45 @@ pub mod args {
         }
     }
 
+    impl Args for ShieldedSync<CliTypes> {
+        fn parse(matches: &ArgMatches) -> Self {
+            let ledger_address = LEDGER_ADDRESS_DEFAULT.parse(matches);
+            let last_query_height = BLOCK_HEIGHT_OPT.parse(matches);
+            let viewing_keys = VIEWING_KEYS.parse(matches);
```
batconjurer, line 5637: do we not want to support passing in spending keys here?

crates/apps/src/lib/cli/client.rs
```diff
@@ -298,6 +300,42 @@ impl CliApi {
...
+                            .copied()
+                            .collect();
+                        let vks = args
+                            .viewing_keys
+                            .into_iter()
+                            .map(|vk| ExtendedFullViewingKey::from(vk).fvk.vk)
+                            .collect::<Vec<_>>();
+                        let mut shielded = namada.shielded_mut().await;
+                        let _ = shielded.load().await;
+                        for vk in wallet_keys {
```
batconjurer, line 324: Let's leave the internals of vk heights inside of the `sync` function. Instead lets concatenate all of the keys together and pass them into the shielded sync function.

crates/sdk/src/masp.rs
```diff
@@ -583,112 +589,136 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-                .await?;
-            tx_iter = txs.iter();
-        }
-        // Now that we possess the unspent notes corresponding to both old and
-        // new keys up until tx_pos, proceed to scan the new transactions.
-        for (indexed_tx, (epoch, tx, stx)) in &mut tx_iter {
-            self.scan_tx(*indexed_tx, *epoch, tx, stx)?;
+            self.vk_heights.entry(*vk).or_default();
         }
+        // the latest block height which has been added to the witness Merkle
```
batconjurer, line 676: I'm actually surprised that the integration tests pass. There is bug here that one failed for me on. We need to save the shielded context here to persist the keys.

batconjurer (commented):
A couple of small changes. Also, I'm not seeing marco's changes in here. I think there is a divergence between base and main.

crates/apps/src/lib/cli.rs
murisi: Whenever a spending key is added to the wallet, its corresponding viewing key is also added under the same name. So I'm thinking that that this code should be sufficient for now. No?

batconjurer: Sure, I just wondered if we considered it important.

grarco (approved):
Looks good to me, thank you!

PR #2479, opened 2024-01-31
Murisi/masp separate parallel sync rebased

murisi:
## Describe your changes
This PR creates a separate synchronization command and heavily draws code from #2422 . More specifically, this PR's changes are as follows:
* Parallelized the downloading of transactions with the scanning of transactions and updating of the note Merkle tree
  * This reduces the wall-clock time required to synchronize the shielded context
  * This allows us to avoid having to cache downloaded transactions until transaction scanning starts
* Created a separate command to synchronize the shielded context  called `shielded-sync`
* Adjusted all other commands to no longer implicitly synchronize the shielded pool before executing their actions
* Modified the synchronization function to save state frequently so that the interrupted and resumed without data loss
* Adjusted the integration tests so that they now explicitly do `shielded-sync`s before doing any actions

Namada v0.30.3


brentstone:
Is this still relevant? @batconjurer @grarco

batconjurer:
> Is this still relevant? @batconjurer @grarco

No

PR #2482, opened 2024-01-31
refactor finalize block

tzemanovic:
## Describe your changes

Refactored token, gov and PoS sub-systems integration in finalize block. The token and PoS logic is moved into their respective crates. Gov proposal exec depends on the shell as well as IBC and PoS so this is kept in `crates/apps/src/lib/node/ledger/shell/governance.rs` for now.

0.31.5


grarco (dismissed):
Looks good to me, thank you!

batconjurer (dismissed):

brentstone:
Can we also move https://github.com/anoma/namada/blob/fdafaa685f3cc0da98dc84a443a93314feffea7e/crates/apps/src/lib/node/ledger/shell/finalize_block.rs#L647-L653 out of `apply_inflation`? I suspect it may have been incorrectly placed into this function, it can at least go right into `finalize_block::finalize_block` or some other function, but it has nothing to do with inflation.

brentstone:
Would it also be safer to call `governance::finalize_block` before `token::finalize_block` in case we want governance to execute changes before MASP `update_allowed_conversions` is called ever?

brentstone:
Also, perhaps we could move `proof_of_stake::log_block_rewards` to `proof_of_stake::rewards::log_block_rewards` and `proof_of_stake::record_slashes_from_evidence` to `proof_of_stake::slashing::record_slashes_from_evidence`.

tzemanovic:
> Would it also be safer to call `governance::finalize_block` before `token::finalize_block` in case we want governance to execute changes before MASP `update_allowed_conversions` is called ever?

Yeah, good call

tzemanovic:
@yito88 it looks like the ibc events were in the wrong place and they would have only been applied on new epochs

brentstone (approved):

PR #2499, opened 2024-01-31
add missing memo in become validator and change consensus keys

Fraccaman:

### DIFF
--- a/crates/apps/src/lib/client/tx.rs
+++ b/crates/apps/src/lib/client/tx.rs
@@ -412,6 +412,10 @@ pub async fn submit_change_consensus_key(
     )
     .add_data(data);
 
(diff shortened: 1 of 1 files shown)


tzemanovic (commented):
since any tx can include a memo, it would be good to do this in some common tx function so that we don't have to do this manually for every tx type

Fraccaman:
this is done in the `build_*` function of every tx by passing `tx_args` but both `init_validator` and `change_consensus_key` dont have it :(

tzemanovic:
> this is done in the `build_*` function of every tx by passing `tx_args` but both `init_validator` and `change_consensus_key` dont have it :(

ah, I see

tzemanovic (approved):

brentstone (approved):

PR #2503, opened 2024-02-01
core refactor

### LINKED ISSUE #2483: core refactor
There are couple left over things in  `namada_core::ledger` mod that can be moved out:
- eth (only addresses that can go in `core/src/types/address.rs` or `ethereum_bridge` crate)
- inflation can probably go into token crate - #2484
- replay protection is spread out - it could maybe be its own crate
- Flatten the core `types`  module so that we're just importing e.g. `namada_core::dec`  etc. 

Also check and remove any unused deps.

tzemanovic:
## Describe your changes

closes #2483

#2493 - diff https://github.com/anoma/namada/pull/2503/files/00cddd4c2585829d7e1b4e9bbded892b3b80d2d2..94d1a0b2048bb33fd2043aa27a69d388d965163a


crates/state/src/wl_storage.rs
```diff
@@ -572,13 +573,27 @@ macro_rules! impl_storage_traits {
 impl_storage_traits!(WlStorage<D, H>);
 impl_storage_traits!(TempWlStorage<'_, D, H>);
 
+impl<D, H> WithConversionState for WlStorage<D, H>
```
batconjurer, line 576: Does this really need a separate trait?

tzemanovic: I think so - it is used to replace the `state` dep in `shielded_token` with `storage`, which restricts what the crate can access. The idea is to use a similar approach (with mocking when needed) to remove inter-sub-system dependencies

PR #2524, opened 2024-02-05
Governance proposals for SE

batconjurer:

### DIFF
--- a/wasm/Cargo.toml
+++ b/wasm/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "wasm_source",
     "tx_template",
     "vp_template",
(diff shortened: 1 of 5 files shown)


brentstone:
Are we going to include NAAN in the shielded set rewards? What would we want the max inflation and target amount to be if so? @cwgoes @adrianbrink @batconjurer

cwgoes:
> Are we going to include NAAN in the shielded set rewards? What would we want the max inflation and target amount to be if so? @cwgoes @adrianbrink @batconjurer

Yes. Target amount 10 million, max inflation per annum 1%.

PR #2543, opened 2024-02-06
Add internal addresses to wallet upon `join-network`

### LINKED ISSUE #2497: Wallet creation should contain native addresses with aliases
It would be nice for the wallet created upon joining a network to also hold some native addresses with aliases to things like the PoS, MASP, governance, etc. addresses.

brentstone:
## Describe your changes
Closes #2497.

v0.31.2

### DIFF
--- a/crates/sdk/src/wallet/store.rs
+++ b/crates/sdk/src/wallet/store.rs
@@ -534,11 +534,6 @@ impl Store {
         address: Address,
         force: bool,
     ) -> Option<Alias> {
-        // abort if the alias is reserved
-        if Alias::is_reserved(&alias).is_some() {
-            println!("The alias {} is reserved.", alias);
-            return None;
-        }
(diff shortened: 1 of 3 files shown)


crates/sdk/src/wallet/store.rs
```diff
@@ -534,11 +534,6 @@ impl Store {
         address: Address,
         force: bool,
     ) -> Option<Alias> {
-        // abort if the alias is reserved
-        if Alias::is_reserved(&alias).is_some() {
-            println!("The alias {} is reserved.", alias);
-            return None;
-        }
```
brentstone, line 541: @batconjurer is it ok to remove these lines? I don't see an issue doing so but want to make sure

batconjurer: Yes, since we don't use aliases in genesis files anymore.

PR #2558, opened 2024-02-07
Fix query of protocol params

brentstone:
## Describe your changes
Several PoS params were missing from this query, and the general output is reformatted to look nicer as well.

v0.31.2


crates/apps/src/lib/client/rpc.rs
```diff
@@ -1538,15 +1538,33 @@ pub async fn query_protocol_parameters(
     let pos_params = query_pos_parameters(context.client()).await;
```
tzemanovic, line 1538: a small improvement - we could use `let PosParams { } = query ...` syntax to destruct the fields to make sure we're using all so if new one is added we don't miss it here

PR #2561, opened 2024-02-08
Removes custom established VPs

### LINKED ISSUE #2554: Remove the ability to update an account's VP
We do not need the ability for an account to update their VP in the initial version of Namada, and we should remove it.

grarco:
## Describe your changes

Closes #2554.

Modifies `tx_init_account` and `tx_update account` to not write the vp key. Updates of it are not supported for now and when initializing a new established account (also for validators) the protocol automatically assigns it the `vp_user`.

The host function `update_validity_predicate` has been removed.

The accounts' vps have been removed from the genesis files.

v0.31.8


wasm/wasm_source/src/tx_update_account.rs
```diff
@@ -17,27 +17,28 @@ fn apply_tx(ctx: &mut Ctx, tx: Tx) -> TxResult {
...
-                err
-            })?;
-
-        ctx.update_validity_predicate(
-            owner,
-            vp_code_sec.code.hash(),
-            &vp_code_sec.tag,
-        )?;
-    }
+    // TODO: commented out for now cause we don't want to support vp changes
```
cwgoes, line 20: Remove the code, don't comment it out. This is what Git is for.

wasm/wasm_source/src/vp_user.rs
```diff
@@ -135,9 +135,12 @@ fn validate_tx(
             KeyType::PgfSteward(address) => address != &addr || *valid_sig,
             KeyType::GovernanceVote(voter) => voter != &addr || *valid_sig,
             KeyType::Vp(owner) => {
-                let has_post: bool = ctx.has_key_post(key)?;
                 if owner == &addr {
-                    has_post && *valid_sig
+                    // TODO: disabled cause we don't want to support vp changes
```
cwgoes, line 139: Ditto

cwgoes (changes requested):

tzemanovic (dismissed):

wasm/wasm_source/src/vp_user.rs
```diff
@@ -134,14 +134,7 @@ fn validate_tx(
...
             KeyType::GovernanceVote(voter) => voter != &addr || *valid_sig,
-            KeyType::Vp(owner) => {
-                let has_post: bool = ctx.has_key_post(key)?;
-                if owner == &addr {
-                    has_post && *valid_sig
-                } else {
-                    true
-                }
-            }
+            KeyType::Vp(owner) => owner != &addr,
```
grarco, line 137: @cwgoes a note on this. This will prevent any vp update even if we wanted to allow it later. To do so we'd need to hard fork and modify this line so that vp updates are enabled. We's also need to modify all the accounts `?` storage key to point to a different vp hash. At the same time we can't allow vp changes in here (even though we removed them from the tx) cause it will be possible to to change VP via a sequence of governance proposals

Fraccaman:
@cwgoes do we really want to remove the possibility to update vps via governance? can't we just remove the possibility for usersd to update their acccou t vp and have just the default one? vps are whitelisted anyway and if we just remove the possibility from CLI and wasm transaction to update the validity predicate we can retain the possibility to create multiple of them via governance.

cwgoes:
> @cwgoes do we really want to remove the possibility to update vps via governance? can't we just remove the possibility for usersd to update their acccou t vp and have just the default one? vps are whitelisted anyway and if we just remove the possibility from CLI and wasm transaction to update the validity predicate we can retain the possibility to create multiple of them via governance.

Yeah - I think we can retain the possibility to update VPs via governance, I just want to remove the user-submitted txs.

tzemanovic (approved):

grarco:
After #2770 we probably need to rethink this a bit since we might need to let user update their vps

grarco:
About my previous comment: given the way we chose to remove vps from the allowlist, I believe we need to revert the `tx_update_account` tx to allow the update of the vp in case it was removed from the allowlist. We can still prevent custom vps in genesis instead

brentstone:
is this something we still want to include in the near future?

grarco:
> Is this PR still relevant? @grarco

I don't think so. I guess we can close this and also the related issue

PR #2575, opened 2024-02-09
Refactor PD Controller

### LINKED ISSUE #2553: Cleanup / DRY PD controller
From a review of `crates/core/src/ledger/inflation.rs`:
- We should be able to have a standard `PDController` which takes standard parameters and outputs the control input.
- This PD controller should be wrapped by a small function for proof-of-stake which needs to calculate the locked ratio before running the controller.
- We should double-check that we're using checked arithmetic (with the various numeric types and possible operator overloading, it's not clear to me).
- We do not need to store the locked ratio in storage; we can simply recalculate it when needed (we do need to store the _old_ locked ratio, just as we need to store the _old_ target amount, but this should be done by the PD controller with no knowledge of proof-of-stake specific semantics).
- This file should be renamed to `pd-controller` or something more indicative, a PD controller alone has nothing to do with inflation. It should also be moved to a standalone crate instead of being in `core`.

brentstone:
## Describe your changes
Closes #2553.

v0.31.2


crates/core/src/ledger/mod.rs
tzemanovic, line 1: this can be rm'd

tzemanovic (approved):

PR #2582, opened 2024-02-09
0.31.3 rc2

tzemanovic:


tzemanovic:
failing unit test is a false positive, see https://github.com/anoma/namada/issues/2584

PR #2588, opened 2024-02-09
Murisi/test vectors 0.30.2 rev 1

murisi:
## Describe your changes
Various changes in order to increase the detail presented in test vectors, increase the variety of test vectors, and ease the implementation of hardware wallet apps. Specifically the changes are as follows:
* Test vector generator now sometimes generates signed transactions
* Greatly increased the level of detail presented in Init Proposal transactions
* Increased the coverage PGF steward transactions to include the IBC variants
* Started including memos in generated transactions sometimes, both in hashed and full form
* Now print out the section hashes in debug vectors to facilitate constructing signing test vectors
* Improved the formatting of Update Steward Commission transactions for the hardware wallet

Namada v0.30.2

### DIFF
--- a/crates/governance/src/storage/proposal.rs
+++ b/crates/governance/src/storage/proposal.rs
@@ -589,16 +589,55 @@ pub mod testing {
     }
 
     prop_compose! {
-        /// Generate an arbitrary PGF target
-        pub fn arb_pgf_target()(
+        /// Generate an arbitrary PGF internal target
+        pub fn arb_pgf_internal_target()(
             target in arb_non_internal_address(),
             amount in arb_amount(),
-        ) -> PGFTarget {
-            PGFTarget::Internal(PGFInternalTarget {
+        ) -> PGFInternalTarget {
+            PGFInternalTarget {
                 target,
                 amount,
-            })
+            }
         }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary port ID
+        pub fn arb_ibc_port_id()(id in "[a-zA-Z0-9_+.\\-\\[\\]#<>]{2,128}") -> PortId {
+            PortId::new(id).expect("generated invalid port ID")
+        }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary channel ID
+        pub fn arb_ibc_channel_id()(id: u64) -> ChannelId {
+            ChannelId::new(id)
+        }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary PGF IBC target
+        pub fn arb_pgf_ibc_target()(
+            target in "[a-zA-Z0-9_]*",
+            amount in arb_amount(),
+            port_id in arb_ibc_port_id(),
+            channel_id in arb_ibc_channel_id(),
+        ) -> PGFIbcTarget {
+            PGFIbcTarget {
+                target,
+                amount,
+                port_id,
+                channel_id,
+            }
+        }
+    }
+
+    /// Generate an arbitrary PGF target
+    pub fn arb_pgf_target() -> impl Strategy<Value = PGFTarget> {
+        prop_oneof![
+            arb_pgf_internal_target().prop_map(PGFTarget::Internal),
+            arb_pgf_ibc_target().prop_map(PGFTarget::Ibc),
+        ]
     }
 
     /// Generate an arbitrary PGF action
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -1002,20 +1002,11 @@ pub mod testing {
             wrapper in arb_wrapper_tx(),
             transfer in arb_transfer(),
             code_hash in arb_hash(),
-            memo in option::of(arb_utf8_code()),
         ) -> (Tx, TxData) {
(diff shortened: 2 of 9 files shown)


crates/governance/src/storage/proposal.rs
```diff
@@ -589,18 +589,57 @@ pub mod testing {
...
                 target,
                 amount,
-            })
+            }
+        }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary port ID
+        pub fn arb_ibc_port_id()(id in "[a-zA-Z0-9_+.\\-\\[\\]#<>]{2,128}") -> PortId {
```
sug0, line 606: I think ports should be alphanumeric. the wallet might not be checking this, so it's probably irrelevant

crates/sdk/src/lib.rs
```diff
@@ -835,15 +838,18 @@ pub mod testing {
...
+        Custom,
     }
 
     prop_compose! {
         // Generate an arbitrary commitment
         pub fn arb_commitment()(
-            hash in arb_hash(),
+            commitment in prop_oneof![
+                arb_hash().prop_map(Commitment::Hash),
+                arbitrary::any::<Vec<u8>>().prop_map(Commitment::Id),
```
sug0, line 849: should we emplace an upper bound on the size of these commitments?

crates/sdk/src/lib.rs
```diff
@@ -862,6 +868,33 @@ pub mod testing {
         }
     }
 
+    prop_compose! {
+        // Generate an arbitrary uttf8 commitment
+        pub fn arb_utf8_commitment()(
+            commitment in prop_oneof![
+                arb_hash().prop_map(Commitment::Hash),
+                "[a-zA-Z0-9_]*".prop_map(|x| Commitment::Id(x.into_bytes())),
```
sug0, line 876: same q here

crates/sdk/src/lib.rs
```diff
@@ -862,6 +868,33 @@ pub mod testing {
...
+            commitment
+        }
+    }
+
+    prop_compose! {
+        // Generate an arbitrary code section
+        pub fn arb_utf8_code()(
+            salt: [u8; 8],
+            code in arb_utf8_commitment(),
+            tag in option::of("[a-zA-Z0-9_]*"),
```
sug0, line 888: similar q

crates/sdk/src/signing.rs
```diff
@@ -1714,6 +1845,30 @@ pub async fn to_ledger_vector(
...
+                tv.output_expert
+                    .push(format!("Memo Hash : {}", HEXLOWER.encode(&hash.0)));
+            }
+            Commitment::Id(id) => {
+                let memo = String::from_utf8(id).map_err(|err| {
+                    Error::from(EncodingError::Conversion(err.to_string()))
+                })?;
+                tv.output.push(format!("Memo : {}", memo));
+                tv.output_expert.push(format!("Memo : {}", memo));
+            }
```
sug0, line 1868: since there is no bound on the size of memos (IIRC), we might run over the length of the columns available to print on in the ledger's screen, no?

sug0 (commented):
a couple of comments

crates/governance/src/storage/proposal.rs
murisi: Thanks for reviewing. Either way works for me since I don't know enough about IBC specs. That being said, the ICS-24 spec seems to allow port numbers like this. See https://github.com/cosmos/ibc/blob/main/spec/core/ics-024-host-requirements/README.md#paths-identifiers-separators . Also note that the regex pattern above was the most general pattern I could find for which `PortId::new`'s internal validation (based on ICS-24) returned `Result::Ok`.

crates/sdk/src/signing.rs
murisi: Good question. Yes you're correct about there being potentially insufficient columns. But this issue is mitigated in the hardware by paging long outputs, i.e. printing only a part of the memo at a time and allowing the user to press "next". To reflect this reality, paging is implemented for the test vectors too. See the function `format_outputs` in `signing.rs` that does this string splitting.

crates/sdk/src/lib.rs
murisi: Yeah this is a possibility, do you have any numbers in mind for the limit/cut-off? But note that the (unrestricted) commitment sizes have not caused practical issues in the test vectors or their use in testing the Ledger app (yet). The Ledger app has been able to handle all the vectors generated so far, even though these vectors are probably being drawn from a superset of what might actually occur in practice.

crates/governance/src/storage/proposal.rs
sug0: if it's been working against a validation predicate, then it's probably ok to keep it as is

crates/sdk/src/lib.rs
sug0: if commitments generated here only represents the happy path, something like 1 KiB at most I think; that said, testing larger commitments would also be great, in case they are too large for the ledger device to handle

PR #2601, opened 2024-02-12
Move ConversionState token map to the storage for inflation proposal

yito88:
## Describe your changes
- Move the token map in `ConversionState` to the storage to be updated by wasm txs
- Store the hash of the asset map in `ConversionState` to be merklized
- Merklize these storage keys and the hash of the commitment tree

~~#2618~~ `v0.31.9`


brentstone:
We will need to change `namada_apps::node::ledger::shell:is_merklized_storage_key` because it currently excludes all MASP keys.

brentstone (approved):
Could be nice to transform `TokenMap` and its usage into lazy data, but other than that LGTM

tzemanovic (approved):

grarco:
Since this adds the masp `convert_anchor` to the merkle tree it partially addresses #2794

PR #2602, opened 2024-02-12
Disable Ethereum Bridge functionality at compile time

### LINKED ISSUE #2598: Disable all Ethereum bridge code (don't run the broadcaster). We can keep the code in the codebase, and still compile it, but it should never be called by the node.

sug0:
## Describe your changes

Closes #2598

`v0.31.4`

### DIFF
--- a/crates/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/crates/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -859,6 +859,7 @@ mod test_finalize_block {
     use namada::eth_bridge::storage::bridge_pool::{
         self, get_key_from_hash, get_nonce_key, get_signed_root_key,
     };
+    use namada::eth_bridge::storage::eth_bridge_queries::is_bridge_comptime_enabled;
     use namada::eth_bridge::storage::min_confirmations_key;
     use namada::ethereum_bridge::storage::wrapped_erc20s;
     use namada::governance::storage::keys::get_proposal_execution_key;
@@ -1445,6 +1446,11 @@ mod test_finalize_block {
     where
         F: FnOnce(&mut TestShell) -> (Tx, TestBpAction),
     {
+        if !is_bridge_comptime_enabled() {
+            // NOTE: this test doesn't work if the ethereum bridge
+            // is disabled at compile time.
+            return;
+        }
         let (mut shell, _, _, _) = setup_at_height(1u64);
         namada::eth_bridge::test_utils::commit_bridge_pool_root_at_height(
             &mut shell.wl_storage,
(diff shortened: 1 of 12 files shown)


crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -5080,6 +5086,11 @@ mod test_finalize_block {
     /// Test that updating the ethereum bridge params via governance works.
     #[tokio::test]
     async fn test_eth_bridge_param_updates() {
+        if !is_bridge_comptime_enabled() {
```
tzemanovic, line 5089: could we keep the bridge enabled for tests? i.e. `#[cfg(any(feature = "namada-eth-bridge", test))]`

tzemanovic:
it would be good to add the feature to `Cargo.toml`s and propagate it through the deps

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
sug0: I was trying to avoid sprinkling `cfg` flags all over the place, instead centralizing them under the predicate `is_bridge_comptime_enabled()`. Either way, most of these tests would simply not work if they ran with `#[cfg(not(feature = "namada-eth-bridge"))]` verifying (i.e. with the Bridge disabled by a compile time flag). The things we should test are that protocol txs get rejected by VPs/the mempool, which we already do with the changes in this PR.

sug0:
> it would be good to add the feature to `Cargo.toml`s and propagate it through the deps

I left it out of this PR on purpose. Figured we won't be enabling the Bridge any time soon, anyway. I can add it to the manifest files, though.

crates/apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -693,6 +697,69 @@ mod test_process_proposal {
 
     const GAS_LIMIT_MULTIPLIER: u64 = 100_000;
 
+    /// Check that we reject a validator set update protocol tx
+    /// if the bridge is not active.
+    #[test]
+    fn check_rejected_valset_upd_bridge_inactive() {
```
batconjurer, line 703: should we also check that votes on bridge pool roots and eth events are rejected?

batconjurer (changes requested):
I think we need to still run the unit tests with bridge enabled in the ci. Otherwise they won't be maintained and will rot and potentially be a big headache later.

crates/apps/src/lib/node/ledger/shell/process_proposal.rs
sug0: we do already, in `check_rejected_bp_roots_bridge_inactive` and `check_rejected_eth_events_bridge_inactive`

sug0:
@batconjurer I agree, but idk how to accomplish that

sug0:
added a CI task to run eth bridge tests

batconjurer (approved):

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: Ah, I don't want to nitpick on this, but I think if we enable the feature in the apps' `dev-deps` for `namada`, we would only have to use `#[cfg(any(feature = "namada-eth-bridge", test))]` (and its negation) in the apps crate. Then
- the rest could stay the same as is
- we can remove `is_bridge_comptime_enabled` altogether as the tests would always run
- we wouldn't have to run unit tests 2 times in CI

tzemanovic (approved):

PR #2617, opened 2024-02-15
Fix vp_user

### LINKED ISSUE #2663: Possible bug in `vp_user`
There have been numerous reports of users failing to unjail their validators after, following a consensus key change, the validators did not properly update their CometBFT keys, resulting in jailing for downtime. The transactions fail largely due to VP rejection from another validator that would be demoted to the below-capacity validator set after the unjailed validator would be reinserted into consensus. Such issues have been reported in #2640 and #2642.

There may have been other reports of unjailing failures at the VP-level that do not involve a consensus key change, though it is unclear if the issue is the same as the above.

Lastly, a small bug in the `vp_user` has already been found and addressed in #2617.

brentstone:
## Describe your changes
Fix the unjailing bug observed in the SE, when unjailing that involves demoting another consensus validator the the below-capacity set is involved.

Closes #2663.

v0.31.9

### DIFF
--- a/wasm/wasm_source/src/vp_user.rs
+++ b/wasm/wasm_source/src/vp_user.rs
@@ -225,15 +225,15 @@ fn validate_pos_changes(
                                     Consensus | BelowCapacity | BelowThreshold
                                 ) && post == Inactive)
                             // Reactivation case
-                            || pre == Inactive && post != Inactive
+                            || (pre == Inactive && post != Inactive)
                             // Unjail case
-                            || pre == Jailed
+                            || (pre == Jailed
                                 && matches!(
                                     post,
                                     Consensus
                                         | BelowCapacity
                                         | BelowThreshold
-                                )
+                                ))
                         {
                             address == owner && **valid_sig
                         } else if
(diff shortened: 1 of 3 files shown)


tzemanovic (approved):

brentstone:
@tzemanovic I think we are safe not to update the implicit VP since validator accounts are only Established. Lmk if you can think of another reason to update that too.

wasm/wasm_source/src/vp_user.rs
```diff
@@ -628,6 +629,153 @@ mod tests {
...
+
+        let pks_map = AccountPublicKeysMap::from_iter(vec![ck3]);
+
+        let mut vp_env = vp_host_env::take();
+        let mut tx_data = Tx::from_type(TxType::Raw);
+        tx_data.set_data(Data::new(vec![]));
+        tx_data.set_code(Code::new(vec![], None));
+        tx_data.add_section(Section::Signature(Signature::new(
+            vec![tx_data.raw_header_hash()],
+            pks_map.index_secret_keys(vec![key::testing::keypair_4()]),
```
tzemanovic, line 742: just a little nit about this test - it would make it clearer to use the `ck3` for this directly to see that it's being signed by `validator3` (the `ck`s are converted to PKs with `ref_to` so we should keep them as SKs and turn them into PKs as needed)

brentstone: done in fc4e7488183c5f4718236cf5a682ba38889b4b5b

tzemanovic (approved):

PR #2619, opened 2024-02-15
Gas accounting for fee unshielding

grarco:


crates/apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -1486,67 +1473,95 @@ where
...
-    let result = apply_wasm_tx(
-        unshield,
-        &TxIndex::default(),
-        ShellParams::new(
-            &mut TxGasMeter::new(fee_unshielding_gas_limit),
-            temp_wl_storage,
-            vp_wasm_cache,
-            tx_wasm_cache,
-        ),
+    if namada::ledger::protocol::run_fee_unshielding(
```
grarco, line 1478: I'm now using for `mempool_validate`, `prepare_proposal` and `process_proposal` the same function used in `finalize_block`. The logic though is a bit different: in the first three cases, if the fee unshielding operation fails, for any reason, the transaction is deemed invalid (won't make it into the block). In `finalize_block`, instead, if the fee unshielding fails we still try to get the fee payment from the transparent address of the fee payer: if the funds are not enough we take what we can and reject the tx (won't get executed) otherwise, at least for now, the transaction can go ahead.

It has to be noted that this scenario in `finalize_block` should never happen, the check in `process_proposal` should prevent such a transaction from being included into the block. We still need to handle this case in case there was a bug and a transaction of this kind could indeed be included. We could also choose, for this case, to still try to get the payment from the transparent address but reject the transaction even if the payment was successful.

murisi (dismissed):
Looks good to me. Thanks!

PR #2621, opened 2024-02-15
Fix nullifiers check

### LINKED ISSUE #2594: Fix check on nullifiers
Currently the masp VP verifies the nullifiers only if `source == MASP`.  This means that a shielding transaction could write arbitrary nullifiers and still be accepted. We should instead do this check all the times, regardless of the source of the transaction.

grarco:
## Describe your changes

Closes #2594.

Call `valid_nullifiers_reveal` regardless of the transaction source to make sure that a shielding tx doesn't reveal any nullifier. Enforces that at least one shielded input is present if the source of the tx is the MASP.

Refactors the checks on the bundles and fixes the check on the shielded outputs when the target is MASP which was previously wrong (was not taking into account the case in which `shielded_outputs` was `None`.

v0.33.0

### DIFF
--- a/crates/namada/src/ledger/native_vp/masp.rs
+++ b/crates/namada/src/ledger/native_vp/masp.rs
@@ -80,7 +80,6 @@ where
         // than once in the same tx
         let mut revealed_nullifiers = HashSet::new();
 
-        // FIXME: test this check in some way if possible
         for description in transaction
             .sapling_bundle()
             .map_or(&vec![], |bundle| &bundle.shielded_spends)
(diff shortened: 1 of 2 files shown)


crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -563,33 +557,43 @@ where
...
-        // in storage with the new output descriptions
-        if !self.valid_note_commitment_update(&shielded_tx)? {
+        // in storage with the new output descriptions and also reveal the
+        // nullifiers correctly (only if needed) NOTE: these two checks
+        // validate the keys that the transaction write in storage and therefore
+        // must be done regardless of the type of transaction (shielding,
+        // shielded, unshielding) since a malicious tx could try to write keys
+        // in an invalid way
+        if !(self.valid_note_commitment_update(&shielded_tx)?
+            || self.valid_nullifiers_reveal(keys_changed, &shielded_tx)?)
```
murisi, line 595: Shouldn't this OR be an AND? We want both the note commitment update and the nullifier reveal to be valid. No?

murisi (changes requested):
Looks good to me. Just one minor query...

crates/namada/src/ledger/native_vp/masp.rs
cwgoes: ... and we should definitely have tests for this if we don't already

PR #2625, opened 2024-02-15
Fix the balance query for IBC tokens

yito88:
## Describe your changes
The shielded balance query sometimes doesn't know the received IBC tokens.
By modifing `precompute_asset_types` to get all tokens received over IBC, the query can get the balances of these tokens.

### DIFF
--- a/crates/sdk/src/masp.rs
+++ b/crates/sdk/src/masp.rs
@@ -1228,14 +1228,14 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
 
     /// Use the addresses already stored in the wallet to precompute as many
     /// asset types as possible.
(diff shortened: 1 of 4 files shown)


crates/sdk/src/masp.rs
```diff
@@ -1231,9 +1231,10 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
     pub async fn precompute_asset_types<N: Namada>(
         &mut self,
         context: &N,
```
grarco, line 1233: Should we just pass in the client at this point, since we don't need the rest of the context?

grarco (dismissed):
Minor comment but looks good to me!

grarco (approved):

PR #2628, opened 2024-02-15
Fix validator voting period

### LINKED ISSUE #2626: Gov VP: buggy `is_valid_validator_voting_period`
The function `is_valid_validator_voting_period` in the governance VP incorrectly does division on u64 before multiplication, which sometimes unintentionally leads to a computation of 0. It is also does not align with the same computation in `StorageProposal::can_be_voted`.

The offending lines are:
```
let duration = voting_end_epoch - voting_start_epoch;
let two_third_duration = (duration / 3) * 2;
current_epoch <= voting_start_epoch + two_third_duration
```

Consider an example of a proposal with start epoch of 4, end epoch of 6, and a validator voting in epoch 5. In the client for `vote-proposal`, the vote is deemed valid by `StorageProposal::can_be_voted`. However, in the VP, `two_third_duration = 0` due to `2 / 3`, and then the vote is deemed invalid and rejected.

brentstone:
## Describe your changes
Closes #2626.

v0.31.9

### DIFF
--- a/.changelog/unreleased/bug-fixes/2628-fix-validator-voting-period.md
+++ b/.changelog/unreleased/bug-fixes/2628-fix-validator-voting-period.md
@@ -1,0 +1,2 @@
+- Fixes the computation of the valid validator voting period.
+  ([\#2628](https://github.com/anoma/namada/pull/2628))
--- a/crates/apps/src/lib/client/rpc.rs
+++ b/crates/apps/src/lib/client/rpc.rs
@@ -1271,7 +1271,7 @@ pub async fn query_proposal_result(
             (proposal_result, proposal_query)
         {
             display_line!(context.io(), "Proposal Id: {} ", proposal_id);
-            if current_epoch >= proposal_query.voting_end_epoch {
+            if current_epoch > proposal_query.voting_end_epoch {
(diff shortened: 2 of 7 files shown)


Krewedk0:
There is also inconsistency with voting period
`namadac query-proposal-result` assumes voting period ends on start of an `End epoch` of a prop:
https://github.com/anoma/namada/blob/9181fbe7533e524e83ef7bb77333a0efc9b8847d/crates/apps/src/lib/client/rpc.rs#L1241-L1244

for `namadac query-proposal` it ends on the next epoch after `End epoch`:
https://github.com/anoma/namada/blob/9181fbe7533e524e83ef7bb77333a0efc9b8847d/crates/governance/src/storage/proposal.rs#L500-L501

And same logic for voting txs:
https://github.com/anoma/namada/blob/9181fbe7533e524e83ef7bb77333a0efc9b8847d/crates/governance/src/storage/proposal.rs#L485-L486

I assume voting should end after an end epoch. If so,  condition check in `query_proposal_result` needs to be fixed

brentstone:
@Krewedk0 thx for catching this, the client log is fixed now

PR #2630, opened 2024-02-15
Refactor loading of gov proposals for execution

brentstone:
## Describe your changes
Governance proposals were previously loaded every single block, which was a pointless operation. This logic is now coupled more tightly with the execution of the proposals, occurring only once per epoch within a function that loads and then executes.

This logic is removed from the `Shell` and reimplemented in the governance crate.

v0.31.4

### DIFF
--- a/crates/apps/src/lib/node/ledger/mod.rs
+++ b/crates/apps/src/lib/node/ledger/mod.rs
@@ -9,14 +9,11 @@ pub mod tendermint_node;
 use std::convert::TryInto;
 use std::net::SocketAddr;
 use std::path::PathBuf;
-use std::str::FromStr;
 use std::thread;
 
 use byte_unit::Byte;
 use futures::future::TryFutureExt;
 use namada::eth_bridge::ethers::providers::{Http, Provider};
-use namada::governance::storage::keys as governance_storage;
-use namada::types::storage::Key;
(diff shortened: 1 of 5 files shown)


adrianbrink:
I think the current behavior is correct. The changes should be activated on the epoch after the grace epoch ends.

brentstone:
@adrianbrink this is kind of arbitrary though right, since we could just increase the `grace_epoch` by 1 to achieve the same behavior. I like the idea of renaming to `activation_epoch` to be more intuitive. I think the code would end up clearer if execution occurred at the grace epoch (or wtv we name it) as well.

brentstone:
@Fraccaman @adrianbrink what do you think of the changes in [9a354f4](https://github.com/anoma/namada/pull/2630/commits/9a354f49ba78deb40dd0ade454075d5dc3784199)? I think renaming makes things much clearer to understand upon reading code and in general going forward, but we can remove this commit if there are objections.

grarco (approved):

tzemanovic:
looks like the e2e tests are actually broken here

tzemanovic:
> looks like the e2e tests are actually broken here

it just needed `make -C wasm_for_tests/wasm_source ` (can be done on merge)

PR #2634, opened 2024-02-15
Fix the token address when shielded transfer back to Namada

yito88:
## Describe your changes
When trying to transfer a token back to the source Namada, the token address was always `IbcToken` in `ibc-gen-shielded`. If the destination Namada is the origin of the token, the address should be an established address.
And, added a test to transfer back over IBC shielded transfer.

`v0.31.4`


sug0 (approved):
lgtm. hm, but wouldn't this also be an issue for non-masp transfers? any namada to namada transfers (regardless of being shielded or not) would be affected I think.

sug0:
btw, rebase on `origin/base` again to get code coverage to run (thereby fixing CI)

yito88:
@sug0 This issue doesn't happen in non-masp transfers because `received_ibc_token` is called only in `ibc-gen-shielded` for now. The prefix is handled in `ibc-rs` for actual transfers.

grarco (approved):

Issue #2624, opened 2024-02-15
Add check in ``Update-account``  to not allow increasing ``--threshold above number`` of active public keys


spidey-169:
## I think I have found a bug which renders a Multisig account unusable. 
This is when trying to increase the number of threshold for a Multisignature account. I created a multi-signatuere account with a single signer and then went on to do txns with it. Then I updarte the number of threshold of the account to 2 and this action though it shouldn't be allowed, went through. Now I cannot use this account at all for any transactions, as I can send no transactions through it as **threshold < total public keys** is violated. This is SEVERE as the user cannot access his funds in this wallet.

## DESCRIPTION:

The description for ``update-account -h`` says

``--threshold <threshold>
          The minimum number of signature to be provided for authorization. Must be less then the maximum number of public keys provided ``

To check, I created a multisig wallet with a single signer using this

```
namada client init-account --alias $WALLET_MULTISIG --public-keys $WALLET --signing-keys $WALLET --memo $MEMO
```

Successful

```
Wrapper transaction hash: 18A56855E8324279A88E190BEDDE5D0C3F97C116AD3AE6C32E400E7E3890608F
Inner transaction hash: 50DC8E7C183619086114227D7DC0B6A6E25DF9411616BBB56874999D59C34912
```

Then I tried to update-this account to have a threshold of 2. Now since this has only 1 public keys, this shouldn't be allowed as **threshold < total public keys**

However I was able to do it using this transaction, where I am providing two wallet keys for signing.

```
namada client update-account --address $WALLET_MULTISIG --threshold 2 --signing-keys $WALLET,$WALLET_2 --memo  $MEMO
```

```
Transaction added to mempool.
Wrapper transaction hash: C468C810B9A0191AC4A1E3513A74035F2A1DE99F73D3448D9E61863F5200A496
Inner transaction hash: A285AFB43F8A063C6EFB17D67D091CF21EC09C44C1F01C16E790BFBDC4F95831
Wrapper transaction accepted at height 22514. Used 23 gas.
Waiting for inner transaction result...
Transaction was successfully applied at height 22515. Used 7130 gas.
```

**The problem is that now I have an account with a threshold of 2 and Only one public Key.**

## Every Transaction from this wallet is rejected no matter what

To confirm account details

``
namada client query-account --owner $WALLET_MULTISIG 
``

```
Address: tnam1q8veg8w2v0azkzetch3ev92y0d6rg3a0su2x297m
Threshold: 2
Public keys:
[...]

PR #2677, opened 2024-02-21
BugFix: Adding Client side check For Requiring threshold < Noof public keys in ``update-account``

spidey-169:
Adding Client side check For Requiring threshold < Noof public keys in ``update-account`` cli call.  

## Describe your changes

Throwing a panic error if condition isn't met so the user accidentally doesn't end up changing threshold to be above the number of public keys. This addresses the existing issue https://github.com/anoma/namada/issues/2624 where the user was able to change threshold to be greater than number of public keys and that caused the wallet to become unusable.

Currently only added the check in ``update-account``, as that is mainly used to update the threshold value.

Also minor grammatical error corrections to correctly use "than" instead of "then"

Based on the Bug pointed out in https://github.com/anoma/namada/issues/2624


phy-chain:
@spidey-169 I think you are missing a crate import and <T> is not correct I think, unless you are implementing some generic.
Try running `make build` on the project, it should give you pointer on how to fix

spidey-169:
> @spidey-169 I think you are missing a crate import and <T> is not correct I think, unless you are implementing some generic.
> 
> Try running `make build` on the project, it should give you pointer on how to fix

Thanks, I think the clippy build tools gave suggestions which I was trying to incorporate. Let me try using make build and see if I can identify the issues. Have gone through a few iterations to fix this already, let me see if I can correctly find it.

sug0:
favoring #2747 over this pr, therefore I'm closing it

PR #2682, opened 2024-02-21
Remove borsh schemas and the encoding spec

sug0:
## Describe your changes

Borsh schemas were only depended upon by the `namada_encoding_spec` crate. Therefore, we can remove schemas, which were an unstable feature in borsh, anyway.

`v0.31.5`


sug0:
might need to rebuild wasms for tests

tzemanovic (approved):

batconjurer (approved):

murisi:
Would keeping the `BorshSchema`'s in Namada be harmful? I ask because these schemas would be extremely useful in describing our serialized formats to external collaborators like Zondax. Providing them with the (automatically derived) schemas would reduce the time we spend manually specifying serializations and simultaneously ease collaborators' process of writing and auditing parsers for our transactions in languages like C. The other more niche use case for these schemas is to automate the construction of parsers like in: https://github.com/anoma/namada/pull/2732 .

sug0:
we'll be keeping borsh schemas in Namada, as they might be handy in the future

PR #2698, opened 2024-02-22
IBC rate limit

### LINKED ISSUE #2552: Overall and throughput-based IBC rate limits
We should have IBC rate limits for deposits into and withdrawals out of Namada, as follows:
- All asset denominations have a deposit limit (by default: 0, meaning that no deposits are allowed). No deposit which would increase the supply on Namada to more than the deposit limit is allowed.
- All asset denominations have a per-epoch net throughput limit (by default: 0, meaning that no throughput is allowed). This net throughput limit tracks the absolute value of `deposits - withdrawals` during the epoch. Deposits or withdrawals which would cause the throughput to exceed the limit are not allowed. Tracking net throughput is necessary to prevent DoS.

These limits serve to limit risk in the case of state machine bugs on Namada. [Context](https://github.com/osmosis-labs/osmosis/blob/main/x/ibc-rate-limit/README.md).

yito88:
## Describe your changes
closes #2552

#2631


crates/ibc/src/context/nft_transfer.rs
```diff
@@ -0,0 +1,372 @@
...
+        token: &Address,
+        is_minted: bool,
+    ) -> Result<(), NftTransferError> {
+        let mint = self.inner.borrow().mint_amount(token)?;
+        let updated_mint = if is_minted {
+            mint.checked_add(Amount::from_u64(1)).ok_or_else(|| {
+                NftTransferError::Other(
+                    "The mint amount overflowed".to_string(),
+                )
+            })?
```
sug0, line 53: realistically this error should be impossible, no? we can only hold 1 unit at a time. we should be checking that `mint` is equal to 0, then return `Amount::from_u64(1)` or smth

crates/ibc/src/context/nft_transfer.rs
```diff
@@ -0,0 +1,372 @@
...
+        } else {
+            mint.checked_sub(Amount::from_u64(1)).ok_or_else(|| {
+                NftTransferError::Other(
+                    "The mint amount underflowed".to_string(),
+                )
+            })?
```
sug0, line 59: inverse condition here

crates/namada/src/ledger/native_vp/ibc/mod.rs
```diff
@@ -209,6 +225,90 @@ where
...
         Ok(())
     }
+
+    fn check_limits(&self, keys_changed: &BTreeSet<Key>) -> VpResult<bool> {
+        let mut tokens: Vec<&Address> = keys_changed
+            .iter()
+            .filter_map(|k| is_any_token_balance_key(k).map(|[key, _]| key))
+            .collect();
+        tokens.sort();
+        tokens.dedup();
```
sug0, line 235: collect into a `BTreeMap` or `HashSet` instead

crates/namada/src/ledger/native_vp/ibc/mod.rs
```diff
@@ -209,6 +225,90 @@ where
...
+                .ctx
+                .read_post(&mint_amount_key)
+                .map_err(Error::NativeVpError)?
+                .unwrap_or_default();
+            if mint_limit < minted {
+                return Err(Error::RateLimit(format!(
+                    "Transfer exceeding the mint limit is not allowed: Mint \
+                     limit {mint_limit}, minted amount {minted}"
+                )));
+            }
```
sug0, line 278: let's avoid doing storage reads directly, to facilitate isolating this kind of logic to the IBC crate. at the API level, we should only request something that implements `StorageRead` (which `ctx.pre()` does IIRC)

crates/namada/src/ledger/native_vp/ibc/mod.rs
```diff
@@ -209,6 +225,90 @@ where
...
+            let diff = if deposit < withdraw {
+                withdraw
+                    .checked_sub(deposit)
+                    .expect("withdraw should be bigger than deposit")
+            } else {
+                deposit
+                    .checked_sub(withdraw)
+                    .expect("deposit should be bigger than withdraw")
+            };
+            if throughput_limit < diff {
```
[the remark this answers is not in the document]
yito88: I think it checks the total value of them. Txs are executed and validated one by one. The `read` in each tx retrieves the value after the previous txs were applied from `block_write_log`. In this case, the first tx exceeding the limit will fail, but the previous txs will be committed.
https://github.com/anoma/namada/blob/d765bbf931519157986476dde4a65ead58e84b5c/crates/state/src/write_log.rs#L182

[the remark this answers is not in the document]
sug0: oh, I just failed to realized and `withdraw` and `deposit` contained. there shouldn't be any problem here

sug0 (changes requested):
benches are broken and I think at least one e2e test

crates/namada/src/ledger/native_vp/ibc/mod.rs
```diff
@@ -207,6 +222,63 @@ where
...
+                .read_post(&deposit_key)
+                .map_err(Error::NativeVpError)?
+                .unwrap_or_default();
+            let withdraw_key = withdraw_key(token);
+            let withdraw: Amount = self
+                .ctx
+                .read_post(&withdraw_key)
+                .map_err(Error::NativeVpError)?
+                .unwrap_or_default();
+            let diff = if deposit < withdraw {
```
cwgoes, line 263: This whole throughput calculation should be abstracted out into a function

cwgoes (commented):
This logic looks correct to me, one small nit.

sug0 (dismissed):

PR #2699, opened 2024-02-22
Murisi/simplify namada masp rewards

murisi:
## Describe your changes
Experimented with modifying the reward distribution algorithm so that native token rewards are distributed in the same way as non-native tokens. Doing this causes a slight variation in the shielded rewards distributed for the default set of parameters and hence necessitated slight changes in the expectations of the integration tests.

At a high level, these changes mean that now all native tokens with an epoch other than 0 receive their rewards as epoch 0 native tokens (just the same as it's done for non-native tokens). The exception to this rule is epoch 0 native tokens, which have a direct conversion to the current epoch (i.e. a conversion term like 100NAM@ep30 - 10NAM@ep0). Hence reward claiming for everything except NAM@ep0 involves claiming a NAM@ep0 reward and then converting that to NAM@epN.

The following list describes the benefits/drawbacks of this approach:
* Benefit: simpler/shorter reward distribution code (due to native token branch being fused with non-native branch)
* Benefit: ability to claim rewards even for small amounts of NAM
  * Note: this benefit is small because these small amounts are insiginificant after inflation adjustment
* Drawback: the rewards become significantly less accurate over time as a "denominator" approaches 0
  * This "denominator" is the negative term of the native token conversion from epoch 0 to the current epoch

#2695 


brentstone:
what's the status of this? Is it still desired or necessary?

murisi:
> what's the status of this? Is it still desired or necessary?

It may be necessary. I'll be looking at integrating it soon.

brentstone:
@murisi Is this PR still relevant or can it be closed?

PR #2735, opened 2024-02-27
Fix temporary values read

### LINKED ISSUE #2683: Wrong storage API in VPs
The read methods (`read`, `read_bytes` and `has_key`) return the value of a temp write if found. This means the VPs cannot tell the difference between an actual and a temporary storage write (also considered that we pass the entire set of changed keys to the vp, including temporary writes, but we don’t pass the `StorageModification`). Therefore a malicious tx could mock some storage changes with temporary writes, the VP would then deem the transaction valid but in the end the actual storage modifications won't match what the VP validated.

We have a `read_temp`  method on `VpEnv` that reads only the temporary modifications so I wonder if we should prevent them from being red from `StorageRead` and explicitly call `read_temp` when needed.

grarco:
## Describe your changes

Closes #2683.

Fixes the `has_key` method provided to the vp env to not return an error on temporary write values (the `read` method was already fine).

Adds a new `read_persistent` method on `WriteLog` that ignore temporary writes and modifies the `read` and `has_key` implementation of `impl_storage_read` to call this new method.
Modifies `tx_iter_next` to skip temporary writes.
Refactors `fetch_or_compile` to call `state.read` (which should include possible updated in the write log of the wasm codes).

Updates `write_temp` in `WriteLog` to return an error when trying to overwrite a `StorageModification::Write` with a `StorageModification::Temp`.
Updates `iter_prefix_post` in `WriteLog` to iterate also on the precommit bucket.
Updates `write`, `write_temp` and `delete` in `WriteLog` to take the precommit bucket into account.

v0.32.0

### DIFF
--- a/crates/namada/src/ledger/vp_host_fns.rs
+++ b/crates/namada/src/ledger/vp_host_fns.rs
@@ -188,7 +188,9 @@ where
             Ok(false)
         }
         Some(&write_log::StorageModification::InitAccount { .. }) => Ok(true),
-        Some(&write_log::StorageModification::Temp { .. }) => Ok(true),
+        Some(&write_log::StorageModification::Temp { .. }) => {
+            Err(RuntimeError::ReadTemporaryValueError)
+        }
         None => {
             // When not found in write log, try to check the storage
             let (present, gas) =
@@ -220,7 +222,9 @@ where
             Ok(false)
         }
         Some(&write_log::StorageModification::InitAccount { .. }) => Ok(true),
-        Some(&write_log::StorageModification::Temp { .. }) => Ok(true),
+        Some(&write_log::StorageModification::Temp { .. }) => {
+            Err(RuntimeError::ReadTemporaryValueError)
+        }
         None => {
             // When not found in write log, try to check the storage
             let (present, gas) =
--- a/crates/namada/src/ledger/vp_host_fns.rs
+++ b/crates/namada/src/ledger/vp_host_fns.rs
@@ -37,8 +37,6 @@ pub enum RuntimeError {
     NumConversionError(TryFromIntError),
     #[error("Memory error: {0}")]
     MemoryError(Box<dyn std::error::Error + Sync + Send + 'static>),
-    #[error("Trying to read a permanent value with read_temp")]
-    ReadPermanentValueError,
     #[error("Invalid transaction code hash")]
     InvalidCodeHash,
     #[error("No value found in result buffer")]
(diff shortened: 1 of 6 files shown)


grarco:
A thing I noticed is that here the type system doesn't really help us: the `block_write_log` will (and should) never contain a `StorageModification::Temp` because we prune them, but since we are not using a different type this isn't enforced leading to some places in the code were we might match on a variant that doesn't really exist. We could think about coming up with a new type here

https://github.com/anoma/namada/blob/78519dd174200998c3885abb1cd714f566c4a648/crates/state/src/write_log.rs#L84

crates/namada/src/ledger/vp_host_fns.rs
```diff
@@ -188,7 +188,9 @@ where
             Ok(false)
         }
         Some(&write_log::StorageModification::InitAccount { .. }) => Ok(true),
-        Some(&write_log::StorageModification::Temp { .. }) => Ok(true),
+        Some(&write_log::StorageModification::Temp { .. }) => {
```
grarco, line 191: The API we provide to the VPs is different to the one we have in protocol or expose to transactions. For VPs we report an error whereas in protocol and for transactions we simply ignore temporary writes and try to look for other storage modifications or directly from storage. I tend to like this latter but I can understand that returning an error might be more explicit. Should we try to unify the api and apply the same logic everywhere (either errors or read-through?)

tzemanovic: I agree that the latter is better and I'm for unifying this in VPs. If a VP allows temp values, it has to read them explicitly so ignoring them for non-temp reads is fine

crates/state/src/write_log.rs
```diff
@@ -168,6 +170,39 @@ impl WriteLog {
         }
     }
 
+    /// Read a non-temporary value at the given key and return the value and the
+    /// gas cost, returns [`None`] if the key is not present in the write
+    /// log
+    pub fn read_persistent(
+        &self,
+        key: &storage::Key,
+    ) -> (Option<&StorageModification>, u64) {
```
tzemanovic, line 179: I think in general this all looks great, but this function return type is a bit strange as in type in can return `StorageModification::Temp` so it's a bit weird that we then have to match on it after calling `read_persistent`. It will be a bit more verbose here but perhaps we should add e.g. another `enum PersistentModification`

tzemanovic (approved):

tzemanovic:
rebased on 0.32.0

Issue #2671, opened 2024-02-20
Does it make sense to be able to pass threshold=0 on init-account ?


phy-chain:
```
namada client init-account      \
    --threshold 0   \
    --alias test-account-to-update2   \
    --signing-keys validatorade     \
    --memo tpknam1qpayc0rxtuulknrvygcwvlhehuw2tvmxhvhxmjm8s43hxj3vdxjlv62jzcq
```
This works without any warning. What would be the consequences of having a 0 threshold account ? Possibility of not signing tx ?

sug0:
hm, this shouldn't make sense. the currently consequences seem to be that no signatures are required to validate an action within that account

phy-chain:
Yep, I confirm. Tested this morning, and you can also lower it to 0 through update-account : 
`namadac update-account --address test-establied-0 --threshold 0 --signing-keys $KEY_ALIAS --memo $NAMADA_MEMO --node http://rpc.namada.validatorade.org`

Indeed, no signing required afterward, I can transfer funds with this account without signing

phy-chain:
I have also fixed "init-validator" to enforce a threshold > 0

PR #2747, opened 2024-02-27
fix: client side check to prevent threshold=0 on init/update

phy-chain:
Adding Client side check to ensure `threshold` > 0 in `init-account` and `update-account `(also `init-validator`)

## Describe your changes

Accounts (and validator accounts) should not be created/updated with less than 1 as threshold, otherwise this would allow transaction requiring to signature.

I initially first reported this in #2671.

### DIFF
--- a/crates/apps/src/lib/cli.rs
+++ b/crates/apps/src/lib/cli.rs
@@ -4183,8 +4183,8 @@ pub mod args {
                 ))
                 .arg(THRESHOLD.def().help(
                     "The minimum number of signature to be provided for \
-                     authorization. Must be more than 0 and less than the maximum number of \
-                     public keys provided.",
+                     authorization. Must be greater than 0 and less than or equal \
+                     to the maximum number of public keys provided.",
                 ))
         }
     }
@@ -4458,8 +4458,8 @@ pub mod args {
                 ))
                 .arg(THRESHOLD.def().help(
                     "The minimum number of signature to be provided for \
-                     authorization. Must be more than 0 and less than the maximum number of \
-                     public keys provided.",
+                     authorization. Must be greater than 0 and less than or equal \
+                     to the maximum number of public keys provided.",
                 ))
         }
     }
@@ -4524,8 +4524,8 @@ pub mod args {
                 ))
                 .arg(THRESHOLD.def().help(
                     "The minimum number of signature to be provided for \
-                     authorization. Must be more than 0 and less than the maximum number of \
-                     public keys provided.",
+                     authorization. Must be greater than 0 and less than or equal \
+                     to the maximum number of public keys provided.",
                 ))
         }
     }
--- a/crates/apps/src/lib/cli.rs
+++ b/crates/apps/src/lib/cli.rs
@@ -4155,6 +4155,12 @@ pub mod args {
             let tx_code_path = PathBuf::from(TX_INIT_ACCOUNT_WASM);
             let public_keys = PUBLIC_KEYS.parse(matches);
             let threshold = THRESHOLD.parse(matches);
+
+            if threshold <= Some(0) {
+                eprintln!("Threshold must be higher than 0.");
+                std::process::exit(1);
+            }
+
             Self {
                 tx,
                 vp_code_path,
@@ -4177,7 +4183,7 @@ pub mod args {
                 ))
                 .arg(THRESHOLD.def().help(
                     "The minimum number of signature to be provided for \
-                     authorization. Must be less then the maximum number of \
+                     authorization. Must be more than 0 and less than the maximum number of \
                      public keys provided.",
                 ))
         }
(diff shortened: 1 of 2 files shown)


Fraccaman:
Thanks for the PR! I think the logic is almost correct, but I would say it better to check if the threshold is `<=` than the number of public-keys associated with the new account. Would you be able to make the change?

phy-chain:
@Fraccaman I think it's 2 differents issues. Checking if the threshold <= public-keys would not prevent it from begin 0, right ?

phy-chain:
Also, I can see that some of the CI check are not passing. Is it expected, or should I fix them ? (I've fixed the namada-release already, Im not used to Rust - make build works fine locally though)

Fraccaman:
to get CI to work, just rebase this branch on `base`. For the logic I think we can do something like;

Fraccaman:
for `update-account` the logic is more complex because you could modify only the threshold but we should check on the client that the new threshold won't "burn" the account. As an example, lets say that the current threshold for account X is 1, and the number of associated public keys is 1. If a user tries to modify the threshold of account X to 2, it should fail (without `--force`).

spidey-169:
> for `update-account` the logic is more complex because you could modify only the threshold but we should check on the client that the new threshold won't "burn" the account. As an example, lets say that the current threshold for account X is 1, and the number of associated public keys is 1. If a user tries to modify the threshold of account X to 2, it should fail (without `--force`).

Hi, looks like this is actually what I am trying to implement in my PR. Will try to see I can get this implemented for update-account and currently planning to throw and exception and exit if threshold > num-public-keys is ever detected

crates/apps/src/lib/cli.rs
```diff
@@ -4156,6 +4156,12 @@ pub mod args {
             let tx_code_path = PathBuf::from(TX_INIT_ACCOUNT_WASM);
             let public_keys = PUBLIC_KEYS.parse(matches);
             let threshold = THRESHOLD.parse(matches);
+
+            if threshold <= Some(0) {
```
sug0, line 4160: `threshold` is optional. if it happens to be equal to `None` (i.e. it wasn't provided as a CLI arg), I think this condition will be equal to `true`, effectively making the `threshold` required, but the error message somewhat inaccurate. it comes down to the way `Ord` is implemented for `Option<T>` (`None < Some(x)` for all `x`).

even if it was `Some`, I think it would make more sense to compare it for equality with `Some(0)`
```suggestion
            if threshold == Some(0) {
```

in the end, the better solution would be to make `THRESHOLD` return an `Option<NonZeroU8>` IMO, but this requires more code changes (not many to be fair, and you just have to follow the errors returned by `make clippy`)

crates/apps/src/lib/cli.rs
```diff
@@ -4507,7 +4525,7 @@ pub mod args {
                 ))
                 .arg(THRESHOLD.def().help(
                     "The minimum number of signature to be provided for \
-                     authorization. Must be less then the maximum number of \
+                     authorization. Must be more than 0 and less than the maximum number of \
```
sug0, line 4528:
```suggestion
                     authorization. Must be greater than 0 and less than or equal to the maximum number of \
```

crates/apps/src/lib/cli.rs
```diff
@@ -4447,7 +4459,7 @@ pub mod args {
...
```
sug0, line 4462:
```suggestion
                     authorization. Must be greater than 0 and less than or equal to the maximum number of \
```

crates/apps/src/lib/cli.rs
```diff
@@ -4178,7 +4184,7 @@ pub mod args {
...
```
sug0, line 4187:
```suggestion
                     authorization. Must be greater than 0 and less than or equal to the maximum number of \
```

sug0 (changes requested):

sug0:
also, please rebase on the latest `base` branch from `origin`
```
$ git fetch -p
$ git rebase origin/base
```

phy-chain:
> git remote add upstream https://github.com/anoma/namada

Am i not supposed to continue working on my fork ? Im not used to open source contributions, so I'd rather not mess up my branche :D

sug0:
> > git remote add upstream https://github.com/anoma/namada
> 
> Am i not supposed to continue working on my fork ? Im not used to open source contributions, so I'd rather not mess up my branche :D

you won't mess it up xd

since your changes are pretty trivial, you shouldn't run into any conflicts by rebasing. the `git remote-add` command adds a new location to fetch git commits/references from, since your forked branch won't have our latest releases.

crates/apps/src/lib/cli.rs
phy-chain: I've looked at the usages of threshold in the codebase and there are around 30 usages/definitions of `threshold: u8` or `threshold: Option<u8>`

I really have no idea which one are relevant for that. This seems like the cleaner solution indeed, but can you guide me into where it's relevant to fix ?
Im open to do pair programming on that if you wish / it's worth your time.

phy-chain: @sug0 Im not sure this is going the right way. I've tried since yesterday to replace most/all usage of u8 for threshold, with NonZeroU8, which leads to a lot of unwrapping/converting to compare with other number (most vec.len() ). And very time I fix one, I have 3 more errors that crop out.
Seems like some simple defensive programming at the entrypoint would work better.

brentstone:
Closing this in favor of #3154.

PR #2777, opened 2024-02-29
Refactored shielded sync to perform decryptions in parallel

batconjurer:
## Describe your changes
Shielded sync now tries to decrypt fetched masp txs in parallel


crates/sdk/src/masp.rs
```diff
@@ -609,10 +657,81 @@ pub struct ShieldedContext<U: ShieldedUtils> {
...
+}
+
+unsafe impl<'shielded, U: ShieldedUtils + MaybeSend + MaybeSync> Send
+    for ThreadHandle<'shielded, U>
+{
+}
+unsafe impl<'shielded, U: ShieldedUtils + MaybeSend + MaybeSync> Sync
+    for ThreadHandle<'shielded, U>
+{
+}
```
sug0, line 682: this might be the source of a nasty bug if `ShieldedUtils` is `!Send` or `!Sync` I think, no?

batconjurer: The main reason, this kind of thing isn't send or sync is because you could copy the &mut, or take the inner value. As long we never do that, this really shouldn't be an issue I think. The inner pointer is private and this thing should only be interacted with via the type api.

sug0:
I think this should be deployed on non-wasm targets. `rayon` is almost certainly not available on wasm

batconjurer:
> I think this should only be compiled on non-wasm targets. `rayon` is almost certainly not available on wasm

That is a probably true. Good point

crates/sdk/src/masp.rs
```diff
@@ -2353,20 +2515,32 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
             },
         );
         self.sync_status = ContextSyncStatus::Speculative;
+        for vk in &vks {
+            self.vk_heights.entry(*vk).or_default();
+        }
+        let ctx = ThreadHandle::new(self);
```
sug0, line 2521: instead of taking `self` directly, which is a `&mut` pointer, maybe steal the content of the pointer for the duration of these calls on `ctx`. then you don't have to wrap a `&mut` pointer with `Arc<Mutex<...>>` which is kind of hacky, and the `unsafe Send/Sync` impls will be safer.

crates/sdk/src/masp.rs
sug0: see https://github.com/anoma/namada/pull/2777#discussion_r1507568036

crates/sdk/src/masp.rs
batconjurer: I mean, I can easily make this change, but it won't be any safer. A type `T` is `send` if and only if `&mut T` is `send`.  I guess the real benefit is that in this particular case `T` is not `clone` but `&mut T` is,

batconjurer: Actually, this is basically impossible to change since I can't always get ownership of a shielded context from the NamadaImpl due to the RwLocks.

crates/sdk/src/masp.rs
```diff
@@ -753,33 +881,54 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-            std::mem::swap(&mut vk_heights, &mut self.vk_heights);
-            let _ = self.save().await;
         }
+        _ = self.utils.save(self).await;
+
+        let sync_status = self.sync_status;
+        let txs = logger.scan(unscanned);
+        let vk_heights = self.vk_heights.clone();
+        let ctx = ThreadHandle::new(self);
+        txs.par_bridge()
```
grarco, line 896: Is it possible to interrupt this process? Or the interrupt between the `update_witness_map` and the call to rayon here? Cause it seems to me like we could call `merge_scanned_data` out of order and maybe this could be a problem if the process gets interrupted cause we might have an inconsistent state of the shielded context?

brentstone:
Is this still relevant? @batconjurer @grarco

batconjurer:
> Is this still relevant? @batconjurer @grarco

Probably not as we are going to refactor the shit out of this now anyway

brentstone:
> > Is this still relevant? @batconjurer @grarco
> 
> Probably not as we are going to refactor the shit out of this now anyway

I'll close then. Feel free to re-open if you think it should stay that way.

PR #2779, opened 2024-02-29
Murisi/multi tx masp vp experiments

murisi:
## Describe your changes
Tried to increase the readability, strength, and generality of the MASP validity predicate whilst decreasing its verbosity. The changes are as follows:
* `Transaction` inputs and outputs that exceed pre-balances and post-balances respectively are no longer allowed.
* Now use the Sapling value balance in order to verify that movement in and out of the shielded pool is correct
* Now handle all `AssetType` decoding in one place before starting to check the transparent bundles
* The VP should now support `Transaction`s and unrelated `Transfer`s touching the same addresses/tokens
* Tried to remove the code that gave IBC minting transactions special treatment now that Sapling value balance is used
* Reduced dependency on the multitoken VP by directly checking the correctness of MASP VP balance changes
* Now check that all `Transaction`s and the balancing transaction are sufficiently authorized; prevents malleability

What would still need to be handled/fixed if this code turns out to be relevant:
* Overflows when trying to construct `ValueSum::new` objects
* Underflows when subtracting `Amount`s during the value balance computations in `verify_sapling_balancing_value`
* Add integration tests. Like checking that an unshielding `Transaction` to Bertha is embeddable inside a `Transfer` to Christel only if Bertha has signed the entire transaction (in addition to the usual MASP checks).

#2690 


crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -315,325 +300,272 @@ where
...
+                &changed_balances.tokens[&out.asset_type];
+            let amount =
+                token::Amount::from_masp_denominated(out.value, *digit);
+            *bal_ref = bal_ref
+                .checked_sub(
+                    &ValueSum::from_pair(token.clone(), amount).unwrap(),
+                )
+                .ok_or_else(|| {
+                    Error::NativeVpError(native_vp::Error::SimpleMessage(
+                        "Overflow in bundle balance",
```
grarco, line 469: This is probably underflow

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -315,325 +300,272 @@ where
...
+        {
+            let amount =
+                token::Amount::from_masp_denominated(out.value, *digit);
+            *bal_ref = bal_ref
+                .checked_sub(
+                    &ValueSum::from_pair(address.clone(), amount).unwrap(),
+                )
+                .ok_or_else(|| {
+                    Error::NativeVpError(native_vp::Error::SimpleMessage(
+                        "Overflow in bundle balance",
```
grarco, line 452: Underflow?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -315,325 +300,272 @@ where
...
+        {
+            let amount =
+                token::Amount::from_masp_denominated(vin.value, *digit);
+            *bal_ref = bal_ref
+                .checked_sub(
+                    &ValueSum::from_pair(address.clone(), amount).unwrap(),
+                )
+                .ok_or_else(|| {
+                    Error::NativeVpError(native_vp::Error::SimpleMessage(
+                        "Overflow in bundle balance",
```
grarco, line 379: Underflow?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -315,325 +300,272 @@ where
...
+                // Otherwise note the contribution to this transparent input
+                let amount =
+                    token::Amount::from_masp_denominated(vin.value, *digit);
+                *bal_ref = bal_ref
+                    .checked_sub(
+                        &ValueSum::from_pair(token.clone(), amount).unwrap(),
+                    )
+                    .ok_or_else(|| {
+                        Error::NativeVpError(native_vp::Error::SimpleMessage(
+                            "Overflow in bundle balance",
```
grarco, line 407: Underflow?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -676,25 +633,61 @@ where
...
+                crate::account::public_keys_index_map(&self.ctx.pre(), signer)?;
+            let threshold = crate::account::threshold(&self.ctx.pre(), signer)?
+                .unwrap_or(1);
+            let max_signatures_per_transaction =
+                crate::parameters::max_signatures_per_transaction(
+                    &self.ctx.pre(),
+                )?;
+            let mut gas_meter = self.ctx.gas_meter.borrow_mut();
+            if tx_data
+                .verify_signatures(
```
grarco, line 676: I'm not completely sure about this. I tend to believe that this kind of validation should be up to the other vps and not the masp. Also this implies that we'd validate signatures more than once which I don't believe is necessary or we might look for signatures which are not there (someone might come up with a vp that does not requires signatures, bad idea but you know...). In general I'd remove this entire `signers` thing

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -285,27 +280,17 @@ where
...
-        for (token, _) in masp_balances {
-            // NOTE: no need to extract the changes of the masp balances too,
-            // we'll examine those of the other transparent addresses and we'll
-            // check the match with the transparent bundle. The rest of the
-            // validation is up to the multitoken vp
-            result.masp.insert(token);
-        }
+        let counterparts_balances: Vec<_> = keys_changed
+            .iter()
+            .filter_map(is_any_token_balance_key)
```
grarco, line 285: Does this also include the ibc minted key?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -315,325 +300,272 @@ where
...
+// using the decodings in tokens and conversion_state for assistance.
+fn verify_sapling_balancing_value(
+    pre: &ValueSum<Address, Amount>,
+    post: &ValueSum<Address, Amount>,
+    sapling_value_balance: &I128Sum,
+    target_epoch: Epoch,
+    tokens: &BTreeMap<AssetType, (Address, token::Denomination, MaspDigitPos)>,
+    conversion_state: &ConversionState,
+) -> Result<bool> {
+    let mut acc = pre.clone();
```
grarco, line 530: Maybe we can directly pass in a mutable owned version of `pre`

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -665,9 +597,34 @@ where
...
-        let changed_balances =
+        let mut changed_balances =
             self.validate_state_and_get_transfer_data(keys_changed)?;
 
+        let masp_address_hash = TransparentAddress(<[u8; 20]>::from(
+            ripemd::Ripemd160::digest(sha2::Sha256::digest(
+                &Address::Internal(Masp).serialize_to_vec(),
+            )),
+        ));
+        if !verify_sapling_balancing_value(
```
grarco, line 608: Is it ok to run this check only on the balance of the masp? Shouldn't we also check that shieldings/unshieldings happened from/to the addresses declared in the transparent bundle?

murisi: I believe/hope that this should be okay, though a second pair of eyes would be appreciated...

So let `a` be the state changes being checked by the VP. `a` can be represented as a pair whose first element is all account balances before the `Tx` is applied, and whose second element is all account balances after the `Tx` is applied. Let `b` be the transformation induced by the inner `Transaction`.  `b` can be represented as a pair whose first element is all the transparent inputs of the `Transaction`, and whose second element is all the transparent outputs of the `Transaction`. Let `a=b`, `a-b`, `a+b` represent element-wise equality, subtraction, and addition respectively.

The old algorithm in `main` prevents malleability by essentially checking that `a=b`. I.e. `Transfer::source` has the same address as the `Transaction::vin`s, that `Transafer::amount` is the sum of the `Transaction::vin`s, that `Transfer::token` is the token of the `Transaction::vin`s, and so on.

This PR instead does the following. It splits the validation of the state changes into two parts: one part is the validation of state changes caused by the inner `Transaction`, the other part is the validation of state changes not caused by the inner `Transaction`. The idea is that is that if both parts pass the validation, then the entire `Tx` must pass the validation.

So algebraically we first check that `a>=b`, i.e. that the inputs and outputs of the inner `Transaction` do not exceed the account balances before or after the entire `Tx`'s application. Then we break the entire `Tx` into two components using the identity `a=b+(a-b)` and check `b` separately from `a-b`. `b` is essentially the MASP `Transaction`, so we check that its transparent transaction value pool is exactly 0, that each transparent input is signed/authorized, and all the other cryptographic checks. As for `a-b`, we essentially want to check that all the VPs would accept the state change described by `a-b` (which represents all state changes excluding those induced by the inner `Transaction`). Since I don't think the latter check is possible, this PR instead ensures that the accounts whose balances would decrease in `a-b` (i.e. those for which their balance in the pair's second element is less than their balance in the first element) have also signed the entire `Tx`.

So back to your question: instead of checking that unshieldings happen to the addresses given in the transparent bundles, we merely check that all accounts that whose balances go down in `a-b` have signed off on the transaction. So if `a` represents an unshielding of 50 BTC to Alice and `b` represents an unshielding of 50 BTC to Alice, then `a-b=0` and no additional authorization/signing is required. But if an attacker embeds the `Transaction` inside a different `Tx` so that `a` is now an unshielding of 50 BTC to Bertha, then: `a-b` is a transfer of 50 BTC from Alice to Bertha (since `b+(a-b)=(unshield 50 BTC to Alice)+(transfer 50 BTC from Alice to Bertha)=(unshield 50 BTC to Bertha)=a`). Since Alice's balance goes down in `a-b`, the VP would check that the entire `Tx` has been signed/authorized by Alice (something an attacker would not be able to do, otherwise they would just steal the funds from Alice after she receives them) and reject everything.

grarco: Thanks for this summary, it helps a lot! I have another question, say that we have a tx embedding two transfers:

1. An unshielding to Alice
2. A transparent transfer from Bob to Alice

More specifically the modifications in storage are the following (assume just one token, NAM):

| Address  | Pre  | Post  | Delta  |
|---|---|---|---|
| MASP  | 10  | 9  | -1  |
| Alice  | 0  | 4  | +4  |
| Bob  | 5  | 2  | -3  |

So the unshielding moved 1 NAM to Alice and Bob transferred 3 NAMs to Alice. The check on `a>=b` should be fine in this case (I guess). Now, the only negative diff is Bob's one so we require his signature on the tx. This is fine but now my question: could Bob redirect the unshielding from Alice to himself? So that the new changes in storage are:

| Address  | Pre  | Post  | Delta  |
|---|---|---|---|
| MASP  | 10  | 9  | -1  |
| Alice  | 0  | 3  | +3  |
| Bob  | 5  | 3  | -2  |

I guess the `a>=b` is still fine and the only negative diff is still Bob's one, so we require only Bob's signature which Bob can provide. Is this correct?

murisi: Yes, only Bob's signature would be required. 2 NAM would be transferred from Bob's account to Alice's, and 1 NAM would be unshielded from the MASP to Alice. So from the perspective of the algorithm, Bob merely sent 1 less NAM to Alice - he did not appropriate/seize Alice's 1 NAM. Does this answer your question?

crates/namada/src/ledger/native_vp/masp.rs
murisi: You're right, I think this a weakness/limitation of this approach. In the past, we assumed that the inner `Transaction` was the "same" as the outer `Tx`, so if the other VPs accepted the `Tx` then we could assume that the inner `Transaction` was also accepted. But now that the inner `Transaction` can be completely different from the outer `Tx`, I'm not sure how to validate `b` or `a-b` without either directly checking the signatures or somehow calling the other VPs on them...

grarco: The vps of the accounts involved should be called anyway right? And the accounts that saw a decrease in their balance will check the signature. If we want to be extra safe maybe we can just call `insert_verifier` in the loop, this way we enforce that the appropriate vps will verify the tx and we'll leave the signature check (if any) up to them

grarco: Ignore the last comment, this is resolved per discussion

crates/namada/src/ledger/native_vp/masp.rs
grarco: Sounds good, one note, currently the replay protection mechanism has some logic based on the presence of valid or invalid signatures that we are not handling or native vps cause those usually don't check signatures, so we'll need to update that too

PR #2787, opened 2024-03-03
Type safe event wrappers

sug0:
## Describe your changes

Slight refactor of the events code in Namada. Now, we emit events with type safe wrappers around the events' attributes.

`v0.31.8`


crates/core/src/event/extend.rs
```diff
@@ -0,0 +1,242 @@
...
+            attrs.insert("log".to_string(), "dejavu".to_string());
+            attrs
+        };
+
+        let mut base_event = Event::applied_tx();
+        base_event.extend(
+            WithDeferredData::begin(WithLog("dejavu".to_string()))
+                .compose(WithLog("dejavu".to_string()))
+                .compose(WithLog("dejavu".to_string()))
+                .compose(WithLog("dejavu".to_string())),
```
grarco, line 201: Should we also add a test where we calle compose with different `WithLog` messages and we check that only the last one si committed?

brentstone (approved):

crates/core/src/event.rs
```diff
@@ -13,13 +15,30 @@ use crate::ibc::IbcEvent;
...
 pub trait EmitEvents {
-    /// Emit an event
-    fn emit(&mut self, value: Event);
+    /// Emit a single [event](Event).
+    fn emit<E>(&mut self, event: E)
+    where
+        E: Into<Event>;
+
+    /// Emit a batch of [events](Event).
+    fn emit_many<B, E>(&mut self, event_batch: B)
```
grarco, line 24: Since we have only two implementation of this trait, and one of the two overrides this default, I wonder if we should avoid this default here and maybe implement this one on `Vec<Event>` using `std::iter::Extend::extend`

grarco (approved):
Minor comments but looks good to me!

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -192,15 +204,16 @@ where
...
-            {
-                let mut tx_event = new_tx_event(&tx, height.0);
-                tx_event["code"] = processed_tx.result.code.to_string();
-                tx_event["info"] =
-                    format!("Tx rejected: {}", &processed_tx.result.info);
-                tx_event["gas_used"] = "0".into();
-                response.events.push(tx_event);
+            if result_code != ResultCode::Ok {
+                response.events.emit(
+                    new_tx_event(&tx, height.0)
```
batconjurer, line 209: not a deal breaker, but since everything begins with "with", why not just 
```
new_tx_event(&tx, height.0)
    .with(ResultCode(result_code)
    .with(Info(format!("Tx rejected: {}", &processed_tx.result.info)
    .with(GasUsed(0.into()),
```

crates/core/src/event/extend.rs
```diff
@@ -0,0 +1,242 @@
+//! Extend [events](Event) with additional fields.
+
+use super::*;
+use crate::hash::Hash;
+use crate::storage::BlockHeight;
+
+/// Provides event composition routines.
+pub trait ComposeEvent {
+    /// Compose an [event](Event) with new data.
+    fn compose<NEW>(self, data: NEW) -> CompositeEvent<NEW, Self>
```
batconjurer, line 10: Why do we need this? It seems like we could just use the `extend` method instead.

crates/core/src/event/extend.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// Begin composing a batch of [`ExtendEvent`] implementations.
+    pub const fn begin(data: THIS) -> Self {
+        Self {
+            this: data,
+            next: WithNoOp,
+        }
+    }
+}
+
+impl<THIS, NEXT> WithDeferredData<THIS, NEXT> {
```
batconjurer, line 141: I really don't understand the need for this and I find it hard to read. It seems this could also be handled with extends.

crates/core/src/event/extend.rs
sug0: to preserve the type of the base events. it will come in handy in a latter pr

crates/core/src/event/extend.rs
sug0: same reasoning here, albeit I admit the usefulness of `WithDeferredData` might be way more questionable lmao

crates/tx/src/event.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+        _ => unreachable!(),
+    };
+    base_event
+        .compose(WithBlockHeight(height.into()))
+        .compose(WithLog(String::new()))
+        .into()
+}
+
+/// Extend an [`Event`] with info data.
```
sug0, line 33:
```suggestion
/// Extend an [`Event`] with result code data.
```

sug0:
**TODO:**
- https://github.com/anoma/namada/pull/2787#discussion_r1510284251
- https://github.com/anoma/namada/pull/2787#discussion_r1510282745

crates/gas/src/event.rs
```diff
@@ -0,0 +1,17 @@
+//! Gas related events.
+
+use namada_core::event::extend::ExtendEvent;
+use namada_core::event::Event;
+
+use super::Gas;
+
+/// Extend an [`Event`] with gas used data.
+pub struct WithGasUsed(pub Gas);
```
batconjurer, line 9: tiny: Just call it `GasUsed`

batconjurer (approved):
One nit

tzemanovic (approved):

crates/gas/src/event.rs
sug0: did this in another branch. is it fine?

PR #2803, opened 2024-03-04
Refactor governance

Fraccaman:
## Describe your changes
- [x] Remove offline governance
- [x] Introduce `DefaultWithWasm` proposal type


crates/proof_of_stake/src/lib.rs
```diff
@@ -303,6 +303,7 @@ where
         amount.change(),
         current_epoch,
```
Fraccaman, line 304: thinking about this, I think we want to check if the validator will be jailed between `current` and `proposal_end_epoch`

Issue #2800, opened 2024-03-04
No validation of size or format on CLI parameters for the validator metadata


phy-chain:
CLI endpoint change-metadata does not seem to enforce any format or size limit for the fields email and discord handle.
Given that these fields are displayed in a lot of webapps, it could break their UI or lead to security issues maybe.

Discord actually has some rules on what's allowed in that field. Same for emails, their is a RFC for that.

Transaction is accepted :

brentstone:
@phy-chain feel free to leave any comments on #2845. Figure we don't need anything complicated, can restrict it quite a bit, and character length is sufficient vs data size.

PR #2805, opened 2024-03-04
fix: add validation on email and discord handle in cli

phy-chain:
This has been reported in : #2800 

- [ ] Add a test for email validation
- [ ] Add a test for discord handle validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4553,6 +4553,7 @@ dependencies = [
  "pretty_assertions",
  "proptest",
  "proptest-state-machine",
+ "regex",
  "serde 1.0.193",
  "test-log",
  "thiserror",
--- a/crates/apps/src/lib/cli.rs
+++ b/crates/apps/src/lib/cli.rs
@@ -17,6 +17,7 @@ use clap::{ArgGroup, ArgMatches, ColorChoice};
 use color_eyre::eyre::Result;
 use namada::io::StdIo;
 use utils::*;
+use regex::Regex;
 pub use utils::{safe_exit, Cmd};
 
 pub use self::context::Context;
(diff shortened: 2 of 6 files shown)


brentstone:
@phy-chain thanks for starting this. It is more correct and appropriate to enforce these constraints in the user VP (`vp_user.rs`) rather than exclusively in the client, however.

We do a similar thing already in the native Governance VP with the restriction of the size of a governance proposal content. Check out `is_valid_content_key` insde the Gov VP (`crates::namada::src::ledger::governance::mod.rs`) for some guidance.

phy-chain:
> @phy-chain thanks for starting this. It is more correct and appropriate to enforce these constraints in the user VP (`vp_user.rs`) rather than exclusively in the client, however.
> 
> We do a similar thing already in the native Governance VP with the restriction of the size of a governance proposal content. Check out `is_valid_content_key` insde the Gov VP (`crates::namada::src::ledger::governance::mod.rs`) for some guidance.
> 
> Make sure to include a test as well.

Ah yes, I did see that, That was my source of inspiration for the MetadataValidation structure actually. But I didnt get how that worked with VP thing.
Ok for the test, I'll keep that in mind to implement it

brentstone:
@phy-chain I'm happy to help at some point soon since this is important, but I don't have the bandwidth for probably a couple days.

phy-chain:
> @phy-chain I'm happy to help at some point soon since this is important, but I don't have the bandwidth for probably a couple days.

@brentstone No rush, Im glad to help, even though Im a total newby at Rust (Im a frontend dev first and foremost). 
I've pushed a new commit to try and go in the right direction. Let me know if it looks better or if Im heading into troubles :D 

I've started a test case but on that, it's a complete mystery how to generate "test data" that's relevant, and test against it. So, i've copied and existing test and added some TODOs as a reflexion basis. 

Feel free to ping me when you have time (better to do so on Discord cause I might not see notifications here). It can also be after the SE if you're too busy.

brentstone:
hey @phy-chain, in some planning today we decided that the priority for this is higher than previously thought, so I went ahead and made a quick PR #2845 to complete this. I'll close this PR for now, but perhaps keep your branch around in case. Thanks for the effort and for identifying #2800!

phy-chain:
Lol ok, that's a way to go at it :sweat_smile: 500 chars limit for all. Still better than nothing I guess :D 
Feel free to keep that around and ping me if need be.

PR #2814, opened 2024-03-05
Derive global map of type deserializers

batconjurer:
## Describe your changes
This PR adds a proc macro that registers the deserializer of a type in a hashmap. This allows us to verify that data blob deserializes as correctly if where are in possession of the hash map key.


crates/account/src/lib.rs
```diff
@@ -10,13 +10,22 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
 pub use storage_key::*;
 pub use types::*;
 
 #[derive(
-    Debug, Clone, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+    Debug,
+    Clone,
+    BorshSerialize,
+    BorshDeserialize,
+    BorshDeserializer,
```
sug0, line 26:
```suggestion
```
the macro is only derived if the `migrations` feature is activated

crates/account/src/lib.rs
```diff
@@ -10,13 +10,22 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
+    Serialize,
+    Deserialize,
 )]
 /// Account data
```
sug0, line 30:
```suggestion
#[cfg_attr(feature = "migrations", derive(BorshDeserializer))]
/// Account data
```
this would need to be repeated throughout all the subsequent derives

crates/macros/Cargo.toml
```diff
@@ -15,10 +15,17 @@ version.workspace = true
 [lib]
 proc-macro = true
 
+[features]
+migrations = ["namada_migrations/migrations"]
+
```
sug0, line 20:
```suggestion

```
this might not need to be feature gated, I think

crates/macros/Cargo.toml
```diff
@@ -15,10 +15,17 @@ version.workspace = true
...
 [dependencies]
+namada_migrations = {path = "../migrations"}
```
sug0, line 22:
```suggestion
```
only the code emitted is using the dep, not us, directly, I think

crates/macros/src/lib.rs
```diff
@@ -332,6 +336,83 @@ where
...
+            "Cannot derive BorshDeserializer on a parameterized type. This \
+             can be done manually for concrete instantiations via the \
+             derive_borshdeserializer! macro."
+        );
+    }
+
+    let deserializer_ident =
+        syn::Ident::new(&format!("{}_DESERIALIZER", ident), Span::call_site());
+    quote!(
+        #[cfg(feature = "migrations")]
```
sug0, line 385:
```suggestion
```
shouldn't need feature gating here if you only derive when the feature is active with `cfg_attr`

crates/macros/src/lib.rs
```diff
@@ -332,6 +336,83 @@ where
...
+        };
+
+        impl namada_migrations::TypeHash for #ident {
+            const HASH: [u8; 32] = #hash;
+        }
+    )
+}
+
+#[allow(unused_macros)]
+macro_rules! derive_borshdeserializer {
```
sug0, line 401: if we promote `derive_borshdeserializer!()` to a procmacro, the hashing can also be computed with `sha2` and we can eliminate the `const-sha2` dep I think. moreoever, we'll also be able to hash the struct fields, and `derive_borsh_deserializer_inne` should be able to be implemented (partially) via `derive_borshdeserializer!()`

crates/migrations/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+#[cfg(feature = "migrations")]
+#[distributed_slice]
+pub static REGISTER_DESERIALIZERS: [fn()];
+
+pub trait TypeHash {
+    const HASH: [u8; 32];
+}
+
+/// Calls all of the regeistered callbacks which place type
+/// deserializes into [`TYPE_DESERIALIZERS`].
```
sug0, line 24:
```suggestion
/// Calls all of the registered callbacks which place type
/// deserializers into [`TYPE_DESERIALIZERS`].
```

crates/migrations/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+use lazy_static::lazy_static;
+#[cfg(feature = "migrations")]
+pub use linkme::distributed_slice;
+
+lazy_static! {
+    pub static ref TYPE_DESERIALIZERS: Mutex<HashMap<[u8; 32], fn(Vec<u8>) -> Option<String>>> =
+        Mutex::new(HashMap::new());
```
sug0, line 12:
```suggestion
/// Predicate that checks if an arbitrary byte array deserializes as some type `T`
/// erased inside of the callback. If the serialization is correct, the full path of `T`
/// is returned as a string (via [`std::any::type_name`]).
type CbFromByteArrayToTypeName = fn(Vec<u8>) -> Option<String>;

lazy_static! {
    pub static ref TYPE_DESERIALIZERS: Mutex<HashMap<[u8; 32], CbFromByteArrayToTypeName>> =
        Mutex::new(HashMap::new());
```

crates/macros/src/lib.rs
```diff
@@ -332,6 +336,83 @@ where
...
+             can be done manually for concrete instantiations via the \
+             derive_borshdeserializer! macro."
+        );
+    }
+
+    let deserializer_ident =
+        syn::Ident::new(&format!("{}_DESERIALIZER", ident), Span::call_site());
+    quote!(
+        #[cfg(feature = "migrations")]
+        #[namada_migrations::distributed_slice(REGISTER_DESERIALIZERS)]
```
sug0, line 386:
```suggestion
        #[::namada_migrations::distributed_slice(REGISTER_DESERIALIZERS)]
```

crates/macros/src/lib.rs
```diff
@@ -332,6 +336,83 @@ where
...
+        static #deserializer_ident: fn() = || {
+            let mut locked = namada_migrations::TYPE_DESERIALIZERS.lock().unwrap();
```
sug0, line 388:
```suggestion
            let mut locked = ::namada_migrations::TYPE_DESERIALIZERS.lock().unwrap();
```

crates/macros/src/lib.rs
```diff
@@ -332,6 +336,83 @@ where
...
+            locked.insert(#hash, |bytes| {
+                #ident::try_from_slice(&bytes).map(|val| format!("{:?}", val)).ok()
+            });
+        };
+
+        impl namada_migrations::TypeHash for #ident {
```
sug0, line 394:
```suggestion
        impl ::namada_migrations::TypeHash for #ident {
```

crates/merkle_tree/src/eth_bridge_pool.rs
```diff
@@ -10,6 +10,9 @@ use namada_core::hash::Hash;
 use namada_core::keccak::{keccak_hash, KeccakHash};
 use namada_core::storage;
 use namada_core::storage::{BlockHeight, DbKeySeg};
+use namada_macros::BorshDeserializer;
+#[cfg(feature = "migrations")]
+use namada_migrations::*;
```
sug0, line 15:
```suggestion
```
I think this might not be needed

sug0 (dismissed):

sug0 (dismissed):

.changelog/unreleased/improvements/2814-global-deserializer-map.md
```diff
@@ -0,0 +1,2 @@
+ - This PR adds a proc macro that registers the deserializer of a type in a hashmap. This allows us to verify that 
+   data blob deserializes as correctly if where are in possession of the hash map key. ([\#2814](https://github.com/anoma/namada/pull/2814))
```
tzemanovic, line 2:
```suggestion
   data blob deserializes correctly if we are in possession of the hash map key. ([\#2814](https://github.com/anoma/namada/pull/2814))
```

tzemanovic (approved):
nice use case for the dist slice!

PR #2627, opened 2024-02-15
Bat/feat/remove tx queue

batconjurer:
## Describe your changes
Instead of having every user tx be executed across two blocks, the first executing a wrapper and the second executing the main payload, this change makes it so that the entire tx is executed in a single block (or rejected).

Closes #2656

NOTE: This requires changes to hermes and so the ci will not pass until that is done.
## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #2819, opened 2024-03-05
Fix allowlist checks

grarco:


crates/tx/src/data/mod.rs
```diff
@@ -60,30 +60,24 @@ pub enum ResultCode {
...
+    InvalidVoteExtension = 10,
     /// Tx is too large
-    TooLarge = 14,
-    /// Decrypted tx is expired
-    ExpiredDecryptedTx = 15,
+    TooLarge = 11,
+    /// Tx code is not allowlisted
+    TxNotAllowlisted = 12,
     // =========================================================================
     // WARN: These codes shouldn't be changed between version!
```
tzemanovic, line 82: this is very breaking

tzemanovic: maybe we can mark the removed ones as deprecated instead

grarco: Ah I know, I still went for it because this PR is scheduled for some versions after 32.0 so it doesn't need to be compatible with the SE according to what we decided in the past few days

tzemanovic (approved):

tzemanovic:
I rebased on updated #2627

PR #2838, opened 2024-03-07
Runtime gas meter

grarco:


crates/namada/src/vm/wasm/run.rs
```diff
@@ -602,18 +604,215 @@ where
...
+struct GasRules;
+
+impl wasm_instrument::gas_metering::Rules for GasRules {
+    fn instruction_cost(
+        &self,
+        instruction: &wasm_instrument::parity_wasm::elements::Instruction,
+    ) -> Option<u32> {
+        // NOTE: costs set to 0 don't actually trigger the injection of a call
+        // to the gas host function (no useless instructions are
+        // injected) NOTE: none of these costs should be zero
```
tzemanovic, line 616: could we leave a note here about where we get the gas numbers?

tzemanovic (approved):

crates/namada/src/vm/wasm/run.rs
```diff
@@ -602,18 +604,217 @@ where
...
+        &self,
+        instruction: &wasm_instrument::parity_wasm::elements::Instruction,
+    ) -> Option<u32> {
+        // NOTE: costs set to 0 don't actually trigger the injection of a call
+        // to the gas host function (no useless instructions are
+        // injected)
+        // NOTE: these costs are taken from the benchmarks crate. None of them
+        // should be zero
+        let gas = match instruction {
+            Unreachable => 129_358,
```
sug0, line 620: this makes aborting txs quite expensive, right? in the current scheme, we abort erroneous tx execution with a `panic!()`, that is passed back to the host as the guest entering unreachable code, since we do not have stack unwinding enabled in wasm

crates/namada/src/vm/wasm/run.rs
```diff
@@ -1283,6 +1485,169 @@ mod tests {
...
+            &outer_tx,
+            &tx_index,
+            &addr,
+            &state,
+            &gas_meter,
+            &keys_changed,
+            &verifiers,
+            vp_cache.clone(),
+        );
+
```
sug0, line 1647: do we also test native vp out of gas errors?

wasm_for_tests/wasm_source/src/lib.rs
```diff
@@ -37,12 +37,41 @@ pub mod main {
...
+#[cfg(feature = "tx_infinite_guest_gas")]
+pub mod main {
+    use namada_tx_prelude::*;
+
+    #[transaction]
+    fn apply_tx(_ctx: &mut Ctx, _tx_data: Tx) -> TxResult {
+        let mut _a = 0u64;
+        loop {
+            _a += 1;
+        }
```
sug0, line 65: does execution halt if we only have an empty loop, like:
```rs
loop {}
```
??

sug0 (commented):
some comments

crates/namada/src/vm/wasm/run.rs
sug0: I've written a new test for this in https://github.com/anoma/namada/pull/2936

crates/benches/vps.rs
```diff
@@ -290,11 +146,11 @@ fn vp_implicit(c: &mut Criterion) {
             .write_log()
             .verifiers_and_changed_keys(&BTreeSet::default());
 
+        let gas_meter = RefCell::new(VpGasMeter::new_from_tx_meter(
+            &TxGasMeter::new_from_sub_limit(u64::MAX.into()),
+        ));
```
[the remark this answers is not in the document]
grarco: Correct but it shouldn't be a problem as long as we don't run out of gas (which should make the benchmark fail)

crates/namada/src/vm/wasm/run.rs
grarco: Well if you look at the other gas costs you'll see that a tx on average consumes some hundreds millions unit of gas (which we then reduce for the actual fee payment), so even a few hundred thousand is not that much in the end. Also, fees are paid upfront, so it shouldn't metter how much gas is consumed by unreachable since the tx will fail anyway

wasm_for_tests/wasm_source/src/lib.rs
grarco: Ah nice one. We do charge gas for branch instructions too so I'd say yes, but let me try to rework this tx (and the vp one) and see if the tests keep working even with empty loops

crates/namada/src/ledger/native_vp/ibc/context.rs
```diff
@@ -73,9 +75,18 @@ where
     fn read_bytes(&self, key: &Key) -> Result<Option<Vec<u8>>> {
         match self.store.get(key) {
             Some(StorageModification::Write { ref value }) => {
+                let gas = key.len() + value.len();
+                self.ctx
+                    .ctx
+                    .charge_gas(gas as u64 * MEMORY_ACCESS_GAS_PER_BYTE)?;
```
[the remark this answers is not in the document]
grarco: We do that in `tx_emit_ibc_event`

https://github.com/anoma/namada/blob/61a07598c35c4c8efd928b6e43124d4a7705ac45/crates/namada/src/vm/host_env.rs#L997-L1018

wasm_for_tests/wasm_source/src/lib.rs
grarco: Ok it seems to be working even with an empty loop, I've pushed the updated test wasms

sug0 (approved):
lgtm

PR #2845, opened 2024-03-07
Limit metadata size

### LINKED ISSUE #2800: No validation of size or format on CLI parameters for the validator metadata

brentstone:
## Describe your changes
Closes #2800.

v0.31.9


tzemanovic (dismissed):

sug0 (commented):
we should also validate genesis metadata
(`crates/apps/src/lib/config/genesis/transactions.rs`)

sug0 (approved):

PR #2846, opened 2024-03-07
Add a param to enable/disable nam transfers

### LINKED ISSUE #2842: Parameter to disable NAM transfers
We should add a governance-controlled parameter that enables/disables specifically native token (NAM) transfers. This can be accomplished by reading this parameter in the multitoken VP, and rejecting unless it is 1 (enabled).

yito88:
## Describe your changes
closes #2842 

`v0.31.9`

### DIFF
--- a/crates/parameters/src/lib.rs
+++ b/crates/parameters/src/lib.rs
@@ -64,6 +64,7 @@ where
         minimum_gas_price,
         fee_unshielding_gas_limit,
         fee_unshielding_descriptions_limit,
+        is_native_token_transferable,
     } = parameters;
 
     // write max tx bytes parameter
@@ -146,6 +147,11 @@ where
 
     let gas_cost_key = storage::get_gas_cost_key();
     storage.write(&gas_cost_key, minimum_gas_price)?;
+
+    let native_token_transferable_key =
+        storage::get_native_token_transferable_key();
+    storage
+        .write(&native_token_transferable_key, is_native_token_transferable)?;
 
     Ok(())
 }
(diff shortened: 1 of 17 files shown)


grarco (dismissed):
There's an issue with `make check-crates` but apart from that looks good to me, thank you!

crates/parameters/src/lib.rs
```diff
@@ -474,3 +488,33 @@ where
 pub fn native_erc20_key() -> Key {
     storage::get_native_erc20_key_at_addr(ADDRESS)
 }
+
+/// Initialize parameters to the storage for testing
+#[cfg(any(test, feature = "testing"))]
```
grarco, line 493: I believe `test` is not propagated to the dependency files and so we end up not finding this function

cwgoes (dismissed):
lgtm

Fraccaman:
will this prevent bonding? I think we need to be able to transfer nam to the PoS account if we need to bond.

cwgoes:
> will this prevent bonding? I think we need to be able to transfer nam to the PoS account if we need to bond.

Ah, good point - in that case we should allow transfers to (but not from) the PoS account.

brentstone:
> > will this prevent bonding? I think we need to be able to transfer nam to the PoS account if we need to bond.
> 
> Ah, good point - in that case we should allow transfers to (but not from) the PoS account.

More particularly, perhaps we should allow transfers between `Internal` addresses and `Implicit / Established` addresses, since we need the same functionality with Governance and then the MASP before transfers are enabled (and maybe others I'm not considering)

cwgoes:
> > > will this prevent bonding? I think we need to be able to transfer nam to the PoS account if we need to bond.
> > 
> > 
> > Ah, good point - in that case we should allow transfers to (but not from) the PoS account.
> 
> More particularly, perhaps we should allow transfers between `Internal` addresses and `Implicit / Established` addresses, since we need the same functionality with Governance and then the MASP before transfers are enabled (and maybe others I'm not considering)

I don't quite follow - why would we need to transfer NAM (specifically) to and from the MASP or governance, before NAM transfers are enabled?

cwgoes:
> Gov proposals require that the author send NAM to the gov address, which is kept if the proposal fails I believe (and refunded if passed). And if you want to shield NAM, don't we send it to the MASP address?

Governance proposal deposits are OK. We do not want to allow shielding NAM until transfers are enabled.

Issue #2900, opened 2024-03-14
Shielded sync improvement


brentstone:
Several possible improvements to be made to shielded sync

- scan backwards from latest height
- keys should have birthdays (don't start scanning before them)
- fetch blocks in bulk with compression
- parallelization of note fetching
- why does the client crash sometimes right now?

HackMD for planning: https://hackmd.io/kiob5_XEQw6M90hqcq4dZw#Index-Crawler--Server

Some related issues opened by others:
- #2905 
- #2957
- #2874
- #2711

phy-chain:
From what I've read on Discord, lots of crashes happen on machine without enough RAM. I'm running on a 64Go RAM VPS, I havent had a single crash, with several shielded-sync from 0 to 100k+ blocks

opsecx:
We're discussing amongst some of us in discord now. For me restarting the validator seemed to do the trick. For others it did not. Unsure if RAM-related but definitely a possibility. This is the error we get though:

Querying error: No response given in the query: 
0: HTTP error
1: error sending request for url (http://127.0.0.1:26657/): connection closed before message completed

Fraccaman:
are you guys using remote or local nodes to shield-sync ?

thousandsofthem:
remote node don't work at all - 0% sync and already getting errors. 5 minutes sync time at most, usually ~1min until error. always starts from scratch

Rigorously:
> remote node don't work at all - 0% sync and already getting errors. 5 minutes sync time at most, usually ~1min until error. always starts from scratch

I have had no problems fetching blocks from a remote node. Might depend on the node or network interface.

In my experience fetching blocks is the least slowest part of the process, because it is network I/O bound. Can it be optimized? Sure.

Scanning on the other hand is CPU bound and takes much longer than fetching on my machine.

Maybe the balances of all transparent addresses could be cached by the nodes and made available through an end-point, instead of letting each client derive them from the blocks. Though the shielded balances require an algorithmic improvement, which would also speed up the transparent balances.

opsecx:
> are you guys using remote or local nodes to shield-sync ?

Local. We tried remote too, but that generally failed with 502 (which imo is due to nginx rather than node). Was solved for me when restarting the validator. Another user had same success after first reporting the opposite.

Rigorously:
> Local. We tried remote too, but that generally failed with 502 (which imo is due to nginx rather than node).

You jinxed it! 

```
Fetched block 130490 of 144363
[...]

PR #2906, opened 2024-03-17
Update masp.rs

chimmykk:
Fetch block & sync faster upon new spending key

## Implement a mech so that it will fetch and re-sync with an example as provided

#2900 


cwgoes:
Thanks for the contribution here!

We've since embarked on a more serious rework of shielded sync, so closing this, but appreciated nonetheless.

PR #2931, opened 2024-03-21
Generate Tx Borsh schema

murisi:
## Describe your changes
Made the necessary changes to successfully generate an accurate `Tx` Borsh schema. This is useful for describing the transaction format to external collaborators. More specifically, the changes were as follows:
* Now using the latest MASP crate, which has the correct Borsh schemas
* Renaming certain structures to avoid name collisions in the Schema
* Adding an example to the repo allowing the generation of a `Tx` schema to a file

Note: Because the MASP crate update fixes serialization of a certain `Transaction` component in corner cases, this PR is breaking.

Namada v0.32.0

### DIFF
--- a/crates/core/src/key/common.rs
+++ b/crates/core/src/key/common.rs
@@ -36,12 +36,15 @@ use crate::{impl_display_and_from_str_via_format, string_encoding};
     BorshDeserializer,
     BorshSchema,
 )]
-pub enum PublicKey {
+pub enum CommonPublicKey {
     /// Encapsulate Ed25519 public keys
     Ed25519(ed25519::PublicKey),
     /// Encapsulate Secp256k1 public keys
     Secp256k1(secp256k1::PublicKey),
 }
+
+/// Public key
+pub type PublicKey = CommonPublicKey;
 
 const ED25519_PK_PREFIX: &str = "ED25519_PK_PREFIX";
 const SECP256K1_PK_PREFIX: &str = "SECP256K1_PK_PREFIX";
@@ -311,12 +314,15 @@ impl FromStr for SecretKey {
     BorshDeserializer,
     BorshSchema,
 )]
-pub enum Signature {
+pub enum CommonSignature {
     /// Encapsulate Ed25519 signatures
     Ed25519(ed25519::Signature),
     /// Encapsulate Secp256k1 signatures
     Secp256k1(secp256k1::Signature),
 }
+
+/// Signature
+pub type Signature = CommonSignature;
 
 impl string_encoding::Format for Signature {
     type EncodedBytes<'a> = Vec<u8>;
--- a/.changelog/unreleased/features/2931-generate-tx-schema.md
+++ b/.changelog/unreleased/features/2931-generate-tx-schema.md
@@ -1,0 +1,2 @@
+- Added a utility to generate Borsh schemas for external collaborators.
+  ([\#2931](https://github.com/anoma/namada/pull/2931))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3898,7 +3898,7 @@ dependencies = [
 [[package]]
 name = "masp_note_encryption"
 version = "1.0.0"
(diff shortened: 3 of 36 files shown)


crates/core/src/key/common.rs
```diff
@@ -22,6 +24,16 @@ use crate::ethereum_events::EthAddress;
...
 
+#[derive(BorshSchema)]
+enum CommonPublicKey {
+    /// Encapsulate Ed25519 public keys
+    #[allow(unused)]
+    Ed25519(ed25519::PublicKey),
+    /// Encapsulate Secp256k1 public keys
+    #[allow(unused)]
+    Secp256k1(secp256k1::PublicKey),
+}
```
sug0, line 35: hm I guess the only reason you created this enum is to avoid sharing a `PublicKey` alias with other public key types in the borsh schema definitions. however, this is an error prone and hacky approach IMO. I'd suggest either doing something like this:
```suggestion
#[derive(BorshSchema)]
#[repr(transparent)]
struct CommonPublicKey(PublicKey);
```
or, even better, using the attributes available to the `BorshSchema` derive macro: https://docs.rs/borsh/latest/borsh/derive.BorshSchema.html#4-borshschemaparams---field-level-attribute

crates/core/src/key/common.rs
```diff
@@ -43,6 +54,18 @@ pub enum PublicKey {
...
+    fn add_definitions_recursively(
+        definitions: &mut BTreeMap<Declaration, Definition>,
+    ) {
+        CommonPublicKey::add_definitions_recursively(definitions)
+    }
+
+    fn declaration() -> Declaration {
+        CommonPublicKey::declaration()
+    }
+}
```
sug0, line 67: if we go with the second approach, we can simply derive `BorshSchema` and remove this `impl` block, I think

crates/core/src/key/common.rs
```diff
@@ -295,6 +318,16 @@ impl FromStr for SecretKey {
...
 
+#[derive(BorshSchema)]
+enum CommonSignature {
+    /// Encapsulate Ed25519 signatures
+    #[allow(unused)]
+    Ed25519(ed25519::Signature),
+    /// Encapsulate Secp256k1 signatures
+    #[allow(unused)]
+    Secp256k1(secp256k1::Signature),
+}
```
sug0, line 329: similar suggestion

crates/core/src/key/common.rs
```diff
@@ -318,6 +350,18 @@ pub enum Signature {
...
+    fn add_definitions_recursively(
+        definitions: &mut BTreeMap<Declaration, Definition>,
+    ) {
+        CommonSignature::add_definitions_recursively(definitions)
+    }
+
+    fn declaration() -> Declaration {
+        CommonSignature::declaration()
+    }
+}
```
sug0, line 363: similar suggestion

sug0 (changes requested):
a couple of suggestions

crates/core/src/key/common.rs
sug0: nvm just realized the first solution wouldn't work :P we'd be using a recursive definition

murisi: Agreed - this was hacky and error prone. I just looked at the `BorshSchema` derive macro attributes. Probably I'm missing something, but the relevant attributes all seem to be at field level. So while it's possible to manipulate the schemas of the subcomponents of a `PublicKey`, it doesn't seem to be possible to modify, say, the declaration of a `PublicKey` itself. No?

That being said, I just pushed an approach of modifying the declaration alias without duplicating the `PublicKey` `struct`. The new approach just uses a type alias instead. Would this be acceptable?

sug0: > That being said, I just pushed an approach of modifying the declaration alias without duplicating the PublicKey struct. The new approach just uses a type alias instead. Would this be acceptable?

I think this is fine, yeah! We had a similar approach while developing the Ethereum Bridge, to avoid name collisions w/ borsh schemas.

sug0 (approved):

tzemanovic (approved):

PR #2940, opened 2024-03-21
Tiago/wasm errs ux

sug0:

`v0.31.9`


sug0:
`vp_stack_limiter` failing on x86_64 linux :|

.changelog/unreleased/improvements/2940-wasm-errs-ux.md
```diff
@@ -0,0 +1,4 @@
+- Change the return type of a VP's predicate function to a Result of unit to
+  some error. In case Namada users perform invalid state changes, they should
+  be met with more descriptive error messages explaining the cause their tx's
+  rejection. ([\#2940](https://github.com/anoma/namada/pull/2940))
```
tzemanovic, line 4:
```suggestion
- Change the return type of a VP's predicate function to a Result of unit or
  some error. In case Namada users perform invalid state changes, they should
  be met with more descriptive error messages explaining the cause of their tx's
  rejection. ([\#2940](https://github.com/anoma/namada/pull/2940))
```

crates/namada/src/ledger/governance/mod.rs
```diff
@@ -37,14 +38,12 @@ pub const MAX_PGF_ACTIONS: usize = 20;
...
-    NativeVpError(#[from] native_vp::Error),
-    #[error("Proposal field should not be empty: {0}")]
-    EmptyProposalField(String),
-    #[error("Vote key is not valid: {0}")]
-    InvalidVoteKey(String),
-}
+#[error("Governance VP error: {0}")]
+pub struct Error(
+    #[from]
+    #[source]
```
tzemanovic, line 44:
```suggestion
```
nit - implied by `#[from]`

tzemanovic (dismissed):

tzemanovic:
auto-squashed
- 9a1f98765b2675a0797c2c205634e372ddb28894
- a9b89cc6a716f5c12ac98a6de326e8a7631eaddd
- f205de83408c89f66539ba7679fbb5da940019e4

crates/namada/src/vm/host_env.rs
```diff
@@ -2003,17 +1991,20 @@ where
...
+        Ok(_) => Ok(()),
         Err(err) => match err {
             namada_tx::VerifySigError::Gas(inner) => {
-                sentinel.borrow_mut().set_out_of_gas();
                 Err(vp_host_fns::RuntimeError::OutOfGas(inner))
             }
-            namada_tx::VerifySigError::InvalidSectionSignature(_) => {
-                sentinel.borrow_mut().set_invalid_signature();
-                Ok(HostEnvResult::Fail.to_i64())
+            namada_tx::VerifySigError::InvalidWrapperSignature => {
```
grarco, line 1999: I believe this case can never occur since the wrapper signature is only verified in protocol, right?

crates/core/src/validity_predicate.rs
```diff
@@ -1,36 +1,32 @@
...
+}
 
-    /// Set the sentinel for an out of gas error
-    pub fn set_out_of_gas(&mut self) {
-        *self = Self::OutOfGas
+impl<T, E> VpErrorExtResult<T> for Result<T, E>
+where
+    E: core::fmt::Display,
+{
+    // FIXME: we can't erase gas errors!
```
grarco, line 17: Are we erasing them at the moment?

sug0: we are, but I found out accidentally that it doesn't matter. when we exit out of a vp (either native or wasm), we add back the gas used by the vp to the tx's gas meter, which will cause that meter to also run out of gas. in the end, it doesn't really matter for us to distinguish between gas errors and other kinds of errors

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -930,64 +936,47 @@ where
...
-                        result.rejected_vps.insert(addr.clone());
-                        // Don't push the error since this is just a flag error
-                    }
-                    _ => {
-                        result.rejected_vps.insert(addr.clone());
-                        result.errors.push((addr.clone(), err.to_string()));
-                    }
+            tx_accepted.map_or_else(
+                |err| {
+                    result.invalid_sig = err.is_invalid_section_signature(); // required by replay protection
```
grarco, line 964: I believe we could remove this `invalid_sig` field from `VpsResult`. I think it was there before because previously an invalid signature would lead to an `Ok(false)` outcome but if everything is an error now we might just push the error to `VpsResult.errors` and then look for invalid signatures in `finalize_block`

crates/core/src/validity_predicate.rs
grarco: I think this works in case of a vp exceeding the transaction's gas limit but does it work even if the error is a gas (u64) overflow? I believe in this case the second addition would add the result of the overflow which probably would not trigger another overflow and could go unnoticed (even though we can expect an overflow error to be rare)

tzemanovic:
ah, the `make test-wasm` are broken in vp_implicit and vp_user

tzemanovic:
> ah, the `make test-wasm` are broken in vp_implicit and vp_user

This was due to the change in host_env `fn vp_verify_tx_section_signature` that now returns an `Err` instead of `HostEnvResult` on invalid sig and because the `native_host_fn!` test helpers unwrap results of host env fns it leads to panics in tests - this is very hard to do differently because the function has to be `extern "C"`

fixed in 6cad262f7a4f7c5415e2c6bb49ac747d7998f46e - the tests now have to catch panics instead

tzemanovic (approved):

PR #2945, opened 2024-03-22
create basic disperse command for cli

colomvax:
## Describe your changes

This PR is a suggestion of addition for a new command in the CLI for Namada, which could be very useful for a data protection oriented blockchain.
It gets its inspiration from the "disperse" contract on Ethereum, which allows a user to send X to N addresses in one single transaction.

I know the implementation is not perfect/complete, because as it is, it sends N txs, instead of a single one, but Im not advanced enough in Rust or my comprehension of Namada codebase to go further than that. But I hope this provides a useful starter to add this command !

This would probably benefit from having its own VP and wasm, as it looks a lot like a simple transfer but to be able to run it in a single tx, it requires to wrap multiple tx inside the inner/wrapper txs. Not sure how to do that, way beyond me :smile: 

### DIFF
--- a/.changelog/unreleased/2945-create-disperse-command-cli.md
+++ b/.changelog/unreleased/2945-create-disperse-command-cli.md
@@ -1,0 +1,2 @@
+- Add a Disperse command in the CLI
+  ([\#2945](https://github.com/anoma/namada/pull/2945))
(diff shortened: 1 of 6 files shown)


cwgoes:
Very nice, thank you for the contribution! We've recently added batch transaction support (many inner transactions in one) - I'm wondering, would that address the use-case you have in mind here?

grarco:
I believe this was implemented in #3459

Fraccaman:
As @grarco said, this was implemented in https://github.com/anoma/namada/pull/3459, closing this! Thanks for the PR anyway!

PR #2946, opened 2024-03-22
Remove Epoch from WrapperTx

### LINKED ISSUE #2691: Remove `Epoch` from `WrapperTx`
Since we've abandoned DKG we could think to remove the `epoch` field of `WrapperTx` which has there to determine the decryption key to be used.

cpp-phoenix:
## Describe your changes
Closes #2691 

v0.34.0


crates/light_sdk/src/transaction/mod.rs
```diff
@@ -80,7 +80,7 @@ pub(in crate::transaction) fn attach_fee(
     fee: DenominatedAmount,
     token: Address,
     fee_payer: common::PublicKey,
-    epoch: Epoch,
+    _epoch: Epoch,
```
batconjurer, line 83: I think it's safe to remove this argument here.

crates/sdk/src/tx.rs
```diff
@@ -183,17 +183,15 @@ pub fn dump_tx<IO: Io>(io: &IO, args: &args::Tx, tx: Tx) {
 /// to it.
 #[allow(clippy::too_many_arguments)]
 pub async fn prepare_tx<C: crate::queries::Client + Sync>(
-    client: &C,
+    _client: &C,
```
batconjurer, line 186: this argument should also be removed

batconjurer (changes requested):
Couple of small things, and it obviously needs a rebase. Otherwise looks good.

Fraccaman:
@cpp-phoenix thanks for ur PR! Do u think u can think the issues in the code review and rebase on `base` ?

crates/sdk/src/masp.rs
```diff
@@ -2478,8 +2478,7 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
                         let mut transfer = None;
                         extract_payload(tx, &mut wrapper, &mut transfer)?;
                         // Epoch data is not needed for transparent transactions
-                        let epoch =
-                            wrapper.map(|x| x.epoch).unwrap_or_default();
+                        let epoch = Epoch(0);
```
grarco, line 2481: I believe in here we should extract the `Epoch` at which the transaction was applied

grarco: Ah nevermind, ignore my last comment

yito88:
We need to update Hermes with this change. E2E tests should pass with https://github.com/heliaxdev/hermes/tree/1.7.4-namada-2946.

crates/sdk/src/tx.rs
batconjurer: I don't know why this comment was marked as resolved. I would still like to have this argument removed. Especially since it removes the need for the generic, which should simplify other parts of the code base.

batconjurer (changes requested):

brentstone:
@batconjurer thx, I resolved the conversation because I thought I had remove the `_client` but somehow it slipped through again. Thanks for the note!

PR #2956, opened 2024-03-25
Replay protection refactor

### LINKED ISSUE #2922: Refactor replay protection
After #2627 the replay protection implementation can be simplified by removing the `ReProtStorageModification::Finalize` variant and the associated methods both in the write log and in storage.

grarco:
## Describe your changes

Closes #2922.

Since there's no more need to keep track and possibly act on tx cases from different block, this PR removes the `ReProtStorageModification` enum and uses a simple `HashSet` to keep track of the transaction hashes.

Also improves replay protection unit tests.

v0.33.0

### DIFF
--- a/crates/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/crates/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -2576,6 +2576,138 @@ mod test_finalize_block {
         }
     }
 
+    // Test two identical txs in the same block. The first one fails but doesn't
+    // write the hash (because of invalid signature). The second one must be
+    // able to execute and pass
+    #[test]
+    fn test_duplicated_tx_same_block_with_failure() {
+        let (mut shell, _, _, _) = setup();
+        let keypair = crate::wallet::defaults::albert_keypair();
+        let keypair_2 = crate::wallet::defaults::bertha_keypair();
+
+        let tx_code = TestWasms::TxWriteStorageKey.read_bytes();
+        let mut wrapper =
+            Tx::from_type(TxType::Wrapper(Box::new(WrapperTx::new(
+                Fee {
+                    amount_per_gas_unit: DenominatedAmount::native(1.into()),
+                    token: shell.state.in_mem().native_token.clone(),
+                },
+                keypair.ref_to(),
+                Epoch(0),
+                GAS_LIMIT_MULTIPLIER.into(),
+                None,
+            ))));
+        wrapper.header.chain_id = shell.chain_id.clone();
+        wrapper.set_code(Code::new(tx_code, None));
+        let key = Key::from(Address::from(&keypair_2.ref_to()).to_db_key())
+            .push(&"test".to_string())
+            .unwrap();
+        wrapper.set_data(Data::new(
+            TxWriteData {
+                key,
+                value: "test".as_bytes().to_vec(),
+            }
+            .serialize_to_vec(),
+        ));
+
+        let mut new_wrapper = wrapper.clone();
+        new_wrapper.update_header(TxType::Wrapper(Box::new(WrapperTx::new(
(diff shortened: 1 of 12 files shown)


crates/apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -36,9 +34,9 @@
 //!     - `epoch`: block epoch
 //!     - `address_gen`: established address generator
 //!     - `header`: block's header
-//! - `replay_protection`: hashes of processed tx
-//!     - `all`: the hashes included up to the last block
-//!     - `last`: the hashes included in the last block
+//! - `replay_protection`: hashes of processed tx for replay protection purposes
+//!     - `current/{hash}`: an hash included in the current block
```
batconjurer, line 38:
```suggestion
//!     - `current/{hash}`: a hash included in the current block
```

crates/apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -36,9 +34,9 @@
...
+//!     - `{hash}`: an hash included in previous blocks
```
batconjurer, line 39:
```suggestion
//!     - `{hash}`: a hash included in previous blocks
```

crates/apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -554,26 +544,9 @@ impl RocksDB {
...
-            let last_key = replay_protection::last_key(&hash);
-            // Restore "buffer" bucket to "last"
-            batch.put_cf(reprot_cf, last_key.to_string(), vec![]);
-
-            // Remove anything in the buffer from the "all" prefix. Note that
-            // some hashes might be missing from "all" if they have been
-            // deleted, this is fine, in this case just continue
-            let all_key = replay_protection::all_key(&hash);
-            batch.delete_cf(reprot_cf, all_key.to_string());
+        // Remove the "current" tx hashes
```
batconjurer, line 547: Don't we also need to remove the `replay_protection::key(_)` entries committed by the previous block as well?

grarco: I don't think so. Let's say we are at block 10 and we have tx1, tx2, tx3 committed in the previous blocks and tx4, tx5 committed in 10. Say that we want to rollback to 9, then we need to remove the transactions committed in 10 (tx4, tx5) which live under the `current` subkey.

We could think about bringing the tx committed in block 9 back under the new `current` subkey but I believe this is not needed because these keys are known anyway and the validation doesn't change if the hash lives under the `current` subkey or not. Also, replay protection is not merklized so this is not needed

brentstone:
is this still just awaiting a final review?

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -2564,25 +2558,143 @@ mod test_finalize_block {
...
+
+        let event = &shell
+            .finalize_block(FinalizeBlock {
+                txs: processed_txs,
+                ..Default::default()
+            })
+            .expect("Test failed");
+
+        // the merkle tree root should not change after finalize_block
+        let root_post = shell.shell.state.in_mem().block.tree.root();
```
batconjurer, line 2677: This is probably me forgetting details, but since the second tx should pass, wouldn't its hash be added to replay prot and change the merkle tree?

grarco: The reprot entries of the db are not merklized

batconjurer: ah right.

batconjurer (approved):

PR #3021, opened 2024-04-07
governance proposals end at the start of the designated Epoch

### LINKED ISSUE #2884: governance proposals end at the end of the designated Epoch instead of the start
the voting period for a governance proposal is not intuitive

for example, the Shielded Expedition's Prop266 governance proposal was set to run from Epoch 24 to Epoch 28, but didn't end until Epoch 29 started. it's confusing to me, because if i specify that i want a proposal to start at Epoch 24, it should not start at the end of Epoch 24, and so when i specify that i want the proposal to end at Epoch 28, i expect that it should be when Epoch 28 starts.

this should either be updated in the code or in the CLI

aarambh-audits:
closes #2884

### DIFF
--- a/crates/apps/src/lib/client/rpc.rs
+++ b/crates/apps/src/lib/client/rpc.rs
@@ -1271,7 +1271,7 @@ pub async fn query_proposal_result(
             (proposal_result, proposal_query)
         {
             display_line!(context.io(), "Proposal Id: {} ", proposal_id);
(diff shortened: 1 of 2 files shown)


brentstone:
Thanks for trying to tackle this. There are a handful of other needed changes needed that you've missed here. Going to include all necessary changes inside of #3087 and close this instead.

PR #3029, opened 2024-04-08
Functions to update token balance and supply

### LINKED ISSUE #2287: Replace ad-hoc token transfer code with storage API code
As part of the modularization effort of Namada, we should strive to replace all custom token transfer logic with the `storage_api` equivalents. Here is an example (from the `ethereum_bridge` crate):

```rs
/// Reads the `Amount` from key, applies update then writes it back
pub fn amount<D, H>(
    wl_storage: &mut WlStorage<D, H>,
    key: &storage::Key,
    update: impl FnOnce(&mut Amount),
) -> Result<Amount>
where
    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
    H: 'static + StorageHasher + Sync,
{
    let mut amount = super::read::amount_or_default(wl_storage, key)?;
    update(&mut amount);
    wl_storage.write_bytes(key, borsh::to_vec(&amount)?)?;
    Ok(amount)
}
```
Functionally, it is a more generic version of `token::transfer`, `token::credit_tokens` and friends.

### LINKED ISSUE #3003: Add functions to update balances and token supplies
This would make the code much more compact and cleaner in some places, rather than getting lots of balance keys and then calling `storage.write` all over the place. Would put these functions in `crates::trans_token::storage`.

brentstone:
## Describe your changes
- Closes #3003
- Closes #2287

v0.34.0

### DIFF
--- a/crates/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
+++ b/crates/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
@@ -23,8 +23,8 @@ use namada_parameters::read_epoch_duration_parameter;
 use namada_state::{DBIter, StorageHasher, WlState, DB};
 use namada_storage::{StorageRead, StorageWrite};
 use namada_trans_token::storage_key::{balance_key, minted_balance_key};
+use token::{burn_tokens, decrement_total_supply, increment_total_supply};
 
-use crate::protocol::transactions::update;
 use crate::storage::bridge_pool::{
     get_nonce_key, is_pending_transfer_key, BRIDGE_POOL_ADDRESS,
 };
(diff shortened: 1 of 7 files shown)


crates/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
```diff
@@ -162,48 +162,9 @@ where
...
-        tracing::debug!(
-            %native_werc20_supply_key,
-            ?balance,
-            "New value calculated",
-        );
-        Ok(())
-    })?;
+    let native_token = state.in_mem().native_token.clone();
+    token::transfer(state, &native_token, &BRIDGE_ADDRESS, receiver, *amount)?;
+    decrement_total_supply(state, &erc20_token_address(native_erc20), *amount)?;
```
sug0, line 167: I think the storage key this is updating is missing from the returned `BTreeSet` of changed keys, no?

sug0 (approved):
nice work! lgtm

crates/ethereum_bridge/src/protocol/transactions/ethereum_events/events.rs
brentstone: In the diff, I didn't remove any storage keys, and the returned `BTreeSet` still contains three storage keys changed, which I think is the right number. Unless you think there is a bug that existed before my changes?

sug0: nvm, the affected key is `native_werc20_supply_key` which is def in the returned sed

tzemanovic (approved):
neat!

PR #3032, opened 2024-04-09
Generalize the core API of events

sug0:
## Describe your changes

Erase protocol specific details from the core API of events in Namada. Submodules should be responsible for defining custom events, which can be converted into core events for emission and querying.

~~**NOTE:** IBC tests involving hermes are expected to break with this PR. [This release](https://github.com/heliaxdev/hermes/releases/tag/v1.7.4-namada-beta9-rc) of hermes fixes the e2e tests (thanks Yuji!).~~

`v0.34.0`


yito88:
I think there are some conflicts with `draft-mainnet` because I have changed to get the IBC info from an IBC message instead of an IBC event in #2631.

brentstone:
> I think there are some conflicts with `draft-mainnet` because I have changed to get the IBC info from an IBC message instead of an IBC event in #2631.

This is ok, since no more PRs are allowed in `draft-mainnet` and we are close to cutting 0.33. Just will have to rebase this PR after that.

crates/core/src/ethereum_structs.rs
```diff
@@ -29,13 +32,60 @@ use crate::keccak::KeccakHash;
     Serialize,
     Deserialize,
 )]
+// TODO: move to `namada_ethereum_bridge::event` or
```
batconjurer, line 35: is there any reason not to resolve this TODO in this PR?

sug0: some defs in core still depend on this I think. I tried moving this stuff before and it was a bit tricky

crates/core/src/event.rs
```diff
@@ -13,21 +14,86 @@ use thiserror::Error;
...
+/// ```
+#[macro_export]
+macro_rules! event_type {
+    ($($tt:tt)*) => {
+        $crate::event::EventType::new($crate::__event_type_impl!($($tt)*))
+    };
+}
+
+/// An event to be emitted in Namada.
+pub trait EventToEmit: Into<Event> {
```
batconjurer, line 66: I'm probably losing my mind, but I went to find the `From<IbcEvent> for Event` and couldn't find it anywhere

sug0: line 399 of this file

batconjurer: I thought the point of all this traits was to localize implementations to the appropriate crate. Seems funny to do this here.

sug0: it is yeah, but core still depends on `IbcEvent` :( eventually we want to move `IbcEvent` to the `ibc` crate

crates/core/src/event.rs
```diff
@@ -13,21 +14,86 @@ use thiserror::Error;
...
+    /// This may be used to group events of a certain kind.
+    const DOMAIN: &'static str;
+}
+
+impl EventToEmit for Event {
+    const DOMAIN: &'static str = "unknown";
+}
+
+impl EventToEmit for IbcEvent {
+    const DOMAIN: &'static str = "ibc";
```
yito88, line 78: The event check in Hermes assumes that all IBC events don't have a prefix.
The check is implemented independently from chain-specific implementation.
We could support these events in our fork Hermes, but it is difficult for the upstream.

sug0:
wip: fixing hermes so ibc e2e tests pass

yito88:
@sug0 Can you try https://github.com/heliaxdev/hermes/releases/tag/v1.7.4-namada-beta9-rc?

crates/sdk/src/masp.rs
```diff
@@ -924,23 +927,12 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-                        }
-                    })
-                    .ok_or_else(|| {
-                        Error::Other(
-                            "Missing required tx result in event".to_string(),
-                        )
-                    })?;
-                let result = TxResult::from_str(tx_result_str)
-                    .map_err(|e| Error::Other(e.to_string()))?;
+                let result = InnerTxAttr::read_from_event_attributes(
```
batconjurer, line 930: very naayyicce

crates/core/src/event/extend.rs
```diff
@@ -52,72 +157,364 @@ where
...
+/// Extend a [`HashMap`] of string to string with event attributed
+/// related methods.
+pub trait ExtendAttributesMap: Sized {
+    /// Insert a new attribute into a map of event attributes.
+    fn with_attribute<DATA>(&mut self, data: DATA) -> &mut Self
+    where
+        DATA: ExtendEventAttributes;
+}
+
+impl<A: AttributesMap> ExtendAttributesMap for A {
```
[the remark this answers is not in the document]
sug0: it's used in #3088 to extend the `QueryMatcher` with new attributes

crates/core/src/event/extend.rs
```diff
@@ -52,78 +177,421 @@ where
...
+        Self::read_opt_from_event_attributes(attributes)?.ok_or(
+            EventError::MissingAttribute(
+                <Self as EventAttributeEntry<'value>>::KEY,
+            ),
+        )
+    }
+}
+
+/// Read a raw (string encoded) attribute from an [event](Event)'s attributes.
+pub trait RawReadFromEventAttributes<'value> {
```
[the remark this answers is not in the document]
sug0: `RawReadFromEventAttributes` doesn't require the inner value of the attribute to implement `FromStr`. it's mostly separated to allow greater flexibility. from an API consumer's perspective, it's just a matter of importing an extra trait

yito88:
Can the CI be updated with the new Hermes by updating `.github/workflows/scripts/hermes.txt`?

sug0:
> Can the CI be updated with the new Hermes by updating `.github/workflows/scripts/hermes.txt`?

yeah for sure! thanks for the suggestion

sug0:
> Can the CI be updated with the new Hermes by updating `.github/workflows/scripts/hermes.txt`?

@yito88 done in [2127a43](https://github.com/anoma/namada/pull/3032/commits/2127a438eaa7e8a0ae21fcc344b2394a796d3942)

crates/core/src/event/extend.rs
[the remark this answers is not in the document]
batconjurer: Given all the traits flying around here, is there a concrete use case that makes this worth it, or can we simplify and re add later if it becomes necessary?

yito88 (dismissed):

Issue #1732, opened 2023-07-17
Refactor epoch pairs


brentstone:
The PoS code has lots of instances of an (Epoch, Epoch) being passed around, and it is not always intuitive as to what this epoch pair is exactly. It would be cleaner and enhance readability to instead pass around a struct, perhaps named `EpochBounds` or similar, that contains explicit `start` and `end` epoch fields that can be used for various operations.

PR #3050, opened 2024-04-10
Refactor epoch pairs

honeygoyal120:
## Describe your changes
The PoS code has lots of instances of an (Epoch, Epoch) being passed around, and it is not always intuitive as to what this epoch pair is exactly. It would be cleaner and enhance readability to instead pass around a struct, perhaps named EpochBounds or similar, that contains explicit start and end epoch fields that can be used for various operations.

#1732

### DIFF
--- a/crates/proof_of_stake/src/lib.rs
+++ b/crates/proof_of_stake/src/lib.rs
@@ -1381,7 +1381,7 @@ where
     }
 
     let mut unbonds_and_redelegated_unbonds: BTreeMap<
-        (Epoch, Epoch),
+        EpochBounds,
         (token::Amount, EagerRedelegatedBondsMap),
     > = BTreeMap::new();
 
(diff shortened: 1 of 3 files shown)


brentstone:
Hi, thanks for your interest in developing. I assume you've seen #1731, which started to address this issue #1732 a while ago on a much older version of the software. Properly addressing this issue will likely require lots of changes to the proof-of-stake code.

Feel free to ask questions if needed, though I may not be able to help guide you very much as we are quite busy working on other higher-priority development. This issue is pretty low-priority, which is why #1731 has become quite stale.

Also, under `Describe your changes` you can simply write `Closes #1732` instead of copying the text from #1732. Then, for the base of the PR, please put a release version or other PR (like 0.32.1 or whatever it is) rather than an issue.

If I see development on this PR become stale in the next week or two, I'll close it.

honeygoyal120:
Okay Thanks @brentstone

PR #3070, opened 2024-04-11
Add a panic when calling namadac reveal-pk and the passed public key is already revealed

### LINKED ISSUE #3069: namadac reveal-pk has to output a message when the public key for the address is already revealed
When namadac reveal-pk is called with passing a public key that has already been revealed, it doesn't output anything. There should be some indication.

TerzoMillenio:
## Describe your changes
Closes #3069
When the public key is already revealed, the program panics with message "Public key for address {} already revealed."


cwgoes:
Hi, thank you for the contribution! 🙌 - would you mind rebasing, and we'll include it in the next release?

Fraccaman:
@cwgoes we addresses this in #3245 I think

cwgoes:
Ah gotcha.

Issue #826, opened 2022-11-25
Audit the codebase for erroneous usage of architecture dependent types


sug0:
Values of type `usize` and many pointer types vary in representation, depending on the underlying architecture of the host system. During deterministic protocol steps, Namada shouldn't rely on expressions utilizing `usize`, unless we set an artificial upper limit on the values of these types. For instance, use `u32::MAX` as the max value of a `usize`; then, the access of index `i` of a vector `v` might look like `&v[i & (u32::MAX as usize)]`.

This is important, to preserve the safety properties of the BFT system (i.e. each node executes the same sequence of steps).

cwgoes:
We decided just to add a warning on non-64 bit systems for the time being.

PR #3074, opened 2024-04-12
core: checked arithmetics

### LINKED ISSUE #2555: Audit for all usage of unchecked arithmetic and update nomenclature
Looking through the codebase for unchecked arithmetic I still find quite a bit, e.g.:
- [here](https://github.com/anoma/namada/blob/4209a7413c21df3f9b1190189990f4146c00dda3/crates/core/src/types/dec.rs#L368)
- [here](https://github.com/anoma/namada/blob/4209a7413c21df3f9b1190189990f4146c00dda3/crates/core/src/types/uint.rs#L667)
- many more, just searching...

We should:
- Remove ALL support for unchecked arithmetic. We should never use it.
- Update all arithmetic to use `checked_sub`, `checked_add` etc. names (don't overload `+` / `-`) to make it clear that we're using checked arithmetic (and so a search for only `+` or `-`, for example, should return only internal implementation code instead decimal types and no usage other than that).

tzemanovic:
## Describe your changes

first part of #2555 addressing core crate

v0.34.0

### DIFF
--- a/crates/core/src/dec.rs
+++ b/crates/core/src/dec.rs
@@ -4,8 +4,6 @@
 //! precision.
 
 use std::fmt::{Debug, Display, Formatter};
-use std::iter::Sum;
-use std::ops::{Add, AddAssign, Div, Mul, Neg, Sub};
 use std::str::FromStr;
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
@@ -13,11 +11,11 @@ use eyre::eyre;
 use namada_macros::BorshDeserializer;
 #[cfg(feature = "migrations")]
 use namada_migrations::*;
-use num_traits::CheckedMul;
 use serde::{Deserialize, Serialize};
 
 use super::token::NATIVE_MAX_DECIMAL_PLACES;
-use crate::token::{Amount, Change};
+use crate::arith::{self, checked};
+use crate::token;
 use crate::uint::{Uint, I256};
 
 /// The number of Dec places for PoS rational calculations
@@ -99,7 +97,7 @@ impl Dec {
             Some(res) => {
                 let res = I256::try_from(res).ok()?;
                 if is_neg {
-                    Some(Self(-res))
+                    Some(Self(res.checked_neg()?))
                 } else {
                     Some(Self(res))
                 }
@@ -121,13 +119,17 @@ impl Dec {
     /// The representation of 1
     pub fn one() -> Self {
         Self(I256(
-            Uint::one() * Uint::exp10(POS_DECIMAL_PRECISION as usize),
+            Uint::one()
+                .checked_mul(Uint::exp10(usize::from(POS_DECIMAL_PRECISION)))
+                .expect("Cannot overflow"),
         ))
     }
 
     /// The representation of 2
     pub fn two() -> Self {
-        Self::one() + Self::one()
+        Self::one()
+            .checked_add(Self::one())
+            .expect("Cannot overflow")
     }
 
     /// Create a new [`Dec`] using a mantissa and a scale.
@@ -136,12 +138,15 @@ impl Dec {
             None
         } else {
             let abs = u64::try_from(mantissa.abs()).ok()?;
-            match Uint::exp10((POS_DECIMAL_PRECISION - scale) as usize)
+            // Cannot underflow
+            #[allow(clippy::arithmetic_side_effects)]
+            let scale_diff = POS_DECIMAL_PRECISION - scale;
+            match Uint::exp10((scale_diff) as usize)
                 .checked_mul(Uint::from(abs))
             {
                 Some(res) => {
                     if mantissa.is_negative() {
-                        Some(Self(-I256(res)))
+                        Some(Self(I256(res).checked_neg()?))
                     } else {
                         Some(Self(I256(res)))
                     }
@@ -152,11 +157,14 @@ impl Dec {
     }
 
     /// Get the non-negative difference between two [`Dec`]s.
-    pub fn abs_diff(&self, other: &Self) -> Self {
-        if self > other {
-            *self - *other
+    pub fn abs_diff(
+        &self,
+        other: Self,
+    ) -> std::result::Result<Self, arith::Error> {
+        if self > &other {
+            checked!(self - other)
         } else {
-            *other - *self
+            checked!(other - *self)
         }
     }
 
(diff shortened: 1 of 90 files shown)


cwgoes:
@tzemanovic Is the title of this PR still accurate? Looks like a bunch of crates beyond core are changed. Does this cover everything, everywhere or are there other sections still remaining to transition to checked arithemetic?

crates/core/src/dec.rs
```diff
@@ -285,17 +335,18 @@ impl TryFrom<String> for Dec {
...
-            None => Self::zero(),
-        }
+impl TryFrom<token::Amount> for Dec {
+    type Error = Error;
+
+    fn try_from(amt: token::Amount) -> std::result::Result<Self, Self::Error> {
+        let raw = I256::try_from(amt.raw_amount())
+            .map_err(|e| eyre!("Invalid raw amount: {e}"))?;
+        let denom = I256(Uint::exp10(
+            (POS_DECIMAL_PRECISION - NATIVE_MAX_DECIMAL_PLACES) as usize,
```
brentstone, line 345: I'm sure we're safe to do this operation but figured I'd point it out as possibly unchecked

crates/core/src/token.rs
```diff
@@ -670,17 +734,17 @@ impl From<Amount> for U256 {
     }
 }
 
-impl From<Dec> for Amount {
-    fn from(dec: Dec) -> Amount {
+impl TryFrom<Dec> for Amount {
+    type Error = ();
```
brentstone, line 738: why `()` for the `Error` type?

crates/core/src/token.rs
```diff
@@ -870,7 +800,7 @@ pub enum AmountParseError {
         "Error decoding token amount, too many decimal places: {0}. Maximum \
          {1}"
     )]
-    ScaleTooLarge(u32, u8),
+    ScaleTooLarge(usize, u8),
```
brentstone, line 803: should we avoid this in light of #826?

crates/core/src/token.rs
```diff
@@ -1032,6 +965,78 @@ pub mod testing {
...
+        type Output = Self;
+
+        fn sub(self, rhs: Self) -> Self::Output {
+            self.checked_sub(rhs).unwrap()
+        }
+    }
+
+    impl std::ops::SubAssign for Amount {
+        fn sub_assign(&mut self, rhs: Self) {
+            *self = *self - rhs;
```
brentstone, line 992: why not checked?

crates/core/src/token.rs
```diff
@@ -1032,6 +965,78 @@ pub mod testing {
...
+        fn mul(self, rhs: T) -> Self::Output {
+            self.checked_mul(rhs.into()).unwrap()
+        }
+    }
+
+    impl std::ops::Mul<Amount> for u64 {
+        type Output = Amount;
+
+        fn mul(self, rhs: Amount) -> Self::Output {
+            rhs * self
```
brentstone, line 1011: checked? And same for the ones below this

crates/proof_of_stake/src/epoched.rs
```diff
@@ -106,7 +107,7 @@ where
             Some(last_update) => {
                 let data_handler = self.get_data_handler();
                 let future_most_epoch =
-                    last_update + FutureEpochs::value(params);
+                    last_update.unchecked_add(FutureEpochs::value(params));
```
brentstone, line 110: why unchecked add?

brentstone (changes requested):
Just left some comments and questions, but otherwise LGTM

tzemanovic:
> @tzemanovic Is the title of this PR still accurate? Looks like a bunch of crates beyond core are changed. Does this cover everything, everywhere or are there other sections still remaining to transition to checked arithemetic?

@cwgoes  It is accurate - the lints to warn about unchecked arithmetics and conversions are only applied in the core crate, i.e.:

```rust
#![warn(
    clippy::cast_sign_loss,
    clippy::cast_possible_truncation,
    clippy::cast_possible_wrap,
    clippy::cast_lossless,
    clippy::arithmetic_side_effects
)]
```

The changes outside of core are only due to removal of unchecked API in core types (except for a few reasonable exceptions), which are extensive but not exhaustive. We should also add these lints to other relevant crates which can still do unchecked arithmetics and conversions with other non-core types.

crates/core/src/dec.rs
tzemanovic: it's ok as they're both consts and the fn is also covered by unit tests

crates/core/src/token.rs
tzemanovic: ah, I'll switch it to use arith err

crates/core/src/token.rs
tzemanovic: it's ok in error type - it doesn't affect the state machine

crates/core/src/token.rs
tzemanovic: these are all within the `mod testing` where it's ok and convenient

crates/core/src/token.rs
tzemanovic: as above

crates/core/src/uint.rs
```diff
@@ -593,224 +597,212 @@ impl I256 {
...
+                Self(inner)
             }
             (false, false) => {
-                let this = self.abs();
-                let that = other.abs();
-                that.cmp(&this)
+                let inner = self.abs().checked_add(rhs.abs())?;
+                if inner > MAX_SIGNED_VALUE {
+                    return None;
+                }
```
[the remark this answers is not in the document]
tzemanovic: I think this is already limited in the `MAX_SIGNED_VALUE` (`"Note the the last digit is 2^63 - 1"`)

crates/proof_of_stake/src/epoched.rs
tzemanovic: This is for convenience when working with epochs - it's u64 so we should never overflow on additions with system params. I added this doc on the method: Panics on overflow. Care must be taken to only use this with trusted values that are known to be in a limited range (e.g. system parameters but not e.g. transaction variables).

crates/core/src/token.rs
tzemanovic: fixed in [70daab5](https://github.com/anoma/namada/pull/3074/commits/70daab5da619e8028203c93d7197ef9d5970b184)

cwgoes:
Understood @tzemanovic - thanks!

Issue #3068, opened 2024-04-11
MASP VP should support governance changes


grarco:
After #2601, governance proposals are allowed to change the asset map of masp for the conversions. At the moment though, the masp vp rejects these changes. We need to update it to support proposals.

In doing so, we could also refactor the VP itself to iterate on the keys as variants of an enum `KeyType` to make it more similar to the other VPs we have.

cwgoes:
We need to do this in phase 1 so we can test the phase transitions.

murisi:
> @murisi do you know how long this will take? any chance you could do it quickly?

I expect this to take maybe 2 or 3 days. I'd already started on making the MASP VP more similar to other VPs. And I don't see why allowing the MASP asset map to be changed should be difficult.

brentstone:
Any chance we could prioritize this, especially for next release? I ask because we have an e2e test `proposal_ibc_token_inflation` that is failing without this, and don't want to be in the dark about that for too long. @murisi

PR #3087, opened 2024-04-16
fix governance proposal execution

### LINKED ISSUE #2884: governance proposals end at the end of the designated Epoch instead of the start

### LINKED ISSUE #3081: Drop write log changes on failed governance proposal
https://github.com/anoma/namada/blob/621cce94399e3ce5088e07dfff842f438b216103/crates/apps/src/lib/node/ledger/shell/governance.rs#L367-L387 

Here we match the result of executing the proposals but it looks like we don’t drop the state in the case the execution was not accepted meaning that the changes are kept in the write log. These could be committed later in `finalize_block` leading to the governance proposal being effectively executed even if the actual wasm code was rejected. If instead the next call to `shell.state` is a `drop` instead of a `commit`  then we should be fine but it means that there is some non-determinism based on what the next call will be to the write log will be.

### LINKED ISSUE #3082: Avoid reading wasm code for non-wasm governance proposal
https://github.com/anoma/namada/blob/621cce94399e3ce5088e07dfff842f438b216103/crates/apps/src/lib/node/ledger/shell/governance.rs#L122

Here it looks like we read the proposal code from storage even though we know that the transaction doesn’t have on. Also, we ignore the actual result of the proposal and always log that it was successful

Fraccaman:
## Describe your changes
- Closes #3081 
- Closes #3082 
- Closes #2884

### DIFF
--- a/crates/namada/src/ledger/governance/mod.rs
+++ b/crates/namada/src/ledger/governance/mod.rs
@@ -81,13 +81,21 @@ where
             .into());
         };
 
+        // Is VP triggered by a governance proposal?
+        let is_governance_proposal = is_proposal_accepted(
+            &self.ctx.pre(),
+            tx_data.data().unwrap_or_default().as_ref(),
+        )
+        .unwrap_or_default();
+
         let native_token = self.ctx.pre().get_native_token()?;
 
         // Find the actions applied in the tx
         let actions = self.ctx.read_actions()?;
 
         // There must be at least one action if any of the keys belong to gov
-        if actions.is_empty()
+        if !is_governance_proposal
+            && actions.is_empty()
             && keys_changed.iter().any(gov_storage::is_governance_key)
         {
             tracing::info!(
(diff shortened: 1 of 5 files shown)


crates/namada/src/ledger/governance/mod.rs
```diff
@@ -81,13 +81,21 @@ where
...
+
         let native_token = self.ctx.pre().get_native_token()?;
 
         // Find the actions applied in the tx
         let actions = self.ctx.read_actions()?;
 
         // There must be at least one action if any of the keys belong to gov
-        if actions.is_empty()
+        if !is_governance_proposal
+            && actions.is_empty()
```
Fraccaman, line 98: @tzemanovic I had to add this condition

grarco (approved):
Looks good to me! The `proposal_ibc_token_inflation` will keep failing until we fix #3068 so we can ignore it for the time being

brentstone (approved):

PR #3093, opened 2024-04-17
Improve UX of validator's address

### LINKED ISSUE #2822: Better UX and distinction between the multiple addresses of a validator
There has been much confusion from some validators regarding what each address they have is. Validators can have an implicit address, which is used to generate a different established address, which is then used in `become_validator` to generate a validator account with that established address.

We need to make it clearer somehow, through logging and docs, that validator actions cannot be performed with the implicit address. Effectively, this implicit address is unrelated to the validator address. Perhaps we can help with automatic naming in a validator's wallet to help them with this distinction.

brentstone:
## Describe your changes
Closes #2822 

v0.33.0


crates/apps/src/lib/config/genesis/chain.rs
```diff
@@ -152,10 +152,21 @@ impl Finalized {
...
+                alias.clone(),
                 validator_wallet,
-            )
+            );
+            let address_alias = alias::validator_address(&alias);
+            wallet.insert_address(
+                address_alias.normalize(),
+                address.clone(),
+                false,
+            );
```
sug0, line 164: I think instead of storing the validator address this way, we could modify `extend_from_pre_genesis_validator` to store the validator address in the wallet.

something like this:
https://github.com/anoma/namada/blob/621cce94399e3ce5088e07dfff842f438b216103/crates/sdk/src/wallet/store.rs#L648-L649

sug0 (changes requested):

crates/sdk/src/wallet/alias.rs
```diff
@@ -161,3 +161,8 @@ pub fn validator_consensus_key(validator_alias: &Alias) -> Alias {
 pub fn validator_tendermint_node_key(validator_alias: &Alias) -> Alias {
     format!("{validator_alias}-tendermint-node-key").into()
 }
+
+/// The bech32m address of a validator
```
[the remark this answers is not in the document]
brentstone: Fixed in [705ccd7](https://github.com/anoma/namada/pull/3093/commits/705ccd79423c74a25ff3450e91fe032bb79e0b9b)

crates/apps/src/lib/config/genesis/chain.rs
brentstone: Done in [705ccd7](https://github.com/anoma/namada/pull/3093/commits/705ccd79423c74a25ff3450e91fe032bb79e0b9b)

crates/sdk/src/wallet/store.rs
```diff
@@ -654,6 +655,13 @@ impl Store {
         ];
         self.pkhs.extend(pkhs);
 
+        self.addresses
+            .extend([(address_alias.clone(), validator_address.clone())]);
+        println!(
+            "\nAdded the validator address {} to the wallet with alias \"{}\".",
+            &validator_address, address_alias
+        );
```
sug0, line 663: the `println!()` can remain in `crates/apps/src/lib/client/utils.rs`

crates/sdk/src/wallet/store.rs
```diff
@@ -654,6 +655,13 @@ impl Store {
         ];
         self.pkhs.extend(pkhs);
 
+        self.addresses
+            .extend([(address_alias.clone(), validator_address.clone())]);
```
sug0, line 659:
```suggestion
        self.addresses
            .insert(address_alias.clone(), validator_address.clone());
```

sug0 (changes requested):
sry just a couple more things

crates/sdk/src/wallet/store.rs
brentstone: done in [03006d8](https://github.com/anoma/namada/pull/3093/commits/03006d891727ec1a624b55cdb8d1fa85ab227347)

crates/sdk/src/wallet/store.rs
brentstone: done in [03006d8](https://github.com/anoma/namada/pull/3093/commits/03006d891727ec1a624b55cdb8d1fa85ab227347)

brentstone: Actually would like to keep this here, since the validator address alias is in scope here but not in the previous location yet. Seems more consistent to keep it here to me, but if I'm open to reconsidering if you felt more strongly about the other option.

tzemanovic: There are unfortunately already some "naked" prints in this module (and probably elsewhere too), but ideally in the SDK crate they should only be using the `(e)display_line` macro, for which we'd need the generic `sdk::io::Io` in scope which would probably be the best option to enable us to replace the other ones too. I'd say for this PR, one more in here won't make much difference, so I'll open an issue to address this separately

tzemanovic: #3106

tzemanovic (dismissed):

sug0 (approved):

PR #3088, opened 2024-04-16
Emit generic events from tx wasms

sug0:
## Describe your changes

Emit core events (i.e. `namada_core::event::Event`) from tx wasms. To this end, we leverage IBC's existing wasm infrastructure.

~~**NOTE:** IBC e2e tests are expected to fail, just like in the base PR.~~

## Indicate on which release or other PRs this topic is based on

#3032

Diff for review: https://github.com/anoma/namada/compare/tiago/generalize-events..tiago/emit-generic-events-wasm
(https://github.com/anoma/namada/pull/3088/files/2fc3cc8bbb6302c1bbb3ccdee26d271ed6bae092..0ac23e8a2d514d30a94abb42eafc8c0d6f5ccebd)

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #3102, opened 2024-04-20
Move events to submodules

sug0:
## Describe your changes

Move `Event` and associated types from `namada_core` to the new `namada_events` crate. Module-specific events (such as `IbcEvent` and `EthBridgeEvent`) were moved to their respective crates (`namada_ibc` and `namada_ethereum_bridge`).

~~**NOTE:** IBC e2e tests are expected to fail, just like in the base PR.~~

#3088

Diff for review: https://github.com/anoma/namada/compare/tiago/emit-generic-events-wasm..tiago/move-events-to-submodules (https://github.com/anoma/namada/pull/3102/files/0ac23e8a2d514d30a94abb42eafc8c0d6f5ccebd..44d0cdbce373910043013bc75da8cd91a4408cee)


sug0:
ci is temporarily borked due to `sccache`

crates/governance/src/event.rs
```diff
@@ -0,0 +1,265 @@
...
+                    proposal_id,
+                    false,
+                    result,
+                ),
+            ),
+        };
+
+        let mut event = Self::new(event_type, EventLevel::Block);
+
+        #[allow(deprecated)]
```
tzemanovic, line 181: should we open an issue for replacing these?

tzemanovic (dismissed):

crates/governance/src/event.rs
sug0: yeah, ideally

crates/ibc/src/event.rs
```diff
@@ -60,6 +61,28 @@ pub mod types {
...
+    const DOMAIN: &'static str = "ibc";
+}
+
+impl From<IbcEvent> for Event {
+    fn from(ibc_event: IbcEvent) -> Self {
+        let mut event = Self::new(
+            EventTypeBuilder::new_of::<IbcEvent>()
+                .with_segment(ibc_event.event_type.0)
+                .build(),
+            EventLevel::Tx,
```
grarco, line 74: Do we emit the ibc events at the tx or block level?

grarco (approved):
Just one comment but looks good to me!

crates/ibc/src/event.rs
sug0: most are tx level events. pgf over ibc is at the block level, but idk a clean solution to change the event level there lol

crates/ibc/src/event.rs
```diff
@@ -0,0 +1,521 @@
...
+fn validate_ibc_event_type(
+    namada_event: &Event,
+) -> Result<IbcEventType, EventError> {
+    if namada_event.kind().domain() != IbcEvent::DOMAIN {
+        return Err(EventError::InvalidEventType);
+    }
+
+    let event_type = namada_event.kind().sub_domain();
+
+    // if !matches!(
```
batconjurer, line 149: what's going on here?

sug0: I'm not sure if you're referring to the let expr, or the commented `matches!()`. the let expr is to strip the `ibc/` prefix from the event type, which `IbcEvent` instances don't contain. the `matches!()` was to check if the sub-domain was valid, but I'm not sure which ones are currently in use in namada...

batconjurer (dismissed):

PR #3103, opened 2024-04-21
Tx batching

### LINKED ISSUE #1356: Handle multiple WASMs in a single transaction
We want to put multiple IBC messages into a transaction so that an IBC relayer (Hermes) can request IBC transactions efficiently.

P.S. By supporting a transaction with multiple WASMs

grarco:
## Describe your changes

Closes #1356.

Implements transaction batching by extending the `Tx` type to carry a set of commitments to `data`, `code` and `memo`. Every tx now is a batch (the default case is a batch of a single inner tx) and can be set to be an atomic batch if needed.  In summary the changes are:

- Iterate on all the commitments and execute every inner tx in sequence. Precommit or drop the changes of an inner tx (based on vps/errors) before launching the next one to ensure that every tx sees the correct state
- If the batch is atomic and a single inner tx fails, we discard all the changes, even those coming from inner txs that were successful
- A single gas meter is used for the entire batch
- Replay protection and inner tx signatures are managed at the batch `Header` level, as in we sign the entire batch header (which contains the commitments of all the txs). This way the logic is simplified and the transaction can carry less signatures
- `WriteLog` and `TxResult` have been updated to support this new logic
- The `Event` emitted by the transaction has been updated to carry the results of all the inner txs of the batch
- Added a new method in the sdk to construct a batch

This is a first attempt to achieve tx batching but the feature should not be considered stabilized yet. More specifically I believe more work will be needed to:

- Refactor some parts of the code base that might have become too intricated (especially the `evalute_tx_result` function in `finalize_block`)
- Improve the logic to build transactions in the sdk and add some support in the client (currently lacking)
- Add at least one integration test to test the entire loop
- More unit tests (ideally one testing gas metering in a batch and more unit tests for the updated write log)
- Review/refactor the events
- Update Hermes (all the tests failing)
- Investigate supporting batching from cli

v0.35.1


crates/tx/src/types.rs
```diff
@@ -1101,26 +1171,30 @@ impl Tx {
...
-        &self.header.memo_hash
+        self.header.batch.insert(item);
     }
 
-    /// Get the memo designated by the memo hash in the header
-    pub fn memo(&self) -> Option<Vec<u8>> {
-        if self.memo_sechash() == &namada_core::hash::Hash::default() {
+    /// Get the memo designated by the memo hash in the header for the specified
+    /// commitment
+    pub fn memo(&self, cmt: &TxCommitments) -> Option<Vec<u8>> {
```
grarco, line 1192: I've modified these methods to require a reference to the commitments that identify an inner tx. we could to this also with just an integer but I believe this is a bit safer and more clear in terms of how an inner tx is identified inside the batch. Sill we can change this logic if that's the case

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -217,19 +237,45 @@ where
...
+                        vp_wasm_cache,
+                        tx_wasm_cache,
+                    },
+                ) {
+                    Err(e @ Error::GasError(_)) => {
+                        // Gas error aborts the execution of the entire batch
+                        state.write_log_mut().drop_tx();
+                        return Err(e);
+                    }
+                    res => {
```
grarco, line 265: Currently we continue in case of other errors even for atomic batches which is not needed since the batch will fail anyway. The reason here was to mimic what we do when evaluating VPs, which is to continue until we have gas at our disposable even if one of the VPs has already rejected the tx

crates/tx/src/types.rs
```diff
@@ -909,6 +899,61 @@ pub struct Header {
...
+)]
+pub struct Header {
+    /// The chain which this transaction is being submitted to
+    pub chain_id: ChainId,
+    /// The time at which this transaction expires
+    pub expiration: Option<DateTimeUtc>,
+    /// A transaction timestamp
+    pub timestamp: DateTimeUtc,
+    /// The commitments to the transaction's sections
+    pub batch: HashSet<TxCommitments>,
```
grarco, line 954: This set could be empty (no inner txs, just wrapper) and I didn't place any check on this. Should we add something?

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -179,31 +178,52 @@ pub fn dispatch_tx<'a, D, H, CA>(
...
+                gas_used: tx_gas_meter.borrow().get_tx_consumed_gas(),
+                wrapper_changed_keys: Default::default(),
+                batch_results: BatchResults(
+                    [(cmt.get_hash(), Ok(result))].into_iter().collect(),
+                ),
+            })
+        }
         TxType::Protocol(protocol_tx) => {
-            apply_protocol_tx(protocol_tx.tx, tx.data(), state)
+            // No bundles of protocol transactions, only take the first one
```
grarco, line 211: Should we support batched of protocol transactions? I'm not even sure if that can be a thing. In case I believe it would require more rework because of the `Protocol` enum

crates/apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -1108,6 +1110,20 @@ where
                     );
                     return response;
                 }
+                let batch_tx_hash = &tx.raw_header_hash();
```
grarco, line 1113: Replay protection is managed at the batch level, as in we check the `raw_header_hash` just like before. This hash contains the commitments of all the txs in the batch. This simplifies the logic and allows to (potenitally) reduce the number of signatures attached to the transaction. Txs in the batch are guaranteed to be different since the commitments are stored in a `HashSet`. Only drawback of this is that it's impossible to rewrap a single tx of the batch

crates/namada/src/ledger/protocol/mod.rs
cwgoes: We don't need to worry about this for now.

crates/apps/src/lib/node/ledger/shell/mod.rs
cwgoes: What do you mean by potentially reduce the number of signatures - as opposed to the wrapper tx fee payer signing over individual transactions within the batch in separate signatures?

crates/tx/src/types.rs
cwgoes: As in, permitting a batch of size 0? There's nothing semantically wrong with that, I think.

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                                ) && !matches!(
+                                    e,
+                                    protocol::Error::MissingSection(_)
+                                ) {
+                                    commit_batch_hash = true;
+                                }
+                                true
+                            }
+                        }
+                    })
```
sug0, line 588: can you move this closure to a binding to improve readability? or better yet, the entire boolean expression.
```rs
let is_any_tx_invalid = /* ... */;

if is_atomic_batch && is_any_tx_invalid { /* ... */ }
```

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                // Atomic batches need custom handling when even a single tx
+                // fails, since we need to drop everything
+                if is_atomic_batch
+                    && tx_result.batch_results.0.iter().any(|(_, result)| {
+                        match result {
+                            Ok(res) => {
+                                // If an inner tx failed for any reason but
+                                // invalid signature, commit its hash to
+                                // storage,
+                                // otherwis allow for a replay
```
sug0, line 560:
```suggestion
                                // storage, otherwise allow for a replay
```

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                                    protocol::Error::MissingSection(_)
+                                ) {
+                                    commit_batch_hash = true;
+                                }
+                                true
+                            }
+                        }
+                    })
+                {
+                    // TODO: nee to increment the rejected txs too
```
sug0, line 590:
```suggestion
                    // TODO: need to increment the rejected txs too
```

sug0: btw for this I guess you can do another pass over the batch of tx results. hm but this has raised another question: if the tx is atomic, shouldn't we be short circuiting execution? we don't need to exec all txs

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                    }
+                    self.state.write_log_mut().drop_batch();
+                    tx_event.extend(Code(ResultCode::InvalidTx));
+                } else {
+                    for (cmt_hash, batched_result) in &tx_result.batch_results.0
+                    {
+                        match batched_result {
+                            Ok(result) => {
+                                if result.is_accepted() {
+                                    if is_masp_tx(&result.changed_keys) {
```
sug0, line 602: if possible, we should move this branch to another helper method. it gets kinda confusing to read, given the many indentation levels

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                                        result.vps_result.rejected_vps
+                                    );
+
+                                    // If an inner tx failed for any reason
+                                    // but
+                                    // invalid
+                                    // signature, commit its hash to
+                                    // storage,
+                                    // otherwise
+                                    // allow for a replay
```
sug0, line 662: man these comments got fucked over by `rustfmt`

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +519,293 @@ where
...
+                            .header_hash;
+                        self.state
+                            .redundant_tx_hash(&header_hash)
+                            .expect("Error while marking tx hash as redundant");
+                    }
+                }
+
+                for _ in 0..commitments_len {
+                    stats.increment_errored_txs();
+                }
```
sug0, line 773:
```suggestion
                stats.increment_errored_txs_by(commitments_len);                
```
something like this to avoid looping

sug0 (changes requested):
here's my first pass over this code. I mainly only looked at the new tx defs in `core` and `FinalizeBlock`

crates/namada/src/ledger/protocol/mod.rs
sug0: I missed this comment in relation to the atomic batches.

> which is to continue until we have gas at our disposable even if one of the VPs has already rejected the tx

sorry I didn't get this

crates/tx/src/types.rs
grarco: I like this!

crates/apps/src/lib/node/ledger/shell/mod.rs
grarco: No I meant that by signing the entire header of the batch we need a signature per signer. If the signer of the inner txs we need only one signature for the entire batch. I still need to review if this is completely safe

cwgoes:
Just a few quick clarifications on your summary:

> Precommit or drop the changes of an inner tx before launching the next one to ensure that every tx sees the correct state

I'm assuming that we precommit if the inner tx was successful and drop if it failed, right?

> If the batch is atomic we discard all the changes

If the batch is atomic and one of the inner txs failed, right?

grarco:
> Just a few quick clarifications on your summary:
> 
> > Precommit or drop the changes of an inner tx before launching the next one to ensure that every tx sees the correct state
> 
> I'm assuming that we precommit if the inner tx was successful and drop if it failed, right?
> 
> > If the batch is atomic we discard all the changes
> 
> If the batch is atomic and one of the inner txs failed, right?

Both are correct, I've updated the description 👍🏻

crates/namada/src/ledger/protocol/mod.rs
grarco: Yes I'm not sure about this logic. I'd like to hear other opinions on this, we could also decide to short circuit the evaluation in this case

crates/tx/src/types.rs
```diff
@@ -1048,6 +1092,32 @@ impl Tx {
...
+    /// commitment is already contained in the set
+    #[cfg(any(test, feature = "testing"))]
+    pub fn push_default_inner_tx(&mut self) -> bool {
+        self.header.batch.insert(TxCommitments::default())
+    }
+
+    /// Add a new inner tx to the transaction. Returns `false` if the
+    /// commitments already existed in the collection. This function expects a
+    /// transaction carrying a single inner tx as input
+    pub fn add_inner_tx(&mut self, other: Tx, cmt: TxCommitments) -> bool {
```
yito88, line 1105: When the same tx code or data is added, they are not inserted but the commitment still points to the sections, and the sections don't exist.
I had an error when multiple ibc txs were added, i.e. batching with the same tx code.
```
"Mempool validation failed: Wrapper transaction code didn't pass the allowlist checks Tx is not allowed in allowlist parameter"
```

crates/apps/src/lib/cli.rs
```diff
@@ -3034,8 +3034,6 @@ pub mod args {
             Err(_) => config::get_default_namada_folder(),
         }),
     );
-    pub const BATCH_SIZE_OPT: ArgDefault<u64> =
```
batconjurer, line 3037: we will still want this for shielded sync, no?

crates/tx/src/types.rs
grarco: You're right, I've pushed a new commit that removes that broken optimization for now. We can bring it back when we'll rework the tx builder

crates/apps/src/lib/cli.rs
grarco: Yes my bad, removed this by mistake

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: I've reworked the entire `evaluate_tx_result` function

crates/tx/src/types.rs
yito88, line 954: Can we use `indexmap::IndexSet` insread of `HashSet`?
I think we need to preserve the order to execute batched transactions in the same order of the insertion to the batch.

crates/tx/src/types.rs
yito88: Thank you! It has resolved the issue.

crates/sdk/src/tx.rs
```diff
@@ -2940,6 +2921,41 @@ async fn used_asset_types<P, R, K, N>(
     Ok(asset_types)
 }
 
+/// Constructs the batched tx from the provided list. Returns also the data for
+/// signing
+pub fn build_batch(
+    mut txs: Vec<(Tx, SigningTxData)>,
+) -> Result<(Tx, Vec<SigningTxData>)> {
+    let (mut batched_tx, sig_data) = txs.pop().ok_or_else(|| {
```
yito88, line 2929: As per the previous comment, we want to execute transactions in the batch in the same order of the insertion.
Can we take the first one here?

crates/tx/src/types.rs
brentstone: I believe this `HashSet` (and probably every instance of `HashSet` now in the codebase) is indeed just an alias for `indexmap::set::IndexSet`. Check `namada_core::collections` where it is declared.

yito88: ah, I see. thank you for telling me.
The `HashSet` is OK. This issue is caused by the `build_batch`.

crates/state/src/write_log.rs
```diff
@@ -546,9 +629,16 @@ impl WriteLog {
     pub fn iter_prefix_pre(&self, prefix: &storage::Key) -> PrefixIter {
         let mut matches = BTreeMap::new();
 
-        for (key, modification) in &self.block_write_log {
-            if key.split_prefix(prefix).is_some() {
-                matches.insert(key.to_string(), modification.clone());
+        let mut buckets = vec![&self.block_write_log];
+        for tx_log in self.batch_write_log.iter().rev() {
```
batconjurer, line 633: same thing here. Why create a new vec?

crates/state/src/write_log.rs
```diff
@@ -561,11 +651,14 @@ impl WriteLog {
     pub fn iter_prefix_post(&self, prefix: &storage::Key) -> PrefixIter {
         let mut matches = BTreeMap::new();
 
-        for bucket in [
-            &self.block_write_log,
-            &self.tx_precommit_write_log,
-            &self.tx_write_log,
-        ] {
+        let mut buckets = vec![&self.block_write_log];
```
batconjurer, line 654: and here

batconjurer (changes requested):

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -832,6 +973,76 @@ mod test_finalize_block {
         )
     }
 
+    // Make a transaction batch with two transactions. Optionally make the batch
+    // atomic and request the failure of the first transaction
+    fn mk_tx_batch(
```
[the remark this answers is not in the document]
grarco: Actually, in some cases we test that the the other changes are committed, but I will another tx anyway

crates/state/src/write_log.rs
```diff
@@ -56,33 +56,69 @@ pub enum StorageModification {
...
-    pub(crate) tx_precommit_write_log:
-        HashMap<storage::Key, StorageModification>,
-    /// The IBC events for the current transaction
-    pub(crate) ibc_events: BTreeSet<IbcEvent>,
-    /// Storage modifications for the replay protection storage, always
-    /// committed regardless of the result of the transaction
+    /// The write log of the transactions of the current batch
+    /// INVARIANT: this has to be sorted by the insertion
+    /// order to correctly read values
+    pub(crate) batch_write_log: Vec<BatchedTxWriteLog>,
```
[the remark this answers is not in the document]
grarco: It is, but it's also guaranteed by a simple vector. I left that comment just as a remainder in cases we needed to change this struct (I wanted to avoid starting to play with marker traits to enforce this thing)

crates/state/src/write_log.rs
```diff
@@ -126,10 +160,19 @@ impl WriteLog {
...
-                self.tx_precommit_write_log.get(key)
+                self.tx_write_log.precommit_write_log.get(key)
+            })
+            .or_else(|| {
+                // If not found, then try to read from batch write log,
+                // following the insertion order
+                self.batch_write_log
+                    .iter()
+                    .rev()
+                    .find_map(|log| log.write_log.get(key))
```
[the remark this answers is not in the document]
grarco: We do, the batched transactions are part of this hierarchy, so we start from the current tx, than the precommit of the current tx, than the previous txs in the same batch in reverse order and finally the write log of the block

crates/sdk/src/tx.rs
grarco: This should be fixed in 47ccfd35d8ed43a23bf2c66a85df34f1c8af4112

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +497,301 @@ where
...
+                            // eth bridge events
+                            .chain(
+                                result
+                                    .eth_bridge_events
+                                    .iter()
+                                    .map(Event::from),
+                            ),
+                    );
+                }
+                Ok(result) => {
```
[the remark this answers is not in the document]
grarco: Reverted to the previous version in 7874576834b9daa1c53b8fa389748f6c796bd7db

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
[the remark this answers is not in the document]
grarco: Updated in c1e7f7e4b23e62ff6b994d00dfa4ad2d02b6a1a4

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+    }
+
+    fn check_inner_results(
+        &mut self,
+        tx_result: &namada::tx::data::TxResult<protocol::Error>,
+        tx_header: &namada::tx::Header,
+        tx_index: usize,
+        height: BlockHeight,
+        commit_batch_hash: &mut bool,
+        is_any_tx_invalid: &mut bool,
```
tzemanovic, line 788: it's a small style issue, but passing these as mut refs instead of return value is not very readable because you don't know what the value is when it's being called without looking up the callers - you can just make some struct for these if you want to return them named

wasm/tx_become_validator/src/lib.rs
```diff
@@ -6,12 +6,19 @@ use namada_tx_prelude::transaction::pos::BecomeValidator;
...
+        ref cmt,
+    } = tx_data;
+    let data =
+        signed
+            .data(cmt)
+            .ok_or_err_msg("Missing data")
+            .map_err(|err| {
+                ctx.set_commitment_sentinel();
+                err
+            })?;
```
tzemanovic, line 21: we should abstract this a bit as it's repeated in every tx, maybe something like:

```rust
let data = ctx.get_tx_data(tx_data)?;
```

in tx_prelude e.g.:
```rust
impl Ctx {
  pub fn get_tx_data(&mut self, tx_data: BatchedTx) -> TxResult {
    ...
  }
}
```

crates/core/src/event/extend.rs
```diff
@@ -111,13 +111,21 @@ impl ExtendEvent for Info {
...
-        event["is_valid_masp_tx"] = masp_tx_index.to_string();
+        let Self((masp_tx_index, cmt_hash)) = self;
+        let attribute = if let Some(hash) = cmt_hash {
+            format!("cmt/{hash}/is_valid_masp_tx")
+        } else {
+            "is_valid_masp_tx".to_string()
+        };
+        // TODO: should we put the index of the tx in the block at the root
+        // level of the even attributes instead of as the value of the masp
+        // attribute?
```
tzemanovic, line 127: @sug0

grarco: More on this: we need the tx index in the client when fetching masp transactions but if we need the index for other things we can think about extracting it to another attribute of the event

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+    // update stats and event, manage replay protection.
+    fn evaluate_tx_result(
+        &mut self,
+        response: &mut shim::response::FinalizeBlock,
+        dispatch_result: &Result<()>,
+        tx_result: namada::tx::data::TxResult<protocol::Error>,
+        tx_data: TxData,
+        mut tx_logs: TxLogs,
+    ) {
+        // Check the commitment of the fee unshielding regardless of the
```
cwgoes, line 517: This will change somewhat as part of the MASP fee rework, right?

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+        tx_logs: &mut TxLogs,
+    ) {
+        // Track the need to commit the batch hash for replay protection. Hash
+        // must be written if at least one of the txs in the batch requires so
+        let mut commit_batch_hash = false;
+        // Track if any of the inner txs failed or was rejected
+        let mut is_any_tx_invalid = false;
+        let mut temp_log = TempTxLogs::default();
+
+        temp_log.check_inner_results(
```
cwgoes, line 593: This is a nit, but I much prefer returning results rather than passing mutable references as arguments.

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+            tx_data.tx_index,
+            tx_data.height,
+            &mut commit_batch_hash,
+            &mut is_any_tx_invalid,
+        );
+
+        if tx_data.is_atomic_batch && is_any_tx_invalid {
+            // Atomic batches need custom handling when even a single tx fails,
+            // since we need to drop everything
+            let unrun_txs = tx_data.commitments_len
```
cwgoes, line 605: Why do we need to do this subtraction? I don't quite follow.

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+                - tx_result.batch_results.0.len() as u64;
+            temp_log.stats.set_failing_atomic_batch(unrun_txs);
+            temp_log.commit_stats_only(tx_logs);
+            self.state.write_log_mut().drop_batch();
+            tx_logs.tx_event.extend(Code(ResultCode::WasmRuntimeError));
```
cwgoes, line 610: Why this error code?

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -652,6 +503,372 @@ where
...
+            tx_data.height,
+            &mut commit_batch_hash,
+            &mut false,
+        );
+
+        let unrun_txs =
+            tx_data.commitments_len - tx_result.batch_results.0.len() as u64;
+
+        if tx_data.is_atomic_batch {
+            tx_logs.stats.set_failing_atomic_batch(unrun_txs);
```
cwgoes, line 665: Seems like some duplicate code here and above which could be abstracted.

crates/apps/src/lib/node/ledger/shell/mod.rs
cwgoes: I think it's fine to bind the signatures for the inner transactions to the batch, if that's helpful.

cwgoes (commented):
Some comments/questions on the `finalize_block` portion.

wasm/tx_become_validator/src/lib.rs
grarco: Agree, this would be a nice improvement!

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Correct, this logic will be removed

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Some transactions of the batch might not be run (because we run out of gas before or if it is an atomic batch because one of the previous txs failed). Here we get how many inner txs fall under this category so that we can update the stats correctly.

Actually about this,

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: We've always used this error except for wrapper errors

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
cwgoes: I see, understood.

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
cwgoes: Is there not a more specific error returned from one of the inner transactions inside the batch which failed?

crates/namada/src/ledger/protocol/mod.rs
grarco: Updated the logic in 8ab0f38e6596fdf3da3d2492c866c992fb7f18b4

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: I've tried to address this in 525d044c624cc4c189a845ed96eb5b7c584715b3

wasm/tx_become_validator/src/lib.rs
grarco: Done in 0a40071e8fddbf813703f9131af0772ac4a98e91

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Done in 525d044c624cc4c189a845ed96eb5b7c584715b3

crates/core/src/event/extend.rs
sug0: I think it's better to emit two separate attributes in the same event, like marco was suggesting above.

for instance:
```
event["is_valid_masp_tx"] = $tx_index
event["commitment-hash"] = hash($tx_commitments)
```

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -308,192 +297,54 @@ where
                     None
                 };
             let tx_gas_meter = RefCell::new(tx_gas_meter);
-            let tx_result = protocol::dispatch_tx(
-                tx.clone(),
+            let mut tx_event = new_tx_event(&tx, height.0);
+            let is_atomic_batch = tx.header.atomic;
+            let commitments_len = tx.commitments().len() as u64;
+            let tx_hash = tx.header_hash();
+            let mut tx_result = namada::tx::data::TxResult::default();
```
grarco, line 304: @tzemanovic any suggestions here? Because we might need to commit some txs of the batch even in case of an error I need to always have a `TxResult` to analyze. Here I modified `dispatch_tx` to take it as an argument instead of returning it. I thought about embedding the `TxResult` even inside the error type but that might load that type too much

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Ah yes the specific error can be found in the `Event` we emit under the `"batch"` attribute: here we serialize the entire `TxResult` object which carries information about all the inner txs we ran including possible errors or the vps that rejected them

crates/core/src/event/extend.rs
grarco: Just a note, a batch could contain more than one masp tx so we need to index that at the inner tx level. At the moment I'm using a custom attribute string which contains a reference to the specific inner tx like this:

`event["cmt/$hash/is_valid_masp_tx"] = $tx_index`

So, in this case we end up with a duplicated transaction index (which is the same for the entire batch).
I believe this is related to https://github.com/anoma/namada/pull/3103#discussion_r1590999811

sug0: one of the attributes refers to the index in the block (`is_valid_masp_tx`), the other, refers to one of the txs in the batch (`commitment-hash`); this way, we identify the masp tx by the batch tx index in the block, and then by the index in the batch itself. I think it suffices

grarco: Ah you want to use the `commitment-hash` attribute in conjunction with the `is_valid_masp_tx` one. We still have one problem cause more than one inner tx in the batch could be a valid masp transaction and this solution would not support that case. We should probably encode a vector of hashes as the value of the `commitment-hash` attribute

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
cwgoes: Hmm OK, I still think it would be good to have an error code which actually specifies the semantics in this case, e.g. `ResultCode::AtLeastOneInnerTxFailed` or something - but this can be handled later if it's a pain to do right now.

crates/core/src/event/extend.rs
sug0: I see yeah, that makes sense!

sug0: since we're changing these events anyway, could we also rename the attribute `is_valid_masp_tx`? the name implies it's a boolean, but it's a tx index. something like `masp-tx-block-index`. as for the commitments, something like `masp-tx-batch-indices`, or `masp-tx-batch-commitments`.

wasm/tx_become_validator/src/lib.rs
tzemanovic: nice!

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: personally I'd still put it in the return type

grarco: Tried to fix it in db8fa3625e194be1a4f2d367ed358ff4fa6d6bd4

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
grarco: Sounds good, I've opened [this](https://github.com/anoma/namada/issues/3199) issue

crates/tx/src/types.rs
grarco: Opened #3200 to address this separately

crates/state/src/write_log.rs
```diff
@@ -158,8 +201,14 @@ impl WriteLog {
         &self,
         key: &storage::Key,
     ) -> (Option<&StorageModification>, u64) {
-        match self.block_write_log.get(key) {
-            Some(v) => {
+        let mut buckets = vec![];
+        for tx_log in self.batch_write_log.iter().rev() {
```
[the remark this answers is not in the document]
batconjurer: yes, I think that is better.

crates/apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: looks good, thx!

murisi (dismissed):
The changes to the `Tx` format look good to me. Given that the only change  in the format is that the data, code, and memo hashes are vectorized, hardware wallet support for singleton batch transactions (where `batch.len() == 1`) should be quickly achievable. But it will probably take some time to implement support non-singleton batch transactions because the hardware wallet codebase wasn't structured for holding, printing, and signing multiple transactions in one go.

crates/tx/src/types.rs
```diff
@@ -1039,6 +1084,27 @@ impl Tx {
...
+    /// Add new default commitments to the transaction. Returns false if the
+    /// commitment is already contained in the set
+    #[cfg(any(test, feature = "testing"))]
+    pub fn push_default_inner_tx(&mut self) -> bool {
+        self.header.batch.insert(TxCommitments::default())
+    }
+
+    /// Add a new inner tx to the transaction. Returns `false` if the
+    /// commitments already existed in the collection. This function expects a
+    /// transaction carrying a single inner tx as input
```
murisi, line 1096: Is there a reason this function is limited to a single inner tx?

crates/tx/src/types.rs
```diff
@@ -1039,6 +1084,27 @@ impl Tx {
...
+    pub fn add_inner_tx(&mut self, other: Tx, cmt: TxCommitments) -> bool {
```
murisi, line 1097: If this function is limited to a single inner tx, then I guess the pair `(other, cmt)` is conceptually a `BatchedTx`...

crates/tx/src/types.rs
grarco: At the moment the support to build batched transactions is very limited. We'll probably want to rework the way we construct txs so for now I've implemented something minimal. The caller can invoke this function multiple times if needed

crates/tx/src/types.rs
grarco: Ah you are right! Still, I think it would just be better to rework this function completely in the near future

wasm_for_tests/tx_invalid_data/src/lib.rs
```diff
@@ -1,11 +1,18 @@
...
-        ctx.set_commitment_sentinel();
-        err
-    })?;
+fn apply_tx(ctx: &mut Ctx, tx_data: BatchedTx) -> TxResult {
+    let BatchedTx {
+        tx: signed,
+        ref cmt,
+    } = tx_data;
+    let _data =
+        signed
```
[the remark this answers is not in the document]
grarco: I believe this will be addressed by #3200

PR #3110, opened 2024-04-23
state/write_log: refactor temp key-vals for better type-safety

### LINKED ISSUE #2915: Type safety for `block_write_log`
Currently the `block_write_log` field of the write log is defined as:

https://github.com/anoma/namada/blob/cc3edde4d572ff41f19d24b5b7da60406acdfdf4/crates/state/src/write_log.rs#L84

where the storage modification is:

https://github.com/anoma/namada/blob/cc3edde4d572ff41f19d24b5b7da60406acdfdf4/crates/state/src/write_log.rs#L43-L64

The block write log though, can and should never contain an entry of the `StorageModification::Temp` variant. Something similar has been introduced in #2735 with a `PersistentStorageModification` enum. We could think about utilizing this also in the block write log to enforce safety via the type system.

tzemanovic:
## Describe your changes

closes #2915

v0.33.0

### DIFF
--- a/crates/state/src/write_log.rs
+++ b/crates/state/src/write_log.rs
@@ -32,10 +32,6 @@ pub enum Error {
     WriteTempAfterWrite,
     #[error("Replay protection key: {0}")]
     ReplayProtection(String),
-    #[error(
-        "Trying to cast a temporary write to a persistent storage modification"
-    )]
-    TempToPersistentModificationCast,
 }
 
 /// Result for functions that may fail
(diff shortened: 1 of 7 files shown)


crates/state/src/write_log.rs
```diff
@@ -222,7 +172,7 @@ impl WriteLog {
     pub fn read_persistent(
```
grarco, line 172: Should this me named just `read`? Just to match the naming scheme of the write methods which are `write` and `write_temp`

grarco (dismissed):
Minor comment but looks good to me, thank you!

brentstone:
let me know if this is rdy for merge or will have another change

crates/state/src/write_log.rs
tzemanovic: good call, will do!

tzemanovic: done in [791d172](https://github.com/anoma/namada/commit/791d17282c294b6cb7d0b0be1caf318df4886bcb) - also removed an unused err case

tzemanovic:
> let me know if this is rdy for merge or will have another change

should be good to go now

grarco (approved):

crates/state/src/write_log.rs
tzemanovic: [172ba75](https://github.com/anoma/namada/commit/172ba7553dc472044d4933116e44da06220b14b5)

PR #3120, opened 2024-04-24
Improve governance voting logic

### LINKED ISSUE #2523: VoteTx should fetch all the voter delegations on wasm
Right now, the `vote_proposal` tx is expecting from the client a list of `validators` for which the voter has a delegation to. The Governance VP is then validating the list, but the validation logic is redundant and not optimal. We should instead fetch the voter delegation list directly from the wasm transactions.

Fraccaman:
## Describe your changes
Closes #2523.

Summary of changes w.r.t the delegation validators and governance voting:
- client:
  - if voter is validator, client will return error if validator is inactive or jailed **in the current epoch**
  - if delegator, client just ensures it has delegations. **Does not filter based on the bond validator's state** 
- wasm tx:
  - fetches the delegation validators **in the current epoch**, passes them into `vote_proposal`, which writes a vote key to storage for each. **Does not filter based on the bond validator's state** 
- gov VP: `is_valid_vote_key` **does not filter based on the bond validator's state**.

Almost no filtering is done in when a vote tx is submitted and vote keys are written into a storage for a proposal. This is because the state of a validator at the tally epoch, which is the voting end epoch, cannot necessarily be known at the epoch of vote submission. We should allow, for example, a delegator to submit a vote for a bond to an inactive validator, which may be active by the time the tally epoch comes around.

Ultimately in `compute_proposal_votes`, a vote will only be counted if the validator is active at the end epoch - in consensus, below-capacity, or below-threshold. We rely on this step to do the proper filtering.

Filtering is performed, however, by preventing a delegator with no delegations from submitting a valid vote tx and by preventing a validator that is currently jailed or inactive from submitting a valid vote tx.

v0.34.0


crates/proof_of_stake/src/tests/test_pos.rs
```diff
@@ -385,7 +385,7 @@ fn test_bonds_aux(params: OwnedPosParams, validators: Vec<GenesisValidator>) {
         assert_eq!(
             details.bonds.len(),
             2,
-            "Contains genesis and newly added self-bond"
+            "contains_key genesis and newly added self-bond"
```
grarco, line 388: Are these modifications indeed correct? They look like the result of a search&replace but I'm not sure they were intended

grarco:
There's a failing unit test and I left a minor comment. Overall it looks good to me but I'd like a second opinion from @brentstone especially for the query and sdk files

brentstone:
@Fraccaman @grarco I fixed up the logic in here and also summarized the changes in the PR description. Lmk your thoughts

brentstone (approved):

crates/namada/src/ledger/governance/mod.rs
```diff
@@ -303,37 +302,65 @@ where
...
+        // hashmap if !matches!(
+        //     delegations_targets.get(validator).unwrap(),
+        //     ValidatorState::Consensus
+        // ) {
+        //     return Err(native_vp::Error::new_alloc(format!(
+        //         "The vote key is not valid due to {validator} being not in \
+        //          the active set"
+        //     ))
+        //     .into());
+        // }
```
Fraccaman, line 363: remove this comment

PR #3141, opened 2024-04-26
Emit balance change events from the protocol

### LINKED ISSUE #84: protocol balance change events
The ledger should attach an event to blocks in which some accounts' balances are modified by the protocol (e.g. wrapper tx fee charges, PoS slashing, future governance proposal refunds), to make the changes externally visible and index-able.

sug0:
## Describe your changes

Closes #84

Emits balance change events for the following protocol actions:

- Fee payments to validators
- PGF payments (including over IBC)
- Governance refunds
- IBC token minting/burning
- Proof-of-stake validator slashing
- WASM transfers

#3102

Diff for review: https://github.com/anoma/namada/compare/tiago/move-events-to-submodules..tiago/balance-change-events (https://github.com/anoma/namada/pull/3141/files/68420a57f251ca8e68ae6960a4a0081e57c16403..406f715b416b069b310410fec20f0ba0cb7c4a06)


crates/state/src/wl_state.rs
```diff
@@ -1183,6 +1232,30 @@ where
     }
 }
 
+impl<D, H> EmitEvents for TempWlState<'_, D, H>
```
grarco, line 1235: The `TempWlState` is only used for validation/simulation in `mempool_validate` and `prepare/process_proposal`. Do we need to support events in those cases too?

crates/trans_token/src/event.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Transfer token event.
+    pub const TRANSFER: EventType = event_type!(TokenEvent, "transfer");
+}
+
+/// A user account.
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum UserAccount {
+    /// Internal chain address in Namada.
+    Internal(Address),
```
grarco, line 33: Can this cause some confusion with the enum variant we have on the `Address` type?

grarco (dismissed):
A couple of minor comments. LGMT!

crates/state/src/wl_state.rs
sug0: ah, I guess not. I implemented `EmitEvents` on pretty much every type that also implemented `State`. I'll remove this impl

crates/trans_token/src/event.rs
sug0: hmm.. any other name suggestions?

grarco: I'm terrible at naming. Let's keep it like that 👍🏻

brentstone:
FWIW, currently slashing does not actually involve any balance changes. The tokens are held in the PoS account and they are kept there, only other voting power-related data is manipulated. In PoS, only bonding and withdrawing involve balance changes.

I don't think I know enough about the event emission in general to say whether or not we should still emit events for slashing or if they should strictly be done for balance changes in the context of this PR though.

sug0:
@brentstone yeah, slashing events were loosely grouped together with other balance change events in this pr, but they're defined as Proof-of-Stake events, rather than Token events. I think it's still useful to emit them in the protocol, as long as we distinguish them from balance change events.

Fraccaman:
my only comment is that we could move each `TokenTransfer` event to a static method to avoid having constant string around the code, but we can also do this refactor later. Anything else looks good to me!

Fraccaman (approved):

crates/trans_token/Cargo.toml
```diff
@@ -20,8 +20,10 @@ migrations = [
 
 [dependencies]
 namada_core = { path = "../core" }
+namada_events = { path = "../events", default-features = false }
```
tzemanovic, line 23:
```suggestion
namada_events = { path = "../events" }
```
tiny (there seem to be no default features)

crates/token/src/lib.rs
```diff
@@ -9,6 +9,10 @@ pub mod storage_key {
     pub use namada_trans_token::storage_key::*;
 }
 
+pub mod event {
+    pub use namada_trans_token::event::*;
+}
```
tzemanovic, line 14:
```suggestion
pub use namada_trans_token::event;
```

tzemanovic (dismissed):

crates/trans_token/Cargo.toml
sug0: I added this because eventually I'd like to move the event log out of `namada_sdk` to this crate

PR #3142, opened 2024-04-26
Remove MASP pin key

### LINKED ISSUE #2381: Broken `query_tx_deltas`
I believe there's an issue with `query_tx_deltas`.
https://github.com/anoma/namada/blob/50caebb38c07fb43dbf102e3817eafd4ffd693fd/sdk/src/masp.rs#L1909

The function relies on calling `tx_search` which is only available if the node has indexing turned on (which we disable by default given the increase in storage consumption).

Even if the indexer is indeed available, the function produces a query based on the attributes `transfer.source`, `transfer.target`, `transfer.token`, but we never actually produce these attributes in `finalize_block`. Indeed, from some tests I ran, the query always returns an empty response.

### LINKED ISSUE #2675: Remove MASP pin key
Currently, the masp optional pin key is used in an inconsistent way: when a user generates a payment address there’s the chance to pin this address, if this is the case than all transactions with this payment address as the receiver will produce a pin key with the hash of the address, otherwise no pin key will be produced. If no pin is requested when generating the address, then no pin key will be produced.

Also, since we produce the pin key in a deterministic way (hash of the payment address) I believe we end up rewriting the same key in storage.

grarco:
## Describe your changes

Closes #2675.
Closes #2381.

Removes the MASP pin key and the associated data/methods. Recomputes the shielded keys and addresses for tests given the changes in the encoding. Removes `query_tx_deltas` as it is unused now.

v0.34.0


Fraccaman:
`run_ledger_ibc_with_hermes` e2e is failing but the test is making some shielded transfers over IBC. Can this be because we have to update hermes?

Fraccaman (approved):

grarco:
@yito88 do you think this error could be do to Hermes?

yito88:
@grarco yes, Hermes makes a request for a shielded transfer now. `token::Transfer` returned from `gen_ibc_shielded_transfer` was changed.

yito88:
I tried to update Hermes with this branch: https://github.com/heliaxdev/hermes/tree/1.7.4-namada-3142
The e2e test `run_ledger_ibc_with_hermes` passed with it in my env.

crates/apps/src/lib/cli/context.rs
```diff
@@ -621,8 +621,5 @@ impl ArgFromMutContext for BalanceOwner {
                 ExtendedViewingKey::arg_from_mut_ctx(ctx, raw)
                     .map(Self::FullViewingKey)
             })
-            .or_else(|_| {
```
batconjurer, line 624: why was this removed?

batconjurer (approved):

crates/apps/src/lib/cli/context.rs
grarco: I believe it was only used for the pinned balance

PR #3164, opened 2024-05-02
Set height for abci_query

### LINKED ISSUE #3065: Fixed height in `abci_query` response
Seems like when directly querying the rpc endpoint `abci_query` of comet, the response shows `height: 1` regardless of the actual height that the user requested.

yito88:
## Describe your changes
closes #3065

`v0.34.0`

### DIFF
--- a/crates/sdk/src/queries/shell/eth_bridge.rs
+++ b/crates/sdk/src/queries/shell/eth_bridge.rs
@@ -248,13 +248,13 @@ where
         return Ok(Default::default());
     }
 
+    let last_committed_height = ctx.state.in_mem().get_last_block_height();
     let mut status = TransferToEthereumStatus {
-        queried_height: ctx.state.in_mem().get_last_block_height(),
+        queried_height: last_committed_height,
         ..Default::default()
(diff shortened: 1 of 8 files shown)


crates/sdk/src/queries/shell/eth_bridge.rs
```diff
@@ -254,12 +254,10 @@ where
     };
 
     // check which transfers in the Bridge pool match the requested hashes
+    let last_committed_height = ctx.state.in_mem().get_last_block_height();
```
grarco, line 257: I believe the behavior that users want is to get back the height that heir query is targeting. So I think it's ok to return the last block height if the query contains a height of `0` but for all the other cases we should set the response height to be the same as the query one

yito88: ah, you're right. I fixed them. https://github.com/anoma/namada/pull/3164/commits/801a115c388b2b9b514702914ba5db965996644a

grarco (approved):
LGTM, thank you!

PR #3183, opened 2024-05-07
wasm: switch to cranelift compiler

tzemanovic:
## Describe your changes

Taking advantage of tx allowlist that prohibit execution of untrusted wasm, we can switch to cranelift compiler to improve wasm performance

### DIFF
--- a/crates/benches/wasm_opcodes.rs
+++ b/crates/benches/wasm_opcodes.rs
@@ -406,13 +406,13 @@ impl Display for WatBuilder {
     }
 }
 
-// Use singlepass compiler (the same one used in protocol) to prevent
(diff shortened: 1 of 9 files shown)


crates/benches/wasm_opcodes.rs
```diff
@@ -406,13 +406,13 @@ impl Display for WatBuilder {
     }
 }
 
-// Use singlepass compiler (the same one used in protocol) to prevent
+// Use Cranelift compiler (the same one used in protocol) to prevent
```
sug0, line 409: cranelift might actually do a couple of optimizations; nothing too aggressive, but probably enough to optimize things like `loop {}`

PR #3186, opened 2024-05-07
Murisi+grarco/multi tx masp vp rebased

murisi:
## Describe your changes
Tried to increase the readability, strength, and generality of the MASP validity predicate whilst decreasing its verbosity. The changes are as follows:
* `Transaction` inputs and outputs that exceed pre-balances and post-balances respectively are no longer allowed.
* Now use the Sapling value balance in order to verify that movement in and out of the shielded pool is correct
* Now handle all `AssetType` decoding in one place before starting to check the transparent bundles
* The VP should now support `Transaction`s and unrelated `Transfer`s touching the same addresses/tokens
* Tried to remove the code that gave IBC minting transactions special treatment now that Sapling value balance is used
* Reduced dependency on the multitoken VP by directly checking the correctness of MASP VP balance changes
* Now check that all `Transaction`s and the balancing transaction are sufficiently authorized; prevents malleability

What would still need to be handled/fixed if this code turns out to be relevant:
* Add integration tests. Like checking that an unshielding `Transaction` to Bertha is embeddable inside a `Transfer` to Christel only if Bertha has signed the entire transaction (in addition to the usual MASP checks).

Namada 0.34.0
https://github.com/anoma/masp/pull/79


crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -253,10 +252,10 @@ where
         Ok(())
     }
 
-    fn validate_state_and_get_transfer_data(
-        &self,
-        keys_changed: &BTreeSet<Key>,
-    ) -> Result<TransparentTransferData> {
+    fn validate_state_and_get_transfer_data<'vp>(
```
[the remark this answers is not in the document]
murisi: You're right. I think this is a holdover from the merge process. Fixed.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,319 @@ where
...
+        }
+        // Maybe the asset type has no attached epoch
+        None if changed_balances.tokens.contains_key(&vin.asset_type) => {
+            let (token, denom, digit) =
+                &changed_balances.tokens[&vin.asset_type];
+            // Determine what the asset type would be if it were epoched
+            let epoched_asset_type =
+                encode_asset_type(token.clone(), *denom, *digit, Some(epoch))
+                    .wrap_err("unable to create asset type")?;
+            if conversion_state.assets.contains_key(&epoched_asset_type) {
```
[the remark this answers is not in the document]
murisi: This check is here to prevent an unepoched asset from entering the shielded pool in the case where its underlying token address is participating in the shielded rewards program. But it is possible for an unepoched asset to enter the shielded pool and then later have its underlying token address participate in the shielded pool. In this case we need to allow users to withdraw these assets even though their epoched equivalents are in the conversion state. This was my reasoning for why this check is not repeated in the context of the transparent output. But please let me know if I'm missing something!

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -184,22 +195,10 @@ where
...
-                    ));
-                tracing::debug!("{error}");
-                return Err(error);
-            }
-        };
-
-        for description in shielded_spends {
+        for description in transaction
+            .sapling_bundle()
+            .map_or(&vec![], |bundle| &bundle.shielded_spends)
```
sug0, line 200:
```suggestion
        for description in transaction
            .sapling_bundle()
            .iter()
            .flat_map(|bundle| &bundle.shielded_spends)
```

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -275,26 +274,19 @@ where
...
+                    if height == self.ctx.get_block_height()?
+                        && index == self.ctx.get_tx_index()? => {}
+                Some(_) => {
+                    return Err(Error::NativeVpError(
+                        native_vp::Error::SimpleMessage("Invalid MASP pin key"),
+                    ));
                 }
-            }
+                None => (),
+            },
```
sug0, line 289: move to `validate_pin_key` helper method

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
 
-        // Verify the changes to balance keys and return the transparent
-        // transfer data Get the token from the balance key of the MASP
-        let balance_addresses: Vec<[&Address; 2]> = keys_changed
+        let mut result = ChangedBalances::default();
+        // Get the changed balance keys
+        let counterparts_balances: Vec<_> = keys_changed
             .iter()
             .filter_map(is_any_shielded_action_balance_key)
             .collect();
```
sug0, line 304: shouldn't need to collect I think
```suggestion
        // Get the changed balance keys
        let counterparts_balances = keys_changed
            .iter()
            .filter_map(is_any_shielded_action_balance_key);
```

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
-            1 => masp_balances[0][0].to_owned(),
-            _ => {
-                // Only one transparent balance of MASP can be updated by the
-                // shielding or unshielding transaction
-                return Err(Error::NativeVpError(
-                    native_vp::Error::SimpleMessage(
-                        "More than one MASP transparent balance was modified",
-                    ),
-                ));
+        for (token, counterpart) in counterparts_balances {
```
sug0, line 306: [`Iterator::try_fold`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.try_fold) with `ChangedBalances`?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
+        .pre
+        .entry(vin.address)
+        .or_insert(ValueSum::zero());
+
+    match conversion_state.assets.get(&vin.asset_type) {
+        // Note how the asset's epoch must be equal to the present: users
+        // must never be allowed to backdate transparent inputs to a
+        // transaction for they would then be able to claim rewards while
+        // locking their assets for negligible time periods.
+        Some(((address, _asset_denom, digit), asset_epoch, _, _))
```
sug0, line 394: this tuple is super complex. IMO it would be better to extract some struct with named fields

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
+        )
+        .ok_or_err_msg("Underflow in output subtraction")?;
+
+    let bal_ref = changed_balances
+        .post
+        .entry(out.address)
+        .or_insert(ValueSum::zero());
+
+    match conversion_state.assets.get(&out.asset_type) {
+        Some(((address, _asset_denom, digit), asset_epoch, _, _))
```
sug0, line 473: same comment

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
+    sapling_value_balance: &I128Sum,
+    target_epoch: Epoch,
+    tokens: &BTreeMap<AssetType, (Address, token::Denomination, MaspDigitPos)>,
+    conversion_state: &ConversionState,
+) -> Result<()> {
+    let mut acc = ValueSum::<Address, SignedAmount>::from_sum(pre.clone());
+    for (asset_type, val) in sapling_value_balance.components() {
+        // Only assets with at most the target timestamp count
+        match conversion_state.assets.get(asset_type) {
+            Some(((address, _, digit), asset_epoch, _, _))
```
sug0, line 564: same comment

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -304,110 +296,327 @@ where
...
+                        "Overflow in MASP value balance",
+                    ))
+                })?;
+                let decoded_change =
+                    ValueSum::from_pair(token.clone(), decoded_change);
+                acc = acc.checked_sub(&decoded_change).ok_or_else(|| {
+                    Error::NativeVpError(native_vp::Error::SimpleMessage(
+                        "Overflow in MASP value balance",
+                    ))
+                })?;
```
sug0, line 599: this code looks nearly identical to me. can it be abstracted into a method/fn?

sug0 (changes requested):
mostly code quality related comments

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -437,356 +646,140 @@ where
...
+            &mut transparent_tx_pool,
+            epoch,
+            conversion_state,
+            &mut signers,
+        )?;
+
+        // Ensure that every account for which balance has gone down has
+        // authorized this transaction
+        for (addr, pre) in changed_balances.pre {
+            if changed_balances.post[&addr] < pre && addr != masp_address_hash {
```
grarco, line 705: Could this check be done only if the balances decrease more than implied by the `Transaction`? As in, if we expect a balance to go down by 10 but we see a decrease of 15 we request a signature since this means that something else happened in parallel to the masp tx that we cannot verify in here. Otherwise, if the decrease matches (or is less than) the exepected value of `Transaction` we don't require/verify the signature. Could this also apply to `validate_transparent_input`? 

. This (if correct) could lead to less signatures needed on the transaction and less verifications.

crates/namada/src/ledger/native_vp/mod.rs
```diff
@@ -637,3 +643,201 @@ pub(super) mod testing {
...
+pub struct SignedAmount(SignedAmountInt);
+
+impl<T> Add<T> for SignedAmount
+where
+    T: Into<SignedAmount>,
+{
+    type Output = Self;
+
+    fn add(self, other: T) -> Self {
+        Self(self.0.overflowing_add(other.into().0).0)
```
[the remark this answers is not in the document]
murisi: Resolved. `murisi+grarco/multi-tx-masp-vp-rebased2` implements `OverflowingAdd` instead.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks, good point. Done.

crates/core/src/token.rs
```diff
@@ -292,6 +293,26 @@ impl Amount {
...
+}
+
+impl CheckedSub for Amount {
+    fn checked_sub(&self, amount: &Self) -> Option<Self> {
+        self.raw
+            .checked_sub(amount.raw)
+            .map(|result| Self { raw: result })
+    }
+}
+
```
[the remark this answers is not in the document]
murisi: Thanks, understood. By the way, I've moved this checked arithmetic logic/traits into a separate crate so that the MASP crate can also use it.

crates/namada/src/ledger/native_vp/masp.rs
murisi: I think you have a point here. I'll try and sketch this change in a separate PR.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks. I've made a structure with named fields to represent this nested tuple in #3264 .

crates/namada/src/ledger/native_vp/masp.rs
murisi: Resolved in https://github.com/anoma/namada/pull/3186#discussion_r1606996403 .

crates/namada/src/ledger/native_vp/masp.rs
murisi: Resolved in https://github.com/anoma/namada/pull/3186#discussion_r1606996403 .

crates/namada/src/ledger/native_vp/masp.rs
murisi: Done, thanks.

crates/namada/src/ledger/native_vp/mod.rs
```diff
@@ -637,3 +643,201 @@ pub(super) mod testing {
...
+    Copy,
+    Clone,
+    Eq,
+    PartialEq,
+    BorshSerialize,
+    BorshDeserialize,
+    BorshSchema,
+    Default,
+)]
+pub struct SignedAmount(SignedAmountInt);
```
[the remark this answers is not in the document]
murisi: Thanks. I've renamed `SignedAmount` to `I320` and moved it into the `core` file `core/src/uint.rs`. Probably we can replace usage of `I256` with `I320` in a separate PR.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks. In #3264, this is no longer an issue because pinned key support has been removed.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks, good catch. Addressed this in #3264 .

PR #3214, opened 2024-05-09
more checked arith and other lints

### LINKED ISSUE #2555: Audit for all usage of unchecked arithmetic and update nomenclature

tzemanovic:
## Describe your changes

closes #2555

0.35.1

### DIFF
--- a/crates/governance/src/event.rs
+++ b/crates/governance/src/event.rs
@@ -71,7 +71,7 @@ impl GovernanceEvent {
         }
     }
 
-    /// Create a new proposal event for defaultwithwasm proposal
+    /// Create a new proposal event for default with wasm proposal
     pub fn passed_proposal(
         proposal_id: u64,
         has_proposal_code: bool,
@@ -86,6 +86,7 @@ impl GovernanceEvent {
         }
     }
 
+    /// Event for a reject proposal
     pub fn rejected_proposal(
         proposal_id: u64,
         has_proposal_code: bool,
@@ -101,14 +102,22 @@ impl GovernanceEvent {
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum ProposalEventKind {
     /// New proposal event
-    NewProposal { proposal_type: GovProposalType },
+    NewProposal {
+        /// Type of proposal
+        proposal_type: GovProposalType,
+    },
     /// Passed proposal
     Passed {
+        /// Does the proposal contain code?
         has_proposal_code: bool,
+        /// Did the proposal code run successfully?
         is_proposal_code_successful: bool,
     },
     /// Rejected proposal
-    Rejected { has_proposal_code: bool },
+    Rejected {
+        /// Does the proposal contain code?
+        has_proposal_code: bool,
+    },
 }
 
 impl From<GovernanceEvent> for Event {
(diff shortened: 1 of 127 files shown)


crates/governance/src/event.rs
```diff
@@ -101,14 +102,22 @@ impl GovernanceEvent {
...
 pub enum ProposalEventKind {
     /// New proposal event
-    NewProposal { proposal_type: GovProposalType },
+    NewProposal {
+        /// Type of proposal
+        proposal_type: GovProposalType,
+    },
     /// Passed proposal
     Passed {
+        /// Does the proposal contain code
```
brentstone, line 111: add a `?` at the end

crates/namada/src/ledger/native_vp/ibc/context.rs
```diff
@@ -80,25 +81,27 @@ where
     fn read_bytes(&self, key: &Key) -> Result<Option<Vec<u8>>> {
         match self.store.get(key) {
             Some(StorageModification::Write { ref value }) => {
-                let gas = key.len() + value.len();
+                let gas = checked!(key.len() + value.len())? as u64;
```
brentstone, line 84: fine not to use a `u64::try_from` as in other places? Same for below

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -405,9 +405,8 @@ where
             return Err(error);
         }
 
-        let mut transparent_tx_pool = I128Sum::zero();
+        let mut transparent_tx_pool = shielded_tx.sapling_value_balance();
         // The Sapling value balance adds to the transparent tx pool
```
brentstone, line 409: put the comment above the code? Or remove?

crates/proof_of_stake/src/epoched.rs
```diff
@@ -121,7 +121,12 @@ where
                                 && epoch
                                     > Self::sub_past_epochs(params, last_update)
                             {
-                                epoch = Epoch(epoch.0 - 1);
+                                epoch = Epoch(
+                                    epoch
+                                        .0
+                                        .checked_sub(1)
+                                        .expect("Cannot underflow"),
```
brentstone, line 128: perhaps use `epoch.prev()` instead

crates/namada/src/ledger/native_vp/ibc/context.rs
tzemanovic: I think it's ok to assume 64bit arch - we'll have a warning if not (https://github.com/anoma/namada/pull/3215)

tzemanovic: also it won't fail on 32bit

tzemanovic:
> Looks good to me in general, just left a couple questions.

thx! I fixed all the issues in https://github.com/anoma/namada/compare/ec36ae148190e183c70cc0c56171aa3bb8b1d359..31e54b9fcccb58d5628cc0362597d5848b576796

PR #3215, opened 2024-05-09
Emit CLI warning log msg on non 64-bit hardware

### LINKED ISSUE #826: Audit the codebase for erroneous usage of architecture dependent types

sug0:
## Describe your changes

Closes #826

`v0.35.1`

### DIFF
--- a/.changelog/unreleased/miscellaneous/3215-warn-on-non-64bit-arch.md
+++ b/.changelog/unreleased/miscellaneous/3215-warn-on-non-64bit-arch.md
@@ -1,0 +1,2 @@
+- Emit CLI warning log msg on non 64-bit node hardware.
+  ([\#3215](https://github.com/anoma/namada/pull/3215))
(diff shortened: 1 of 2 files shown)


cwgoes (dismissed):
Covers the condition we want, though it would be nice to abstract into a function.

brentstone:
Is this ready for draft in its current state?

tzemanovic (approved):

PR #3220, opened 2024-05-10
Harden max proposal bytes configuration and deserialization

### LINKED ISSUE #3189: Inconsistent block size limit handling in proposal preparation
Reported by Informal Systems.

# Description
The process of preparing proposals involves the `BlockAllocator`, which sets the maximum size of all transactions.
This maximum size, determined by the parameter `max_proposal_bytes` in the `BlockAllocator` configuration (/ block_alloc.rs#L141), should align with the `max_tx_bytes` parameter in the `RequestPrepareProposal` .
However, there lacks a validation step to ensure this alignment, potentially leading to discrepancies between these parameters over time.

# Problem Scenarios
The absence of consistency checking between `max_proposal_bytes` and `max_tx_bytes` poses several risks. Firstly, it violates Requirement 2 [PrepareProposal, tx-size] of the ABCI++ 0.37 specification, which mandates that the total size of transactions in a proposal does not exceed `max_tx_bytes` when calling `ResponsePrepareProposal` . Secondly, such discrepancies compromise liveness, as proposals from honest validators exceeding the maximum allowed would be rejected by the consensus engine.

# Recommendation
To address these issues, it would be desirable to implement a validation step to ensure that `max_proposal_bytes` does not exceed `max_tx_bytes` , or that the total size of transaction bytes in the proposal does not exceed the `max_tx_bytes` specified in the `RequestPrepareProposal` .

sug0:
## Describe your changes

Closes #3189

Indeed, Namada ignores the `max_tx_bytes` parameter from CometBFT, which may shrink dynamically based on the presence of evidence of misbehavior data in blocks.

This parameter is only passed as an argument to `PrepareProposal` ABCI calls, and (crucially) never to `ProcessProposal` calls, forcing it to be available under `ProcessProposal` via some other means.

As such, `max_proposal_bytes` (a semantically equivalent Namada-only parameter) was introduced in #843. To ensure the application never exceeds `max_tx_bytes`, CometBFT was configured with a much larger maximum block size (100 MiB, and 90 MiB max for `max_proposal_bytes`).

In #2187, we modified these bounds, setting the max block size to 16 MiB and the max value of `max_proposal_bytes` to 6 MiB. Evidence data is the only reason the space reserved for txs may shrink; as 10 MiB is plenty of available space for misbehavior reports, we should never exceed the maximum block size.

It is also important to state CometBFT faces a problem of the same nature. [During CometBFT block validation](https://github.com/cometbft/cometbft/blob/v0.37.6/state/validation.go#L15), `MaxTxBytes` is never checked. The only place where it is checked is [during block construction](https://github.com/cometbft/cometbft/blob/v0.37.6/state/execution.go#L139-L141), leaving the burden of this validation to ABCI apps. Unfortunately, as we stated, `ProcessProposal` is not called with `max_tx_bytes`...

That said, this PR does the following:

* Hardens `BorshDeserialize` for `ProposalBytes`.
* Avoids using "magic numbers" to configure CometBFT, and uses `ProposalBytes::MAX` instead.

`v0.35.1`

### DIFF
--- a/crates/apps/src/lib/node/ledger/tendermint_node.rs
+++ b/crates/apps/src/lib/node/ledger/tendermint_node.rs
@@ -4,7 +4,7 @@ use std::process::Stdio;
 use std::str::FromStr;
 
 use borsh_ext::BorshSerializeExt;
-use namada::core::chain::ChainId;
+use namada::core::chain::{ChainId, ProposalBytes};
 use namada::core::key::*;
 use namada::core::storage::BlockHeight;
 use namada::core::time::DateTimeUtc;
@@ -418,10 +418,7 @@ async fn update_tendermint_config(
         config.mempool.max_tx_bytes = 1024 * 1024;
 
         // Hold 50x the max amount of txs in a block
-        //
-        // 6 MiB is the default Namada max proposal size governance
-        // parameter -> 50 * 6 MiB
-        config.mempool.max_txs_bytes = 50 * 6 * 1024 * 1024;
+        config.mempool.max_txs_bytes = 50 * ProposalBytes::MAX.get();
 
         // Hold up to 4k txs in the mempool
         config.mempool.size = 4000;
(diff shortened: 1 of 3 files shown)


batconjurer (approved):

crates/apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -475,12 +472,13 @@ async fn write_tm_genesis(
             .try_into()
             .expect("Failed to convert initial genesis height");
     }
+    const EVIDENCE_AND_PROTOBUF_OVERHEAD: u64 = 10 * 1024 * 1024;
     let size = block::Size {
         // maximum size of a serialized Tendermint block.
```
cwgoes, line 477: Comment is now out of date

cwgoes (approved):
Changes are fine, but comment needs to be updated

tzemanovic (approved):

PR #3222, opened 2024-05-10
Murisi/masp governance changes

murisi:
## Describe your changes
Addresses issue #3068 . This PR allows changes to the MASP token map when authorized by governance.

#3186 


crates/shielded_token/src/storage_key.rs
```diff
@@ -121,6 +121,15 @@ pub fn is_masp_commitment_anchor_key(key: &storage::Key) -> bool {
         ] if *addr == address::MASP && prefix == MASP_NOTE_COMMITMENT_ANCHOR_PREFIX)
 }
 
+/// Check if the given storage key is a masp token map key
+pub fn is_masp_token_map_key(key: &storage::Key) -> bool {
+    matches!(&key.segments[..],
+    [DbKeySeg::AddressSeg(addr),
+             DbKeySeg::StringSeg(prefix),
+             ..
```
grarco, line 129: I think we can avoid this line since the token map key only contains the first two segments

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -816,3 +541,321 @@ where
...
+            .any(|key| is_masp_token_map_key(key))
+        {
+            self.is_valid_parameter_change(tx_data)?;
+        }
+        // The MASP transfer keys can only be changed by a valid Transaction
+        if masp_keys_changed
+            .iter()
+            .any(|key| is_masp_transfer_key(key))
+        {
+            self.is_valid_masp_transfer(tx_data, keys_changed)?;
```
grarco, line 857: Do you think we should prevent normal masp transfers from governance proposals? It shouldn't be possible anyway given the way we construct the transaction to execute a proposal, but maybe for extra safety. If this became possible, since we don't emit events for governance transactions, the clients could not fetch this data leading to an impossibility to submit valid masp transfers

crates/shielded_token/src/storage_key.rs
murisi: Thanks. Rectified.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks, understood. Rectified in #3264 .

PR #3226, opened 2024-05-13
Add line wrapping to the cli commands

### LINKED ISSUE #1796: Improve formatting of help messages in the CLI
Some CLI commands have huge help messages, with no line wrapping. For instance:
image

It would be helpful to:

- [ ] Wrap the output of help lines to 80 characters or less.
- [ ] For longer help messages, use something like...
    - [`Arg::long_help`](https://docs.rs/clap/4.3.4/clap/struct.Arg.html#method.long_help)
    - [`Arg::next_line_help`](https://docs.rs/clap/4.3.4/clap/struct.Arg.html#method.next_line_help)

batconjurer:
## Describe your changes
Closes #1796 

Extends claps `Arg` and `Command` types with namada traits. These add configurable line wrapping and hyphenation policies. Also sets help to always be on the next line, but this is easily changeable later.

### DIFF
--- a/crates/apps/src/lib/cli.rs
+++ b/crates/apps/src/lib/cli.rs
@@ -33,11 +33,11 @@ const WALLET_CMD: &str = "wallet";
 const RELAYER_CMD: &str = "relayer";
 
 pub mod cmds {
-
     use super::utils::*;
     use super::{
         args, ArgMatches, CLIENT_CMD, NODE_CMD, RELAYER_CMD, WALLET_CMD,
     };
+    use crate::wrap;
 
     /// Commands for `namada` binary.
     #[allow(clippy::large_enum_variant)]
@@ -148,7 +148,7 @@ pub mod cmds {
         fn def() -> App {
             <Self as Cmd>::add_sub(
                 App::new(Self::CMD)
-                    .wrapped_about("Node sub-commands.")
+                    .about(wrap!("Node sub-commands."))
                     .subcommand_required(true)
                     .arg_required_else_help(true),
             )
@@ -190,7 +190,7 @@ pub mod cmds {
         fn def() -> App {
             <Self as Cmd>::add_sub(
                 App::new(Self::CMD)
-                    .wrapped_about("Relayer sub-commands.")
+                    .about(wrap!("Relayer sub-commands."))
                     .subcommand_required(true),
             )
         }
@@ -428,7 +428,7 @@ pub mod cmds {
         fn def() -> App {
             <Self as Cmd>::add_sub(
                 App::new(Self::CMD)
-                    .wrapped_about("Client sub-commands.")
+                    .about(wrap!("Client sub-commands."))
                     .subcommand_required(true)
                     .arg_required_else_help(true),
             )
@@ -565,7 +565,7 @@ pub mod cmds {
         fn def() -> App {
             <Self as Cmd>::add_sub(
                 App::new(Self::CMD)
-                    .wrapped_about("Wallet sub-commands.")
+                    .about(wrap!("Wallet sub-commands."))
                     .subcommand_required(true)
                     .arg_required_else_help(true),
             )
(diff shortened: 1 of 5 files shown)


crates/apps/src/lib/cli.rs
```diff
@@ -3299,7 +3323,7 @@ pub mod args {
                      Unix,`$HOME/Library/Application Support/Namada` on \
                      Mac,and `%AppData%\\Namada` on Windows.",
```
sug0, line 3324:
```suggestion
                     Mac, and `%AppData%\\Namada` on Windows.",
```

sug0 (approved):
quite pleased with this change, hm but there's a small issue with text under backticks that contains spaces, for instance:
```
Options:
      --chain-id <chain-id>
          The chain ID.
      --base-dir <base-dir>
          The base directory is where the nodes, client and wallet
          configuration and state is stored. This value can also
          be set via `NAMADA_BASE_DIR` environment variable, but
          the argument takes precedence, if specified. Defaults to
          `$XDG_DATA_HOME/namada` (`$HOME/.local/share/namada` where
          `XDG_DATA_HOME` is unset) on Unix,`$HOME/Library/Application
          Support/Namada` on Mac,and `%AppData%\Namada` on Windows.
```
there was a line break on a space in text inside of backticks under the description of `--base-dir`, which might be a bit confusing. I'm fine living with this quirk, tbh, so lgtm!

crates/apps/src/lib/cli/utils.rs
```diff
@@ -154,6 +160,41 @@ pub const fn arg_multi<T, K>(name: &'static str) -> ArgMulti<T, K> {
...
+    fn wrapped_help(self, help: &str) -> Self;
+    #[allow(dead_code)]
+    fn wrapped_long_help(self, help: &str) -> Self;
+}
+
+impl WrappedHelp for ClapArg {
+    fn wrapped_help(self, help: &str) -> Self {
+        let dictionary = Standard::from_embedded(Language::EnglishUS).unwrap();
+        let options = Options::new(LINE_WRAPPING)
+            .word_splitter(WordSplitter::Hyphenation(dictionary));
```
sug0, line 173: maybe I'd also make these `options` a `LazyStatic` global, because I've noticed a small decrease in printing speed of the help msgs, which might partially be due to this being constructed for every flag (though I'm unsure if it does any transformations on the data structures at this stage)

batconjurer: It's probably the fact that it is recomputing the wrapping every time maybe? If I use macros, I can make the whole string formatting happen at compile time. I'll make a reference implementation and we'll see if we like it.

sug0: yeah, compile time formatting would solve this I guess. try to use `konst`. we have it in scope in `apps` I think

batconjurer: there were problems trying to use that unless I wanted to implement line wrapping from scratch. But I was able to use a macros version of the same crate. The first time the help is accessed, it's noticably slower, but every time after that it's not.

batconjurer: I must admit I don't like the way it looks as much though.

sug0: I reran the modified code. the help msgs do indeed print much faster; in fact, now they print instantly (or feel like that anyway), whereas before they took like 100ms to fully load. cosmetically, I don't notice any difference on the help msgs themselves. the code looks a bit uglier for sure, but I prefer better ux over better dx.

sug0 (dismissed):

crates/apps/src/lib/cli/utils.rs
```diff
@@ -154,6 +154,12 @@ pub const fn arg_multi<T, K>(name: &'static str) -> ArgMulti<T, K> {
     }
 }
 
+#[macro_export]
+macro_rules! wrap {
+    ($text:literal) => {
+        textwrap_macros::fill!($text, 60)
```
tzemanovic, line 160: I can haz 80 plz?

tzemanovic (dismissed):
neat!

sug0:
@brentstone fyi your merge didn't include the last commit

PR #3233, opened 2024-05-13
Bump to tendermint-rs 0.36

yito88:
## Describe your changes
- `tendermint-rs` v0.36
- `tower-abci` v0.14
- `ibc-rs` v0.53 (not released yet) [blocked]
  - Refactored with `ibc-derive` to use `#[derive(IbcClientState)]`
- `masp` with `bip0039` v0.12 [Need to be reviewed]


Fraccaman:
is this ready?

crates/events/src/extend.rs
```diff
@@ -88,18 +88,18 @@ impl AttributesMap for Vec<namada_core::tendermint::abci::EventAttribute> {
...
     fn retrieve_attribute(&self, key: &str) -> Option<&str> {
         self.iter().find_map(|attr| {
-            if attr.key == key {
-                Some(attr.value.as_str())
+            if attr.key_str().expect("Attribute key is malformed UFT-8") == key
+            {
+                Some(
+                    attr.value_str()
+                        .expect("Attribute value is malformed UTF-8"),
+                )
```
sug0, line 102: hm, this should be fine, as long as we always trust the input source of these events. either way, it looks like this is only fallible if someone constructs a v034 event, which contained base64 encoded strings. maybe we should fail silently, though? returning `None` on these conversion failures, and (perhaps) emitting a `tracing::debug` log msg

crates/ibc/src/event.rs
```diff
@@ -233,7 +233,16 @@ impl TryFrom<RawIbcEvent> for IbcEvent {
...
-            .map(|tag| (tag.key.to_string(), tag.value.to_string()))
+            .map(|tag| {
+                (
+                    tag.key_str()
+                        .expect("Attribute key is malformed UFT-8")
+                        .to_string(),
+                    tag.value_str()
+                        .expect("Attribute value is malformed UTF-8")
+                        .to_string(),
+                )
```
sug0, line 244: similar comment here about untrusted inputs

crates/events/src/extend.rs
yito88: thanks, it makes sense. I fixed it at 9087cac

sug0 (approved):

PR #3268, opened 2024-05-20
Track caller's emitted events

sug0:
## Describe your changes

Track the source code origin of some emitted event in Namada.

`v0.36.1`

### DIFF
--- a/crates/events/Cargo.toml
+++ b/crates/events/Cargo.toml
@@ -14,6 +14,7 @@ version.workspace = true
 
 [features]
 default = []
+debug = []
 mainnet = []
 migrations = [
   "namada_migrations",
--- a/crates/events/Cargo.toml
+++ b/crates/events/Cargo.toml
@@ -20,7 +20,7 @@ migrations = [
   "namada_migrations",
   "linkme",
 ]
-testing = []
+testing = ["debug"]
 
 [dependencies]
 namada_core = {path = "../core"}
--- a/crates/events/src/tracer.rs
+++ b/crates/events/src/tracer.rs
@@ -1,0 +1,271 @@
+//! Trace the origin of emitted events.
+//!
+//! ## Example
+//!
+//! ```
+//! #[track_caller]
+//! fn emit_event(event: crate::Event, events: &mut impl EmitEvents) {
+//!     let mut tracer = EventTracer::trace(events);
+//!     tracer.emit(event);
+//! }
+//! ```
+
+use std::borrow::Cow;
+use std::fmt;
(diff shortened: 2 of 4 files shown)


crates/events/src/tracer.rs
```diff
@@ -227,6 +228,29 @@ mod tests {
...
+            emit_event(ev, &mut events);
+            events.pop().unwrap()
+        };
+
+        let trace = event.read_attribute::<EventOrigin>().unwrap();
+
+        assert!(
+            trace.line > _EMIT_EVENT_START_LINE
+                && trace.line < _EMIT_EVENT_END_LINE
+        );
```
sug0, line 248: just realized this is wrong :(

we must check that the event originated...

crates/events/src/tracer.rs
```diff
@@ -227,6 +228,29 @@ mod tests {
...
 
+    #[test]
+    fn test_event_trace_emit_event() {
+        let event = {
+            let ev = Event::new(
+                EventTypeBuilder::new_with_type("test").build(),
+                EventLevel::Tx,
+            );
+            let mut events = Vec::with_capacity(1);
+            emit_event(ev, &mut events);
```
sug0, line 239: ...here instead

tzemanovic: it's right, no? I think it's more like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d5cf2e4b13d8cd82d7d8046657a9243b

crates/events/Cargo.toml
```diff
@@ -14,6 +14,7 @@ version.workspace = true
 
 [features]
 default = []
+debug = []
```
tzemanovic, line 17: we can probably add this to `testing = ["debug"]`

tzemanovic (approved):

crates/events/src/tracer.rs
```diff
@@ -0,0 +1,301 @@
...
+            const START_LINE: u32 = line!();
+            emit_event(ev, &mut events);
+            const END_LINE: u32 = line!();
+
+            (events.pop().unwrap(), START_LINE, END_LINE)
+        };
+
+        let trace = event.read_attribute::<EventOrigin>().unwrap();
+
+        assert!(trace.line > start_line && trace.line < end_line);
```
brentstone, line 255: nit: maybe `trace.line == start_line + 1 && trace.line == END_LINE - 1`?

sug0: doesn't matter to be precise, we only care to check if the event was emitted within a certain region of the source code

PR #3284, opened 2024-05-21
Replay protection contributes to App Hash

batconjurer:
## Describe your changes
Replay protection entries need to be verifiable and thus should contribute to the app hash. This PR makes a cryptographic commitment to all replay protection entries (the root of some implicit merkle tree) which is itself merklized.

### DIFF
--- a/crates/core/src/address.rs
+++ b/crates/core/src/address.rs
@@ -143,7 +143,7 @@ impl From<raw::Address<'_, raw::Validated>> for Address {
             raw::Discriminant::TempStorage => {
                 Address::Internal(InternalAddress::TempStorage)
             }
-            Discriminant::ReplayProtection => {
+            raw::Discriminant::ReplayProtection => {
                 Address::Internal(InternalAddress::ReplayProtection)
             }
         }
--- a/crates/core/src/address.rs
+++ b/crates/core/src/address.rs
@@ -15,6 +15,7 @@ use namada_migrations::*;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
+use crate::address::raw::Discriminant;
 use crate::ethereum_events::EthAddress;
 use crate::ibc::primitives::Signer;
 use crate::ibc::IbcTokenHash;
@@ -143,6 +144,9 @@ impl From<raw::Address<'_, raw::Validated>> for Address {
             raw::Discriminant::TempStorage => {
                 Address::Internal(InternalAddress::TempStorage)
             }
+            Discriminant::ReplayProtection => {
+                Address::Internal(InternalAddress::ReplayProtection)
+            }
         }
     }
 }
@@ -241,6 +245,13 @@ impl<'addr> From<&'addr Address> for raw::Address<'addr, raw::Validated> {
                 raw::Address::from_discriminant(raw::Discriminant::TempStorage)
                     .validate()
                     .expect("This raw address is valid")
+            }
+            Address::Internal(InternalAddress::ReplayProtection) => {
+                raw::Address::from_discriminant(
+                    raw::Discriminant::ReplayProtection,
+                )
+                .validate()
+                .expect("This raw address is valid")
             }
         }
     }
@@ -575,6 +586,8 @@ pub enum InternalAddress {
     Pgf,
     /// Masp
     Masp,
+    /// Replay protection
+    ReplayProtection,
     /// Address with temporary storage is used to pass data from txs to VPs
     /// which is never committed to DB
     TempStorage,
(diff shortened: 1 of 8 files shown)


crates/core/src/address.rs
```diff
@@ -15,6 +15,7 @@ use namada_migrations::*;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
+use crate::address::raw::Discriminant;
```
sug0, line 18:
```suggestion
```

crates/core/src/address.rs
```diff
@@ -143,6 +144,9 @@ impl From<raw::Address<'_, raw::Validated>> for Address {
             raw::Discriminant::TempStorage => {
                 Address::Internal(InternalAddress::TempStorage)
             }
+            Discriminant::ReplayProtection => {
+                Address::Internal(InternalAddress::ReplayProtection)
+            }
```
sug0, line 149:
```suggestion
            raw::Discriminant::ReplayProtection => {
                Address::Internal(InternalAddress::ReplayProtection)
            }
```
yes, this is uber OCD

crates/core/src/address/raw.rs
```diff
@@ -65,8 +65,10 @@ pub enum Discriminant {
     IbcToken = 13,
     /// MASP raw address.
     Masp = 14,
+    /// Replay protection
+    ReplayProtection = 15,
     /// Temporary storage address.
-    TempStorage = 15,
+    TempStorage = 16,
```
sug0, line 71: we should avoid doing this. when adding new addrs, append a new entry with the value of the previous entry + 1. this has less chances of breaking existing networks (I think the shielded expedition net is still up)

crates/core/src/hash.rs
```diff
@@ -156,6 +157,23 @@ impl Hash {
     }
 }
 
+impl<'a> Add<&'a Hash> for Hash {
```
sug0, line 160: :joy: this is some proper magic shit. I'm somewhat opposed to this `Add` impl, especially because it is not too clear what should be happening when adding two hashes

crates/state/src/wl_state.rs
```diff
@@ -227,14 +230,24 @@ where
...
+                .iter()
+                .try_fold(commitment, |mut acc, hash| {
+                    self.write_replay_protection_entry(
+                        batch,
+                        &replay_protection::current_key(hash),
+                    )?;
+                    acc = acc + hash;
+                    Ok::<_, Error>(acc)
+                })?;
+        self.write(&replay_prot_key, new_commitment)?;
```
sug0, line 249: I'm more in favor of manually updating a `Sha256` state here with `try_for_each` vs. `try_fold`, then writing the resulting hash after `try_for_each`, returned from `hasher.finalize()`

crates/core/src/address/raw.rs
batconjurer: Easy fix, I just went for alphabetic order

crates/core/src/hash.rs
batconjurer: I also wondered if this would be confusing, but then I tried to think what other operations people would guess it to be and I couldn't come up with any.

crates/state/src/wl_state.rs
batconjurer: That wouldn't quite work. We would need to finalize each iteration and create a new hasher. It ends up being needlessly noisy, hence the Add impl.

crates/core/src/hash.rs
batconjurer: I googled adding hashes and realized xor-ing hashes is incredibly common. So yeah, this is confusing

crates/state/src/wl_state.rs
sug0: hm I see. maybe we can define/reuse an existing merkle tree hasher fn for two hashes?

sug0: nvm just saw your last commit!

sug0 (changes requested):
from my end: basically only missing a rebase on `v0.36.1`, and running `make fmt`, which CI was complaining about

crates/core/src/hash.rs
```diff
@@ -155,12 +154,11 @@ impl Hash {
...
-
-impl<'a> Add<&'a Hash> for Hash {
-    type Output = Self;
 
-    fn add(self, rhs: &'a Hash) -> Self::Output {
+    /// Given hashes A and B, compute Sha256(A||B),
+    /// but if one value is the zero hash, the other
+    /// value is returned.
+    pub fn concat(self, rhs: &Hash) -> Self {
         let mut hasher = Sha256::default();
```
sug0, line 162: maybe we can defer the construction of the hash state to the if-branch where it is actually used

sug0 (approved):

grarco (approved):
LGTM!

PR #3293, opened 2024-05-22
Merklized but not diffed

yito88:
## Describe your changes
- Update some data in the subspace and the merkle tree, but not update the diffs
    - Add `diff_key_filter` to check if the data should be stored to diffs even if `merkle_tree_key_filter` is true
- Add `NoDiff` subtree in Merkle tree
    - Need it because we can't rebuild each subtree when some data aren't in diffs

`v0.36.1`

### DIFF
--- a/crates/node/src/shell/mod.rs
+++ b/crates/node/src/shell/mod.rs
@@ -353,13 +353,6 @@ where
     event_log: EventLog,
 }
 
-/// Merkle tree storage key filter. Return `false` for keys that shouldn't be
-/// merklized.
-pub fn is_merklized_storage_key(_key: &namada_sdk::storage::Key) -> bool {
-    // We need to merklize all keys for now
-    true
-}
-
 /// Storage key filter to store the diffs into the storage. Return `false` for
 /// keys whose diffs shouldn't be stored.
 pub fn is_key_diff_storable(key: &namada_sdk::storage::Key) -> bool {
@@ -448,7 +441,6 @@ where
             chain_id.clone(),
             native_token,
             config.shell.storage_read_past_height_limit,
-            is_merklized_storage_key,
             is_key_diff_storable,
         );
         let vp_wasm_cache_dir =
--- a/crates/node/src/shell/mod.rs
+++ b/crates/node/src/shell/mod.rs
@@ -355,7 +355,14 @@ where
 
 /// Merkle tree storage key filter. Return `false` for keys that shouldn't be
 /// merklized.
-pub fn is_merklized_storage_key(key: &namada_sdk::storage::Key) -> bool {
+pub fn is_merklized_storage_key(_key: &namada_sdk::storage::Key) -> bool {
+    // We need to merklize all keys for now
+    true
+}
+
+/// Storage key filter to store the diffs into the storage. Return `false` for
+/// keys whose diffs shouldn't be stored.
+pub fn is_key_diff_storable(key: &namada_sdk::storage::Key) -> bool {
     !(token::storage_key::is_masp_key(key)
         && *key != token::storage_key::masp_convert_anchor_key()
         && *key != token::storage_key::masp_token_map_key()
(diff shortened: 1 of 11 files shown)


batconjurer (dismissed):

crates/node/src/shell/mod.rs
```diff
@@ -355,7 +355,14 @@ where
 
 /// Merkle tree storage key filter. Return `false` for keys that shouldn't be
 /// merklized.
-pub fn is_merklized_storage_key(key: &namada_sdk::storage::Key) -> bool {
+pub fn is_merklized_storage_key(_key: &namada_sdk::storage::Key) -> bool {
```
cwgoes, line 358: We should just get rid of this function, honestly.

crates/node/src/storage/mod.rs
```diff
@@ -544,6 +563,41 @@ mod tests {
...
+            if key != client_counter_key() {
+                continue;
+            }
+            let merkle_key =
+                Key::from(NO_DIFF_KEY_PREFIX.to_string().to_db_key())
+                    .join(&key);
+            let tree =
+                state.get_merkle_tree(height, Some(StoreType::NoDiff))?;
+            assert_eq!(tree.root().0, roots.get(&height).unwrap().0);
+            match write_type {
```
cwgoes, line 582: Needs comments, this logic is not clear.

crates/node/src/shell/mod.rs
yito88: Removed cd441a2

crates/node/src/storage/mod.rs
yito88: added a comment cd441a2

crates/node/src/storage/mod.rs
```diff
@@ -437,15 +445,23 @@ mod tests {
             address::testing::nam(),
             None,
             is_merklized_storage_key,
+            is_key_diff_storable,
         );
+        let make_key = |suffix: u64| {
```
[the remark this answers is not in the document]
cwgoes: Can you explain what this whole `suffix % 3u64` logic is doing, though?

crates/node/src/storage/mod.rs
cwgoes: Can you explain the semantics of these 0/1/3 write types?

cwgoes (changes requested):
I still want clearer comments 😁

crates/node/src/storage/mod.rs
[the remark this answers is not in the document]
yito88: 44c9201

crates/node/src/storage/mod.rs
yito88: 44c9201

crates/node/src/storage/mod.rs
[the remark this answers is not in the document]
brentstone: Also what does "probable" mean in this context?

brentstone:
@yito88 @cwgoes do we need to use this for replay protection too or do we only need what is already in #3284?

crates/node/src/storage/mod.rs
```diff
@@ -513,37 +527,84 @@ mod tests {
...
+            match write_type {
+                0 => {
+                    // data was not updated
+                    if *current_state.get(&key).unwrap() {
+                        assert!(tree.has_key(&merkle_key)?);
+                    } else {
+                        assert!(!tree.has_key(&merkle_key)?);
+                    }
+                }
+                1 | 3 => {
```
cwgoes, line 595: but what do these numbers 1 or 3 mean?

cwgoes (changes requested):

cwgoes:
> @yito88 @cwgoes do we need to use this for replay protection too or do we only need what is already in #3284?

We do not need to use this for replay protection, but we do need to merge this to include everything in the app hash.

crates/node/src/storage/mod.rs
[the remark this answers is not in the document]
yito88: these data require proofs of the existence or non-existence

cwgoes (approved):

PR #3294, opened 2024-05-22
Enforce fee payment

### LINKED ISSUE #3075: Enforcing fee payment on transactions
After merging #2627 the transactions are now executed in order without the need to split the wrappers from the wasm transactions which poses an issue when it comes to fee payment that wasn't there in the previous execution model. In `process_proposal` we verify that the signer of a tx has enough balance to pay the fees for that tx: unfortunately though, with the new logic, the fee payment executed by the wrapper txs are now interspersed with the execution of the wasm codes. These could be transactions modifying the balance of the fee payers of the next transactions, meaning that the result of the check performed in `process_proposal` isn't valid anymore.

A possible solution could be to somewhat mimic the previous execution model by first collecting all the fees due by the transactions in the block and only then execute the wasm codes. This though would require some major changes (once more) in the execution model.

grarco:
## Describe your changes

Closes #3075.

Restructures `finalize_block` to collect the fees from all the transactions before executing the inner txs, putting this back in line with the validation logic of `process_proposal` and removing an attack vector.

v0.37.0

### DIFF
--- a/crates/namada/src/ledger/protocol/mod.rs
+++ b/crates/namada/src/ledger/protocol/mod.rs
@@ -180,7 +180,7 @@ impl From<Error> for DispatchError {
 
 /// Arguments for transactions' execution
 pub enum DispatchArgs<'a, CA: 'static + WasmCacheAccess + Sync> {
-    /// Protocoli tx data
+    /// Protocol tx data
     Protocol(&'a ProtocolTx),
     /// Raw tx data
     Raw {
--- a/crates/namada/src/ledger/protocol/mod.rs
+++ b/crates/namada/src/ledger/protocol/mod.rs
@@ -18,7 +18,7 @@ use namada_state::StorageWrite;
 use namada_token::event::{TokenEvent, TokenOperation, UserAccount};
 use namada_tx::data::protocol::{ProtocolTx, ProtocolTxType};
 use namada_tx::data::{
-    BatchResults, BatchedTxResult, TxResult, TxType, VpStatusFlags, VpsResult,
+    BatchResults, BatchedTxResult, TxResult, VpStatusFlags, VpsResult,
     WrapperTx,
 };
 use namada_tx::{BatchedTxRef, Tx};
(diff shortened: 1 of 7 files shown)


crates/tests/src/integration/ledger_tests.rs
```diff
@@ -1578,3 +1580,218 @@ fn change_validator_metadata() -> Result<()> {
...
+            "--source",
+            ALBERT_KEY,
+            "--target",
+            BERTHA,
+            "--token",
+            NAM,
+            "--amount",
+            // We want this transaction to consume all the remaining available
+            // balance. If we executed the inner txs right after the
+            // corresponding wrapper's fee paymwent this would succeed (but
```
batconjurer, line 1629:
```suggestion
            // corresponding wrapper's fee payment this would succeed (but
```

batconjurer (dismissed):

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -178,51 +178,150 @@ impl From<Error> for DispatchError {
     }
 }
 
+/// Arguments for transactions' execution
+pub enum DispatchArgs<'a, CA: 'static + WasmCacheAccess + Sync> {
+    /// Protocoli tx data
```
tzemanovic, line 183:
```suggestion
    /// Protocol tx data
```
(but I do like protocoli better tbh)

crates/tests/src/integration/ledger_tests.rs
```diff
@@ -1578,3 +1580,218 @@ fn change_validator_metadata() -> Result<()> {
 
     Ok(())
 }
+
+// Test that fee payment is enforced and aligned with process proposal. The test
+// generates a tx that subtract funds from the fee payer of a following tx. Test
+// that wrappers (and fee payments) are evaluated before the inner transactions.
+#[test]
+fn enforce_fee_payment() -> Result<()> {
```
tzemanovic, line 1588: update `.github/workflows/scripts/e2e.json` to run this in CI

tzemanovic: ah sry, nvm it's integration

tzemanovic (dismissed):

crates/tests/src/integration/ledger_tests.rs
grarco: Isn't that file for e2e tests only? I see from the CI logs that the test runs

crates/tx/src/data/protocol.rs
```diff
@@ -56,6 +56,7 @@ impl ProtocolTx {
 }
 
 #[derive(
+    Copy,
     Clone,
     Debug,
     BorshSerialize,
```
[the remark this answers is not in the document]
grarco: Fixed in 69857d68144ebb34bb2afadc37b2f81ec9293311

PR #3297, opened 2024-05-23
split transfer tx

### LINKED ISSUE #1677: Refactor transfer transaction creation and WASM to split shielded and transparent transfers
I audited the transfer transaction creation for any data leakage, and I didn't find any, but the way things are structured is very confusing, e.g. setting amount and token to sentinel amounts for MASP transactions. I think it would be clearer if we refactor transfers into four separate transactions:
- Transparent transfer
- Shielding
- Unshielding
- Shielded transfer

These should have separate WASM, separate functions, and separate CLI commands, and it should be clear what data is kept private and what data is sent to the ledger in each case (and this should be reflected by the WASM function's arguments).

### LINKED ISSUE #3345: Avoid unnecessary loops in client for masp
Relevant after merging #3297 

We don't need to loop in client when submitting shielded or unshielding transactions cause those do not fail based on the submission epoch.

See https://github.com/anoma/namada/pull/3297#discussion_r1620405811.

tzemanovic:
## Describe your changes

closes #1677, closes #3345

The transfer CLI command is split into:
- `transfer` (shielded)
- `transparent-transfer`
- `shield` (from transparent to shielded)
- `unshield` (from shielded to transparent)

With a dedicated wasm tx for each case, respectively:
- `tx_shielded_transfer`
- `tx_transparent_transfer`
- `tx_shielding_transfer`
- `tx_unshielding_transfer`

#3232 (based on 0.37) 

- diff https://github.com/anoma/namada/pull/3297/files/ef33d6289d91689ab67d8f61dd0f5b2641f0f9fd..aa45a7813f41faf7bfc0d3e1a48ac038dacba921


crates/core/src/ibc.rs
```diff
@@ -71,422 +47,3 @@ impl FromStr for IbcTokenHash {
...
-            BorshDeserialize::deserialize_reader(reader)?;
-        let message = IbcMsgNftTransfer::decode_vec(&msg)
-            .map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
-        Ok(Self { message, transfer })
-    }
-}
-
-/// IBC shielded transfer
-#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshDeserializer)]
-pub struct IbcShieldedTransfer {
```
grarco, line 159: This wasn't used right?

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -363,22 +361,6 @@ where
     Ok(extended_tx_result)
 }
 
-/// Load the wasm hash for a transfer from storage.
-///
-/// # Panics
-/// If the transaction hash is not found in storage
-pub fn get_transfer_hash_from_storage<S>(storage: &S) -> Hash
```
grarco, line 328: Ah this might have been my bad. I believe this was used to construct the tx for fee unshielding and I forgot to remove it, thanks!

crates/sdk/src/tx.rs
```diff
@@ -2884,60 +2882,274 @@ pub async fn build_transfer<N: Namada>(
...
+            source,
+            validated_amount.amount(),
+            check_balance,
+            args.tx.force,
+            context,
+        )
+        .await?;
+    }
+
     // TODO(namada#2597): this function should also take another arg as the fees
```
grarco, line 3038: This comment can be removed from here, for shielding transactions we'll pay fees from the transparent balance

crates/apps_lib/src/client/tx.rs
```diff
@@ -760,22 +772,59 @@ pub async fn submit_transfer(
...
+    }
+    Ok(())
+}
+
+pub async fn submit_shielding_transfer(
+    namada: &impl Namada,
+    args: args::TxShieldingTransfer,
+) -> Result<(), error::Error> {
+    // Repeat once if the tx fails on a crossover of an epoch
+    for _ in 0..2 {
```
grarco, line 806: I believe this loop is only needed for shielding transactions but we don't need it when doing shielded or unshielding. Maybe @murisi can confirm this

crates/core/src/ibc.rs
tzemanovic: yeah, I couldn't find any use

grarco (approved):

crates/apps_lib/src/client/tx.rs
murisi: > I believe this loop is only needed for shielding transactions but we don't need it when doing shielded or unshielding. Maybe @murisi can confirm this

Yes, agreed. You cannot submit a shielding transaction in later epochs because users would exploit this by pretending the transactions were submitted much earlier and claim a large reward. But late submissions are permitted for shielded and unshielding because there's nothing to be gained by the user by doing so.

tzemanovic: thx! Fixed in [3e61a3a](https://github.com/anoma/namada/pull/3297/commits/3e61a3aa8aff3fb12674b6e7c5aecee5c1ec71df)

tzemanovic: added one more commit to remove this from shielded transfer too - https://github.com/anoma/namada/pull/3297/files/838cec13b9d1d4b71f864b526be7addbab1d80e5..7ca37179029644c06a4f8f003e23fc3cb98c6503

tzemanovic:
auto-squashed fixups

PR #3310, opened 2024-05-24
Scheduled migrations

batconjurer:
Allow nodes to schedule a migrations json to be read and run to facilitate hard-forking. This is done by taking a migrations json and passing the path, a hash of the contents, and a block height to the node when starting the ledger.

### DIFF
--- a/crates/sdk/src/migrations.rs
+++ b/crates/sdk/src/migrations.rs
@@ -511,12 +511,12 @@ impl Display for UpdateStatus {
 pub fn commit<D: DB>(
     db: &D,
     height: BlockHeight,
-    mut migration: Option<ScheduledMigration<D::Migrator>>,
+    migration: Option<ScheduledMigration<D::Migrator>>,
 ) -> Option<ScheduledMigration<D::Migrator>>
 where
     D::Migrator: DeserializeOwned,
 {
-    let maybe_migration = migration.take();
+    let maybe_migration = migration;
     if let Some(migration) = maybe_migration.as_ref() {
         if height != migration.height {
             return maybe_migration;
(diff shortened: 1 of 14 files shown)


tzemanovic (approved):

crates/sdk/src/migrations.rs
```diff
@@ -431,6 +506,46 @@ impl Display for UpdateStatus {
...
+/// If so, apply it to the DB.
+pub fn commit<D: DB>(
+    db: &D,
+    height: BlockHeight,
+    mut migration: Option<ScheduledMigration<D::Migrator>>,
+) -> Option<ScheduledMigration<D::Migrator>>
+where
+    D::Migrator: DeserializeOwned,
+{
+    let maybe_migration = migration.take();
```
sug0, line 519: don't think you need to do this here as far as I can tell
```suggestion
    migration: Option<ScheduledMigration<D::Migrator>>,
) -> Option<ScheduledMigration<D::Migrator>>
where
    D::Migrator: DeserializeOwned,
{
    let maybe_migration = migration;
```

crates/sdk/src/migrations.rs
```diff
@@ -431,6 +506,46 @@ impl Display for UpdateStatus {
     }
 }
 
+/// Check if a scheduled migration should take place at this block height.
+/// If so, apply it to the DB.
+pub fn commit<D: DB>(
+    db: &D,
+    height: BlockHeight,
+    mut migration: Option<ScheduledMigration<D::Migrator>>,
+) -> Option<ScheduledMigration<D::Migrator>>
```
sug0, line 515: I'm slightly more in favor of mutating the migrations option in-place (i.e. by passing a `&mut Option<...>` ref), which we could steal with `std::mem::take` or `Option::take`

tzemanovic (approved):

PR #3318, opened 2024-05-27
Custom masp epoch tracker

### LINKED ISSUE #2833: Investigate whether we can create a second MASP epoch tracker
h/t @gavinly 

Investigate whether we can create a second MASP epoch, where `masp_epoch = regular_epoch % n` (`mod n`), where `n` could be, for example, `4` - the objective being to allow faster epochs without letting the conversion table grow too much.

grarco:
## Describe your changes

Closes #2833.

This PR tracks masp epochs differently from the usuale `Epoch`. A new protocol parameter defines a multiplicative factor that defines a masp epoch in terms on normal epoch (as in, every `x` normal epochs a new masp epochs starts). This way we can decouple the update logic of the the masp conversion tree from the rest of the protocol.

A new `MaspEpoch` is introduced: this is just a wrapper type around `Epoch` to make sure we don't pass the wrong epoch to masp functions that deal with masp epochs.

`v0.38.1`


crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -579,10 +579,13 @@ fn pgf_over_ibc_with_hermes() -> Result<()> {
 
 #[test]
 fn proposal_ibc_token_inflation() -> Result<()> {
+    const MASP_EPOCH_MULTIPLIER: u64 = 2;
```
grarco, line 582: This param seems to make the test fail (if we set it to 1 instead it works). @yito88 investigated the issue and found out that an update to Hermes solves it and makes this test pass

crates/core/src/masp.rs
```diff
@@ -23,6 +24,72 @@ use crate::string_encoding::{
...
+    }
+
+    /// Change to the previous masp epoch.
+    pub fn prev(&self) -> Option<Self> {
+        Some(Self(self.0.checked_sub(1)?))
+    }
+
+    /// Initialize a new masp epoch from the provided one
+    #[cfg(any(test, feature = "testing"))]
+    pub fn new(epoch: u64) -> Self {
```
sug0, line 88:
```suggestion
    pub const fn new(epoch: u64) -> Self {
```

crates/core/src/masp.rs
```diff
@@ -23,6 +24,72 @@ use crate::string_encoding::{
...
+    ) -> Result<Self, &'static str> {
+        Ok(Self(
+            epoch
+                .checked_div(masp_epoch_multiplier)
+                .ok_or("Masp epoch multiplier cannot be 0")?,
+        ))
+    }
+
+    /// Returns a 0 masp epoch
+    pub fn zero() -> Self {
```
sug0, line 77:
```suggestion
    pub const fn zero() -> Self {
```

crates/core/src/masp.rs
```diff
@@ -23,6 +24,72 @@ use crate::string_encoding::{
...
+        Ok(Self(Epoch(raw)))
+    }
+}
+
+impl MaspEpoch {
+    /// Converts and `Epoch` into a `MaspEpoch` based on the provided conversion
+    /// rate
+    pub fn try_from_epoch(
+        epoch: Epoch,
+        masp_epoch_multiplier: u64,
```
sug0, line 67:
```suggestion
        masp_epoch_multiplier: NonZeroU64,
```
avoids this being a faulty op I think (perhaps not bc of div underflows)

crates/core/src/parameters.rs
```diff
@@ -46,6 +46,9 @@ pub struct Parameters {
     pub implicit_vp_code_hash: Option<Hash>,
     /// Expected number of epochs per year (read only)
     pub epochs_per_year: u64,
+    /// The multiplier for masp epochs (it requires this amount of epochs to
+    /// transition to the next masp epoch)
+    pub masp_epoch_multiplier: u64,
```
sug0, line 51:
```suggestion
    pub masp_epoch_multiplier: NonZeroU64,
```

crates/apps_lib/src/config/genesis/templates.rs
```diff
@@ -293,6 +293,8 @@ pub struct ChainParams<T: TemplateValidation> {
     pub implicit_vp: String,
     /// Expected number of epochs per year
     pub epochs_per_year: u64,
+    /// How many epochs it takes to transition to the next masp epoch
+    pub masp_epoch_multiplier: u64,
```
sug0, line 297: IMO this name is a bit confusing. isn't this supposed to be more like a divider? so that masp epochs are faster than regular protocol epochs

crates/node/src/shell/finalize_block.rs
```diff
@@ -1710,6 +1711,27 @@ mod test_finalize_block {
...
+            )
+            .unwrap();
+
+        assert_eq!(shell.state.get_block_epoch().unwrap(), Epoch::default());
+
+        for _ in 1..masp_epoch_multiplier {
+            shell.start_new_epoch(None);
+            assert!(!shell.state.is_masp_new_epoch(true).unwrap());
+        }
+        shell.start_new_epoch(None);
```
sug0, line 1731: the assertion inside the loop won't be performed when `masp_epoch_multiplier` has a value of 1, because of the open range (`1..masp_epoch_multiplier`). maybe add the same assertion just before the loop starts
```rs
assert!(!shell.state.is_masp_new_epoch(true).unwrap());
```

crates/node/src/shell/finalize_block.rs
```diff
@@ -1710,6 +1711,27 @@ mod test_finalize_block {
...
+    fn test_masp_epoch_progression() {
+        let (mut shell, _broadcaster, _, _eth_control) = setup();
+
+        let masp_epoch_multiplier =
+            namada::ledger::parameters::read_masp_epoch_multiplier_parameter(
+                &shell.state,
+            )
+            .unwrap();
+
+        assert_eq!(shell.state.get_block_epoch().unwrap(), Epoch::default());
```
sug0, line 1725: the default epoch is actually 1 I think, no? or is it 0 :confused:

crates/node/src/shell/testing/node.rs
```diff
@@ -380,6 +380,21 @@ impl MockNode {
...
+        let masp_epoch_multiplier =
+            namada::parameters::read_masp_epoch_multiplier_parameter(
+                &self.shell.lock().unwrap().state,
+            )
+            .unwrap();
+        let mut epoch = Epoch::default();
+
+        for _ in 0..masp_epoch_multiplier {
+            epoch = self.next_epoch();
+        }
```
sug0, line 393: ehhh. can't we just use `checked_add`?

crates/node/src/shell/testing/node.rs
```diff
@@ -380,6 +380,21 @@ impl MockNode {
             .0
     }
 
+    pub fn next_masp_epoch(&mut self) -> Epoch {
```
sug0, line 383: is the mut ref needed?
```suggestion
    pub fn next_masp_epoch(&self) -> Epoch {
```

sug0 (changes requested):

crates/node/src/shell/testing/node.rs
grarco: Yes cause internally it calls `next_epoch` which requires a mutable ref

crates/node/src/shell/testing/node.rs
grarco: I believe we need to run the logic of `next_epoch`, don't we?

crates/node/src/shell/finalize_block.rs
grarco: We derive the `Default` trait and `Epoch` it's just a wrapper around a `u64`, os it should be 0

crates/node/src/shell/finalize_block.rs
grarco: Actually if the multiplier is indeed one the assertion will fail because in that case we would indeed start a new masp epoch too

crates/apps_lib/src/config/genesis/templates.rs
grarco: No we actually want masp epochs to be slower than normal ones so that we don't produce new conversions too often

crates/core/src/parameters.rs
grarco: I like this one but we have also other parameters that could benefit from this. Should this become a wider refactor?

sug0: yeah, replacing primitive integer types with their nonzero equivalents where prudent

sug0 (approved):

murisi (approved):
Looks good to me. Thanks!

crates/core/src/parameters.rs
Rigorously: Definitely helpful, because if max_proposal_bytes_key is set to 0, it will cause a consensus failure, crash the nodes and halt the chain.

Also, setting max_block_gas to 0, or at least lower than the minimum required for a transaction, will make the chain non-interactable, but the protocol people told me that is acceptable risk.

PR #3319, opened 2024-05-27
Store IBC trace when minting

### LINKED ISSUE #3229: Validate IBC event types
IBC event types are currently not being validated. We should check if they are valid event types, using [`ibc-rs`](https://lib.rs/crates/ibc).

https://github.com/anoma/namada/blob/4ed62290d3d3062b655c9bfc7cf08bd1b2230fd8/crates/ibc/src/event.rs#L143-L161

### LINKED ISSUE #3317: Store IBC denom when minting the IBC token
Currently, we store a minted IBC denom after receiving the packet. It requires checking the result by the IBC event.
However, we can store the IBC denom when minting the IBC token and don't need to check emitted IBC events. (We should do that but it wasn't possible with the old ibc-rs.)

yito88:
## Describe your changes
closes #3317, closes #3229 
Also,
- Remove `get_ibc_events` and IBC event conversion functions
- Fix `is_receiving_success` without checking events

`v0.37.0`

### DIFF
--- a/crates/ibc/src/context/token_transfer.rs
+++ b/crates/ibc/src/context/token_transfer.rs
@@ -142,7 +142,7 @@ where
             .map_err(TokenTransferError::from)
     }
 
-    fn store_ibc_denom(
+    fn maybe_store_ibc_denom(
         &self,
         owner: &Address,
         coin: &PrefixedCoin,
@@ -305,7 +305,7 @@ where
 
         // Store the IBC denom with the token hash to be able to retrieve it
         // later
-        self.store_ibc_denom(account, coin)?;
+        self.maybe_store_ibc_denom(account, coin)?;
 
         self.inner
             .borrow_mut()
--- a/crates/ibc/src/context/token_transfer.rs
+++ b/crates/ibc/src/context/token_transfer.rs
@@ -141,6 +141,32 @@ where
             .store_withdraw(token, added_withdraw)
             .map_err(TokenTransferError::from)
     }
+
+    fn store_ibc_denom(
+        &self,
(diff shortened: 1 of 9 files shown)


crates/ibc/src/event.rs
```diff
@@ -140,63 +140,6 @@ pub struct IbcEvent {
...
-            attributes: {
-                let mut attrs: HashMap<_, _> =
-                    namada_event.into_attributes().into_iter().collect();
-                attrs.with_attribute(event_domain_of::<Self>());
-                attrs
-            },
-        })
-    }
-}
-
```
sug0, line 199: I think this could have still been useful, but not to the ledger codebase directly. it could be useful as sdk library code for users to consume. maybe this can be restored in the future, with proper event type validation

sug0 (approved):
lgtm

crates/ibc/src/context/token_transfer.rs
```diff
@@ -141,6 +141,32 @@ where
             .store_withdraw(token, added_withdraw)
             .map_err(TokenTransferError::from)
     }
+
+    fn store_ibc_denom(
+        &self,
+        owner: &Address,
+        coin: &PrefixedCoin,
+    ) -> Result<(), TokenTransferError> {
+        if coin.denom.trace_path.is_empty() {
```
cwgoes, line 150: Why would we call the function in this case, though? If we must do it like this, we should call the function `maybe_store_ibc_denom` or something, since it sometimes doesn't store anything at all.

cwgoes (dismissed):
ACK modulo nit

PR #3322, opened 2024-05-28
pgf: let governance upgrade storage

Fraccaman:

### DIFF
--- a/crates/namada/src/ledger/pgf/mod.rs
+++ b/crates/namada/src/ledger/pgf/mod.rs
@@ -59,8 +59,20 @@ where
         // Find the actions applied in the tx
         let actions = self.ctx.read_actions()?;
 
+        // Is VP triggered by a governance proposal?
+        let is_governance_proposal = is_proposal_accepted(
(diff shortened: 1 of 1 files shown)


cwgoes (changes requested):
Maybe a quick unit test would be nice?

crates/namada/src/ledger/pgf/mod.rs
```diff
@@ -59,8 +59,20 @@ where
...
+                .data(batched_tx.cmt)
+                .unwrap_or_default()
+                .as_ref(),
+        )
+        .unwrap_or_default();
+
         // There must be at least one action if any of the keys belong to PGF
-        if actions.is_empty()
+        if !is_governance_proposal
+            && actions.is_empty()
```
tzemanovic, line 75: I think if we don't need to check anything else for proposals then we should return early - this will continue to the `keys_changed` loop below which might reject a proposal

brentstone:
what's the status here? Mergeable for Friday?

PR #3350, opened 2024-06-03
Remove old disposable keys from the wallet

### LINKED ISSUE #3239: Garbage collect disposable keys from the wallet
Somewhat related to #3053

Disposable keys accumulate in the wallet after performing fee unshielding. We must arrange to automatically clear them from the wallet at some point, to avoid polluting it with keys that should only be used once.

sug0:
## Describe your changes

Closes #3239

Garbage collect disposable signing keys.

`v0.38.1`

### DIFF
--- a/.changelog/unreleased/improvements/3350-diposable-keys-cleanup.md
+++ b/.changelog/unreleased/improvements/3350-diposable-keys-cleanup.md
@@ -1,0 +1,2 @@
+- Remove old disposable keys from the wallet.
+  ([\#3350](https://github.com/anoma/namada/pull/3350))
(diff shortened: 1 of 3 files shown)


tzemanovic (approved):

grarco (approved):
I still believe the strongest solution would be to wait for a confirmation from the client and two blocks on top (in case of a rollback) just to have absolute certainty that the transaction has been executed and we don't need the address anymore. Anyway, this solution is already a nice improvement, thanks!

brentstone:
should we implement @grarco's suggestion or merge as is? @sug0

sug0:
@brentstone I think we should increase the key lifetime a bit; let them live for, let's say 1 week, before deleting them

Issue #2597, opened 2024-02-12
Embed masp transactions inside fee payment


grarco:
The masp client has some intrinsics limitations when it comes to producing transactions which derives from previous txs' output descriptions for which it has not queried a node. More specifically, given the missing notes coming from other txs not produced by this very wallet, it's impossible to construct a merkle tree matching one of those seen by the protocol: the result is that (almost) all txs built as such will be rejected because of an invalid commitment anchor.

To get around this, for shielding and unshielding operations (the ones affected by this), as proposed by @murisi, we could embed them directly inside the fee unshielding `Transaction` object, so that the client can use the same, confirmed notes to produce a single transaction, avoiding the issue altogether.

To achieve this, we need to modify the client to produce a single `Transaction` object in such cases and attach an empty tx as the inner to mock a no-op. We could also support another custom tx as the inner to allow running multiple transactions (the ones embedded in the fee payment and the actual inner).

For safety reasons this depends on #2592 so that the fee unshielding operation cannot be exploited to perform free transactions.
It also depends on #2595 and #2596 for the unshielding case to support different tokens between the actual fee unshielding and the proper unshielding txs and also to avoid an extra transparent transfer from the fee payer to the actual recipient.

Finally, we need to review the logic in `run_fee_unshielding` in terms of error management to see if needs to be changed to support this feature.

cwgoes:
@grarco This will change the transaction format, right? Only for MASP transactions or also for all transactions?

cwgoes:
@grarco Would it be possible to do a special case analysis of an inner transaction with an output from a MASP transfer that pays the fee?

cwgoes:
New logic discussed 2024.04.26:
- When we analyze for "sufficient fee or not", we should first look at the wrapper, and in the special case where the wrapper fee paying address doesn't have enough balance, check if the inner contains a valid unshield to that address. Reject immediately if the inner does not contain an unshield.

cwgoes:
Note with batching: first inner tx in a batch must pay for the fees.

cwgoes:
Order of operations here:
1. Remove fee unshielding bit in wrapper tx (#3217)
2. Vectorize/refactor the transfer struct & inner tx (@grarco is working on this)
3. Add in support for the first inner tx unshielding a fee (@grarco will work on this)

PR #3356, opened 2024-06-03
Vectorize transfers

### LINKED ISSUE #2596: Allow multiple transfers in a single tx
We should allow multiple token transfers in a single transaction, i.e. involving more than one sender/receiver and with different tokens. We could either modify the current `tx_transfer` or come up with a different one. We'll also need to update the client and SDK to produce this new type of transactions.

grarco:
## Describe your changes

Closes #2596.
Required modification to enable #2597.

Modifies transparent and masp transactions to allow for multi-party transfers. Transparent transfers allow for multiple sources, targets token and amounts.

Shielding and unshielding masp transfers are instead vectorized in a slightly different way: shieldings are vectorized in a fan-in fashion: multiple sources, amounts and tokens but the target is a single `PaymentAddress`. Unshieldings are fan-out: multiple targets, tokens and amount but the source is a single `SpendingKey`.

Ibc shielded actions remain unchanged as I believe we can't vectorize them because of the way packets are designed. 

This PR does NOT update the client to support vectorized transfers as at the moment it's unclear to me how to achieve that maintaining a decent UX: for now we only support these transactions via the SDK.

`v0.39.0`

### DIFF
--- a/crates/apps_lib/src/client/tx.rs
+++ b/crates/apps_lib/src/client/tx.rs
@@ -743,7 +743,18 @@ pub async fn submit_transparent_transfer(
     namada: &impl Namada,
     args: args::TxTransparentTransfer,
 ) -> Result<(), error::Error> {
-    submit_reveal_aux(namada, args.tx.clone(), &args.source).await?;
+    submit_reveal_aux(
+        namada,
+        args.tx.clone(),
+        &args
+            .data
+            .first()
+            .ok_or_else(|| {
+                error::Error::Other("Missing transfer data".to_string())
+            })?
+            .source,
+    )
+    .await?;
 
     let (mut tx, signing_data) = args.clone().build(namada).await?;
 
--- a/crates/apps_lib/src/client/tx.rs
+++ b/crates/apps_lib/src/client/tx.rs
@@ -743,6 +743,14 @@ pub async fn submit_transparent_transfer(
     namada: &impl Namada,
     args: args::TxTransparentTransfer,
 ) -> Result<(), error::Error> {
+    if args.data.len() > 1 {
+        // TODO(namada#3379): Vectorized transfers are not yet supported in the
+        // CLI
+        return Err(error::Error::Other(
+            "Unexpected vectorized transparent transfer".to_string(),
+        ));
+    }
+
     submit_reveal_aux(
         namada,
         args.tx.clone(),
--- a/crates/light_sdk/src/transaction/transfer.rs
+++ b/crates/light_sdk/src/transaction/transfer.rs
@@ -25,7 +25,7 @@ impl Transfer {
         amount: DenominatedAmount,
         args: GlobalArgs,
     ) -> Self {
-        let data = namada_sdk::token::TransparentTransfer {
+        let data = namada_sdk::token::TransparentTransferData {
             source,
             target,
             token,
--- a/crates/light_sdk/src/transaction/transfer.rs
+++ b/crates/light_sdk/src/transaction/transfer.rs
@@ -1,7 +1,11 @@
 use namada_sdk::address::Address;
 use namada_sdk::hash::Hash;
 use namada_sdk::key::common;
-use namada_sdk::token::DenominatedAmount;
+use namada_sdk::token::transaction::Transaction;
+use namada_sdk::token::ShieldingTransferData;
+pub use namada_sdk::token::{
+    DenominatedAmount, TransparentTransfer, UnshieldingTransferData,
+};
 use namada_sdk::tx::data::GasLimit;
 use namada_sdk::tx::{
     Authorization, Tx, TxError, TX_SHIELDED_TRANSFER_WASM,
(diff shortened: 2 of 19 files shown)


crates/apps_lib/src/client/tx.rs
```diff
@@ -743,7 +743,18 @@ pub async fn submit_transparent_transfer(
     namada: &impl Namada,
     args: args::TxTransparentTransfer,
 ) -> Result<(), error::Error> {
-    submit_reveal_aux(namada, args.tx.clone(), &args.source).await?;
+    submit_reveal_aux(
+        namada,
+        args.tx.clone(),
+        &args
+            .data
+            .first()
```
yito88, line 751: Do we need to check all sources, not only the first one? Should the source be the same in each transfer?

tzemanovic: good q - I think this should probably collect a set of unique sources and then iterate over them

grarco: For the time being I haven't really implemented any kind of support for vectorized transfers in the cli because I believe the ux would be quite bad. So there's no real way to get more than one source in here

crates/apps_lib/src/client/tx.rs
```diff
@@ -760,22 +783,59 @@ pub async fn submit_transfer(
...
+    for _ in 0..2 {
+        let (mut tx, signing_data, tx_epoch) =
+            args.clone().build(namada).await?;
+
+        if args.tx.dump_tx {
+            tx::dump_tx(namada.io(), &args.tx, tx);
+            break;
+        } else {
+            sign(namada, &mut tx, &args.tx, signing_data).await?;
+            let cmt_hash = tx.first_commitments().unwrap().get_hash();
```
batconjurer, line 826: I guess this relates to your comment above that vectorization is not really supported in the cli atm.

crates/apps_lib/src/client/tx.rs
tzemanovic: maybe we could add some check in here then that they are all the same

crates/light_sdk/src/transaction/transfer.rs
```diff
@@ -1,46 +1,102 @@
...
             args,
-            init_proposal.serialize_to_vec(),
-            TX_TRANSFER_WASM.to_string(),
+            transfers,
+            TX_TRANSPARENT_TRANSFER_WASM.to_string(),
         ))
     }
 
+    /// Build a shielded transfer transaction from the given parameters
+    pub fn shielded(
```
batconjurer, line 37: how out of data is our c bindings example?

grarco: I fear very much 😢

crates/sdk/src/signing.rs
```diff
@@ -703,25 +702,170 @@ fn format_outputs(output: &mut Vec<String>) {
...
+
+    fn update_token_amount_map<'a>(
+        map: &mut HashMap<&'a Address, DenominatedAmount>,
+        token: &'a Address,
+        amount: DenominatedAmount,
+    ) -> Result<(), Error> {
+        match map.get_mut(token) {
+            Some(prev_amount) => {
+                *prev_amount =
+                    prev_amount.checked_add(amount).ok_or_else(|| {
```
batconjurer, line 836: shouldn't we use the `checked!` macro here?

batconjurer (changes requested):
Did I miss or are the new wasms added to the whitelist?

crates/apps_lib/src/client/tx.rs
grarco: That's correct, I can probably add some checks like suggested above

crates/apps_lib/src/client/tx.rs
grarco: I've added a check on the length of the input data in adbc489ed6bb8934ead1b0b0bfdee095aaf157c4

crates/sdk/src/signing.rs
```diff
@@ -703,25 +702,170 @@ fn format_outputs(output: &mut Vec<String>) {
...
+        &self,
+        address: TransferSide<'_>,
+    ) -> Result<HashMap<&Address, DenominatedAmount>, Error> {
+        Ok(match self {
+            TokenTransfer::Transparent(transfers) => {
+                let mut map: HashMap<&Address, DenominatedAmount> =
+                    HashMap::new();
+
+                for transfer in &transfers.0 {
+                    match address {
```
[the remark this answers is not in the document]
grarco: Done in 280b8e245005d7cf8a15d16d650eafe380e72ab7

crates/sdk/src/signing.rs
grarco: Done in 280b8e245005d7cf8a15d16d650eafe380e72ab7

crates/sdk/src/masp.rs
```diff
@@ -1640,235 +1628,305 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-                        .add_transparent_input(TxOut {
-                            asset_type: *asset_type,
-                            value: amount_part,
-                            address: script,
-                        })
-                        .map_err(builder::Error::TransparentBuild)?;
-                }
+                // Load the current shielded context given the spending key we
+                // possess
+                let mut shielded = context.shielded_mut().await;
```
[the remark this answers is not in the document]
grarco: We actually want to load the context only if the source is a spending key, which we can know only from within the loop. But I've added a boolean in 937a972e40b976b1f771daa5097bef7a1c2f75f2 to avoid loading the context more than once

tzemanovic (approved):

murisi:
@grarco Not sure if this is strictly necessary, but do any of the wasms in this PR allow a transfer with transparent inputs, transparent outputs, and shielded inputs /outputs all there simultaneously?

grarco:
> @grarco Not sure if this is strictly necessary, but do any of the wasms in this PR allow a transfer with transparent inputs, transparent outputs, and shielded (inputs or outputs) all there simultaneously?

Not really, it's more focused on vectorizing the 4 different types of transfers: so shielded transfers are already vectorized since they can contain multiple spend/output notes. Transparent transfers have been updated to support multiple sources/targets/tokens. Shielding has been updated to support multiple transparent sources and unshielding has been updated to support multiple transparent outputs

Issue #305, opened 2022-08-09
Add E2E Tests for MASP Wallet


murisi:
Make E2E tests for MASP wallet functionality to prevent future regressions. It should essentially test all subcommands of `anomaw masp`. I.e. under various CLI parameters, it should test the correct functioning of:
* `namadaw masp gen-addr ...`
* `namadaw masp add ...`
* `namadaw masp gen-key ...`
* `namadaw masp list-keys ...`
* `namadaw masp list-addrs ...`
* `namadaw masp find ...`

See transparent wallet E2E tests at `tests/src/e2e/wallet_tests.rs` for guidance on how to do this.

karbyshev:
The wallet CLI structure has been changed #2135

brentstone:
@murisi @batconjurer @grarco @Fraccaman how relevant is this issue still?

PR #3365, opened 2024-06-04
masp e2e test

### LINKED ISSUE #3253: Test MASP proof/sig verifications
We should ensure to have at least one test that checks the correctness of the proofs/signatures verification in the masp vp. All the integration tests run with mocked builder/verifier so we cannot rely on them. The last masp e2e test will be removed in #3217. We still have the shielded action e2e tests but I believe it would be better to have a specific one.

tzemanovic:
## Describe your changes

closes #3253
closes https://github.com/anoma/namada/issues/305

v0.40.0

- ~~[ ] I have added a changelog~~ tests only

### DIFF
--- a/crates/tests/src/e2e/ledger_tests.rs
+++ b/crates/tests/src/e2e/ledger_tests.rs
@@ -2316,3 +2316,153 @@ fn rollback() -> Result<()> {
 
     Ok(())
 }
+
+/// We test shielding, shielded to shielded and unshielding transfers:
+/// 1. Run the ledger node
+/// 2. Shield 20 BTC from Albert to PA(A)
+/// 3. Transfer 7 BTC from SK(A) to PA(B)
+/// 4. Assert BTC balance at VK(A) is 13
+/// 5. Unshield 5 BTC from SK(B) to Bertha
+/// 6. Assert BTC balance at VK(B) is 2
+///
+/// NOTE: We need this test to verify the correctness of the proofs generation
+/// and verification process because integration tests use mocks.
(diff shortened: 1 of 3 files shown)


grarco (dismissed):
Minor comment, but this is good, thank you!

crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2322,3 +2322,150 @@ fn rollback() -> Result<()> {
 
     Ok(())
 }
+
```
grarco, line 2325: Can we just add another line in the comment here to specify that we need this test to verify the correctness of the proofs generation/verification process and that this cannot be achieved in the integration tests because of the mocks? Just saying cause I can see ourselves forgetting about this thing an being tempted to remove this test later down the road 😅

tzemanovic: sure can!

tzemanovic: added [55406a9](https://github.com/anoma/namada/pull/3365/commits/55406a9edee320c9e35a7f2e90366c1d5791e451)

grarco (approved):

Fraccaman:
this is great, thanks! Do u think we could also switch the benches to use the fake proof verification for MASP? it would save a lot of time

brentstone:
what is the status here?

grarco:
> what is the status here?

I believe it only needs to be rebased on the latest release

PR #3366, opened 2024-06-04
Remove max block time param

sug0:
## Describe your changes

Remove the `max_expected_time_per_block` genesis parameter.

`v0.39.0`

### DIFF
--- a/crates/parameters/src/lib.rs
+++ b/crates/parameters/src/lib.rs
@@ -467,7 +467,7 @@ where
     init_storage(&params, storage)
 }
 
-const BLOCK_TIME_ESTIMATE_UPPER_BOUND: DurationSecs = DurationSecs(120);
+const BLOCK_TIME_ESTIMATE_UPPER_BOUND: DurationSecs = DurationSecs(60);
 
 /// Return an estimate of the maximum time taken to decide a block,
 /// by sourcing block headers from up to `num_blocks_to_read`.
--- a/crates/parameters/src/lib.rs
+++ b/crates/parameters/src/lib.rs
@@ -478,14 +478,12 @@ where
     S: StorageRead,
 {
     let ending_height = last_block_height.0;
-    let beginning_height = ending_height.saturating_sub(num_blocks_to_read);
+    let beginning_height =
+        ending_height.saturating_sub(num_blocks_to_read.saturating_sub(1));
 
     let block_timestamps = {
-        let vec_size = checked!(ending_height - beginning_height + 1)
-            .into_storage_result()?;
-
         let mut ts = Vec::with_capacity(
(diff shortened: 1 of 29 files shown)


crates/sdk/src/queries/shell.rs
```diff
@@ -131,6 +138,36 @@ where
...
+/// Return an estimate of the maximum time taken to decide a block
+fn max_block_time<D, H, V, T>(
+    ctx: RequestCtx<'_, D, H, V, T>,
+) -> namada_storage::Result<DurationSecs>
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+{
+    // NB: get max time over this num of blocks
+    const NUM_BLOCKS_TO_READ: u64 = 5;
```
grarco, line 150: Should we allow the caller to pass this value as an argument?

crates/parameters/src/lib.rs
```diff
@@ -459,3 +442,208 @@ where
...
+    S: StorageRead,
+{
+    let EpochDuration {
+        min_num_of_blocks,
+        min_duration: DurationSecs(min_duration),
+    } = read_epoch_duration_parameter(storage)?;
+
+    checked!(min_duration / min_num_of_blocks)
+        .map(|block_time| {
+            // NB: fallback to one block per epoch
```
grarco, line 504: On block per second here?

crates/sdk/src/queries/shell.rs
sug0: no, because this is an unprivileged rpc method. a value too high can cause nodes to block for too long performing storage reads

crates/parameters/src/lib.rs
sug0: one block per `$minimum-epoch-duration`, which you compute from `$epochs-per-year`

yito88:
IBC E2E tests passed with [Hermes](https://github.com/heliaxdev/hermes/tree/1.8.2-namada-3366) with 9b41780

tzemanovic (dismissed):

yito88:
The IBC E2E tests in CI failed when creating channels. I'm investigating it...

yito88:
hmm, flaky tests... These tests didn't fail locally. After some retries, they passed.

sug0:
> hmm, flaky tests... These tests didn't fail locally. After some retries, they passed.

@yito88 I suppose it's a matter of decreasing the max block time, to increase the amount of delta blocks in the masp tx expiration

brentstone:
Is this ready for draft now?

sug0:
still messing around with timeout values to avoid flakes in ci

sug0:
@brentstone I think this is less flakey now. should be ok to merge now

PR #3372, opened 2024-06-05
Outsource masp signature verification

### LINKED ISSUE #3312: Outsource signature verification from masp vp
Currently we check some signatures in the masp vp when the balance changes of a transaction do not match the state transition implied by the masp `Transaction`.

We might be able to move this check back to the involved validity predicated by using a masp `Action`. More specifically, instead of verifying the signatures we can check that their vps have been triggered and let the validation up to them.

However, these vps may not be triggered by the transaction, so the transaction itself should write an `Action` (a temporary value in storage that doesn't get committed to storage) to manually trigger these vps. If the transaction fails to do so, the check in the masp vp on the triggered vps will fail, leading to a rejection of the tx.

This would decouple the validation logic of the involved addresses from the masp vp (allowing custom logics) and could lead to less signatures verification (with the current implementation we could end up verifying the same signature twice, once in the user vp and the second time in the masp vp).

grarco:
## Describe your changes

Closes #3312.

Swaps the signature verification call in the masp vp with a check on the triggered vps. Introduces a new masp `Action` to support signature verification in the affected vps since we expect them to not be triggered by the transaction itself (no changed keys).

#3516 (diffs for review: https://github.com/anoma/namada/pull/3372/files/17b9fcfa9c85f5a115646120f6595e8cf28c38a7..fc63450e3dfc2882c467d96a2600d761fb40853b)

### DIFF
--- a/crates/namada/src/ledger/native_vp/masp.rs
+++ b/crates/namada/src/ledger/native_vp/masp.rs
@@ -908,10 +908,23 @@ where
             } else if let Some(TAddrData::Addr(signer)) =
                 changed_bals_minus_txn.decoder.get(&signer)
             {
-                // Otherwise the owner's vp must have been triggered
+                // Otherwise the owner's vp must have been triggered and the
+                // relative action must have been written
                 if !verifiers.contains(signer) {
                     let error = native_vp::Error::new_alloc(format!(
                         "The required vp of address {signer} was not triggered"
+                    ))
+                    .into();
+                    tracing::debug!("{error}");
+                    return Err(error);
+                }
+
+                if !namada_tx::action::is_required_masp_signer(
+                    &self.ctx, signer,
+                )? {
+                    let error = native_vp::Error::new_alloc(format!(
+                        "The required masp signer action for address {signer} \
+                         is missing"
                     ))
                     .into();
                     tracing::debug!("{error}");
--- a/crates/namada/src/ledger/native_vp/masp.rs
+++ b/crates/namada/src/ledger/native_vp/masp.rs
@@ -21,7 +21,6 @@ use namada_core::ibc::apps::transfer::types::msgs::transfer::MsgTransfer as IbcM
 use namada_core::ibc::apps::transfer::types::packet::PacketData;
 use namada_core::masp::{addr_taddr, encode_asset_type, ibc_taddr, MaspEpoch};
 use namada_core::storage::Key;
-use namada_gas::GasMetering;
 use namada_governance::storage::is_proposal_accepted;
 use namada_ibc::core::channel::types::commitment::{
     compute_packet_commitment, PacketCommitment,
@@ -750,6 +749,7 @@ where
         &self,
         tx_data: &BatchedTxRef<'_>,
         keys_changed: &BTreeSet<Key>,
+        verifiers: &BTreeSet<Address>,
     ) -> Result<()> {
         let masp_epoch_multiplier =
             namada_parameters::read_masp_epoch_multiplier_parameter(
@@ -908,27 +908,15 @@ where
             } else if let Some(TAddrData::Addr(signer)) =
                 changed_bals_minus_txn.decoder.get(&signer)
             {
-                // Otherwise the signer must be decodable so that we can
-                // manually check the signatures
-                let public_keys_index_map =
-                    crate::account::public_keys_index_map(
-                        &self.ctx.pre(),
-                        signer,
-                    )?;
-                let threshold =
-                    crate::account::threshold(&self.ctx.pre(), signer)?
-                        .unwrap_or(1);
-                let mut gas_meter = self.ctx.gas_meter.borrow_mut();
-                tx_data
-                    .tx
-                    .verify_signatures(
-                        &[tx_data.tx.raw_header_hash()],
-                        public_keys_index_map,
-                        &Some(signer.clone()),
-                        threshold,
-                        || gas_meter.consume(crate::gas::VERIFY_TX_SIG_GAS),
-                    )
-                    .map_err(native_vp::Error::new)?;
+                // Otherwise the owner's vp must have been triggered
+                if !verifiers.contains(signer) {
+                    let error = native_vp::Error::new_alloc(format!(
+                        "The required vp of address {signer} was not triggered"
+                    ))
+                    .into();
+                    tracing::debug!("{error}");
+                    return Err(error);
+                }
             } else {
                 // We are not able to decode the signer, so just fail
                 let error = native_vp::Error::new_const(
@@ -1259,7 +1247,7 @@ where
         &self,
         tx_data: &BatchedTxRef<'_>,
         keys_changed: &BTreeSet<Key>,
-        _verifiers: &BTreeSet<Address>,
+        verifiers: &BTreeSet<Address>,
     ) -> Result<()> {
         let masp_keys_changed: Vec<&Key> =
             keys_changed.iter().filter(|key| is_masp_key(key)).collect();
@@ -1289,7 +1277,7 @@ where
             self.is_valid_parameter_change(tx_data)
         } else if masp_transfer_changes {
             // The MASP transfer keys can only be changed by a valid Transaction
-            self.is_valid_masp_transfer(tx_data, keys_changed)
+            self.is_valid_masp_transfer(tx_data, keys_changed, verifiers)
         } else {
             // Changing no MASP keys at all is also fine
             Ok(())
--- a/crates/namada/src/ledger/native_vp/masp.rs
+++ b/crates/namada/src/ledger/native_vp/masp.rs
@@ -703,7 +703,7 @@ where
             )
             .map_err(native_vp::Error::new)?,
         );
-        // And then record thee final state
+        // And then record the final state
         let post_entry = result.post.get(&addr_hash).cloned().unwrap_or(zero);
         result.post.insert(
             addr_hash,
@@ -826,7 +826,7 @@ where
         )?;
 
         // The set of addresses that are required to authorize this transaction
-        let mut signers = BTreeSet::new();
+        let mut authorizers = BTreeSet::new();
 
         // Checks on the sapling bundle
         // 1. The spend descriptions' anchors are valid
@@ -848,7 +848,7 @@ where
             &mut changed_bals_minus_txn,
             masp_epoch,
             conversion_state,
-            &mut signers,
+            &mut authorizers,
         )?;
 
         // Ensure that every account for which balance has gone down as a result
@@ -871,14 +871,14 @@ where
                 (minus_txn_pre, minus_txn_post) != (pre.clone(), post)
             {
                 // This address will need to provide further authorization
-                signers.insert(addr);
+                authorizers.insert(addr);
             }
         }
 
         // Ensure that this transaction is authorized by all involved parties
-        for signer in signers {
+        for authorizer in authorizers {
             if let Some(TAddrData::Addr(IBC)) =
-                changed_bals_minus_txn.decoder.get(&signer)
+                changed_bals_minus_txn.decoder.get(&authorizer)
             {
                 // If the IBC address is a signatory, then it means that either
                 // Tx - Transaction(s) causes a decrease in the IBC balance or
@@ -908,7 +908,7 @@ where
                     }
                 }
             } else if let Some(TAddrData::Addr(signer)) =
-                changed_bals_minus_txn.decoder.get(&signer)
+                changed_bals_minus_txn.decoder.get(&authorizer)
             {
                 // Otherwise the owner's vp must have been triggered and the
                 // relative action must have been written
@@ -927,22 +927,22 @@ where
                 // because of a masp transaction, which might require a
                 // different validation than a normal balance change
                 if !actions.contains(&namada_tx::action::Action::Masp(
-                    namada_tx::action::MaspAction::MaspSigner(
+                    namada_tx::action::MaspAction::MaspAuthorizer(
                         signer.to_owned(),
                     ),
                 )) {
                     let error = native_vp::Error::new_alloc(format!(
-                        "The required masp signer action for address {signer} \
-                         is missing"
+                        "The required masp authorizer action for address \
+                         {signer} is missing"
                     ))
                     .into();
                     tracing::debug!("{error}");
                     return Err(error);
                 }
             } else {
-                // We are not able to decode the signer, so just fail
+                // We are not able to decode the authorizer, so just fail
                 let error = native_vp::Error::new_const(
-                    "Unable to decode a transaction signer",
+                    "Unable to decode a transaction authorizer",
                 )
                 .into();
                 tracing::debug!("{error}");
@@ -973,18 +973,17 @@ fn unepoched_tokens(
     Ok(())
 }
 
-// Handle transparent input
 fn validate_transparent_input<A: Authorization>(
     vin: &TxIn<A>,
     changed_balances: &mut ChangedBalances,
     transparent_tx_pool: &mut I128Sum,
     epoch: MaspEpoch,
     conversion_state: &ConversionState,
-    signers: &mut BTreeSet<TransparentAddress>,
+    authorizers: &mut BTreeSet<TransparentAddress>,
 ) -> Result<()> {
     // A decrease in the balance of an account needs to be
     // authorized by the account of this transparent input
-    signers.insert(vin.address);
+    authorizers.insert(vin.address);
     // Non-masp sources add to the transparent tx pool
     *transparent_tx_pool = transparent_tx_pool
         .checked_add(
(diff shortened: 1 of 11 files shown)


crates/tx/src/action.rs
```diff
@@ -63,11 +63,13 @@ pub enum PgfAction {
...
-/// MASP tx action.
+/// MASP tx actions.
 #[derive(Clone, Debug, BorshDeserialize, BorshSerialize)]
-pub struct MaspAction {
+pub enum MaspAction {
     /// The hash of the masp [`crate::types::Section`]
-    pub masp_section_ref: Hash,
+    MaspSectionRef(Hash),
+    /// A required signer for the transaction
+    MaspSigner(Address),
```
tzemanovic, line 72: do we need this new case? It looks like it's never constructed

tzemanovic (dismissed):
besides the pending question LGTM

crates/tx/src/action.rs
grarco: Ah yes sorry. This is not used at the moment cause it's only needed when a change in the balance of an address does not match what we see in the `TranpsarentBundle` of the `Masp` section. This in turn can only happen if a transaction is both a masp and a transparent transfer which in up until this PR is something that we don't support in the SDK and client. But it should become relevant after #3459

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -486,32 +486,15 @@ where
...
-                        &[tx_data.tx.raw_header_hash()],
-                        public_keys_index_map,
-                        &Some(signer.clone()),
-                        threshold,
-                        max_signatures_per_transaction,
-                        || gas_meter.consume(crate::gas::VERIFY_TX_SIG_GAS),
-                    )
-                    .map_err(native_vp::Error::new)?;
+                // Otherwise the owner's vp must have been triggered
+                if !verifiers.contains(signer) {
```
murisi, line 490: I would almost certainly agree with this logic if `signer`'s VP was being executed on the state change being checked by this MASP VP minus the state changes induced by the inner `Transaction`. My concern here is as follows: imagine Alice constructs an unshielding `Transaction` of 10 BTC to Bertha. Now imagine that Christel intercepts this `Transaction` and embeds it inside a  `Transfer` `Tx` that increases the balance of Christel by 10 BTC (instead of Bertha who is the rightful recipient). In this case, since `Tx` is effectively two transfers (an unshielding `Transaction` of 10 BTC to Bertha followed by a transparent `Transfer` of 10 BTC from Bertha to Christel), the `signers` map would contain Bertha's address. My specific concern here is that Bertha's VP doesn't see this foul play (of rehousing the `Transaction` in a different `Tx`) and always accepts the state change because the net change to Bertha's balance as a result of the entire state change is zero (after all, Bertha is only mentioned in the transparent output of the `Transaction`). To me this approach would only make sense if the verifiers could be made to specifically validate just the `Transfer` of 10 BTC from Bertha to Christel (which would be the result of doing the above "subtraction"), in which case they would certainly demand an authorization signature from Bertha (as was being done before in an admittedly restrictive way). Am I missing something here?

murisi (changes requested):
I agree that outsourcing verification is the most robust approach. But I'm not sure if these changes are robust against replay attacks...

crates/namada/src/ledger/native_vp/masp.rs
grarco: So this PR only removes the signature verification and requests the corresponding VP to be triggered. As you are saying though, Bertha's VP would not reject the tx in the attack scenario you described because the balance change is 0. To make this work I've introduced a new masp `Action` (i.e. a temporary write to write log that is not committed to storage) with the address of the verifier whose vp we want to check this tx (in this case Bertha). Again, the 0 balance change would lead to the tx being accepted, so on top of that, I've updated the implicit and user VPs to handle this `Action` with a signature check (but custom vps could add more checks if desired): so the signature check has been simply moved from the masp vp to the user and implicit vps.

For this logic to work we need some cooperation from the tx which must write the temporary action key(s) required (the ones carrying the addresses whose extra signatures have been attached to the tx).

I believe this logic is ok (but please double check again) but I think that there are other flaws in this this implementation of mine:
1. Even though the `Action` carries the address of the signer this does not trigger its vp, since it's not the leading part of the key and we don't trigger VPs anymore if their address appear in non-leading locations of a key (@tzemanovic  please correct me if I'm wrong). In this case we need a way for the tx to trigger the required VPs
2. Checking that the VPs have been triggered is not enough. Someone could craft a tx that does not write the required actions but for which the relevant VPs are still triggered. In this case we could end up in the scenario you envisioned. On top of checking that VPs have been triggered we should also check that the corresponding `Action` has been written to storage to ensure that the VP will run the correct validation

Note that, as I've already answered Tomas before, at the moment the transactions do not support this `Action` logic cause with the previous design of vectorized transfers these scenario could not happen. It will become relevant after the new vectorized strategy has been merged

tzemanovic: For a single transfer I think we would need do something like this:

1. If a VP of an address who must authorize the tx is not going to be triggered from the changed keys (yes, it is only the leading part) then the tx has to `insert_verifier`
2. The MASP VP should require an `Action` to be written so a tx is not allowed to skip it (i.e. actions must not be empty)

For these to work together the MASP VP should then ensure that when the action is `MaspSigner` then the address in this action is contained in the `verifiers` set.

*However*, with the scenario with multiple transfers @murisi described I think this may not be sufficient. A tx could only attach a single action despite performing multiple actions. We might then need to add more information to the actions to describe the transfers in more detail and the MASP and user/implicit VPs will have to enforce that the storage changes match these actions. In this example, for the unshielding action from Alice we could have e.g. `Action::Unshield { dest, amount }` where `dest == Bertha`` and if `Christel` was to maliciously embed this transfer in another tx, either:

- Only with the original action, an increase in Christel's balance would not be matching the action (the MASP VP has to check the transparent bundle's `vout.address` against the actions to enforce this)
- With the original action and added `Action::Transparent { src, dest, amount } where `src == Bertha`, `dest == Christel` and `amount == 10`, the user/implicit VP must require signature from Bertha (even if the overall balance change of Bertha is 0)
- Without the original action, but with an `Action::Unshield` where `dest == Christel`, the side-effects of the storage changes would not match the actions (the actual unshielding `dest` is Bertha)

murisi: > So this PR only removes the signature verification and requests the corresponding VP to be triggered.

Ah, I completely misread the approach taken by this PR. My bad.
> but custom vps could add more checks if desired

I see. And at the bare minimum, custom VPs must not forget to handle MASP `Action`.
> In this case we need a way for the tx to trigger the required VPs

Nice, I see you've now covered this case.
> On top of checking that VPs have been triggered we should also check that the corresponding Action has been written to storage to ensure that the VP will run the correct validation

Nice, I see you've also covered this case.

murisi: > For a single transfer I think we would need do something like this:
> 
>     1. If a VP of an address who must authorize the tx is not going to be triggered from the changed keys (yes, it is only the leading part) then the tx has to `insert_verifier`
> 
>     2. The MASP VP should require an `Action` to be written so a tx is not allowed to skip it (i.e. actions must not be empty)
> 
> 
> For these to work together the MASP VP should then ensure that when the action is `MaspSigner` then the address in this action is contained in the `verifiers` set.

At least in the current PR, this is all covered right? The current conditional forces the presence of the signer in the verifiers list. The next conditional forces the presence of the corresponding `MaspSigner` `Action`.

> However, with the scenario with multiple transfers @murisi described I think this may not be sufficient.

The `signer` variable in this loop takes on as its value each of the `Address`es whose balances are debited (even if their net change is zero), so if this verifier check and the following `MaspSigner` check is correct then this VP should prevent Christel from committing thievery (since the `Tx` would be forced to trigger the signature checks elsewhere).

wasm/tx_transfer/src/lib.rs
```diff
@@ -53,7 +53,9 @@ fn apply_tx(ctx: &mut Ctx, tx_data: BatchedTx) -> TxResult {
             .wrap_err("Encountered error while handling MASP transaction")?;
         update_masp_note_commitment_tree(&shielded)
             .wrap_err("Failed to update the MASP commitment tree")?;
-        ctx.push_action(Action::Masp(MaspAction { masp_section_ref }))?;
+        ctx.push_action(Action::Masp(MaspAction::MaspSectionRef(
```
murisi, line 56: The MASP integration tests seem to be failing for me. Probably they are failing on at least the shielding transactions because their signers have not been added as actions. So I think we need to push some `MaspSigner` `Action`s in this vicinity - the same set that will be iterated through in the `for signer in signers {` loop if we desire to support the Christel thieving corner case above. But at the least, all of the net debited accounts should at least be added to the `MaspSigner` `Action`s. No?

murisi (changes requested):
Ah, now I understand what you're doing. I think. However the MASP integration tests are currently failing on my machine. Maybe related: is there any place in the `TX_TRANSFER_WASM` code where we push `MaspSigner` `Action`s?

wasm/tx_transfer/src/lib.rs
grarco: That's correct. I'm fixing this issue rn

grarco:
Ok so I've pushed some more commits, to recap:

- In the masp vp we check that the vp of the signer has been triggered and also check that the corresponding action has been pushed (otherwise the vp might run a different validation logic, we have to inform it that it is required to validate a masp transaction)
- I've talked to @tzemanovic for the solution he proposed (multiple actions to describe what's actually happening). We concluded that for now we can go with the current solution of just pushing a single action (@murisi I think you agreed in your previous comment that this is safe and we don't need more detailed data). His suggestion though could be taken into consideration for future updated in case we want to allow for more complicated validation logic from the implicit/user vps side
- `TX_TRANSFER_WASM` has been updated to push the required `MaspAction::MaspAuthorizer` actions
- The `multi_transfer` function has been updated to return the set of addresses whose balance has gone down
- The `Transfer` type has been extended to carry an additional set of masp verifiers. This is because the solution here above doesn't cover all the cases, for example, it does not cover the tamper attack that Murisi has explained before, therefore we need the transaction to trigger these extra vps too

One small annoyance, the masp vp considers required authorizers all the addresses whose balance has decreased: this includes balances that had nothing to do with the MASP in that transaction. Because of this we are forced to push actions for these addresses too: at the moment both the `implicit` and `user` vps just check for the signature which coincides with the check for a normal balance decrease, but in case the verification changed (for example in some custom vp), then this could run a validation logic which is not correct for the actual changes.

murisi (dismissed):
The latest additions would work. But I wonder if there isn't a way to achieve outsourcing without having to augment `Transfer` with a vector of VP addresses... Might it be possible to collect the VP `Address`es from the `Authorization` sections of the `Tx` and insert those into the verifier list instead of embedding them in the `Transfer` structure? Or would this be too constraining/hacky? After all the user and implicit VPs look for the signatures in these sections...

murisi:
> One small annoyance, the masp vp considers required authorizers all the addresses whose balance has decreased: this includes balances that had nothing to do with the MASP in that transaction. Because of this we are forced to push actions for these addresses too: at the moment both the `implicit` and `user` vps just check for the signature which coincides with the check for a normal balance decrease, but in case the verification changed (for example in some custom vp), then this could run a validation logic which is not correct for the actual changes.

Let me work on a separate PR to fix the MASP VP so that addresses unrelated to the MASP `Transaction` do not require a signature when their balances decrease. I think this should be a very minor non-conflicting change.

tzemanovic (dismissed):

grarco:
> The latest additions would work. But I wonder if there isn't a way to achieve outsourcing without having to augment `Transfer` with a vector of VP addresses... Might it be possible to collect the VP `Address`es from the `Authorization` sections of the `Tx` and insert those into the verifier list instead of embedding them in the `Transfer` structure? Or would this be too constraining/hacky? After all the user and implicit VPs look for the signatures in these sections...

I tend to think it would be slightly hacky and I'm not sure we can derive the address from an authorization section in case of a multisig account, or can we?

murisi:
> > The latest additions would work. But I wonder if there isn't a way to achieve outsourcing without having to augment `Transfer` with a vector of VP addresses... Might it be possible to collect the VP `Address`es from the `Authorization` sections of the `Tx` and insert those into the verifier list instead of embedding them in the `Transfer` structure? Or would this be too constraining/hacky? After all the user and implicit VPs look for the signatures in these sections...
> 
> I tend to think it would be slightly hacky and I'm not sure we can derive the address from an authorization section in case of a multisig account, or can we?

Fair enough, probably attempting this is more trouble than it's worth.

On a separate note, I suspect that 65d07fb317fa9f2f4b871de68f9ab87ecf6712dd (i.e. with all the debited accounts as authorizers but without `opt_authorizers` additional authorizers field) combined #3516 (which reduces the set of authorizations required for a MASP `Tx`) would trigger the necessary VPs for all the transfers generated by `namadac` to pass. Though the difference with that approach is that it would probably be impossible to successfully execute the Alice-Bertha-Christel `Tx` above since Bertha's address would never be inserted by the tx as a verifier. But I'm doubting that transactions like this would actually be needed in practice, so it wouldn't be much of a loss. All this to say that I think this PR would also be perfectly fine without the `Adds additional authorizers for masp transactions` commit. It's up to you!

grarco:
Hey @murisi, I've pushed some last updates:

- This branch is now based on #3516
- The `Transfer` data has been reverted to the previous one (no extra signers)
- The transfer transaction pushes actions for the debited accounts but only if these were involved in the masp part of the transfer
- The transfer transaction fails if the vins addresses of the masp transfer have not been debited
- The masp vp now checks that only the required masp actions have been pushed

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -845,32 +842,56 @@ where
...
+            // Transaction alters its balance
+                (minus_txn_pre, minus_txn_post) != (pre.clone(), post)
             {
-                signers.insert(addr);
+                // This address will need to provide further authorization
+                authorizers.insert(addr);
             }
         }
 
+        let mut actions_authorizers: HashSet<&Address> = actions
```
grarco, line 878: `Actions` are actually a vector of `Action`, so it's possible to push the same action more than once. I believe from the perspective of the masp vp this is of no importance: we just want to check that the action for a specific authorizer has been written, it doesn't matter if it's duplicated

murisi (approved):
This PR looks fine as-is, thanks! I just a few minor notes/queries which are probably unimportant.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -900,39 +921,53 @@ where
...
+                    "Unable to decode a transaction authorizer",
                 )
                 .into();
                 tracing::debug!("{error}");
                 return Err(error);
             }
         }
+        // The transaction shall not push masp authorizer actions that are not
+        // needed cause this might lead vps to run a wrong validation logic
+        if !actions_authorizers.is_empty() {
```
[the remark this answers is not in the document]
grarco: For our VPs, that's correct, we'd just recheck the signature (which actually doesn't even happen because of the `Gadget` that caches the result of the signature check and ensures that we run that verification at most once). I'm not sure this holds for incorrectly VPs though: I believe the worst that could happen was the rejection of a valid transaction but not the acceptance of an invalid one. But some really broken VPs could actually end up accepting invalid transactions because of masp actions. So my reasoning here was to be very conservative, also because there's really no reason why someone should push a masp action which isn't required

wasm/tx_transfer/src/lib.rs
```diff
@@ -53,7 +54,38 @@ fn apply_tx(ctx: &mut Ctx, tx_data: BatchedTx) -> TxResult {
...
+                    .collect::<BTreeSet<_>>()
+            },
+        );
+        let masp_authorizers: Vec<_> = debited_accounts
+            .into_iter()
+            .filter(|account| {
+                vin_addresses.contains(&addr_taddr(account.clone()))
+            })
+            .collect();
+        if masp_authorizers.len() != vin_addresses.len() {
```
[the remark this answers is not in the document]
grarco: Correct, this is a current limitation of the transfer transaction, but as you are saying we don't support these kind of transactions in the client at the moment

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
grarco: That's correct, there's some burden on the transaction side to match the set of verifiers. But eventually the transaction has to either match the set of authorizers required by the vps exactly or match a superset of it (in case we wanted to relax this condition). The only thing I can think of to ease the computation on the transaction side would be to push actions for all the involved parties (which need to be computed anyway since, as you are saying, they might not be limited to the transparent bundle). This would lead to an increase in gas cost given by the extra (unneeded) signatures attached to the transaction and by the cost of their verification: at the end this might outweigh the added gas cost for the logic to be placed in the transfer transaction. Another issue is that collecting some of these signatures could be complicated and make the ux more unpleasant

Issue #3307, opened 2024-05-24
Implement state sync


batconjurer:
In order to quickly sync nodes, we need to enable the state sync protocol supported by Cometbft.

An initial write-up has been created [here](https://hackmd.io/xAeRQtArRom-Ti3DUMkBXg).

Tasks:
- [x] Add the ability to create chunkable snapshots to our rocksdb implementation
- [x] Spawn a background task to create snapshots are certain blockheights
- [x] Implement chunking logic for snapshots
- [x] Implement the relevant abci calls
- [x] Write logic to verify and apply chunks to a syncing node.

brentstone:
Is this considered done or just needs #3394 until it is completed?

batconjurer:
> Is this considered done or just needs #3394 until it is completed?

Yeah, #3394 finishes it off

sug0:
Finished in #3701

PR #3383, opened 2024-06-06
Add ability to create db snapshots

batconjurer:
## Describe your changes
Resolves the first three points of Issue #3307.

### DIFF
--- a/crates/node/src/shell/mod.rs
+++ b/crates/node/src/shell/mod.rs
@@ -659,7 +659,7 @@ where
 
     /// Commit a block. Persist the application state and return the Merkle root
     /// hash.
-    pub fn commit(&mut self) -> response::Commit {
+    pub fn commit(&mut self) -> shim::Response {
         self.bump_last_processed_eth_block();
 
         self.state
@@ -679,13 +679,16 @@ where
 
         self.broadcast_queued_txs();
 
-        response::Commit {
-            // NB: by passing 0, we forbid CometBFT from deleting
-            // data pertaining to past blocks
-            retain_height: tendermint::block::Height::from(0_u32),
-            // NB: current application hash
-            data: merkle_root.0.to_vec().into(),
-        }
+        shim::Response::Commit(
+            response::Commit {
+                // NB: by passing 0, we forbid CometBFT from deleting
+                // data pertaining to past blocks
+                retain_height: tendermint::block::Height::from(0_u32),
+                // NB: current application hash
+                data: merkle_root.0.to_vec().into(),
+            },
+            self.state.db().path().into(),
+        )
     }
 
     /// Updates the Ethereum oracle's last processed block.
--- a/crates/node/src/shims/abcipp_shim.rs
+++ b/crates/node/src/shims/abcipp_shim.rs
@@ -8,6 +8,7 @@ use namada::core::hash::Hash;
 use namada::core::key::tm_raw_hash_to_string;
 use namada::core::storage::BlockHeight;
 use namada::proof_of_stake::storage::find_validator_by_raw_hash;
+use namada::state::DB;
 use namada::time::{DateTimeUtc, Utc};
 use namada::tx::data::hash_tx;
 use namada_sdk::migrations::ScheduledMigration;
@@ -26,7 +27,7 @@ use crate::facade::tendermint::v0_37::abci::{
 };
 use crate::facade::tower_abci::BoxError;
 use crate::shell::{EthereumOracleChannels, Shell};
-use crate::storage::SnapshotCreator;
+use crate::storage::DbSnapshot;
 
 /// The shim wraps the shell, which implements ABCI++.
 /// The shim makes a crude translation between the ABCI interface currently used
@@ -210,23 +211,40 @@ impl AbcippShim {
                 ),
                 _ => {}
             }
-
         }
         let TakeSnapshot::Yes(db_path) = take_snapshot else {
             return;
         };
-        // it's important that this block the thread although it will
-        // be incredibly fast.
-        let Ok(snapshot_creator) = SnapshotCreator::new(db_path) else {
-            tracing::error!("Could not open the DB in order to take a snapshot");
-            return
-        };
+        let base_dir = self.service.base_dir.clone();
+
+        let (snap_send, snap_recv) = tokio::sync::oneshot::channel();
         let snapshot_task = std::thread::spawn(move || {
-            snapshot_creator.write_to_file()
-            // do stuff with snapshot;
+            let db = crate::storage::open(db_path, true, None)
+                .expect("Could not open DB");
+            let snapshot = db.snapshot();
+            // signal to main thread that the snapshot has finished
+            snap_send.send(()).unwrap();
+
+            let last_height = db
+                .read_last_block()
+                .expect("Could not read database")
+                .expect("Last block should exists")
+                .height;
+            let cfs = db.column_families();
+            let path = DbSnapshot::path(last_height, base_dir.clone());
+
+            snapshot.write_to_file(cfs, &path)?;
+            DbSnapshot::cleanup(last_height, &base_dir)
         });
-        // TODO: What to do if an old snapshot task is still running?
-        self.snapshot_task.replace(snapshot_task);
+
+        // it's important that the thread is
+        // blocked until the snapshot is created
+        if snap_recv.blocking_recv().is_err() {
+            tracing::error!("Failed to start snapshot task.")
+        } else {
+            // TODO: What to do if an old snapshot task is still running?
+            self.snapshot_task.replace(snapshot_task);
+        }
     }
 }
 
--- a/crates/node/src/shims/abcipp_shim.rs
+++ b/crates/node/src/shims/abcipp_shim.rs
@@ -16,7 +16,9 @@ use tokio::sync::mpsc::UnboundedSender;
 use tower::Service;
 
 use super::abcipp_shim_types::shim::request::{FinalizeBlock, ProcessedTx};
-use super::abcipp_shim_types::shim::{Error, Request, Response, TxBytes};
+use super::abcipp_shim_types::shim::{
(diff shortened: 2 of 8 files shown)


tzemanovic (dismissed):

crates/node/src/shell/mod.rs
```diff
@@ -679,13 +685,26 @@ where
...
-        }
+        let take_snapshot = match self.blocks_between_snapshots {
+            Some(b) => committed_height.0 % b == 0,
+            _ => false,
+        };
+        let take_snapshot = if take_snapshot {
+            self.state.db().path().into()
+        } else {
+            TakeSnapshot::No
+        };
```
sug0, line 696: nit: move this to a method on the shell, `check_snapshot_required()` or smth

crates/node/src/shims/abcipp_shim.rs
```diff
@@ -177,6 +191,62 @@ impl AbcippShim {
...
         }
     }
+
+    fn update_snapshot_task(&mut self, take_snapshot: TakeSnapshot) {
+        if self
+            .snapshot_task
+            .as_ref()
+            .map(|t| t.is_finished())
+            .unwrap_or_default()
+        {
```
sug0, line 201: nit: I think this code is easier to read if we bind this logic to a variable, and invert the condition to do an early return
```rs
if !snapshot_taken {
    return;
}
```

crates/node/src/shims/abcipp_shim.rs
```diff
@@ -177,6 +191,62 @@ impl AbcippShim {
...
+        };
+        let base_dir = self.service.base_dir.clone();
+
+        let (snap_send, snap_recv) = tokio::sync::oneshot::channel();
+        let snapshot_task = std::thread::spawn(move || {
+            let db = crate::storage::open(db_path, true, None)
+                .expect("Could not open DB");
+            let snapshot = db.snapshot();
+            // signal to main thread that the snapshot has finished
+            snap_send.send(()).unwrap();
```
sug0, line 226: can we comment something here to explain why this is required? I believe it is because we can't take a snapshot of some inconsistent state, only of fully committed state.

ideally we wouldn't need to block reads though. we only need to block db writes while snapshots are taking place. we could modify the rocksdb impl to block writes while a snapshot is being taken, but allow reads to proceed as usual. then we could delete this channel, and instead set some flag e.g. a `block_db_writes: AtomicBool` in rocksdb or smth

crates/node/src/storage/rocksdb.rs
```diff
@@ -111,10 +115,16 @@ const ADDRESS_GEN_KEY_SEGMENT: &str = "address_gen";
...
+const MAX_CHUNK_SIZE: usize = 10_000_000;
 
 /// RocksDB handle
 #[derive(Debug)]
-pub struct RocksDB(rocksdb::DB);
+pub struct RocksDB {
+    /// Handle to the db
+    inner: rocksdb::DB,
+    /// Indicates if read only
+    read_only: bool,
```
sug0, line 126: oh? we already have something like this. this should be atomic, though, no? and we should be making use of it, so we don't block further abci calls from making progress (as long as they don't involve writes)

crates/node/src/storage/rocksdb.rs
```diff
@@ -218,21 +229,34 @@ pub fn open(
         REPLAY_PROTECTION_CF,
         replay_protection_cf_opts,
     ));
-
-    rocksdb::DB::open_cf_descriptors(&db_opts, path, cfs)
-        .map(RocksDB)
-        .map_err(|e| Error::DBError(e.into_string()))
+    Ok(if read_only {
```
sug0, line 232: ah... nevermind. rip. I see how this is being used. we would need another field to block db writes of existing instances opened in read-write mode

crates/node/src/storage/rocksdb.rs
```diff
@@ -688,12 +734,148 @@ impl RocksDB {
...
+            ) {
+                let val = HEXLOWER.encode(&raw_val);
+                let bytes = format!("{cf_name}:{key}={val}\n");
+                chunker.add_line(&bytes);
+                buf.write_all(bytes.as_bytes())?;
+            }
+            buf.flush()?;
+        }
+        let chunks = chunker.finalize();
+        let val = HEXLOWER.encode(&chunks.serialize_to_vec());
```
sug0, line 775: I think base64 might be better for this usecase, as it is slightly more efficient than hex encoding in terms of space usage. ideally, snapshots would be stored in a binary format, anyway, to save even more space

crates/node/src/storage/rocksdb.rs
```diff
@@ -688,12 +734,148 @@ impl RocksDB {
...
+        for entry in std::fs::read_dir(base_dir)? {
+            let entry = entry?;
+            if entry.path().is_file() && Some(toml) == entry.path().extension()
+            {
+                if let Some(name) = entry.path().file_name() {
+                    let Some(height) = name
+                        .to_string_lossy()
+                        .strip_prefix("snapshot_")
+                        .and_then(|n| n.strip_suffix(".toml"))
+                        .and_then(|h| BlockHeight::from_str(h).ok())
```
sug0, line 797: this shit is kinda gnarly 😂

crates/node/src/storage/rocksdb.rs
```diff
@@ -688,12 +734,148 @@ impl RocksDB {
...
+                boundary: self.current_boundary,
+                hash: Hash(hash),
+            });
+            self.current_size = 0;
+        }
+
+        self.current_size =
+            checked!(self.current_size + line.as_bytes().len()).unwrap();
+        self.hasher.update(line.as_bytes());
+        self.current_boundary = checked!(self.current_boundary + 1).unwrap();
```
sug0, line 866: if you're based on `v0.39.0` we already have `checked!(x += y)` btw
```suggestion
        checked!(self.current_boundary += 1).unwrap();
```

crates/node/src/storage/rocksdb.rs
```diff
@@ -688,12 +734,148 @@ impl RocksDB {
...
+            let hash: [u8; 32] = hasher.finalize().into();
+            self.chunks.push(Chunk {
+                boundary: self.current_boundary,
+                hash: Hash(hash),
+            });
+            self.current_size = 0;
+        }
+
+        self.current_size =
+            checked!(self.current_size + line.as_bytes().len()).unwrap();
```
sug0, line 864:
```suggestion
        checked!(self.current_size += line.as_bytes().len()).unwrap();
```

crates/node/src/storage/rocksdb.rs
```diff
@@ -2379,4 +2567,257 @@ mod test {
...
+    /// Test that taking a snapshot actually
+    /// freezes the database in time even if
+    /// it is written to.
+    #[test]
+    fn test_snapshot_creation() {
+        let temp = tempfile::tempdir().expect("Test failed");
+        let mut db = open(&temp, false, None).expect("Test failed");
+        db.write_subspace_val(
+            1.into(),
+            &Key::parse("bing/fucking/bong").expect("Test failed"),
```
sug0, line 2699: drawing

sug0 (commented):
left a couple of comments

crates/node/src/shims/abcipp_shim.rs
batconjurer: I'm confused what you mean. Right now, we don't early return if there is current snapshot task or not.

crates/node/src/shims/abcipp_shim.rs
batconjurer: I agree that ideally we wouldn't block state reads. I'm not sure if it's possible, but rocksdb claims making a snapshot is fast.

crates/node/src/storage/rocksdb.rs
batconjurer: No, we have a read only access type to a read/write rocksdb handle. We need to have two rocksdb handles open concurrently and the rocksdb impl enforces that at most one is read/write by keeping an os lock.

batconjurer: I'm not sure what you mean by atomic. And the reason for having two concurrent handles is precisely so we don't block other abci calls.

crates/node/src/storage/rocksdb.rs
batconjurer: I think the snapshot format can be greatly improved. But I wanted to get a working version of state sync first since we can iterate on the snapshot format easily later.

crates/node/src/storage/rocksdb.rs
batconjurer: I agree 😫 Do you have a suggestion?

crates/node/src/storage/rocksdb.rs
batconjurer: But I can change it to base64 for now

crates/node/src/shims/abcipp_shim.rs
sug0: if I understand correctly, a snapshot is being taken when `self.snapshot_task.is_some()` and `!t.is_finished()`.
```rs
let snapshot_taken = 
        self
            .snapshot_task
            .as_ref()
            .map(|t| t.is_finished())
            .unwrap_or_default();
```
to early return, we could check the inverse condition (i.e. `!snapshot_taken`)

crates/node/src/shims/abcipp_shim.rs
sug0: let's leave this pending for now, then

crates/node/src/storage/rocksdb.rs
sug0: well, not really. when I write this kinda shit I typically use `split_once`, which isn't that much better. I follow it up with a but of `let-else` bindings to extract the tokens I'm after. string parsing is always going to look ugly

sug0 (approved):

PR #3389, opened 2024-06-07
Enforce consistent datetime serialization

sug0:
## Describe your changes

Our `DateTimeUtc` type allowed a relaxed representation of RFC3339 strings. We now enforce a string subset of this format, to guarantee deterministic serialization.

`v0.39.0`


tzemanovic (approved):
looks good, just in `crates/apps_lib/src/cli.rs` pls update help for `NAMADA_START_TIME`

murisi (dismissed):
Looks good to me. Thanks!

crates/core/src/time.rs
```diff
@@ -362,3 +375,34 @@ impl From<DurationNanos> for crate::tendermint::Timeout {
         Self::from(std::time::Duration::from(val))
     }
 }
+
+#[cfg(test)]
+mod core_time_tests {
+    use super::*;
+
```
murisi, line 382: These tests demonstrate that `DateTimeUtc::from_rfc3339` and `DateTimeUtc::to_rfc3339` are inverses. They also demonstrate that valid `DateTime`s are not necessarily valid `DateTimeUtc`s. Could we please also have a test to demonstrate that valid `DateTimeUtc`s are valid `DateTime`s? This would increase confidence that `DateTimeUtc` is a subset of `DateTime`, and that the hardware wallet will successfully handle it.

murisi (changes requested):
Just to be sure that hardware wallet will continue to work, could we please demonstrate in a unit test that `DateTimeUtc` is a subset of `DateTime` somewhere?

crates/core/src/time.rs
```diff
@@ -362,3 +375,34 @@ impl From<DurationNanos> for crate::tendermint::Timeout {
...
+    // TODO: if someone wants to take this on, convert this test
```
murisi, line 383: By the way, proptesting these properties might be straightforward. See the code at https://github.com/anoma/namada/blob/879a3268f2edfbdd0da1c98cf0e4bbdac48f5b0d/crates/sdk/src/lib.rs#L1008 .

sug0:
@murisi could you please test this pr with the ledger hw wallet?

sug0:
hermes might need to be updated for the ibc e2e tests to pass. we checked with the [hw wallet's timestamp parsing code](https://github.com/Zondax/ledger-namada/blob/3052d7c4412d9bf10e327451090fe31f292699cf/app/src/parser_print_common.c#L114-L137) and there seems to be no issue there, but we can only be 100% certain with in-device testing

crates/core/src/time.rs
```diff
@@ -324,30 +338,29 @@ impl TryFrom<Rfc3339String> for DateTimeUtc {
...
 }
 
 impl TryFrom<crate::tendermint::time::Time> for DateTimeUtc {
-    type Error = chrono::ParseError;
+    type Error = ();
 
-    fn try_from(t: crate::tendermint::time::Time) -> Result<Self, Self::Error> {
-        Rfc3339String(t.to_rfc3339()).try_into()
+    fn try_from(t: crate::tendermint::time::Time) -> Result<Self, ()> {
+        DateTimeUtc::from_unix_timestamp(t.unix_timestamp()).ok_or(())
```
yito88, line 363: The nanosecond is dropped here. I tried to convert it via `tendermint_proto` Timestamp and it passed.
```suggestion
        crate::tendermint_proto::google::protobuf::Timestamp::from(t).try_into()
```

yito88:
This branch worked without Hermes updating after fixing Tendermint time conversion and the genesis time format in `crate/tests/src/e2e/setup.rs`.

yito88:
`e2e::ibc_tests::run_ledger_ibc` failed due to the failure of the IBC proof verification for the consensus state in CI though the test didn't use Hermes.
Weirdly, the test passed on my local machine (macOS w/ M1 pro).

murisi:
> @murisi could you please test this pr with the ledger hw wallet?

This PR works with the hardware wallet. We just need to make a (hopefully) minor change on the hardware wallet to increase the precision it displays. See for instance the following where `Actual` is hardware wallet output and `Expected` is what's produced in the test vectors.
```
1: Expected: is equal to "2 | Timestamp : 4900-04-22 06:26:45.041754692 UTC"
1:   Actual: "2 | Timestamp : 4900-04-22 06:26:45.041754 UTC"
```

murisi (approved):
Looks good to me. Thanks!

murisi (changes requested):
Could we please change the line https://github.com/anoma/namada/blob/7f90393d0456a8fa4afac71aeb4caa390b99a229/crates/sdk/src/signing.rs#L1917 to `format!("Timestamp : {}", tx.header.timestamp),` ? This will change the expected output on the Ledger app from `2 | Timestamp : 4900-04-22 06:26:45.041754692 UTC` to something like `3 | Timestamp : 4440-02-19T23:31:34.674481+00:00`. This is required because the Borsh serialization does not contain the last 3 decimal places.

Either we do the above, or we increase the number of decimal places in the Borsh serialization of a `DateTimeUtc` so that the hardware wallet is able to meet the precise expectations.

murisi (approved):
Great! This PR does not break the hardware wallet. Thanks!

PR #3391, opened 2024-06-07
Parameterize gas scale

### LINKED ISSUE #1881: Gas `SCALE` as protocol parameter
Currently, the `SCALE` of the gas is a constant. This value has an impact in the way that the gas (and therefore its price) is computed. Since the minimum gas price is already a protocol parameter, it would be slightly better to make the gas scale one as well to give the community complete control on the two factors that determine the gas fee of a transaction.

brentstone:
## Describe your changes
Closes #1881

v0.39.0

### DIFF
--- a/crates/gas/src/event.rs
+++ b/crates/gas/src/event.rs
@@ -17,3 +17,17 @@ impl EventAttributeEntry<'static> for GasUsed {
         self.0
     }
 }
+
+/// Extend a [`namada_events::Event`] with the gas scale data.
+pub struct GasScale(pub u64);
+
+impl EventAttributeEntry<'static> for GasScale {
+    type Value = u64;
+    type ValueOwned = Self::Value;
+
+    const KEY: &'static str = "gas_scale";
+
+    fn into_value(self) -> Self::Value {
+        self.0
+    }
+}
--- a/crates/gas/src/event.rs
+++ b/crates/gas/src/event.rs
@@ -17,17 +17,3 @@ impl EventAttributeEntry<'static> for GasUsed {
         self.0
     }
 }
-
-/// Extend a [`namada_events::Event`] with the gas scale data.
-pub struct GasScale(pub u64);
-
-impl EventAttributeEntry<'static> for GasScale {
-    type Value = u64;
-    type ValueOwned = Self::Value;
-
-    const KEY: &'static str = "gas_scale";
-
-    fn into_value(self) -> Self::Value {
-        self.0
-    }
-}
--- a/crates/gas/src/lib.rs
+++ b/crates/gas/src/lib.rs
@@ -202,7 +202,6 @@ impl FromStr for Gas {
 
     fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
         let raw: u64 = s.parse().map_err(GasParseError::Parse)?;
-        // let gas = Gas::from_whole_units(raw).ok_or(GasParseError::Overflow)?;
         Ok(Self { sub: raw })
     }
 }
--- a/crates/gas/src/lib.rs
+++ b/crates/gas/src/lib.rs
@@ -113,9 +113,6 @@ pub const MASP_PARALLEL_GAS_DIVIDER: u64 = PARALLEL_GAS_DIVIDER / 2;
 /// Gas module result for functions that may fail
 pub type Result<T> = std::result::Result<T, Error>;
 
-/// Decimal scale of Gas units
-const SCALE: u64 = 100_000_000;
-
 /// Representation of gas in sub-units. This effectively decouples gas metering
 /// from fee payment, allowing higher resolution when accounting for gas while,
 /// at the same time, providing a contained gas value when paying fees.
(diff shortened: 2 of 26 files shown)


crates/node/src/shell/prepare_proposal.rs
```diff
@@ -288,7 +289,10 @@ where
     tx.validate_tx().map_err(|_| ())?;
     if let TxType::Wrapper(wrapper) = tx.header().tx_type {
         // Check tx gas limit for tx size
-        let gas_limit = Gas::try_from(wrapper.gas_limit).map_err(|_| ())?;
+        let gas_scale = get_gas_scale(temp_state)
+            .expect("Failed to get gas scale from parameters");
```
grarco, line 293: I see that we panic here while in `process_proposal` we raise an error. Can we use the same strategy in both cases?

crates/node/src/shell/mod.rs
```diff
@@ -1069,7 +1069,10 @@ where
             TxType::Wrapper(wrapper) => {
                 // Validate wrapper first
                 // Tx gas limit
-                let gas_limit = match Gas::try_from(wrapper.gas_limit) {
+                let gas_scale = get_gas_scale(&self.state)
+                    .expect("Failed to get gas scale from parameters");
```
grarco, line 1073: Same here in mempool validation

crates/node/src/shell/mod.rs
```diff
@@ -1089,8 +1092,11 @@ where
                 }
 
                 // Max block gas
+                let gas_scale = namada::parameters::get_gas_scale(&self.state)
```
grarco, line 1095: I think we can avoid reading the same value once again, let's just copy the value from here above

crates/namada/src/ledger/mod.rs
```diff
@@ -57,8 +56,11 @@ mod dry_run_tx {
         // Wrapper dry run to allow estimating the gas cost of a transaction
         let (mut tx_result, tx_gas_meter) = match tx.header().tx_type {
             TxType::Wrapper(wrapper) => {
-                let gas_limit =
-                    Gas::try_from(wrapper.gas_limit).into_storage_result()?;
+                let gas_scale = namada_parameters::get_gas_scale(ctx.state)?;
```
grarco, line 59: I believe we can read the parameter before the match since we need it in any case

crates/gas/src/event.rs
```diff
@@ -17,3 +17,17 @@ impl EventAttributeEntry<'static> for GasUsed {
         self.0
     }
 }
+
+/// Extend a [`namada_events::Event`] with the gas scale data.
+pub struct GasScale(pub u64);
```
grarco, line 22: This seems to not be used

crates/gas/src/lib.rs
```diff
@@ -188,7 +193,7 @@ impl From<Gas> for u64 {
 impl Display for Gas {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         // Display the gas in whole amounts
-        write!(f, "{}", self.get_whole_gas_units())
+        write!(f, "{}", self.sub)
```
grarco, line 196: This doesn't convert the gas to whole units anymore so the comment above is misleading. Shouldn't we keep the same logic as before?

grarco: Ah right this si because we don't have the scale anymore. Probably we should take into consideration removing this trait implementation at this point and create a method with a more descriptive name. Cause I think when we display the gas we really want to display in in whole units, the sub units are of no interest

crates/gas/src/lib.rs
```diff
@@ -197,8 +202,7 @@ impl FromStr for Gas {
 
     fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
         let raw: u64 = s.parse().map_err(GasParseError::Parse)?;
-        let gas = Gas::from_whole_units(raw).ok_or(GasParseError::Overflow)?;
```
grarco, line 200: Same thing here, we should probably remove this implementation and come up with a custom method

crates/node/src/shell/prepare_proposal.rs
brentstone: done in 037f62f73b6c7fd1e9a73b8d485a67caa8ec0767

crates/node/src/shell/mod.rs
brentstone: done in 037f62f73b6c7fd1e9a73b8d485a67caa8ec0767

crates/node/src/shell/mod.rs
brentstone: done in 037f62f73b6c7fd1e9a73b8d485a67caa8ec0767

crates/namada/src/ledger/mod.rs
brentstone: done in 037f62f73b6c7fd1e9a73b8d485a67caa8ec0767

crates/gas/src/event.rs
brentstone: done in 037f62f73b6c7fd1e9a73b8d485a67caa8ec0767

crates/gas/src/lib.rs
brentstone: comment changed in [112da61](https://github.com/anoma/namada/pull/3391/commits/112da6116adeb7a1dd41c726aab0232a6b6efd3f)

grarco (approved):

grarco (dismissed):

crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -2202,7 +2202,7 @@ fn check_tx_height(test: &Test, client: &mut NamadaCmd) -> Result<u32> {
         .split_once(' ')
         .unwrap()
         .1
-        .split_once('.')
+        .split_once(',')
```
tzemanovic, line 2205: pls also change the full stop in the comment above `// Expecting e.g. "height 1337."`

grarco: Done in 16a2c585a240bde7cb6fb6b9a2cc5e11fee03261

PR #3393, opened 2024-06-07
MASP fee payment

### LINKED ISSUE #2597: Embed masp transactions inside fee payment

grarco:
## Describe your changes

Closes #2597.

Introduces fee payment via the MASP. If a transaction can pay fees via a transparent transfer from the wrapper signer's balance than the protocol proceeds like that. Otherwise, we look to see if the first transaction of the batch is a masp transfer and after its changes the transparent fee payment is possible (i.e. we check that the tx unshields enough funds to the fee payer).

Masp fee payment in the SDK is supported for the transactions for which it makes sense: shielded, unshielding and ibc transactions (when shielded action and only on the source chain, where the initial unshield happens).

To limit the DOS changes of an attacker, if masp fee payment is required than we set a custom gas limit (protocol param) to avoid attackers reducing the tps by requesting large gas limit: if fees are successfully paid we bring back the original gas limit set but the user.

Modification have been done to the way we construct masp transaction to account for multiple sources/targets which is required for fee shielded fee payment. Also, to pay fees, the SDK tries to collect the funds in this order:

1. From changes
2. From unused notes of the transaction's sources
3. From the optional, user provided spend notes

There are a few of things that could be improved but are not part of this PR (since I thought it's already quite large). I'm going to open issues for these:
- the `disposable-gas-payer` is still a `Tx` arg but we should instead pull it out of there and move it to the args of the specific transactions for which the SDK supports masp fee payment
- The new `Changes` type used when constructing masp transactions could probably be a `U128Sum` instead of a `I128Sum`
- Currently we only support a single additional spending key for gas payment in the CLI but we should allow for multiples. To do that we need to add a new `ArgMultiOpt` arg which we are currently lacking

#3356 (diffs for review: https://github.com/anoma/namada/pull/3393/files/1e98403560a297f56f38ff67397a4bb9182068aa..e1b5857845bafa1be9f4ee4500b5544b56bf926f)


crates/core/src/token.rs
```diff
@@ -231,13 +231,14 @@ impl Amount {
...
-    /// Given a u128 and [`MaspDigitPos`], construct the corresponding
+    /// Given a i128 and [`MaspDigitPos`], construct the corresponding
     /// amount.
-    pub fn from_masp_denominated_u128(
-        val: u128,
+    pub fn from_masp_denominated_i128(
+        val: i128,
         denom: MaspDigitPos,
     ) -> Option<Self> {
         let lo = u64::try_from(val).ok()?;
```
tzemanovic, line 240: ah, I think this fn is not quite right - I think I messed it up in #3074 and now it will return `None` for anything greater than `u64::MAX` due to this conversion. Previously, it was doing a truncating cast (`let lo = val as u64;` so I think we need do something like that again

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -432,166 +482,168 @@ where
...
+                    try_masp_fee_payment(shell_params, tx, tx_index)
+                {
+                    // NOTE: Even if the unshielding was succesfull we could
+                    // still fail in the transfer (e.g. cause the unshielded
+                    // amount is not enough to cover the fees). In this case we
+                    // want do drop the changes applied by the masp transaction
+                    // and try to drain the fees from the transparent balance.
+                    // Because of this we must NOT propagate errors from within
+                    // this branch
+                    let balance = crate::token::read_balance(
```
batconjurer, line 550: just out of curiosity, what is the flow where this is expected? My surface level understanding seems to think we should just reject the tx at this point.

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -432,166 +482,168 @@ where
...
-                                block_proposer.clone(),
-                            ),
-                            source_post_balance: post_bal.into(),
-                            target_post_balance,
-                        },
-                    }
-                    .with(HeightAttr(current_block_height))
-                    .with(TxHashAttr(wrapper_tx_hash)),
-                );
+                    // Ok to unwrap_or_default. In the default case, the only
```
batconjurer, line 556: maybe its safer to panic. If there is a bug, this would allow it to be silent.

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -621,66 +785,160 @@ fn token_transfer<WLS>(
...
-                        .write(&dest_key, new_dest_balance.serialize_to_vec())
-                    {
-                        Ok(_) => Ok(()),
-                        Err(e) => Err(Error::FeeError(e.to_string())),
+    // Transfer `token` from `src` to `dest`. Returns an `Err` if `src` has
+    // insufficient balance or if the transfer the `dest` would overflow (This
+    // can only happen if the total supply doesn't fit in `token::Amount`).
+    // Contrary to `crate::token::transfer` this function updates the tx
+    // write log and not the block write log.
+    fn inner_fee_token_transfer<WLS>(
```
batconjurer, line 793: why create the inner function? it's not used more than once

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -621,66 +785,160 @@ fn token_transfer<WLS>(
...
-        state,
-        &wrapper.fee.token,
-        &wrapper.fee_payer(),
-    )
-    .unwrap();
-
-    let fees = wrapper
-        .get_tx_fee()
-        .map_err(|e| Error::FeeError(e.to_string()))?;
+    fn inner_check_fees<S, D, H, CA>(
```
batconjurer, line 867: same question here

crates/sdk/src/lib.rs
```diff
@@ -1130,7 +1135,7 @@ pub mod testing {
             let tx_data = match masp_tx_type {
                 MaspTxType::Shielded => {
                     tx.add_code_from_hash(code_hash, Some(TX_SHIELDED_TRANSFER_WASM.to_owned()));
-                    let data = ShieldedTransfer { section_hash: shielded_section_hash };
+                    let data = ShieldedTransfer { fee_unshield: transfers.0.first().map(|transfer| UnshieldingTransferData { target: transfer.target.to_owned(), token: transfer.token.to_owned(), amount: transfer.amount }), section_hash: shielded_section_hash };
```
batconjurer, line 1138: how the hell did this get past cargo fmt?

crates/sdk/src/tx.rs
```diff
@@ -3026,13 +3077,43 @@ pub async fn build_shielded_transfer<N: Namada>(
...
+    let balance_key = balance_key(&args.fee_token, &fee_payer_address);
+    let balance = rpc::query_storage_value::<_, token::Amount>(
+        context.client(),
+        &balance_key,
+    )
+    .await
+    .unwrap_or_default();
+    let total_fee = checked!(fee_amount.amount() * u64::from(args.gas_limit))?;
+
+    Ok(match total_fee.checked_sub(balance) {
```
batconjurer, line 3106: nit: I think it's clearer to check `total_fee > balance` and then calculate the diff afterwards

crates/tests/src/integration/masp.rs
```diff
@@ -2171,3 +2171,1068 @@ fn dynamic_assets() -> Result<()> {
...
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 10000"));
+
+    Ok(())
+}
+
+// Test that when paying gas via masp we select the gas limit as the minimum
+// between the transaction's gas limit and the protocol parameter.
+#[test]
+fn masp_fee_payment_gas_limit() -> Result<()> {
```
batconjurer, line 2391: this test doesn't seem to check that if gas limit provided by the tx is smaller than the protocol parameter, it uses the user supplied gas limit. Is this tested elsewhere?

crates/tests/src/integration/masp.rs
```diff
@@ -2171,3 +2171,1068 @@ fn dynamic_assets() -> Result<()> {
...
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 0"));
+
+    Ok(())
+}
+
+// Test masp fee payement with a custom provided spending key. Check that fees
+// are splitted between the actual source of the payment and this gas spending
```
batconjurer, line 2717:
```suggestion
// are split between the actual source of the payment and this gas spending
```

crates/sdk/src/masp.rs
```diff
@@ -491,15 +540,62 @@ pub fn find_valid_diversifier<R: RngCore + CryptoRng>(
...
+                0
+            };
+
+            let change_amt = I128Sum::from_nonnegative(
+                asset_type.to_owned(),
+                unsigned_change_amt,
+            )
+            .expect("Change is guaranteed to be non-negative");
+            changes = changes
+                .map(|prev| prev + change_amt.clone())
```
batconjurer, line 572: curious that unchecked arithmetic is ok here

tzemanovic: we don't have the strict clippy checks in the sdk :/ - ideally, we shouldn't be implementing any logic here

tzemanovic: imho we should move the sdk masp mod into the shielded_token crate

grarco: It's actually checked if I recall correctly

crates/namada/src/ledger/protocol/mod.rs
grarco: In theory this should never happen because a tx that cannot pay fees cannot be included in a block (we reject `process_proposal`). Here we have this logic to guarantee liveness and some degree of correcteness in case the check in process

grarco:
> The masp.rs file in the sdk crate is too damn large. Can we split out all tx building functionality (of which there is now a lot) into a separate file?

I agree but can we do that in a separate PR that does only that?

crates/namada/src/ledger/protocol/mod.rs
grarco: It's because I want to drop the write log on every error but I wanted to do that in a single place (and also avoid forgetting to call the drop if we need to propagate another error)

crates/tests/src/integration/masp.rs
grarco: Ah it's actually already tested in `masp_fee_payment`

crates/namada/src/ledger/protocol/mod.rs
grarco: I've added an unwrap to the previous operation (the storage read) in 9698bb7a0da969ea45b68de6fa601f5689cd7899

crates/core/src/token.rs
grarco: Adjusted in 6026c4217ec5d87bdc903b09b5812a36ad58d2ea

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -621,66 +785,160 @@ fn token_transfer<WLS>(
...
+    where
+        WLS: State + StorageRead,
+    {
+        if amount.is_zero() {
+            return Ok(());
+        }
+        let src_key = crate::token::storage_key::balance_key(token, src);
+        let src_balance = crate::token::read_balance(state, token, src)
+            .map_err(Error::StorageError)?;
+        match src_balance.checked_sub(amount) {
```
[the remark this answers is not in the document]
grarco: The other transfer function we makes a protocol write, so it writes directly to the block write log but for fee payment we want to write to the tx write log. Anyway, Tomas came up with a solution with which we can reuse the old function so this code here above is outdated. The update is in 4f893352e2b1c561c3b0cc81afcfdd0f3ac838d3

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -600,18 +652,130 @@ where
...
+                    tracing::error!(
+                        "The fee unshielding tx is invalid, some VPs rejected \
+                         it: {:#?}",
+                        result.vps_result.rejected_vps
+                    );
+                }
+
+                let masp_ref = namada_tx::action::get_masp_section_ref(*state)
+                    .map_err(Error::StateError)?;
+                // Ensure that the transaction is actually a masp one, otherwise
```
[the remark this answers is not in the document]
grarco: Not really, even if saw a `Transaction` section in the tx that does not guarantee that the object belongs to the first transaction in the batch (or to any transaction at all). We could deserialize the `Data` of the first tx in the batch and see if it is an `UnshieldingTransfer` but this introduces a dependency on the data structure used in wasm: this would require a hard fork in case we needed to change the data layout and it would prevent using custom transactions to pay masp fees (of course if they are allowlisted).

So to have the absolute certainty without introducing a dependency on the data structure we need to run the tx first

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -600,18 +652,130 @@ where
...
+    );
+    gas_meter
+        .copy_consumed_gas_from(&tx_gas_meter.borrow())
+        .map_err(|e| Error::GasError(e.to_string()))?;
+    let ref_unshield_gas_meter = RefCell::new(gas_meter);
+
+    let valid_batched_tx_result = {
+        // NOTE: A clean tx write log must be provided to this call
+        // for a correct vp validation. Block and batch write logs, instead,
+        // should contain any prior changes (if any). This is to simulate
```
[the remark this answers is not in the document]
grarco: We had an issue with the old implementation of masp fee payment (fee unshielding) for which some keys (the tx hash if I remember right) were in the write log when we launched the wasm execution leading to a vp rejection. I'll check if this is still the case and if not I'll remove the precommit field of the write log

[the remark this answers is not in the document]
cwgoes: I don't quite follow - that sounds like a potential bug?

crates/namada/src/ledger/protocol/mod.rs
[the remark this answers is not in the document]
cwgoes: I see; makes sense.

crates/namada/src/ledger/protocol/mod.rs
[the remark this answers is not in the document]
grarco: So the tx is indeed the first in the batch but before executing it we might write some keys in the write log (at the moment I believe it's only the hash of the tx for replay protection). This was (in the old implementation) leading to one of the vps (the masp one if I recall correctly) rejecting the transaction. I believe this is no more the case because:

1. We write the tx hash directly in the block write log and so this should not appear in the set of changed keys that we pass to the vp
2. The masp vp (and also the other ones) should now allow for foreign keys by simply ignoring them (it wasn't the case before I think)

So we might be able to remove this logic from here, but I'm still checking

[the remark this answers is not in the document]
grarco: Removed in 9e4d4c9693499bf366841db3efb663f81b247c16

crates/namada/src/ledger/protocol/mod.rs
cwgoes: Not sure that I follow - why would we want to subtract from a transparent balance, if the user tried to pay fees by unshielding? Sounds confusing. I feel like we should just bail if the unshielding doesn't unshield enough.

cwgoes (commented):
Re-reviewed the core changes.

crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -600,18 +652,130 @@ where
...
+        + Sync,
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+    CA: 'static + WasmCacheAccess + Sync,
+{
+    // The fee payment is subject to a gas limit imposed by a protocol
+    // parameter. Here we instantiate a custom gas meter for this step and
+    // initialize it with the already consumed gas. The gas limit should
+    // actually be the lowest between the protocol parameter and the actual gas
+    // limit of the transaction
```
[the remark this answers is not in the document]
grarco: Being the gas limit it's the maximum amount of gas the tx can use but here we make sure that we pick the minimum between the protocol imposed gas limit and the one declared by the transaction, so that if the protocol limit is for example 20k and the transaction sets a limit of 15k we pick the latter (otherwise the transaction would exceed its own limit and won't pay for the whole gas used)

crates/namada/src/ledger/protocol/mod.rs
grarco: It's just a fallback in case the check in `process_proposal` was flawed. We shouldn't really ever reach this branch but, in case, the logic here was to try and get at least a partial fee payment from the transparent balance of the wrapper's signer. We can also decide to drop this logic and just do nothing. Of course, the tx is rejected

crates/namada/src/ledger/protocol/mod.rs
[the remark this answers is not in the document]
cwgoes: Yes, and it is this picking the latter that I would argue makes this a _maximum_ gas limit, since the gas limit can never exceed it

brentstone:
What's the status of this PR? Is it rdy for review and just waiting for some?

crates/namada/src/ledger/protocol/mod.rs
grarco: Actually removed in e1b5857845

crates/namada/src/ledger/protocol/mod.rs
grarco: Removed it in e1b5857845

PR #3401, opened 2024-06-11
Test batch tx events

### LINKED ISSUE #3397: Wrong `Batch` event
In `handle_inner_tx_results` we set the value for the `Batch` event which is supposed to contain the results of all the inner transactions of the batch. At the moment though, we set its value to be the one of the current inner tx we are evaluating (since `handle_inner_tx_result` is called in a loop for very tx of the batch).

https://github.com/anoma/namada/blob/879a3268f2edfbdd0da1c98cf0e4bbdac48f5b0d/crates/node/src/shell/finalize_block.rs#L494

This means that we are overwriting the value of this event with just the result of the last inner transaction. We should instead read the current value of the event (if present) and extend the `batch_results` field with the new entry.

sug0:
## Describe your changes

Closes #3397

Adds additional test coverage to batch tx events emission, to make sure we correctly build a batch of inner tx events from a batched tx.

`v0.39.0`

### DIFF
--- a/crates/node/src/shell/finalize_block.rs
+++ b/crates/node/src/shell/finalize_block.rs
@@ -5796,4 +5796,165 @@ mod test_finalize_block {
             assert!(!shell.state.has_key(&key.parse().unwrap()).unwrap());
         }
     }
+
+    #[test]
+    fn test_multiple_events_from_batch_tx_all_valid() {
+        let (mut shell, _, _, _) = setup();
+
+        let sk = wallet::defaults::bertha_keypair();
+
+        let batch_tx = {
+            let mut batch =
+                Tx::from_type(TxType::Wrapper(Box::new(WrapperTx::new(
+                    Fee {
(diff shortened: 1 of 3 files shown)


crates/node/src/shell/finalize_block.rs
```diff
@@ -5796,4 +5796,165 @@ mod test_finalize_block {
...
+                        ),
+                        token: shell.state.in_mem().native_token.clone(),
+                    },
+                    sk.ref_to(),
+                    WRAPPER_GAS_LIMIT.into(),
+                ))));
+            batch.header.chain_id = shell.chain_id.clone();
+            batch.header.atomic = false;
+
+            // append first inner tx to batch (this one is valid)
```
grarco, line 5896: Nit, we have a `mk_tx_batch` helper for unit tests in `finalize_block` that you could use to build the batch

grarco (approved):
LGTM, thank you!

Issue #3385, opened 2024-06-07
Shielded sync refactor


batconjurer:
In order to improve shielded sync, we need to address several issues across several PRs.

Phase 1.
 - [x] When fetching notes, connections and related failures should not halt shielded sync. Instead, the process should be restarted
 - [x] If fetching is interrupted, the data fetched should be persisted locally so that progress isn't lost
 - [x] A trait for fetching behavior should be added to provide modularity
 
Phase 2.
 - [x] finish developing a masp indexer to support faster fetching of notes and to perform preprocessing
 - [x] Integrate the indexer with shielded sync by implementing the trait developed in phase 1 above.
 - [x] Commitment trees and witnesses can be constructed by anyone out of band. An option to download these should be added instead of always constructing them client side. 

Phase 3.
 - [x] The trial decryptions of MASP notes should be performed in parallel.
 - [x] Scanning does not wait for all notes to be fetched but starts scanning them as they become available.

PR #3405, opened 2024-06-12
Phase 1 of Shielded Sync refactor

batconjurer:
## Describe your changes
Implements phase 1 of Issue #3385 
 - When fetching notes, connections and related failures should not halt shielded sync. Instead, the process should be restarted
 -   If fetching is interrupted, the data fetched should be persisted locally so that progress isn't lost
 -  A trait for fetching behavior should be added to provide modularity 

### DIFF
--- a/crates/sdk/src/masp.rs
+++ b/crates/sdk/src/masp.rs
@@ -621,10 +621,10 @@ impl IntoIterator for Unscanned {
 
     fn into_iter(self) -> Self::IntoIter {
         let txs = {
-            let mut txs: IndexedNoteData = Default::default();
+
             let mut locked = self.txs.lock().unwrap();
-            std::mem::swap(&mut txs, &mut locked);
-            txs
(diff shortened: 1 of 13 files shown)


crates/sdk/src/masp.rs
```diff
@@ -548,10 +620,15 @@ impl IntoIterator for Unscanned {
     type Item = IndexedNoteEntry;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.txs.into_iter()
+        let txs = {
+            let mut txs: IndexedNoteData = Default::default();
+            let mut locked = self.txs.lock().unwrap();
+            std::mem::swap(&mut txs, &mut locked);
```
grarco, line 626: Can't we just use `std::mem::take` here?

batconjurer: yep, can fix.

grarco (approved):

PR #3409, opened 2024-06-12
Murisi/masp ibc replay protection using txdata on 0.39.0

### LINKED ISSUE #3300: Replay Protect Shielded Actions
The `Transaction` object embedded inside MASP-IBC transactions only commits to the IBC internal address. This means that an attacker could potentially replay a `Transaction` object inside a different `Tx` where the specific IBC addresses have been changed. Ideally we should get the `Transaction` object to commit to specific IBC addresses instead of just the general IBC internal address.

murisi:
## Describe your changes
An attempt to address issue #3300. The approach taken is to include the IBC receiver in the pre-image of shielded `Transaction` outputs. More specifically, the following changes have been made:
* The target of outgoing shielded actions is the hash of the IBC receiver (plus a 1 byte tag indicating an IBC recipient)
  * For non-IBC receivers, we just prepend a 1 byte tag to the target `Address` indicating a non-IBC recipient
* Modified IBC transactions to use the raw integer representation of amounts rather than the human readable decimal representation. This involved:
  * Modifying the client to be able to  derive the denominations of IBC tokens. This is done by essentially looking up the base token corresponding to an IBC token and finding the denomination of that
  * Making the protocol return no denomination for IBC token instead of a zero
  * This was necessary because MASP uses raw amounts, so we need to make sure that MASP-IBC computations utilize the same units
* Modifying the MASP validity predicate to check MASP transaction inputs and outputs against IBC packets. This involved:
  * Deriving the token `Address` that corresponds to an transfer `PacketData` and checking that the transparent inputs and outputs use that
  * Checking that the transparent outputs of a MASP `Transaction` go to the `receiver` stated in `PacketData` in the case of an outgoing shielded action
  * Checking that the transparent inputs of a MASP `Transaction` are the IBC internal address in the case of a refund or an incoming shielded action

This PR was derived from merging Namada 0.39.0 into #3406 .

Closes #3300 

Namada v0.39.0
Patched Hermes 1.8.2: https://github.com/heliaxdev/hermes/pull/34


crates/core/src/masp.rs
```diff
@@ -478,23 +488,82 @@ impl Display for TransferSource {
     }
 }
 
+/// Represents the pre-image to a TransparentAddress
+#[derive(Debug, Clone, BorshDeserialize, BorshSerialize, BorshDeserializer)]
+pub enum TAddrData {
+    /// A transparent address
```
cwgoes, line 494:
```suggestion
    /// A transparent address within Namada
```

crates/core/src/masp.rs
```diff
@@ -478,23 +488,82 @@ impl Display for TransferSource {
...
+        match self {
+            Self::Addr(x) => x.clone(),
+            // An IBC signer address effectively means that assets are
+            // associated with the IBC internal address
+            Self::Ibc(_) => IBC,
+        }
+    }
+
+    /// Get the contained IBC receiver, if any
+    pub fn payment_address(&self) -> Option<String> {
```
cwgoes, line 512: Name does not match comment? Should this be called `ibc_receiver_address` ?

crates/core/src/masp.rs
```diff
@@ -513,13 +582,23 @@ impl TransferTarget {
             _ => None,
         }
     }
+
+    /// Get the contained MaybeIbcAddress, if any
```
cwgoes, line 586: What is a `MaybeIbcAddress` ?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
         Ok(())
     }
 
+    /// Look up the IBC denomination from a IbcToken.
```
cwgoes, line 298: Should this function be defined in the `ibc` crate instead, and imported here? Seems like we might want to use it in multiple places (and I wonder if it is in fact already defined somewhere...)

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+    fn apply_send_packet(
+        &self,
+        acc: &mut ChangedBalances,
+        msg: &MsgTransfer,
+        keys_changed: &BTreeSet<Key>,
+    ) -> Result<()> {
+        // Compute the packet commitment for this message
+        let Ok(packet_data_bytes) =
+            serde_json::to_vec(&msg.message.packet_data)
+        else {
```
cwgoes, line 346: When would this case happen? Seems concerning, no?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+            }
+        }
+
+        Ok(token.as_ref().to_string())
+    }
+
+    // Apply the given send packet to the changed balances structure
+    fn apply_send_packet(
+        &self,
+        acc: &mut ChangedBalances,
```
cwgoes, line 339: Is there a reason we take `acc` as a `mut` argument instead of simply returning a `ChangedBalances` which can be added or something? In general, I dislike the pattern of taking arguments as `mut` which are effectively return types.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+            *post_entry =
+                checked!(post_entry - &delta).map_err(native_vp::Error::new)?;
+        }
+        Ok(())
+    }
+
+    // Apply the given write acknowledge to the changed balances structure
+    fn apply_write_ack(
+        &self,
+        acc: &mut ChangedBalances,
```
cwgoes, line 438: Same comment as above (prefer returning a `ChangedBalances` to taking it as `mut`)

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+                )?;
+            }
+            // Ignore all other IBC events
+            _ => {}
+        }
+        Result::<_>::Ok(acc)
+    }
+
+    // Apply the balance change to the changed balances structure
+    fn apply_balance_change(
```
cwgoes, line 542: Could all these functions simply be exported by the `ibc` crate? Also, could they be combined into one which takes all the state changes and returns the `ChangedBalances` for this transaction which the MASP VP needs?

cwgoes (changes requested):
The concept makes sense; I'm wondering if we can clean up the `ChangedBalances` calculation somewhat - it's very hard to follow.

crates/core/src/masp.rs
murisi: Ah, I renamed `MaybeIbcAddress` to `TAddrData` and forgot to update this comment. Will fix, thanks.

crates/core/src/masp.rs
murisi: Yes, you're correct. Will fix this.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+                self.apply_write_ack(
+                    &mut acc,
+                    msg,
+                    &packet_data,
+                    keys_changed,
+                    ibc_acks,
+                )?;
+            }
+            // Ignore all other IBC events
+            _ => {}
```
yito88, line 536: IBC messages `AckPacket` and `Timeout` could bring MASP transfer due to the refund when IBC transfer failure too.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,343 @@ where
...
+    ) -> Result<ChangedBalances> {
+        match ibc_msg {
+            // This event is emitted on the sender
+            IbcMessage::Transfer(msg) => {
+                // Get the packet commitment from post-storage that corresponds
+                // to this event
+                let receiver = msg.message.packet_data.receiver.to_string();
+                let addr = TAddrData::Ibc(receiver.clone());
+                acc.decoder.insert(ibc_taddr(receiver), addr);
+                self.apply_send_packet(&mut acc, msg, keys_changed)?;
```
yito88, line 514: We could generalize `apply_send_packet` for `Transfer` and `NftTransfer`

crates/namada/src/ledger/native_vp/masp.rs
murisi: Assuming the correct implementation of `Serialize` on `PacketData`, I can't think of anything that would cause this case to happen. The point of this line had merely been that in this case `Tx` validation could theoretically correctly continue without adding this Packet to `ChangedBalances`. That being said, propagating an error is also a possibility. Would you prefer this or something else?

crates/namada/src/ledger/native_vp/masp.rs
murisi: There's no reason. Returning `ChangedBalances` also works. Will apply this change.

crates/namada/src/ledger/native_vp/masp.rs
murisi: You're right, this function is already implemented in the client (where the queries are sent to an RPC node). I'll investigate whether the protocol version here (which reads storage directly) can somehow reuse this logic.

crates/namada/src/ledger/native_vp/masp.rs
cwgoes: I think it's safer to reject the transaction in these sorts of cases - just defense-in-depth.

crates/namada/src/ledger/native_vp/masp.rs
murisi: I believe that the changes caused by an `AckPacket` message are already captured by `apply_balance_change` . For example see the following excerpt generated by running `run_ledger_ibc_with_hermes`:
```
IBC Message: AckPacket(MsgAcknowledgement { message: MsgAcknowledgement { packet: Packet { sequence: Sequence(3), source_port: PortId(\"transfer\"), source_channel: ChannelId(\"channel-0\"), destination_port: PortId(\"transfer\"), destination_channel: ChannelId(\"channel-0\"), data: [123, 34, 100, 101, 110, 111, 109, 34, 58, 34, 116, 110, 97, 109, 49, 113, 121, 56, 113, 103, 120, 108, 99, 116, 101, 101, 104, 108, 107, 55, 48, 115, 110, 56, 119, 120, 50, 112, 100, 108, 97, 118, 116, 97, 121, 112, 51, 56, 118, 118, 114, 110, 107, 104, 113, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 34, 49, 48, 48, 48, 48, 48, 48, 48, 48, 48, 34, 44, 34, 115, 101, 110, 100, 101, 114, 34, 58, 34, 122, 110, 97, 109, 49, 99, 104, 117, 99, 109, 110, 115, 118, 107, 117, 109, 121, 52, 52, 110, 48, 120, 113, 54, 112, 116, 102, 110, 108, 48, 112, 54, 117, 51, 52, 56, 102, 55, 109, 117, 120, 97, 50, 113, 55, 116, 108, 103, 114, 53, 118, 56, 97, 116, 54, 113, 118, 108, 57, 112, 121, 52, 97, 112, 101, 117, 99, 121, 118, 112, 55, 97, 120, 55, 122, 122, 53, 122, 99, 114, 34, 44, 34, 114, 101, 99, 101, 105, 118, 101, 114, 34, 58, 34, 105, 110, 118, 97, 108, 105, 100, 95, 114, 101, 99, 101, 105, 118, 101, 114, 34, 44, 34, 109, 101, 109, 111, 34, 58, 34, 34, 125], timeout_height: Never, timeout_timestamp: Timestamp { time: Some(Time(2024-06-14 13:52:21.516949312)) } }, acknowledgement: Acknowledgement([123, 34, 101, 114, 114, 111, 114, 34, 58, 34, 102, 97, 105, 108, 101, 100, 32, 116, 111, 32, 112, 97, 114, 115, 101, 32, 97, 115, 32, 65, 99, 99, 111, 117, 110, 116, 73, 100, 34, 125]), proof_acked_on_b: CommitmentProof(\"0A97020A94020A3261636B732F706F7274732F7472616E736665722F6368616E6E656C732F6368616E6E656C2D302F73657175656E6365732F3312200402EE754C5E43142130AC8C370D449337CDF35722F4C1A36F9AF37B10A687B31A2408012A200000000000000000000000000000000000000000000000000000000000000000222408011220E062C9E219D7FBD9D42B8871745637580097DF41B9BBE2C2629884CD7CF5C428222408011A20007159D6694F92C048952FABE29A83CC39BC9526AAD27F0FBB3E142E9C54C854222408011A201D27263E79026BA20EC88F32874CD5589123B3E2B0468657E94B9B07C96C772D222408011A20F3017AA4EF352D4585DD8E16C56F868C71269B3DBAD67873BF4BD38AC8592C250A9E010A9B010A036962631220394803D0AF9DBCD091308DC5BE9DE0BBA22A068EB0CEC05A4134A8ADFD9EA30A1A26080110012A200000000000000000000000000000000000000000000000000000000000000000222408011A203D02BB9C9E71792AEFC7924B284E76883A7043CB5768CE7E33717CCBFA5A7BD8222408011A201E79C6A260250336321A4347CAA7F354672D7FA9839A02CA58762EFE0A9DBB7D\"), proof_height_on_b: Height { revision: 0, height: 124 }, signer: Signer(\"tnam1qz0nvec686e9pks8ynhm5ddq8ke7j2eey50uagtr\") }, transfer: Some(ShieldingTransfer { source: Internal IBC: tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4, token: Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, amount: DenominatedAmount { amount: Amount { raw: 1000000000 }, denom: Denomination(0) }, shielded_section_hash: Hash([65, 170, 255, 182, 66, 245, 59, 220, 144, 51, 109, 85, 74, 210, 181, 200, 182, 203, 57, 224, 222, 167, 45, 26, 94, 17, 60, 80, 101, 34, 103, 187]) }) })
Changed Balances: ChangedBalances { tokens: {AssetType { identifier: [10, 170, 166, 12, 65, 99, 62, 13, 116, 182, 9, 215, 18, 196, 130, 27, 41, 214, 181, 104, 168, 46, 202, 240, 12, 92, 76, 92, 181, 215, 209, 118], nonce: Some(0) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Three), AssetType { identifier: [53, 212, 33, 11, 138, 34, 147, 17, 161, 183, 27, 205, 164, 179, 27, 28, 156, 104, 183, 65, 47, 108, 70, 245, 193, 214, 195, 181, 180, 141, 165, 106], nonce: Some(2) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), One), AssetType { identifier: [96, 72, 232, 131, 106, 52, 7, 194, 177, 23, 31, 184, 184, 50, 168, 110, 65, 206, 104, 151, 235, 80, 90, 101, 169, 126, 226, 45, 185, 89, 59, 212], nonce: Some(1) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Two), AssetType { identifier: [161, 97, 138, 24, 204, 180, 226, 116, 36, 16, 242, 35, 132, 48, 211, 54, 131, 69, 170, 62, 15, 49, 26, 131, 92, 95, 187, 63, 135, 128, 197, 147], nonce: Some(1) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Zero)}, decoder: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): Addr(Internal IBC: tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): Addr(Internal MASP: tnam1pcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzmefah)}, ibc_denoms: {\"tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq\": Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq}, pre: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 2000000000 }}), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 8000000000 }})}, post: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 1000000000 }}), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 9000000000 }})} }
```
At a high level this packet is refunding 10 `tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq` tokens. Note how the IBC internal address balance goes down by 10 tokens and the MASP internal address balance goes up by 10 tokens. This suggests to me that it is not necessary to utilize the `AckPacket` messages here, and that monitoring the balance changes of IBC internal address will suffice.

murisi: I also believe that an argument analogous to the above can be made for the IBC `Timeout` message. For example, consider the following excerpt from the same test:
```
IBC Message: Timeout(MsgTimeout { message: MsgTimeout { packet: Packet { sequence: Sequence(4), source_port: PortId(\"transfer\"), source_channel: ChannelId(\"channel-0\"), destination_port: PortId(\"transfer\"), destination_channel: ChannelId(\"channel-0\"), data: [123, 34, 100, 101, 110, 111, 109, 34, 58, 34, 116, 110, 97, 109, 49, 113, 121, 56, 113, 103, 120, 108, 99, 116, 101, 101, 104, 108, 107, 55, 48, 115, 110, 56, 119, 120, 50, 112, 100, 108, 97, 118, 116, 97, 121, 112, 51, 56, 118, 118, 114, 110, 107, 104, 113, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 34, 49, 48, 48, 48, 48, 48, 48, 48, 48, 48, 34, 44, 34, 115, 101, 110, 100, 101, 114, 34, 58, 34, 122, 110, 97, 109, 49, 118, 112, 50, 102, 104, 121, 57, 101, 56, 100, 119, 56, 53, 51, 122, 99, 97, 50, 116, 55, 112, 120, 120, 53, 120, 52, 113, 115, 108, 50, 121, 118, 103, 114, 109, 97, 107, 55, 101, 114, 48, 56, 57, 53, 118, 115, 57, 51, 114, 115, 55, 101, 118, 102, 114, 112, 122, 116, 97, 109, 108, 122, 51, 115, 109, 121, 53, 57, 54, 104, 103, 115, 100, 97, 112, 34, 44, 34, 114, 101, 99, 101, 105, 118, 101, 114, 34, 58, 34, 122, 110, 97, 109, 49, 122, 120, 116, 56, 101, 50, 50, 117, 122, 54, 54, 54, 99, 101, 55, 104, 120, 113, 112, 99, 54, 57, 121, 102, 106, 51, 116, 112, 100, 57, 118, 50, 54, 101, 112, 50, 101, 112, 119, 110, 51, 52, 107, 118, 121, 117, 119, 106, 104, 57, 56, 104, 104, 114, 101, 57, 56, 57, 55, 115, 104, 99, 106, 106, 52, 99, 110, 113, 117, 103, 119, 108, 118, 52, 113, 34, 44, 34, 109, 101, 109, 111, 34, 58, 34, 34, 125], timeout_height: Never, timeout_timestamp: Timestamp { time: Some(Time(2024-06-14 12:52:47.695933275)) } }, next_seq_recv_on_b: Sequence(4), proof_unreceived_on_b: CommitmentProof(\"0A86041283040A3672656365697074732F706F7274732F7472616E736665722F6368616E6E656C732F6368616E6E656C2D302F73657175656E6365732F3412F9010A3672656365697074732F706F7274732F7472616E736665722F6368616E6E656C732F6368616E6E656C2D302F73657175656E6365732F331201011A2408012A200000000000000000000000000000000000000000000000000000000000000000222408011220FE65AAC7FA573BCAE82431882B3FB689926FC708AA848F0C15852A2A17152625222408011220FB867EFBA26D5059137DFA54C2A8DC00673F55A015E41D9078A8875B571B6BA0222408011A2071376AE7874E57AFDBC1088887A3DB66277BEFF0C3170A79E4C655F63602EFA4222408011220C9A504EFDEDC6E7BA6C7534AE6B493732C04C356E8167A12EFE2BE43EC51F0ED1ACC010A3677697468647261772F746E616D31703575387A79766D667633726C32376D6333677367617870613233396875656839736774637A3874122000743BA40B0000000000000000000000000000000000000000000000000000001A2408012A20000000000000000000000000000000000000000000000000000000000000000022240801122076D3570FAFF45DCDE578C121C9EE89FCBBA38E469D6F2D1F904E4198FF263767222408011220C9A504EFDEDC6E7BA6C7534AE6B493732C04C356E8167A12EFE2BE43EC51F0ED0A9E010A9B010A036962631220394803D0AF9DBCD091308DC5BE9DE0BBA22A068EB0CEC05A4134A8ADFD9EA30A1A26080110012A200000000000000000000000000000000000000000000000000000000000000000222408011A2014AD5A9623535DFFB0A8F1691AB21B9D133D6AF9E56A66AD6C0F93CBE4132190222408011A201E79C6A260250336321A4347CAA7F354672D7FA9839A02CA58762EFE0A9DBB7D\"), proof_height_on_b: Height { revision: 0, height: 150 }, signer: Signer(\"tnam1qz0nvec686e9pks8ynhm5ddq8ke7j2eey50uagtr\") }, transfer: Some(ShieldingTransfer { source: Internal IBC: tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4, token: Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, amount: DenominatedAmount { amount: Amount { raw: 1000000000 }, denom: Denomination(0) }, shielded_section_hash: Hash([214, 51, 205, 51, 14, 5, 69, 21, 36, 81, 90, 101, 230, 220, 29, 243, 81, 31, 203, 134, 134, 120, 15, 247, 71, 210, 170, 44, 7, 40, 14, 60]) }) })
Changed Balances: ChangedBalances { tokens: {AssetType { identifier: [10, 170, 166, 12, 65, 99, 62, 13, 116, 182, 9, 215, 18, 196, 130, 27, 41, 214, 181, 104, 168, 46, 202, 240, 12, 92, 76, 92, 181, 215, 209, 118], nonce: Some(0) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Three), AssetType { identifier: [53, 212, 33, 11, 138, 34, 147, 17, 161, 183, 27, 205, 164, 179, 27, 28, 156, 104, 183, 65, 47, 108, 70, 245, 193, 214, 195, 181, 180, 141, 165, 106], nonce: Some(2) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), One), AssetType { identifier: [96, 72, 232, 131, 106, 52, 7, 194, 177, 23, 31, 184, 184, 50, 168, 110, 65, 206, 104, 151, 235, 80, 90, 101, 169, 126, 226, 45, 185, 89, 59, 212], nonce: Some(1) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Two), AssetType { identifier: [161, 97, 138, 24, 204, 180, 226, 116, 36, 16, 242, 35, 132, 48, 211, 54, 131, 69, 170, 62, 15, 49, 26, 131, 92, 95, 187, 63, 135, 128, 197, 147], nonce: Some(1) }: (Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq, Denomination(8), Zero)}, decoder: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): Addr(Internal IBC: tnam1qcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvtr7x4), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): Addr(Internal MASP: tnam1pcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzmefah)}, ibc_denoms: {\"tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq\": Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq}, pre: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 2000000000 }}), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 8000000000 }})}, post: {TransparentAddress([155, 209, 98, 252, 116, 194, 185, 61, 115, 5, 201, 229, 4, 170, 177, 214, 152, 246, 217, 52]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 1000000000 }}), TransparentAddress([196, 128, 240, 153, 79, 139, 252, 48, 1, 179, 248, 81, 97, 118, 59, 124, 233, 55, 127, 247]): ValueSum({Established: tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq: Amount { raw: 9000000000 }})} }
```
This packet is also refunding 10 `tnam1qy8qgxlcteehlk70sn8wx2pdlavtayp38vvrnkhq` tokens. And note how the `Timeout` message was not utilized.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Agreed. I was just a bit reluctant to do this now because the NFT E2E tests are not a part of the tests crate in the Namada repo and require some additional setup which I've not done yet. Do you think you could get this done more quickly/efficiently than me and open up a separate PR to generalize `Transfer` to `NftTransfer`? If not, then I can start working on this...

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+    pub fn query_ibc_denom(
+        &self,
+        token: impl AsRef<str>,
+        owner: Option<&Address>,
+    ) -> Result<String> {
+        let hash = match Address::decode(token.as_ref()) {
+            Ok(Address::Internal(InternalAddress::IbcToken(hash))) => {
+                hash.to_string()
+            }
+            _ => return Ok(token.as_ref().to_string()),
```
cwgoes, line 312: Are we sure that the `token` in this case is really the `denom` that we're looking for?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+        };
+
+        if let Some(owner) = owner {
```
cwgoes, line 315: Why do we need to know the owner to query the IBC denomination at all?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+        // Compute the packet commitment for this message
+        let packet_data_bytes = serde_json::to_vec(&message.packet_data)
+            .map_err(native_vp::Error::new)?;
+        let packet_commitment =
+            VpValidationContext::<'a, 'a, S, CA>::compute_packet_commitment(
+                &packet_data_bytes,
+                &message.timeout_height_on_b,
+                &message.timeout_timestamp_on_b,
+            );
+        // Try to find a key change with the same port, channel, and commitment
```
cwgoes, line 356: Is there a reason we can't just construct the path directly and read from it (instead of iterating)? What data are we missing?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+                }
+            }
+        }
+
+        Ok(token.as_ref().to_string())
+    }
+
+    // Find the given IBC message in the changed keys and return the associated
+    // sequence number
+    fn search_ibc_transfer(
```
cwgoes, line 342: We should name this something that indicates that it will return a sequence number, e.g. `find_ibc_transfer_sequence` or such

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+    // Try to determine which address would cause query_ibc_denom to yield the
+    // supplied denom
+    fn reverse_query_ibc_denom(
+        denom: &PrefixedDenom,
+        ibc_denoms: &BTreeMap<String, Address>,
+    ) -> Option<Address> {
+        ibc_denoms
+            .get(&denom.to_string())
+            .cloned()
+            // If the reverse lookup failed, then guess the Address
```
cwgoes, line 388: Should we check that `query_ibc_denom` does in fact return the supplied denom? Or is that guaranteed?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+        ) else {
+            return Ok(acc);
+        };
+        let delta = ValueSum::from_pair(
+            token.clone(),
+            Amount::from_uint(Uint(*msg.message.packet_data.token.amount), 0)
+                .unwrap(),
+        );
+        // If there is a transfer to the IBC account, then deduplicate the
+        // balance increase since we already accounted for it above
```
cwgoes, line 433: What do you mean "above" ?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+        if is_sender_chain_source(
+            msg.message.port_id_on_a.clone(),
+            msg.message.chan_id_on_a.clone(),
+            &msg.message.packet_data.token.denom,
+        ) {
+            let post_entry =
+                acc.post.entry(addr_taddr(IBC)).or_insert(ValueSum::zero());
+            *post_entry =
```
cwgoes, line 441: Would prefer that we write this in a less mutating style (e.g. first construct `post_entry`, then add it)

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+    ) -> Result<Option<AcknowledgementStatus>> {
+        // Ensure that the event corresponds to the current changes
+        // to storage
+        let ack_key = storage::ack_key(
+            &message.packet.port_id_on_a,
+            &message.packet.chan_id_on_a,
+            message.packet.seq_on_a,
+        );
+        if !keys_changed.contains(&ack_key) {
+            // Otherwise ignore this event
```
cwgoes, line 472: What do you mean by "otherwise"? In this case there is no acknowledgement right?

crates/sdk/src/masp.rs
```diff
@@ -1777,21 +1751,20 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
                     builder
                         .add_transparent_output(
-                            &TransparentAddress(hash),
+                            &t_addr_data.taddress(),
                             *asset_type,
                             contr,
                         )
                         .map_err(builder::Error::TransparentBuild)?;
+                } else {
+                    unreachable!(
```
grarco, line 1764: Better to return an error from here instead of panicking?

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+            let ibc_trace_key = ibc_trace_key(owner.to_string(), &hash);
+            if let Some(ibc_denom) = self.ctx.read_pre(&ibc_trace_key)? {
+                return Ok(ibc_denom);
+            }
+        }
+
+        // No owner is specified or the owner doesn't have the token
+        let ibc_denom_prefix = ibc_trace_key_prefix(None);
+        let ibc_denoms = self.ctx.iter_prefix(&ibc_denom_prefix)?;
+        for (key, ibc_denom, _gas) in ibc_denoms {
```
grarco, line 325: Would be slightly better to charge the gas we receive back here

yito88 (changes requested):
I think we could make it a bit simpler because we can assume that the validated transaction has only one IBC transfer.

crates/namada/src/ledger/native_vp/masp.rs
murisi: I need to investigate this. For the purposes of validation, this function almost exactly mirrors the computation done during the construction of IBC transactions in the client. Compare this to https://github.com/anoma/namada/blob/08e56643fd65d42de3cf854122ab1fe74698f7ac/crates/sdk/src/rpc.rs#L1406 , which is invoked at https://github.com/anoma/namada/blob/08e56643fd65d42de3cf854122ab1fe74698f7ac/crates/sdk/src/tx.rs#L2583 . cc @yito88

crates/namada/src/ledger/native_vp/masp.rs
murisi: See https://github.com/anoma/namada/pull/3409#discussion_r1642904303 .

crates/namada/src/ledger/native_vp/masp.rs
murisi: We're missing the sequence number. Also, Yuji suggested a different approach:
> packet sequence can be obtained from the state with read_pre and the key next_sequence_*_key because the transaction has only one IBC message.

crates/namada/src/ledger/native_vp/masp.rs
murisi: My mistake, the computation is now below. See here: https://github.com/anoma/namada/blob/08e56643fd65d42de3cf854122ab1fe74698f7ac/crates/namada/src/ledger/native_vp/masp.rs#L451 . Will fix this.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Yes, you're correct. Will rephrase this comment.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
     // Check that transfer is pinned correctly and record the balance changes
     fn validate_state_and_get_transfer_data(
         &self,
         keys_changed: &BTreeSet<Key>,
+        ibc_msgs: &[IbcMessage],
     ) -> Result<ChangedBalances> {
         // Get the changed balance keys
-        let mut counterparts_balances = keys_changed
-            .iter()
-            .filter_map(is_any_shielded_action_balance_key);
```
grarco, line 290: Can we remove the `is_any_shielded_action_balancekey` if we don't need it anymore?

crates/namada/src/ledger/native_vp/masp.rs
murisi: I think this should be guaranteed given that we only ever insert the results of `query_ibc_denom` into the `ibc_denoms` map. My intent had been that nothing other than the outputs of `query_ibc_denom` is ever inserted into the `ibc_denoms` map.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+                checked!(post_entry - &delta).map_err(native_vp::Error::new)?;
+        }
+        // Required for the packet's receiver to get funds
+        let post_entry = acc
+            .post
+            .entry(ibc_taddr(msg.message.packet_data.receiver.to_string()))
+            .or_insert(ValueSum::zero());
+        // Enable funds to be received by the receiver in the
+        // IBC packet
+        *post_entry =
```
grarco, line 451: I guess the idea here is to mock a balance change to an address that we are not aware of on this chain (based on the ibc msg) so that then we can verify the matching between this ibc msg packet and masp transparent output with `validate_transparent_output`, correct?

grarco: Also, are we validating the optional `ShieldingTransfer` that the receiving chain gets? I see that this message is built in `build_ibc_transfer` and we also link the hash of the unshielding masp section of the source chain. The method `execute` called by the ibc native vp returns this data but it seems like we never use it. This is probably a question for @yito88

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+        *post_entry = checked!(
+            post_entry + &ValueSum::from_pair((*token).clone(), post_balance)
+        )
+        .map_err(native_vp::Error::new)?;
+        Result::<_>::Ok(result)
+    }
+
+    // Collect the various packet acknowledgements and store them by their
+    // commitments
+    fn extract_ibc_acknowledgements(
```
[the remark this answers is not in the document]
murisi: Thanks. Given that these acknowledgements are used in the context of an IBC receiving packet message, how would one extract an `AcknowledgementStatus` from a `IbcMsgRecvPacket`?

crates/namada/src/ledger/native_vp/masp.rs
murisi: > I guess the idea here is to mock a balance change to an address that we are not aware of on this chain (based on the ibc msg) so that then we can verify the matching between this ibc msg packet and masp transparent output with `validate_transparent_output`, correct?

Yes, exactly.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
yito88: We can check if receiving is successful like
https://github.com/anoma/namada/blob/08e56643fd65d42de3cf854122ab1fe74698f7ac/crates/ibc/src/lib.rs#L228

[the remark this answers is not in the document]
murisi: Thanks, I see. We can simplify the code by assuming that the acknowledgement status for a success is always `AcknowledgementStatus::success(ack_success_b64())`.

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +299,385 @@ where
...
+            .ctx
+            .read_pre(&counterpart_balance_key)?
+            .unwrap_or_default();
+        let post_balance: Amount = self
+            .ctx
+            .read_post(&counterpart_balance_key)?
+            .unwrap_or_default();
+        // Make it possible to decode IBC tokens
+        result.ibc_denoms.insert(
+            self.query_ibc_denom(token.to_string(), Some(counterpart))
```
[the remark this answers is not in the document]
murisi: Looking at this line: https://github.com/anoma/namada/blob/5cc23b4f43b12ecfc70eeca8d04cbe15f4b11a3a/crates/sdk/src/tx.rs#L2588 . I agree that `Transfer::token` is identical to `PacketData::token::denom` in the case where it's not a `Address::Internal(InternalAddress::IbcToken(_))`. But in the case where a `Address::Internal(InternalAddress::IbcToken(_))` is being sent, I'm not sure how else to determine what this token is given a `PacketData::token::denom`. Is there another way to do this?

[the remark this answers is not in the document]
yito88: We can get the `IbcToken` address from the denom in the packet with `namada_ibc::storage::ibc_token`.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
yito88: Yes, it should be always the same ack when the receiving is successful.

crates/namada/src/ledger/native_vp/masp.rs
yito88: > The method execute called by the ibc native vp returns this data but it seems like we never use it.
Right, IBC actions never use `ShieldingTransfer`.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
murisi: Ah, makes sense. Will fix this, thanks.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
murisi: Agreed that checking for the presence of a packet in storage is already done in the IBC VP. Given that we only want to apply a packet that will be accepted by the IBC VP, this code currently checks that the IBC packet in the `Tx` data section is placed into IBC storage and that the IBC VP has been triggered. Would it be sufficient to only check that the IBC packet is in the `Tx` data and that the IBC VP has been triggered (without checking for changes to storage)? cc @cwgoes . It seems that doing this makes more assumptions (in the MASP VP) about how the IBC VP interprets a given `Tx` and what its acceptance means in relation to the packet in the `Tx`.

crates/namada/src/ledger/native_vp/masp.rs
yito88: I hope this could be helpful: https://github.com/anoma/namada/pull/3422

murisi: Brilliant, thanks! Did you manage to run the NFT E2E tests on this? Just want to clarify https://github.com/anoma/namada/pull/3409#discussion_r1644380375 , then I'll merge this in.

yito88: To be honest, I didn't try the NFT E2E test. I will test it after testing fungible token transfers in this PR.
Also, I'll release the updated Hermes for this PR to test in CI.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
cwgoes: I think we should err on the side of duplication for safety here for now, but this is a good (and general) question as to whether VPs can rely on transaction data semantics only. We should add a comment about this redundancy to clarify for future readers.

crates/namada/src/ledger/native_vp/masp.rs
grarco: @yito88 is this ok? Do we have a way to verify that the relayer is indeed forwarding the correct `ShieldingTransfer`?

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
murisi: Fixed this by integrating changes from #3422 .

crates/namada/src/ledger/native_vp/masp.rs
murisi: Removed the function and `ShieldedActionOwner`. Thanks.

crates/namada/src/ledger/native_vp/masp.rs
murisi: This map has now been removed since the suggestion https://github.com/anoma/namada/pull/3409#discussion_r1644247178 is now being used.

crates/namada/src/ledger/native_vp/masp.rs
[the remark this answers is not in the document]
murisi: Understood, thanks. Added a comment to this effect.

crates/vp_env/src/lib.rs
```diff
@@ -119,6 +120,18 @@ where
     /// Get a tx hash
     fn get_tx_code_hash(&self) -> Result<Option<Hash>, namada_storage::Error>;
 
+    /// Get the masp tx part of the shielded action
+    fn get_shielded_action(
```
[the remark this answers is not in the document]
murisi: Thanks, done.

crates/sdk/src/masp.rs
murisi: Thanks. Resolved.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Ah, I see thanks. I've partially applied #3422 where it pertains to determining a token `Address` from a packet `denom`. However I just wanted to clarify whether or not logic to receive NFT messages is missing before applying the rest of the PR. See https://github.com/anoma/namada/pull/3422/files#r1645836855 . Thanks.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Thanks. Done.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Understood, done.

crates/namada/src/ledger/native_vp/masp.rs
murisi: I've discussed turning this function into an RPC endpoint with @tzemanovic . Doing so seems suboptimal because this function is a essentially a heuristic for guessing IBC denominations from `IbcToken` hashes based on the results the protocol has cached so far. Also it seems to me that somehow using generics to share this logic would be overkill for this limited issue. Do you think there's a way to eliminate the use of this function in the MASP VP @yito88 ?

crates/namada/src/ledger/native_vp/masp.rs
yito88: Ah, the IBC VP has to validate the `ShieldingTransfer` for the given IBC message.
I'll open another PR.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Moving the `ChangedBalances` functions into the `namada_ibc` crate is problematic because those functions depend on the `namada_vp_env` crate (which itself depends on the `namada_ibc` crate). The latter dependency is due to those functions doing pre and post balance reads as well as IBC trace reads and packet commitment reads from storage. I suspect I'd have to restructure these functions in order to reduce their dependence on validity predicate functions first. Should I attempt restructuring those functions in this PR?

crates/core/src/masp.rs
```diff
@@ -467,6 +469,14 @@ impl TransferSource {
...
```
cwgoes, line 473: Comment is still wrong?

crates/namada/src/ledger/native_vp/masp.rs
cwgoes: @yito88 can you clarify here? I'm quite confused about this

crates/namada/src/ledger/native_vp/masp.rs
cwgoes: Ah, I see.

crates/namada/src/ledger/native_vp/masp.rs
cwgoes: (discussed offline - let's do this separately)

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -279,66 +295,346 @@ where
...
+        let mut changed_balances = counterparts_balances
+            .try_fold(ChangedBalances::default(), |acc, account| {
+                self.apply_balance_change(acc, account)
+            })?;
+
+        let ibc_addr = TAddrData::Addr(IBC);
+        // Enable decoding the IBC address hash
+        changed_balances.decoder.insert(addr_taddr(IBC), ibc_addr);
+
+        // Go through the IBC events and note the balance chages they imply
```
cwgoes, line 630:
```suggestion
        // Go through the IBC events and note the balance changes they imply
```

crates/namada/src/ledger/native_vp/masp.rs
yito88: `query_ibc_denom` is for looking for readable IBC denom with `IbcToken` hash.
Using an owner was to avoid looking up all IBC denoms when the balance query didn't require `--token`. (Now the query requires both the token and the owner. The RPC function is left for now.)
I think we don't need the query in MASP VP because the IBC denom should be in the packet and the token address can be calculated from it.

murisi: For context, the token address for a receive packet is computed here in the MASP VP: https://github.com/anoma/namada/blob/4de6577c73e223504b0fcb226257ef7b85e248a7/crates/namada/src/ledger/native_vp/masp.rs#L483C25-L483C55 . This computation was designed to produce the same shielded token as the one produced by Hermes, which can be found at https://github.com/anoma/namada/blob/5cc23b4f43b12ecfc70eeca8d04cbe15f4b11a3a/crates/sdk/src/tx.rs#L3505 

> I think we don't need the query in MASP VP because the IBC denom should be in the packet and the token address can be calculated from it.

Thanks for the clarification @yito88 . Would this same reasoning apply to how Hermes derives the token from the IBC denom (also using `query_ibc_denom` and `received_ibc_token`)? Or is the context of the Hermes computation somehow different?

yito88: `query_ibc_denom` is to look up for stored IBC denoms. `received_ibc_token` returns the IBC denom to be minted/unescrowed. Hermes doesn't derive the token.

crates/core/src/masp.rs
murisi: Apologies. Resolved.

crates/namada/src/ledger/native_vp/masp.rs
murisi: Resolved. Thanks.

PR #3416, opened 2024-06-17
Index batched txs via their wrapper and commitment hashes

### LINKED ISSUE #3403: Uniquely identify user txs
Currently, inner txs in a batch are only uniquely identified by the code, data and memo section hashes, but these hashes are not necessarily unique, since they may only differ by the `salt` parameter used to compute them, which is not decided by the protocol (i.e. users can control the value of `salt`).

In order to uniquely identify an inner tx in a batch, we should concat the hash of the wrapper tx containing the inner txs (i.e. `tx.header_hash()`) with the hash of all the commitments, and compute a new hash which we now define as the actual hash of the inner tx.

sug0:
## Describe your changes

Closes #3403

`v0.39.0`

### DIFF
--- a/.changelog/unreleased/improvements/3416-uniquely-identify-inner-txs.md
+++ b/.changelog/unreleased/improvements/3416-uniquely-identify-inner-txs.md
@@ -1,0 +1,2 @@
+- Index batched txs via their wrapper and commitment hashes.
+  ([\#3416](https://github.com/anoma/namada/pull/3416))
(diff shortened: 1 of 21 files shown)


grarco (approved):
Looks good, thanks!

tzemanovic (approved):
just a small style nit: I'd re-export `either` from `tx` crate since it now appears in its public API so that the other crates don't have to pull it in directly to use it

Issue #1214, opened 2023-03-13
consider using `std::io::Error` in place of `storage_api::Error`


tzemanovic:
The `std::io::Error` has the same design considerations that we want for `storage_api::Error` and is familiar to Rust devs, so it might be better to use it than to re-invent our own. Additionally, `std::io::Error` is well suited for no-std (wasm) env too.

tzemanovic:
a new alternative that improves many things in error handling: https://github.com/hashintel/hash/tree/main/libs/error-stack

tzemanovic:
another new crate that provides nice error handling ergonomics without giving up type-safety https://crates.io/crates/terrors (can be used in combination with some other crate)

PR #3419, opened 2024-06-17
move masp validation from SDK into shielded_token crate

### LINKED ISSUE #3101: Move `verify_shielded_tx` out of the SDK
The functions that verifies the masp signatures and proofs currently lives in the SDK crate. We don't really have a plan to expose this to external developers, so we should move it somewhere else, possibly in the masp native vp file. We can also rework the `consume_verify_gas` argument to just be a `VpGasMeter`

tzemanovic:
## Describe your changes

closes #3101

0.39.0


crates/shielded_token/src/validation.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+fn load_pvks() -> &'static PVKs {
+    &VERIFIYING_KEYS
+}
+
+/// Verify a shielded transaction.
+pub fn verify_shielded_tx<F>(
+    transaction: &Transaction,
+    consume_verify_gas: F,
```
grarco, line 119: Do we have access to the `VpGasMeter` object from this crate. Would be a bit cleaner if we could pass a mutable reference directly to it, but no worries if we don't have it

grarco (dismissed):
One minor comment but LGTM!

crates/shielded_token/src/validation.rs
tzemanovic: yeah we do and I tried to use it, but because the `VpEnv::charge_gas` replaces the error with `meter.consume(used_gas).map_err(|_| { Error::SimpleMessage("Gas limit exceeded in native vp")})` it actually makes it more verbose :/ I'd like to change how to handle composition of error types with #1214 that could make this nicer, but for now I'd leave it as is

grarco (dismissed):

tzemanovic:
The `make check-crates` was still failing, but is now fixed in [b588224](https://github.com/anoma/namada/pull/3419/commits/b58822479095834944b5a20d587bd037ac10f5cd) - I had to re-introduce the `download-params` as the validation must always be available even without it for the SDK to compile

tzemanovic:
> The `make check-crates` was still failing, but is now fixed in [b588224](https://github.com/anoma/namada/pull/3419/commits/b58822479095834944b5a20d587bd037ac10f5cd) - I had to re-introduce the `download-params` feature as the MASP VP must always be available even without it for the SDK to compile

So `download-params` wasn't enough on it's own - I also had to bring back the `validation` feature too in [89eaa9f](https://github.com/anoma/namada/pull/3419/commits/89eaa9f610de6361c4bcd95c77f9dcb8a6d97537) to prevent JS wasm bindgen leaking into our wasm build (I think this is due the combination of `rand_core` which pulls `getrandom` and `masp_proof` which enabled feature `js` in `getrandom`).

tzemanovic:
so the problem was indeed in `masp_proofs` - https://github.com/anoma/masp/pull/85

tzemanovic:
I rm'd the `"validation"` feature again in https://github.com/anoma/namada/compare/89eaa9f610de6361c4bcd95c77f9dcb8a6d97537..2f3abec526148f599a567907d153fd7b9d2e45eb as it's not needed with the masp_proofs fix

grarco (approved):

PR #3420, opened 2024-06-17
test ibc e2e fix for #3389

sug0:
#3389


sug0:
Closing this pr, as it seems #3389's CI is finally passing.

Here are the two additional patches applied to this PR, which was based on #3389:
[0001-Wait-for-block-height-2-in-IBC-tests.patch](https://github.com/user-attachments/files/15886294/0001-Wait-for-block-height-2-in-IBC-tests.patch)
[0002-Avoid-precision-drop-in-timestamp-conversions.patch](https://github.com/user-attachments/files/15886295/0002-Avoid-precision-drop-in-timestamp-conversions.patch)

PR #3428, opened 2024-06-19
Improve gas interface

### LINKED ISSUE #3395: Change gas SCALE parameter
After #3391 we should change the scale to some more sane default (ideally 1 order of magnitude less) cc @grarco

### LINKED ISSUE #3427: Improve usage of `Gas` type
Currently we deal with the `Gas` type in both sub units (for gas tracking) and whole units (to display results to the user). The current interface makes it difficult to understand if the `Gas` we are dealing with is the one or the other. We should create a new type to be used when displaying results to users (with events) while keeping the current type just for tracking purposes.

grarco:
## Describe your changes

Closes #3427.
Closes #3395.

Introduces a new `WholeGas` type to better distinguish between gas in subunits (for tracking) and gas in whole units (for events and fees).

Reduces the gas scale in the localnet genesis file and in tests by an order of magnitude to increase the gas cost of transactions which was too low.

Removes the redundant `gas_used` field of `TxResult` (since we already emit an attribute `GasUsed`) and collapse the previous `BatchResults` directly into `TxResult`.

`v0.40.0`


grarco:
@yito88 Hermes tests failing again (sorry about that)

grarco:
> @yito88 Hermes tests failing again (sorry about that)

Ah actually I believe the failures come from #3391

yito88:
yeah, it seems that the parameter structure change in #3391 causes the failure.

grarco:
> yeah, it seems that the parameter structure change in #3391 causes the failure.

Maybe not, I've just pushed a fix to that branch that was supposed to fix a failing ibc test (non-hermes) but it seems to have fixed all the e2e tests

brentstone:
do we still want this PR?

grarco:
> do we still want this PR?

I think it would be nice if we could add it to the next release

grarco:
> @grarco Looks good. `e2e::ibc_tests::run_ledger_ibc` failed without Hermes. It seems that the default gas limit isn't enough for `tx_ibc.wasm`.

Thanks for catching it! Should be fixed in `fb81246a0e9179384bec1838f8a2229d17d09e64`. I've also fixed the broken integration tests

crates/sdk/src/tx.rs
```diff
@@ -420,11 +420,9 @@ pub fn display_batch_resp(context: &impl Namada, resp: &TxResponse) {
             InnerTxResult::Success(_) => {
                 display_line!(
                     context.io(),
-                    "Transaction {} was successfully applied at height {}, \
-                     consuming {} gas units.",
+                    "Transaction {} was successfully applied at height {}.",
```
brentstone, line 423: why did you decide to remove the gas consumption of the tx in the logs?

crates/tx/src/data/mod.rs
```diff
@@ -500,7 +484,7 @@ pub struct VpsResult {
 impl<T: Serialize> fmt::Display for TxResult<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if f.alternate() {
-            write!(f, "Transaction is valid. Gas used: {}", self.gas_used)
+            write!(f, "Transaction is valid.")
```
brentstone, line 487: My opinion is that keeping this log is important if we are able to still.

grarco: I've had a chat with @Fraccaman on this. The gas is already emitted as an event, so this field in the `TxResult` was essentially duplicated. I'll try to update the logs in the client to retrieve this piece of information from the events

crates/sdk/src/tx.rs
grarco: My bad, brought it back in 209da0c6ad

crates/tx/src/data/mod.rs
grarco: Brought back the logging of the gas used in 209da0c6ad but directly in the SDK when fetching the tx events. It is no longer possible to log it directly in here since the gas is no longer present here

PR #3439, opened 2024-06-25
Early sapling balance check

### LINKED ISSUE #2721: Early rejection of transactions with invalid sapling value balance
We could introduce an early check in the masp vp to reject a transaction carrying an invalid `sapling_value_balance`. These transactions are already rejected by the masp vp but this happens later during the evaluation process.

grarco:
## Describe your changes

Closes #2721.

Moves the check on the sapling value balance slightly up in the masp vp to reject invalid transactions earlier.

#3393 (diffs for review: https://github.com/anoma/namada/pull/3439/files/13b4bcd098d0288c11c90088c6a7b106818f76a8..a3bb632a1f40700abae9027a80cc9fcd88725f70)


murisi (dismissed):
Great, thanks!

crates/namada/src/ledger/native_vp/masp.rs
```diff
@@ -390,7 +390,7 @@ where
         }
 
         // The Sapling value balance adds to the transparent tx pool
-        let mut transparent_tx_pool = shielded_tx.sapling_value_balance();
+        let transparent_tx_pool = shielded_tx.sapling_value_balance();
```
murisi, line 393: Since `transparent_tx_pool` is only used inside `fn validate_transparent_bundle`, maybe this variable declaration could be moved into `fn validate_transparent_bundle`? If doable, this would reduce the number of parameters taken by `fn validate_transparent_bundle`.

grarco: You're right! Done in a3bb632a1f40700abae9027a80cc9fcd88725f70

murisi (approved):

PR #3422, opened 2024-06-18
Refine IBC parts for masp ibc replay protection

yito88:
## Indicate on which release or other PRs this topic is based on
#3409 
Diff: https://github.com/anoma/namada/compare/76fab4c6c7b1c12c65aa02cf4f2923429bd4da1d...5c3b3ab70af77ccdbf7eedec60aa4cc4d3b2d1dd

## Checklist before merging to `draft`
- [ ] I have added a changelog
- [x] Git history is in acceptable state

PR #3444, opened 2024-06-26
Make the sender set MASP transaction to IBC memo

### LINKED ISSUE #3438: Another receiver can get tokens in IBC shielding transfer with an evil IBC relayer
Currently, MASP `Transaction` is built by an IBC relayer(Hermes) and is used at the destination Namada.
However, even if Hermes replaces the actual receiver with another receiver payment address when building MASP `Transaction`, Namada VPs can't check if it is built with the proper receiver.

yito88:
## Describe your changes
closes #3438 

~~WARNING: don't merge this PR for now because `tx_ibc.wasm` failed to be compiled on my Apple silicon machine.~~
```
Transaction runner error: Wasm compilation error: Compilation error: Assembler failed finalization with: ImpossibleRelocation(Dynamic(DynamicLabel(0)))
```
We could workaround this error.

#3422
Diff: https://github.com/anoma/namada/compare/5c3b3ab70af77ccdbf7eedec60aa4cc4d3b2d1dd...ca795542d1b2dfe1b4d3b4fea8f3dfd751672db1


crates/ibc/src/msg.rs
```diff
@@ -91,103 +90,86 @@ impl BorshDeserialize for MsgNftTransfer {
...
+    let bytes = HEXUPPER.decode(memo.as_ref().as_bytes()).ok()?;
+    IbcShieldingData::try_from_slice(&bytes).ok()
+}
+
+/// Extract MASP transaction from IBC packet memo
+pub fn extract_masp_tx_from_packet(
+    packet: &Packet,
+    is_sender: bool,
+) -> Option<MaspTransaction> {
+    let is_ft_packet = if is_sender {
```
cwgoes, line 142: What is `is_ft_packet` - whether or not the packet is for a fungible token? Why do we even need to know this?

crates/ibc/src/msg.rs
```diff
@@ -91,103 +90,86 @@ impl BorshDeserialize for MsgNftTransfer {
...
+        packet.port_id_on_a == PortId::transfer()
+    } else {
+        packet.port_id_on_b == PortId::transfer()
+    };
+
+    let shielding_data = if is_ft_packet {
```
cwgoes, line 148: There could be more packet types in the future; this logic is not very resilient.

cwgoes: We should also separate this out into a function.

crates/ibc/src/msg.rs
yito88: It needs to decode the packet. Unfortunately, the packet data structure isn't the same as the NFT packet data.
https://github.com/cosmos/ibc-rs/blob/7ff41b3804cda649e0d7f7c21389b8d65f8d9b4b/ibc-apps/ics20-transfer/types/src/packet.rs#L28-L33
https://github.com/cosmos/ibc-rs/blob/7ff41b3804cda649e0d7f7c21389b8d65f8d9b4b/ibc-apps/ics721-nft-transfer/types/src/packet.rs#L29-L43

crates/ibc/src/msg.rs
yito88: Fixed: ca795542d1b2dfe1b4d3b4fea8f3dfd751672db1

crates/apps_lib/src/cli.rs
```diff
@@ -6498,7 +6526,7 @@ pub mod args {
                     "The channel ID via which the token is received."
                 )))
                 .arg(REFUND.def().help(wrap!(
-                    "Generate the shielded transfer for refunding."
+                    "Generate the shielding transfer for refunding."
```
grarco, line 6529: This is used to generate both the refunding tx and the actual shielding transfer on the destination chain, correct? If that's the case maybe we can change the message here

crates/sdk/src/args.rs
```diff
@@ -2662,6 +2662,6 @@ pub struct GenIbcShieldedTransfer<C: NamadaTypes = SdkTypes> {
     pub port_id: PortId,
     /// Channel ID via which the token is received
     pub channel_id: ChannelId,
-    /// Generate the shielded transfer for refunding
+    /// Generate the shielding transfer for refunding
```
grarco, line 2665: Same here as in the client

crates/sdk/src/masp.rs
```diff
@@ -864,6 +869,32 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
     }
 
+    /// Extract the relevant shield portions from the IBC messages in [`Tx`]
+    fn extract_masp_tx_from_ibc_message(
+        tx: &Tx,
+    ) -> Result<Vec<Transaction>, Error> {
+        let mut masp_txs = Vec::new();
+        for cmt in &tx.header.batch {
+            let tx_data = tx.data(cmt).ok_or_else(|| {
+                Error::Other("Missing expected masp transaction".to_string())
```
grarco, line 879:
```suggestion
                Error::Other("Missing transaction data".to_string())
```

crates/sdk/src/masp.rs
```diff
@@ -864,6 +869,32 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+                if let Some(masp_tx) = extract_masp_tx_from_envelope(envelope) {
+                    masp_txs.push(masp_tx);
+                }
+            }
+        }
+        if !masp_txs.is_empty() {
+            Ok(masp_txs)
+        } else {
+            Err(Error::Other(
+                "IBC meesage doesn't have masp transaction".to_string(),
```
grarco, line 893:
```suggestion
                "IBC message doesn't have masp transaction".to_string(),
```

crates/tx/src/action.rs
```diff
@@ -27,6 +27,7 @@ pub enum Action {
     Gov(GovAction),
     Pgf(PgfAction),
     Masp(MaspAction),
+    IbcShielding,
```
grarco, line 30: I was thinking that instead of `IbcShielding` perhaps we could just expand the `MaspAction` to be an enum like:

```
enum MaspAction {
    TxSection(Hash),
    InnerTxCmt(Hash)
}
```

The first case is the one we already have and its hash is the hash of the tx's `Masp` section. The second case is an alternative to `IbcShielding` and its hash is that of the `TxCommitments` (i.e. the inner tx hash). The ibc transaction could write the second variant when the tx is an incoming one. In the SDK then, we can look at the specific inner txs of the batch that contain a masp ibc tx and extract only their memos without looking at all of them.

This is a minor suggestion, feel free to ignore it

grarco (dismissed):
Minor comments but looks good to me, thanks!

crates/apps_lib/src/cli.rs
yito88: Thanks, Hermes used the argument. Now `build_ibc_transfer` generates the masp tx for the refunding.
Users doesn't have to generate it with this command anymore.
I'll remove this argument.

crates/tx/src/action.rs
yito88: Thanks, it looks better. However, the change for setting Hash into the action caused wasm compilation error on arm mac. Let's leave it as it is.

grarco:
Actually, I believe there's an issu, my bad for not spotting it earlier. In the sdk we use `extract_masp_tx_from_ibc_message` when we have a valid masp tx which is missing the masp ref event. In this case we iterate over all the inner txs of the batch and try to extract the ibc message from the tx data and if we can we push the masp `Transaction` to the result.

The issue is that we don't know if that specific inner tx was successful: if it wasn't, the client (or indexer) extracts information from a wrong source and spoils the internal state of the `ShieldedContext`, which might prevent it from constructing valid transactions in the future.

We need a way to tell, in the events we emit, which transactions of the batch are actually valid and filter only these ones in `extract_masp_tx_from_ibc_message`

grarco:
> Actually, I believe there's an issue, my bad for not spotting it earlier. In the sdk we use `extract_masp_tx_from_ibc_message` when we have a valid masp tx which is missing the masp ref event. In this case we iterate over all the inner txs of the batch and try to extract the ibc message from the tx data and if we can we push the masp `Transaction` to the result.
> 
> The issue is that we don't know if that specific inner tx was successful: if it wasn't, the client (or indexer) extracts information from a wrong source and spoils the internal state of the `ShieldedContext`, which might prevent it from constructing valid transactions in the future.
> 
> We need a way to tell, in the events we emit, which transactions of the batch are actually valid and filter only these ones in `extract_masp_tx_from_ibc_message`

Opened #3488 to track this separately

PR #3446, opened 2024-06-27
Murisi/generalize vectorized transfers

murisi:
## Describe your changes
Attempt to simplify the combination of split MASP transfers and vectorized transfers. The main change is to make a transfer format that allows shielded operations to be done in the presence of both transparent inputs and transparent outputs. All the other transfer formats are then subsumed into this one.

#3356 

- diff https://github.com/anoma/namada/pull/3446/files/1e98403560a297f56f38ff67397a4bb9182068aa..ddded01e757393f0b3e1868617a57ebcde183c85


cwgoes (approved):
Conceptual ACK - much, much cleaner!

crates/sdk/src/args.rs
```diff
@@ -325,19 +341,26 @@ impl TxShieldedTransfer {
     }
 }
 
+/// Shielded transfer-specific arguments
```
grarco, line 344:
```suggestion
/// Shielding transfer-specific arguments
```

crates/sdk/src/masp.rs
```diff
@@ -1621,235 +1609,309 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
-                        .map_err(builder::Error::SaplingBuild)?;
-                }
+        let mut is_context_loaded = false;
+        for MaspTransferData {
+            source,
+            target,
+            token,
+            amount,
+        } in data
+        {
```
grarco, line 1622: Just a note: I think I've improved this in #3393 by trying to group the same sources/tokens before looking for the available notes

crates/sdk/src/signing.rs
```diff
@@ -702,61 +701,101 @@ fn format_outputs(output: &mut Vec<String>) {
...
+
 impl TokenTransfer<'_> {
-    fn source(&self) -> Option<&Address> {
-        match self {
-            TokenTransfer::Transparent(transfer) => Some(&transfer.source),
-            TokenTransfer::Shielded => None,
-            TokenTransfer::Shielding(transfer) => Some(&transfer.source),
-            TokenTransfer::Unshielding(_) => None,
-        }
+    fn sources(&self) -> Vec<&Address> {
```
grarco, line 712: Question here: should we ensure that in the vectors return by `sources` and `targets` there are no duplicates? Probably not because of the hardware wallet?

grarco (approved):

crates/sdk/src/signing.rs
murisi: > Question here: should we ensure that in the vectors return by `sources` and `targets` there are no duplicates? Probably not because of the hardware wallet?

I think you're right. We should ensure there are no duplicates in `sources` nor `targets`. Otherwise `make_ledger_token_transfer_endpoints` might print the same source or target repeatedly. Would you be able to look at #3459 ? I think this issue is inadvertently fixed there...

crates/token/src/lib.rs
```diff
@@ -82,127 +82,24 @@ where
...
-    Eq,
-    PartialOrd,
-    Serialize,
-    Deserialize,
-)]
-pub struct ShieldingTransfer {
+pub struct Transfer {
     /// Transfer-specific data
-    pub data: ShieldingTransferData,
+    pub data: Vec<TransferData>,
```
tzemanovic, line 87: let's make this slightly more descriptive
```suggestion
    pub transparent: Vec<TransparentTransfer>,
```

tzemanovic (approved):

crates/token/src/lib.rs
murisi: Done, thanks.

crates/sdk/src/args.rs
murisi: Fixed, thanks.

brentstone:
Closing in favor of #3459

Issue #3192, opened 2024-05-07
Code optimization recommendations


grarco:
Reported by Informal Systems.

## Description

Upon reviewing the codebase, several optimizations were identified that could enhance the efficiency and readability of the code.

## Problem Scenarios

1. The `validate_tx` function is redundantly called twice in the `check_proposal_tx` function, during
process proposal, leading to unnecessary overhead. 1st: /process_proposal.rs#L249-L256
2nd: /process_proposal.rs#L265-L270
2. The previously defined `inner_tx_hash` should be used instead of &tx.raw_header_hash() .
`mempool_validation()`
3. Move parameter obtaining from storage outside loop iterations to minimize unnecessary storage calls. In this way, the number of calls to storage would be reduced. /governance.rs#L79
4. Since the `validate_wrapper_bytes` function is specific to Wrapper transactions, the transaction type check should be performed at the beginning of the function. If the transaction is not a WrapperTx, an error
should be immediately returned. /prepare_proposal.rs#L286

## Recommendation

To improve performance, it is recommended to implement the suggestions outlined in the Problem Scenarios section.

PR #3448, opened 2024-06-27
Streamline validation

### LINKED ISSUE #3190: Incorrect error code mapping in `check_proposal_tx` description
Reported by Informal Systems.

# Description

The `check_proposal_tx` function contains error codes that are not correctly mapped to their corresponding IDs in the description comment. This discrepancy can lead to confusion and potential misinterpretation of error conditions. For example, the error code `WasmRuntimeError` is listed as 3 in the function's description comment, but it is defined as 1 in the `ResultCode` enum, not 3 as indicated in the comment. Additionally, some error codes are not listed at all in the function's description comment, leading to incomplete documentation.

# Problem Scenarios

Developers relying on the error code documentation within the description comment of the `check_proposal_tx` function may encounter confusion considering discrepancies between the documented error codes and their actual definitions.

# Recommendation

To ensure consistency and accuracy, it is recommended to align the error code documentation in the description
comment of the `check_proposal_tx` function (/process_proposal.rs#L174-L184) with the definitions in the `ResultCode` enum (/mod.rs#L55-L86).

### LINKED ISSUE #3193: Streamlining validation logic across proposal lifecycle
Reported by Informal Systems.

# Description

Any stateless validation done when adding a transaction to the mempool does not need to be done again in prepare proposal. This is because a proposal from an honest proposer only includes transactions from its mempool. It is then guaranteed that any transaction in such a proposal has passed all those checks.
The same reasoning could be applied to process proposal and finalize block: under the assumption that 2f+1 of
the voting power is controlled by honest validators, CometBFT guarantees that any block in a finalize block call has been validated by at least one honest validator via process proposal. Nevertheless, if a larger set of validators is compromised, this may not be true anymore. In such a case, one may want to replicate any validity checks done in process proposal in finalize block to guarantee the validity of histories under all circumstances.
We have identified one redundant validity check in prepare proposal and one in finalize block.

# Problem Scenarios

- Redundant signature check in mempool validate [code](https://github.com/anoma/namada/blob/f4c838d2c53b28091af9b9a2f10b5b53e55fe65e/crates/apps/src/lib/node/ledger/shell/mod.rs#L940) and prepare proposal [code](https://github.com/anoma/namada/blob/f4c838d2c53b28091af9b9a2f10b5b53e55fe65e/crates/apps/src/lib/node/ledger/shell/prepare_proposal.rs#L285)
- Redundant signature check in process proposal [code](https://github.com/anoma/namada/blob/f4c838d2c53b28091af9b9a2f10b5b53e55fe65e/crates/apps/src/lib/node/ledger/shell/process_proposal.rs#L265-L270) and finalize block [code](https://github.com/anoma/namada/blob/f4c838d2c53b28091af9b9a2f10b5b53e55fe65e/crates/apps/src/lib/node/ledger/shell/finalize_block.rs#L189-L195)

# Recommendation

We recommend removing the duplicated signature check from prepare proposal. Regarding the checks in finalize block, we recommend proceeding differently depending on whether the team wants to guarantee the validity of histories under all circumstances:

-  If required, we recommend making sure that all validity checks in process proposal are replicated in finalize block.
-  If not required, we recommend removing the duplicated checks from finalize block.

grarco:
## Describe your changes

Closes #3190.
Closes #3193.

This PR fixes some wrong documentation for the `finalize_block` and `check_proposal_tx_function`.

It also introduces a new local configuration parameter to allows a node to rerun the `process_proposal` checks before finalize block in case they don't want to blindly trust the majority of validators. This is implemented for both full nodes and validators since the latter could also skip the `process_proposal` call in some cases.

#3192 (diffs for review: https://github.com/anoma/namada/pull/3448/files/176754fbd523ec7d17a1dbbd10f59900b846aeb1..1a704b679b122656aa4ff6d0baa947054c326271)

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -138,54 +138,7 @@ impl Shell {
             Request::FinalizeBlock(finalize) => {
                 tracing::debug!("Request FinalizeBlock");
 
-                let recheck_process_proposal = match self.mode {
-                    shell::ShellMode::Validator {
-                        ref local_config, ..
-                    } => local_config
-                        .as_ref()
-                        .map(|cfg| cfg.recheck_process_proposal)
-                        .unwrap_or_default(),
-                    shell::ShellMode::Full { ref local_config } => local_config
-                        .as_ref()
-                        .map(|cfg| cfg.recheck_process_proposal)
-                        .unwrap_or_default(),
-                    shell::ShellMode::Seed => false,
-                };
-
-                if recheck_process_proposal {
-                    let tm_raw_hash_string =
-                        tm_raw_hash_to_string(&finalize.proposer_address);
-                    let block_proposer = find_validator_by_raw_hash(
-                        &self.state,
-                        tm_raw_hash_string,
-                    )
-                    .unwrap()
-                    .expect(
-                        "Unable to find native validator address of block \
-                         proposer from tendermint raw hash",
-                    );
-
-                    let check_result = self.process_txs(
-                        &finalize
-                            .txs
-                            .iter()
-                            .map(|ptx| ptx.tx.clone())
-                            .collect::<Vec<_>>(),
-                        finalize.header.time,
-                        &block_proposer,
-                    );
-
-                    let invalid_txs = check_result.iter().any(|res| {
-                        let error = ResultCode::from_u32(res.code)
-                            .expect("Failed to cast result code");
-                        !error.is_recoverable()
-                    });
-
-                    if invalid_txs {
-                        return Err(Error::InvalidBlockProposal);
-                    }
-                }
-
+                self.try_recheck_process_proposal(&finalize)?;
                 self.finalize_block(finalize).map(Response::FinalizeBlock)
             }
             Request::Commit => {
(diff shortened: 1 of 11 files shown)


grarco:
**NOTE**: currently, the recheck is done calling `process_txs` but I believe it would be slightly better to call `process_proposal` (which is the handler of the cometbft's call), so that if we changed anything in the handler that would immediately translate to the recheck logic too, while, at the moment, if the modification lies outside of `process_txs` this would not translate to the recheck. On top of that, because of the shim, nodes can currently call `process_txs` twice for the same block proposal. With the addition of the recheck logic this can go up to 3 calls to this function: would be nice if we could keep track of the whether the function already ran so that we can ensure that it gets called at most once.

I'll try to address both of these in another branch

crates/node/src/lib.rs
```diff
@@ -134,6 +137,55 @@ impl Shell {
             }
             Request::FinalizeBlock(finalize) => {
                 tracing::debug!("Request FinalizeBlock");
+
```
tzemanovic, line 140: let's put this new block into e.g. `self.try_recheck_process_proposal(&finalize)?;`

tzemanovic (changes requested):
1 small style recommendation - other than that LGTM

crates/node/src/lib.rs
grarco: Done in fc3241dfd11cebfdbc68cb318f3e9b1469d1fcf7

tzemanovic (approved):

PR #3456, opened 2024-06-28
Phase 2 of Shielded Sync refactor

sug0:
## Describe your changes

Implements phase 2 of Issue #3385

Based on #3498

**Diff for review:** https://github.com/anoma/namada/compare/tiago/masp-ss-client-phase-1..tiago/masp-ss-phase-2

### DIFF
--- a/crates/sdk/src/masp.rs
+++ b/crates/sdk/src/masp.rs
@@ -3590,7 +3590,24 @@ mod test_shielded_sync {
     /// Would love to do this in a less opaque fashion, but
     /// making these things is a misery not worth my time.
     ///
-    /// This a tx sending 1 BTC from Albert to Albert's PA
+    /// This a tx sending 1 BTC from Albert to Albert's PA,
+    /// that was extracted from a masp integration test.
+    ///
+    /// ```ignore
+    /// vec![
+    ///     "shield",
+    ///     "--source",
+    ///     ALBERT,
+    ///     "--target",
+    ///     AA_PAYMENT_ADDRESS,
+    ///     "--token",
+    ///     BTC,
+    ///     "--amount",
+    ///     "1",
+    ///     "--node",
+    ///     validator_one_rpc,
+    /// ]
+    /// ```
     fn arbitrary_masp_tx() -> Transaction {
         Transaction::try_from_slice(&[
             2, 0, 0, 0, 10, 39, 167, 38, 166, 117, 255, 233, 0, 0, 0, 0, 255,
--- a/crates/sdk/src/masp.rs
+++ b/crates/sdk/src/masp.rs
@@ -644,6 +644,72 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
         Ok(last_height_in_witnesses.unwrap_or_else(BlockHeight::first))
     }
 
+    #[cfg(not(target_family = "wasm"))]
+    async fn update_with_pre_built_data<M>(
+        &mut self,
+        client: &M,
+        height: BlockHeight,
+    ) -> Result<(), Error>
+    where
+        M: MaspClient,
+    {
+        trait ThenAwait {
+            #[allow(async_fn_in_trait)]
+            async fn then_await<F>(self, fut: F) -> Option<F::Output>
+            where
+                F: std::future::Future;
+        }
+
+        impl ThenAwait for bool {
+            async fn then_await<F>(self, fut: F) -> Option<F::Output>
+            where
+                F: std::future::Future,
+            {
+                if self { Some(fut.await) } else { None }
+            }
+        }
+
+        let tree_fut = async {
+            client
+                .capabilities()
+                .may_fetch_pre_built_tree()
+                .then_await(client.fetch_commitment_tree(height))
+                .await
+                .transpose()
+        };
+        let notes_map_fut = async {
+            client
+                .capabilities()
+                .may_fetch_pre_built_notes_map()
+                .then_await(client.fetch_tx_notes_map(height))
+                .await
+                .transpose()
+        };
+        let witness_map_fut = async {
+            client
+                .capabilities()
+                .may_fetch_pre_built_witness_map()
+                .then_await(client.fetch_witness_map(height))
+                .await
+                .transpose()
+        };
+
+        let (maybe_tree, maybe_notes_map, maybe_witness_map) =
+            futures::try_join!(tree_fut, notes_map_fut, witness_map_fut)?;
+
+        if let Some(tree) = maybe_tree {
+            self.tree = tree;
+        }
+        if let Some(notes_map) = maybe_notes_map {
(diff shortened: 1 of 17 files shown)


crates/sdk/src/masp.rs
```diff
@@ -622,35 +618,123 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+    async fn fetch_aux<IO, M>(
         &mut self,
         client: M,
         progress: &impl ProgressTracker<IO>,
         start_query_height: Option<BlockHeight>,
         last_query_height: Option<BlockHeight>,
         retry: RetryStrategy,
-        // NOTE: do not remove this argument, it will be used once the indexer
-        // is ready
-        _batch_size: u64,
```
grarco, line 534: Is it ok to remove this arg? I believe @batconjurer had some plans for it but I'm not up to date with the latest decisions for the indexer

batconjurer: I'm good with it because of the phase 3 design that we have now.

crates/sdk/src/masp/utils.rs
```diff
@@ -58,17 +104,64 @@ pub trait MaspClient<'client, C: Client> {
...
         tx_sender: FetchQueueSender,
-        from: u64,
-        to: u64,
+        from: BlockHeight,
+        to: BlockHeight,
     ) -> Result<(), Error>;
+
+    /// Return the capabilities of this client.
+    #[inline(always)]
+    fn capabilities(&self) -> MaspClientCapabilities {
```
grarco, line 113: Do we need these default implementations?

crates/sdk/src/masp/utils.rs
```diff
@@ -153,13 +245,426 @@ where
...
+///
+/// [`namada-masp-indexer`]: <https://github.com/anoma/namada-masp-indexer>
+#[cfg(not(target_family = "wasm"))]
+#[derive(Clone, Debug)]
+pub struct IndexerMaspClient {
+    indexer_api: Arc<reqwest::Url>,
+    client: reqwest::Client,
+}
+
+#[cfg(not(target_family = "wasm"))]
```
grarco, line 265: Do we need this trait?

crates/sdk/src/masp/utils.rs
```diff
@@ -153,13 +245,426 @@ where
...
+        shutdown_signal: &mut ShutdownSignal,
+        mut tx_sender: FetchQueueSender,
+        BlockHeight(mut from): BlockHeight,
+        BlockHeight(to): BlockHeight,
+    ) -> Result<(), Error> {
+        use serde::Deserialize;
+
+        #[derive(Deserialize)]
+        struct TransactionSlot {
+            // masp_tx_index: u64,
```
grarco, line 372: Should we remove this or do we have plans to use it in the upcoming updates?

crates/sdk/src/masp/utils.rs
```diff
@@ -153,13 +245,426 @@ where
...
+
+            if from >= to {
+                break;
+            }
+        }
+
+        // NB: a hack to drain the iterator in case
+        // there were still some items left
+        // TODO: improve this lol
+        while fetch_iter.next().is_some() {}
```
grarco, line 493: Couldn't this be a sign that something went wrong?

crates/sdk/src/masp.rs
```diff
@@ -523,18 +596,145 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedContext<U> {
...
+    #[cfg(not(target_family = "wasm"))]
+    async fn update_with_pre_built_data<M>(
+        &mut self,
+        client: &M,
+        height: BlockHeight,
+    ) -> Result<(), Error>
+    where
+        M: MaspClient,
+    {
+        trait ThenAwait {
```
tzemanovic, line 656: tbh it seems to me that this little abstraction adds more code than it's worth

crates/sdk/src/masp/utils.rs
```diff
@@ -153,13 +245,426 @@ where
...
+
+    async fn fetch_tx_notes_map(
+        &self,
+        BlockHeight(height): BlockHeight,
+    ) -> Result<BTreeMap<IndexedTx, usize>, Error> {
+        use serde::Deserialize;
+
+        #[derive(Deserialize)]
+        struct Note {
+            // is_fee_unshielding: bool,
```
grarco, line 552: Same here, can we remove these or are they needed for future updates?

crates/sdk/src/masp/utils.rs
```diff
@@ -204,6 +709,7 @@ impl Iterator for FetchQueueReceiver {
                 if let Some(entry) = self.cache.pop_first() {
                     return Some(entry);
                 }
+                core::hint::spin_loop();
```
grarco, line 712: Shouldn't this be handled by `flume`?

crates/sdk/src/masp/utils.rs
batconjurer: This will get used in the future

crates/sdk/src/masp/utils.rs
batconjurer: No, this can be refactored.

crates/sdk/src/masp/utils.rs
sug0: this iterator is going over a range of blocks. if we do not find txs in some specific block, the iterator will not be moved forward, resulting in some elements left to yield after we exit the first loop. this is just a hack to drain the iterator in these cases.

we will get rid of this mechanic in the phase 3 code, progress will be kept track of differently

crates/sdk/src/masp/utils.rs
sug0: well, we don't. it's more leaning towards an aesthetic change, rather than a functional one. it does kind of stop you from fucking up the keep-alive headers

crates/sdk/src/masp/utils.rs
```diff
@@ -153,13 +245,426 @@ where
...
+        use serde::Deserialize;
+
+        #[derive(Deserialize)]
+        struct Witness {
+            bytes: Vec<u8>,
+            index: usize,
+        }
+
+        #[derive(Deserialize)]
+        struct WitnessMapResponse {
```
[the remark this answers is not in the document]
sug0: if anyone should be exposing these types, it should be the `namada-masp-indexer` itself.

crates/sdk/src/masp/utils.rs
sug0: the value here is being acquired through an indirection on a flume channel. we're accessing a cache that indirectly pulls from this channel. we could have used a condvar to check when a new value would be available on the cache, but the truth is we're replacing the entirety of this code in phase 3...

grarco (approved):

PR #3459, opened 2024-06-28
Murisi/unconstrained transfers

murisi:
## Describe your changes
An attempt to defragment the inputs and outputs of a transfer to ease the printing of transfers on hardware wallets. More specifically, the following changes have been made:
* The inputs and outputs of transfers have been vectorized individually as opposed to vectorizing the whole transfer object. So the number of inputs to a transfer does not have to match the number of outputs
* Inputs and outputs to a transfer no longer have to occur in balancing pairs where the amount taken from a source in a pair has to be fully transferred to the target in the pair
* The code for constructing a transfer automatically aggregates multiple credits or debits to an account instead of leaving them as separate entries. This also helps with more concise hardware wallet printing
* Introduced a `multi_transfer` function in the transaction prelude that can apply any number of balance changes simultaneously instead of just 2 like the existing `transfer` function
* Generalized the Transfer event emitted by some operations to allow multiple sources and multiple targets. This involved writing some serialization and deserialisation functions to support the conversions of these events.

#3446 

- diff: https://github.com/anoma/namada/pull/3459/commits/424d0f6589ca8e1586c080acbb55ffe46fbb1057


cwgoes (commented):
Conceptual ACK, but this should receive careful review for correct arithmetic.

crates/trans_token/src/storage.rs
```diff
@@ -258,6 +260,65 @@ where
...
+    S: StorageRead + StorageWrite,
+{
+    // Collect all the accounts whose balance has changed
+    let mut accounts = BTreeSet::new();
+    accounts.extend(sources.keys().cloned());
+    accounts.extend(dests.keys().cloned());
+    // Apply the balance change for each account in turn
+    for ref account @ (ref owner, ref token) in accounts {
+        let unexpected_err = || {
+            storage::Error::new_alloc(
```
grarco, line 282: I think this could be a `new_const` and moved out of the loop

crates/token/src/lib.rs
```diff
@@ -108,21 +101,114 @@ impl Transfer {
...
+        owner: Address,
+        token: Address,
+        amount: DenominatedAmount,
+    ) -> Option<Self> {
+        let account = Account { owner, token };
+        let zero = DenominatedAmount::new(Amount::zero(), amount.denom());
+        let source_amount = *self.sources.get(&account).unwrap_or(&zero);
+        let target_amount = *self.targets.get(&account).unwrap_or(&zero);
+        // If this account is already a source, then reduce the source
+        if amount < source_amount {
```
grarco, line 184: If we want to debit an account, shouldn't we increase its source in any case? Or reduce the target if available?

crates/token/src/lib.rs
```diff
@@ -108,21 +101,114 @@ impl Transfer {
...
-    pub source: Address,
-    /// Target address will receive the tokens
-    pub target: Address,
-    /// Token's address
-    pub token: Address,
-    /// The amount of tokens
-    pub amount: DenominatedAmount,
+pub struct Transfer {
+    /// Sources of this transfer
+    pub sources: BTreeMap<Account, DenominatedAmount>,
```
grarco, line 113: Question, do we lose information about the specific transfers with this struct? @sug0 has placed some`TokenEvent`s around the codebase and I think we don't have all the data to construct those events anymore (specifically we don't have the exact mapping between source and target)

crates/token/src/lib.rs
murisi: Oops, my mistake. The credit/debit function names are flipped. Will fix.

crates/token/src/lib.rs
```diff
@@ -108,21 +101,114 @@ impl Transfer {
...
+    /// Transfer assets between accounts
+    pub fn transfer(
+        self,
+        source: Address,
+        target: Address,
+        token: Address,
+        amount: DenominatedAmount,
+    ) -> Option<Self> {
+        self.credit(source, token.clone(), amount)?
+            .debit(target, token, amount)
```
tzemanovic, line 210: isn't this other way around? I.e. transfer debits the source and credits the target

crates/token/src/lib.rs
murisi: Good point, Yes we lose specific information. @sug0 Could we create a multi-transfer event that doesn't require exact mappings between sources and targets?

crates/trans_token/src/storage.rs
murisi: > I think this could be a `new_const` and moved out of the loop

Thanks, done.

crates/tx_prelude/src/token.rs
```diff
@@ -51,3 +52,34 @@ pub fn transfer(
...
+    for (_, token) in dests.keys() {
+        if token.is_internal() {
+            // Established address tokens do not have VPs themselves, their
+            // validation is handled by the `Multitoken` internal address, but
+            // internal token addresses have to verify the transfer
+            ctx.insert_verifier(token)?;
+        }
+    }
+
+    namada_token::multi_transfer(ctx, sources, dests)?;
```
tzemanovic, line 82: this needs to emit an event similar to the above

crates/token/src/lib.rs
tzemanovic: let's change the `TokenOperation::Transfer` (in `crates/trans_token/src/event.rs`) to support this

murisi: > let's change the `TokenOperation::Transfer` (in `crates/trans_token/src/event.rs`) to support this

I've attempted to do this in e4da8c0433ec323310d64dc8f4e106e157a9b0f1 . Would you be able to review again?

tzemanovic (dismissed):

crates/token/src/lib.rs
tzemanovic: the change lgtm, but some tests seem to be affected

grarco: @murisi it looks good to me but tests are failing

murisi: Ah, fixed. Thanks!

crates/events/src/extend.rs
```diff
@@ -666,6 +668,90 @@ where
...
+
+impl<'de> Deserialize<'de> for UserAccount {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = <String as Deserialize>::deserialize(deserializer)?;
+        FromStr::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
```
sug0, line 753: replace with `namada::core::string_encoding::StringEncoded`, which uses `Display` and `FromStr` to encode some value with `serde`

sug0: `StringEncoded<UserAccount>` that is

brentstone: @murisi can we try to address any further items related to this PR in a follow-up PR? This has been merged to draft now.

murisi: Sure, understood. Thanks.

Issue #3440, opened 2024-06-25
Malleable MASP Witness Data


murisi:
The `MaspTx` section is identified by the TxId digest described https://zips.z.cash/zip-0244#txid-digest . This is problematic because the TxId digest does not cover witness data like zero-knowledge proofs, authorization signatures, and binding signatures meaning that those entries are malleable. We should probably hash MASP `Transaction` serialization bytes when trying to prevent tampering and the TxId digest to facilitate hardware wallet signing.

cwgoes:
Good catch. I agree, we need to commit to this data. We should have no malleable data in transactions at all, really, not that I can think of at least.

cwgoes:
@murisi This has since been fixed, right?

grarco:
Closed by #3463

PR #3463, opened 2024-06-28
Murisi/non malleable masp witness data2

murisi:
## Describe your changes
An attempt to address https://github.com/anoma/namada/issues/3440 . The MASP section hash is now based on the `Transaction`'s Borsh serialization instead of the TxId digest. However, unlike #3460 the `Transfer::shielded_section_hash` remains as a `TxId` and the `MaspBuilder::target` also remains as a `TxId`. This should make it easier for the hardware wallet to produce the wrapper signature without having to find and replace section hashes that have been invalidated after the insertion of the witness data like the binding signature and the authorization signatures into the MASP section.

Namada v0.39.0
Potentially supersedes #3460 


cwgoes (approved):

crates/sdk/src/tx.rs
```diff
@@ -3081,7 +3083,7 @@ pub async fn build_shielding_transfer<N: Namada>(
         source: source.clone(),
         token: args.token.clone(),
         amount: validated_amount,
-        shielded_section_hash: Hash::zero(),
+        shielded_section_hash: TxId::from_bytes(Default::default()).into(),
```
grarco, line 3086: Is it enough to use the `TxId` as the commitment to the shielded section? The hash of the `data` section ends up in the `TxCommitments` which in turn is part of the `raw_header_hash` that gets signed by the inner tx signer(s). If this is malleable then it's still possible to tamper with the masp section. Since the hash of the masp `Transaction` is now used when computing the hash of the masp section, once the wrapper has been constructed it's safe but it is still possible for the wrapper signer to modify that section before wrapping the inner batch. Of course this is a less sever issue than tampering with the wrapper since:

- The wrapper cannot be tampered with (as stated above) so the correctness of fee payment is preserved
- We can expect in most cases the wrapper signer to be the same entity that generated the inner batch
- Tampering with the proofs/signatures could only make the transaction fail (to my understanding), so essentially it turns the tx into a no-op since it's going to be rejected. This is exactly what already happens if the wrapper signer tampers with other sections of the inner tx (like code or data): the tx either fails during execution or is rejected by vps

So overall since to me like we are safe (as long as my assumptions that tampering with proofs/signatures can only make the tx fail) but I still wanted to bring up this topic before approving the pr

PR #3470, opened 2024-07-01
Draft+3459+3463+hermes

brentstone:


wasm/tx_ibc/src/lib.rs
```diff
@@ -3,51 +3,59 @@
...
+            .targets
+            .into_iter()
+            .map(|(account, amount)| {
+                ((account.owner, account.token), amount.amount())
+            })
+            .collect::<BTreeMap<_, _>>();
+
+        // Effect the multi transfer
+        token::multi_transfer(ctx, &sources, &targets)
+            .wrap_err("Token transfer failed")?;
```
yito88: This change wasn't included in #3459 or #3463. unexpected change?

murisi: You're right. I made this change when I was trying to merge #3459 and #3463 with the MASP fee payment changes (that were also touching IBC code). This code is basically an extension of what's already in `draft`.

yito88: ah, I see. The transfer has been added for the fee payment in #3393. However, `masp_fee_payment` was discarded.
```rust
    if let Some(UnshieldingTransferData {
        token,
        amount,
        target,
    }) = &masp_fee_payment
    {
        // Transparent unshield for fee payment
        token::transfer(
            ctx,
            &Address::Internal(address::InternalAddress::Masp),
            target,
            token,
            amount.amount(),
        )?;
    }
```

murisi: Yep, exactly. @grarco Let me know if you have any objections to this merge resolution. Essentially it removes the redundant `Transfer` (that isn't executed by `tx_ibc`) from `MsgTransfer` (and similar IBC messages).

grarco: In general it looks ok to me, since the new `Transfer` object is vectorized we can embed the fee payment directly inside it. But the `run_ledger_ibc_with_hermes` test is failing and it contains a masp fee payment over ibc so it's better if we check it

PR #3473, opened 2024-07-02
Avoid multiple calls to `process_proposal`

grarco:
## Describe your changes

This PR tries to improve the extra calls to process proposal requested by Namada (nothing changes for the requests coming from Comet). More specifically:

- The more complete `process_proposal` is now called instead of `process_txs`
- The `InMemory` struct has been expanded with an extra field that caches the requests to `process_proposal` for a given block so that future calls can be avoided. This cache is cleared when handling a `FinalizeBlock` request

`v0.40.0`

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -192,9 +192,10 @@ impl Shell {
         };
 
         if recheck_process_proposal {
-            let process_req = crate::shims::abcipp_shim_types::shim::request::finalize_block_to_process_proposal_req(
-                            finalize.clone(),
-                        ).map_err(|_| Error::InvalidBlockProposal)?;
+            let process_req = finalize
+                .clone()
+                .cast_to_process_proposal_req()
+                .map_err(|_| Error::InvalidBlockProposal)?;
             // FIXME: should instead mock a call to call? Probably yes
             // FIXME: need the processing results? Maybe for caching
             // even though caching here is useless
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -126,9 +126,27 @@ impl Shell {
             Request::ProcessProposal(block) => {
                 tracing::debug!("Request ProcessProposal");
                 // TODO: use TM domain type in the handler
-                let (response, _tx_results) =
-                    self.process_proposal(block.into());
-                // FIXME: here set the flag
+                // NOTE: make sure to put any checks inside process_proposal
+                // since that function is called in other places to rerun the
+                // checks if (when) needed. Every check living outside that
+                // function will not be correctly replicated in the other
+                // locations
+                let (response, tx_results) =
+                    self.process_proposal(block.clone().into());
+                // Cache the response in case of future calls from Namada. If
+                // hash conversion fails avoid caching
+                if let Ok(block_hash) = block.hash.try_into() {
+                    self.state.in_mem_mut().process_proposal_cache.insert(
+                        block_hash,
+                        (
+                            response,
+                            tx_results
+                                .into_iter()
+                                .map(|res| res.into())
+                                .collect(),
+                        ),
+                    );
+                }
                 Ok(Response::ProcessProposal(response))
             }
             Request::RevertProposal(_req) => {
@@ -172,9 +190,9 @@ impl Shell {
     }
 
     // Checks if a run of process proposal is required before finalize block
-    // (recheck) and, in case, performs it
+    // (recheck) and, in case, performs it. Clears the cache before returning
     fn try_recheck_process_proposal(
-        &self,
+        &mut self,
         finalize_req: &shims::abcipp_shim_types::shim::request::FinalizeBlock,
     ) -> Result<(), Error> {
         let recheck_process_proposal = match self.mode {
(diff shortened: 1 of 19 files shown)


crates/node/src/shell/finalize_block.rs
```diff
@@ -1850,7 +1851,10 @@ mod test_finalize_block {
             let req = FinalizeBlock {
                 txs,
                 proposer_address: proposer_address.clone(),
-                votes: votes.clone(),
+                decided_last_commit: tendermint::abci::types::CommitInfo {
+                    round: 0u8.into(),
```
sug0, line 1855: rather than passing 0 here, shouldn't we be passing the round value contained in the request?
https://docs.rs/tendermint/0.37.0/tendermint/abci/request/struct.FinalizeBlock.html#structfield.decided_last_commit

crates/node/src/lib.rs
```diff
@@ -128,8 +126,27 @@ impl Shell {
...
+                // NOTE: make sure to put any checks inside process_proposal
+                // since that function is called in other places to rerun the
+                // checks if (when) needed. Every check living outside that
+                // function will not be correctly replicated in the other
+                // locations
+                let (response, tx_results) =
+                    self.process_proposal(block.clone().into());
+                // Cache the response in case of future calls from Namada. If
+                // hash conversion fails avoid caching
+                if let Ok(block_hash) = block.hash.try_into() {
```
sug0, line 138: copy/clone the block hash before to avoid cloning the entire block. that's the only field required from `block` on line 138

crates/node/src/shims/abcipp_shim.rs
```diff
@@ -114,43 +113,89 @@ impl AbcippShim {
...
+                                        process_resp.to_owned(),
+                                        res.iter()
+                                            .map(|res| res.to_owned().into())
+                                            .collect(),
+                                    ),
+                                    None => {
+                                        // Need to run process proposal to
+                                        // extract the data we need
+                                        // for finalize block (tx results)
+                                        let process_req = super::abcipp_shim_types::shim::request::CheckProcessProposal::from(begin_block_request.clone()).cast_to_tendermint_req(self.delivered_txs.clone());
```
sug0, line 142: can we trry to fix formatting here? it's all fucked

crates/node/src/shims/abcipp_shim.rs
```diff
@@ -114,43 +113,89 @@ impl AbcippShim {
...
+                                    Some((process_resp, res)) => (
+                                        // We already have the result of
+                                        // process proposal for this block
+                                        // cached in memory
+                                        process_resp.to_owned(),
+                                        res.iter()
+                                            .map(|res| res.to_owned().into())
+                                            .collect(),
+                                    ),
+                                    None => {
```
sug0, line 138: so, as far as I can tell, the only reason why a process proposal wouldn't be in cache is because you shut down your node, clearing it from memory, right? under normal circumstances, you always receive a `ProcessProposal` request before you receive an `EndBlock` request. under these assumptions, you would never/seldom hit the `None` branch

crates/node/src/shims/abcipp_shim.rs
```diff
@@ -114,43 +113,89 @@ impl AbcippShim {
...
+
+                                        (process_resp, res)
+                                    }
+                                }
+                            }
+                            Err(_) => {
+                                // Need to run process proposal to extract the
+                                // data we need
+                                // for finalize block (tx results)
+                                let process_req = super::abcipp_shim_types::shim::request::CheckProcessProposal::from(begin_block_request.clone()).cast_to_tendermint_req(self.delivered_txs.clone());
```
sug0, line 172: again, can we try to fix formatting here?

sug0 (changes requested):

crates/node/src/shell/finalize_block.rs
grarco: This is just a unit test, we don't have a real request from which we can extract that information

crates/node/src/shims/abcipp_shim.rs
grarco: I believe there's one case in which a node could receive an `EndBlock` call without having the result cached. It's explained here in the comet docs https://docs.cometbft.com/v0.37/spec/abci/abci++_example_scenarios#scenario-3

crates/node/src/lib.rs
grarco: Fixed in e2132f3c74d07884730e4a63a97b495022ae5228

crates/node/src/shims/abcipp_shim.rs
grarco: Done in 33a4049e0b0b87b6fd68b560d3d2e3a43663ea8e

crates/node/src/shims/abcipp_shim.rs
grarco: Done in 33a4049e0b0b87b6fd68b560d3d2e3a43663ea8e

crates/node/src/shims/abcipp_shim.rs
sug0: ah right, if the system is ahead of you

sug0 (approved):
lgtm!

PR #3490, opened 2024-07-05
Fix IBC MASP tx extraction

### LINKED ISSUE #3488: Shielded action events
> Actually, I believe there's an issue, my bad for not spotting it earlier. In the sdk we use `extract_masp_tx_from_ibc_message` when we have a valid masp tx which is missing the masp ref event. In this case we iterate over all the inner txs of the batch and try to extract the ibc message from the tx data and if we can we push the masp `Transaction` to the result.
> The issue is that we don't know if that specific inner tx was successful: if it wasn't, the client (or indexer) extracts information from a wrong source and spoils the internal state of the `ShieldedContext`, which might prevent it from constructing valid transactions in the future.
> We need a way to tell, in the events we emit, which transactions of the batch are actually valid and filter only these ones in `extract_masp_tx_from_ibc_message`

_Originally posted by @grarco in https://github.com/anoma/namada/issues/3444#issuecomment-2209266402_

On top of this we should also handle masp fee payment coming from IBC in `apply_wrapper_tx` https://github.com/anoma/namada/blob/5cc23b4f43b12ecfc70eeca8d04cbe15f4b11a3a/crates/namada/src/ledger/protocol/mod.rs#L403

something similar to what we do here in the masp fee payment pr https://github.com/anoma/namada/blob/e1b5857845bafa1be9f4ee4500b5544b56bf926f/crates/namada/src/ledger/protocol/mod.rs#L463-L475

So to recap we need to:

1. mark, in protocol, if an ibc transaction was a successful masp tx by emitting some sort of event that points to this tx in the batch (we already have the event to point to the transaction index in the block). The pointer could be something like the inner tx index in the batch or the hash of the `TxCommitments`
2. update `apply_wrapper_tx` to include in its return type `ExtendedTxResult` the pointer to the first transaction in the batch if it was a valid masp over ibc tx that paid fees. This will also require a change to `try_masp_fee_payment` to return the pointer (at the moment it returns the hash of the section but we need something different for an incoming shielded action, as mentioned above)
3. update the sdk to only look for the inner txs of a batch that have been marked as successful. This is done by looking at the emitted events as we already do for masp txs  and outgoing shielded actions

yito88:
## Describe your changes
closes #3488 

`v0.40.0`

### DIFF
--- a/crates/events/src/extend.rs
+++ b/crates/events/src/extend.rs
@@ -519,9 +519,9 @@ impl EventAttributeEntry<'static> for MaspTxBatchRefs {
 }
 
 /// Extend an [`Event`] with data sections for IBC shielding transfer.
-pub struct IbcTxBatchRefs(pub IbcTxDataRefs);
+pub struct IbcMaspTxBatchRefs(pub IbcTxDataRefs);
 
-impl EventAttributeEntry<'static> for IbcTxBatchRefs {
+impl EventAttributeEntry<'static> for IbcMaspTxBatchRefs {
     type Value = IbcTxDataRefs;
     type ValueOwned = Self::Value;
 
--- a/crates/events/src/extend.rs
+++ b/crates/events/src/extend.rs
@@ -525,7 +525,7 @@ impl EventAttributeEntry<'static> for IbcMaspTxBatchRefs {
     type Value = IbcTxDataRefs;
     type ValueOwned = Self::Value;
 
-    const KEY: &'static str = "ibc_tx_batch_refs";
+    const KEY: &'static str = "ibc_masp_tx_batch_refs";
 
     fn into_value(self) -> Self::Value {
         self.0
--- a/crates/events/src/extend.rs
+++ b/crates/events/src/extend.rs
@@ -8,6 +8,7 @@ use std::str::FromStr;
 use namada_core::address::Address;
 use namada_core::collections::HashMap;
 use namada_core::hash::Hash;
+use namada_core::ibc::IbcTxDataRefs;
 use namada_core::masp::MaspTxRefs;
 use namada_core::storage::{BlockHeight, TxIndex};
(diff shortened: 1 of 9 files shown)


crates/namada/src/ledger/protocol/mod.rs
```diff
@@ -740,16 +765,20 @@ where
...
                 }
 
-                let masp_ref = namada_tx::action::get_masp_section_ref(*state)
-                    .map_err(Error::StateError)?;
+                let masp_section_ref =
+                    match namada_tx::action::get_masp_section_ref(*state)
+                        .map_err(Error::StateError)?
+                    {
+                        Some(masp_tx_id) => Either::Left(masp_tx_id),
+                        None => Either::Right(*first_tx.cmt.data_sechash()),
```
grarco, line 773: Should we check if the ibc action has been written in this case to be slightly more safe?

crates/events/src/extend.rs
```diff
@@ -517,6 +518,20 @@ impl EventAttributeEntry<'static> for MaspTxBatchRefs {
     }
 }
 
+/// Extend an [`Event`] with data sections for IBC shielding transfer.
+pub struct IbcTxBatchRefs(pub IbcTxDataRefs);
```
grarco, line 522: Could we rename to something slightly different? This is used to notify the presence of masp over ibc transactions, not ibc txs in general

crates/namada/src/ledger/protocol/mod.rs
yito88: fixed to check it: 45d2798

crates/events/src/extend.rs
yito88: Agree. Renamed to `IbcMaspTxBatchRefs`: 45d2798

PR #3498, opened 2024-07-10
MASP shielded sync client phase 1

sug0:
## Describe your changes

Supersedes #3405, #3426 (rebase of the last one on the latest release)

Implements phase 1 of Issue #3385 
 - When fetching notes, connections and related failures should not halt shielded sync. Instead, the process should be restarted
 -   If fetching is interrupted, the data fetched should be persisted locally so that progress isn't lost
 -  A trait for fetching behavior should be added to provide modularity 

`v0.40.0`

### DIFF
--- a/crates/apps_lib/src/client/masp.rs
+++ b/crates/apps_lib/src/client/masp.rs
@@ -70,6 +70,8 @@ where
     io: &'io IO,
     r#type: ProgressType,
     peeked: Option<T>,
+    #[cfg(not(unix))]
+    num_logs_counter: usize,
 }
 
 impl<'io, T, I, IO> LoggingIterator<'io, T, I, IO>
@@ -95,6 +97,8 @@ where
             io,
             r#type,
             peeked: None,
+            #[cfg(not(unix))]
+            num_logs_counter: 0,
         }
     }
 
@@ -124,6 +128,14 @@ where
         self.peek();
         let next_item = self.peeked.take()?;
         self.advance_index();
+
+        #[cfg(not(unix))]
+        {
+            if self.num_logs_counter % 20 != 0 {
+                return Some(next_item);
+            }
+        }
+
         let (index, length) = {
             let locked = self.progress.lock().unwrap();
             (locked.length, locked.index)
@@ -132,7 +144,12 @@ where
         let percent = std::cmp::min(100, (100 * index) / length);
         let completed: String = vec!['#'; percent].iter().collect();
         let incomplete: String = vec!['.'; 100 - percent].iter().collect();
-        display_line!(self.io, "\x1b[2A\x1b[J");
+
+        #[cfg(unix)]
+        {
+            clear_last_lines(self.io, 2);
+        }
+
         match self.r#type {
             ProgressType::Fetch => display_line!(
                 self.io,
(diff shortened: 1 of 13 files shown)


grarco (approved):

crates/apps_lib/src/client/masp.rs
```diff
@@ -28,137 +30,209 @@ pub async fn syncing<
...
+    I: Iterator<Item = T>,
+    IO: Io,
+{
+    fn new(
+        items: I,
+        io: &'io IO,
+        r#type: ProgressType,
+        progress: Arc<Mutex<IterProgress>>,
+    ) -> Self {
+        let (size, _) = items.size_hint();
```
tzemanovic, line 87: 1. the first thing is a lower bound, does that matter in this context?
2. does it actually work? It looks like the `flume::Receiver` that we wrap in this doesn't override the default impl (which is `(0, None)`)

crates/apps_lib/src/client/masp.rs
```diff
@@ -28,137 +30,209 @@ pub async fn syncing<
...
-            self.items = {
-                let mut new_items = vec![];
-                std::mem::swap(&mut new_items, &mut self.items);
-                new_items.into_iter().rev().collect()
-            };
+    fn advance_index(&mut self) {
+        let mut locked = self.progress.lock().unwrap();
+        locked.index += 1;
+        if let ProgressType::Scan = self.r#type {
+            locked.length = self.items.size_hint().0;
```
tzemanovic, line 105: also here

crates/apps_lib/src/client/masp.rs
```diff
@@ -28,137 +30,209 @@ pub async fn syncing<
...
+        self.advance_index();
+        let (index, length) = {
+            let locked = self.progress.lock().unwrap();
+            (locked.length, locked.index)
+        };
+
+        let percent = std::cmp::min(100, (100 * index) / length);
         let completed: String = vec!['#'; percent].iter().collect();
         let incomplete: String = vec!['.'; 100 - percent].iter().collect();
         display_line!(self.io, "\x1b[2A\x1b[J");
```
tzemanovic, line 135: it would be nice to use something more readable and portable for the escape code - we have `owo-colors` in deps

crates/sdk/src/masp.rs
```diff
@@ -3434,66 +3561,482 @@ pub mod fs {
...
+        test_client, TestUnscannedTracker, TestingMaspClient,
+    };
+    use crate::masp::utils::{DefaultTracker, ProgressTracker, RetryStrategy};
+
+    // A viewing key derived from A_SPENDING_KEY
+    pub const AA_VIEWING_KEY: &str = "zvknam1qqqqqqqqqqqqqq9v0sls5r5de7njx8ehu49pqgmqr9ygelg87l5x8y4s9r0pjlvu6x74w9gjpw856zcu826qesdre628y6tjc26uhgj6d9zqur9l5u3p99d9ggc74ald6s8y3sdtka74qmheyqvdrasqpwyv2fsmxlz57lj4grm2pthzj3sflxc0jx0edrakx3vdcngrfjmru8ywkguru8mxss2uuqxdlglaz6undx5h8w7g70t2es850g48xzdkqay5qs0yw06rtxcpjdve6";
+
+    /// A serialized transaction that will work for testing.
+    /// Would love to do this in a less opaque fashion, but
+    /// making these things is a misery not worth my time.
```
tzemanovic, line 3591: :D it would be good to note how this was generated/obtained so we don't carry the misery into maintenance

crates/apps_lib/src/client/masp.rs
sug0: the `size_hint` function implemented on exact sized iterators (like ranges) is always the proper size of the iterator. we wrap two iterators with this `ProgressTracker` abstraction: one that is a [`Range`](https://doc.rust-lang.org/nightly/std/ops/struct.Range.html) and another that is our custom `FetchQueueReceiver` type. the latter implements `size_hint` based on the number of items [that are left to consume on the `flume` channel](https://docs.rs/flume/latest/flume/struct.Receiver.html#method.len). the values returned by this implementation will be a bit racy, but never 0, unless the channel is empty. in the end, we only care about giving some notion of progress to CLI users.

grarco (approved):

tzemanovic (approved):

PR #3510, opened 2024-07-12
Update gas for storage occupation

grarco:
## Describe your changes

Increases the gas cost component for storage occupation and ties this cost the other ones so that it's not free-floating anymore (which could cause a divergence if the other costs are changed).

Removes the redundant `new_from_sub_limit` constructor for `TxGasMeter`.

#3428 (diff for review: https://github.com/anoma/namada/pull/3510/files/209da0c6ad9cd456a5a6634ec25cc45408e0b72c..33a6cc9437692ccb1ef24e58f2536ac2dfdeb8f3)


grarco:
Let's wait for the CI to come back online before reviewing this cause there's a change we might have a few tests failing due to gas

crates/gas/src/lib.rs
```diff
@@ -54,8 +54,14 @@ const COMPILE_GAS_PER_BYTE: u64 = 1_955;
...
 const WRAPPER_TX_VALIDATION_GAS: u64 = 3_245_500;
+// There's no benchmark to calculate the cost of storage occupation, so we
+// define it as the cost of storage latency (which is needed for any storage
+// operation and it's based on actual execution time), plus the same cost
+// multiplied by an arbitrary factor that represents the higher cost of storage
+// space as a resource. This way, the storage occupation cost is not completely
+// free-floating but it's tied to the other costs
 const STORAGE_OCCUPATION_GAS_PER_BYTE: u64 =
-    100_000 + PHYSICAL_STORAGE_LATENCY_PER_BYTE;
+    PHYSICAL_STORAGE_LATENCY_PER_BYTE * (1 + 10);
```
brentstone, line 64: why did you write the argument as `(1 + 10`? Do each of these numbers account for some different effect?

brentstone (approved):
LGTM, modulo response to the `(1 + 10)` comment

crates/gas/src/lib.rs
grarco: They do: the 1 represent the actual cost of storage latency, the 10 component instead is just an arbitrary multiplier that I use to compute an actual storage occupation gas cost based on the storage latency. They are not really tied but we cannot leave the storage occupation gas cost free-floating because we'd end up with the same issue this Pr tries to solve. So we need a way to tie that cost to some other gas cost that is actually based on a benchmark and the storage latency is the closest one

brentstone: cool would you mind describing this a bit in a comment / docstring?

grarco: It's described right above

brentstone: Ok I think I was having trouble understanding the comment at first and was seeking clarity, but I think it's fine now. Thanks.

PR #3516, opened 2024-07-16
Eliminates MASP VP's requirement for all debited accounts to sign Tx

murisi:
## Describe your changes
This PR attempts to eliminate the MASP VP's requirement (mentioned in https://github.com/anoma/namada/pull/3372#issuecomment-2225421559) for all debited accounts to sign a `Tx`. This is done by requiring only those accounts whose balances are modified by the inner `Transaction` and whose balances decrease to sign the transaction. The justification for this change is that in the case of accounts not modified by the inner `Transaction`, other VPs already validate all changes pertaining to the given address.

Namada 0.40.0

### DIFF
--- a/.changelog/unreleased/improvements/3516-reduce-masp-vp-signatures.md
+++ b/.changelog/unreleased/improvements/3516-reduce-masp-vp-signatures.md
@@ -1,0 +1,2 @@
+- Eliminates the MASP VPs requirement for all debited accounts to sign a Tx.
+  ([\#3516](https://github.com/anoma/namada/pull/3516))
--- a/crates/namada/src/ledger/native_vp/masp.rs
+++ b/crates/namada/src/ledger/native_vp/masp.rs
@@ -687,33 +687,26 @@ where
             .read_post(&counterpart_balance_key)?
             .unwrap_or_default();
         // Public keys must be the hash of the sources/targets
-        let address_hash = addr_taddr(counterpart.clone());
+        let addr_hash = addr_taddr(counterpart.clone());
         // Enable the decoding of these counterpart addresses
         result
             .decoder
-            .insert(address_hash, TAddrData::Addr(counterpart.clone()));
+            .insert(addr_hash, TAddrData::Addr(counterpart.clone()));
+        let zero = ValueSum::zero();
         // Finally record the actual balance change starting with the initial
         // state
-        let pre_entry = result
-            .pre
-            .get(&address_hash)
-            .cloned()
-            .unwrap_or(ValueSum::zero());
+        let pre_entry = result.pre.get(&addr_hash).unwrap_or(&zero).clone();
         result.pre.insert(
-            address_hash,
+            addr_hash,
             checked!(
                 pre_entry + &ValueSum::from_pair((*token).clone(), pre_balance)
             )
             .map_err(native_vp::Error::new)?,
         );
         // And then record thee final state
-        let post_entry = result
-            .post
-            .get(&address_hash)
-            .cloned()
-            .unwrap_or(ValueSum::zero());
+        let post_entry = result.post.get(&addr_hash).cloned().unwrap_or(zero);
         result.post.insert(
-            address_hash,
+            addr_hash,
             checked!(
                 post_entry
                     + &ValueSum::from_pair((*token).clone(), post_balance)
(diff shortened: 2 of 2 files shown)


grarco:
Is this still required?

https://github.com/anoma/namada/blob/17b9fcfa9c85f5a115646120f6595e8cf28c38a7/crates/namada/src/ledger/native_vp/masp.rs#L975-L977

murisi:
> Is this still required? I think with the new logic we don't need this: in fact, this might require some additional authorizers which are not needed
> 
> https://github.com/anoma/namada/blob/17b9fcfa9c85f5a115646120f6595e8cf28c38a7/crates/namada/src/ledger/native_vp/masp.rs#L975-L977

I am not sure if the VP is as correct if we remove this insertion code. It might be, but I'm not sure yet. How are you reasoning about this? My thinking is as follows: so currently the overall logic is to validate a `Tx` by splitting it into two parts: the `Transaction`, and `Tx - Transaction` and validating those separately/independently - if both parts are correct then so is the whole. The signature verifications in this question are a part of validating the `Transaction` part - essentially we're making sure that all the transparent input addresses (let's say that an Albert is amongst them) in the `Transaction` have signed (directly or indirectly) over the entire `Transaction` object. Are there conditions under which we can avoid doing this check? I guess if the overall state change being validated by the MASP VP resulted in Albert's balance decreasing, then maybe we could avoid checking for signatures since Albert's VP is triggered to check the balance change anyway. But it seems the assumption here would be that Albert's VP (whether it's user, implicit, or something else) would always run logic to check that there's a signature (directly or indirectly) over the `Transaction` object. Am I missing something here?

What issues come up when integrating this PR with outsourcing? My thinking is that the above insertion is only triggered when a `Transaction` has transparent inputs (say Albert is amongst them). This means that either the overall state change being validated by the MASP VP debits Albert's account, or the `Transaction` has transparent outputs going to Albert, or the overall `Tx` contains some other credits to Albert that cancels out the transparent input. I want to think that the latter two cases are rare and can always be simplified (by netting) so that Albert exclusively either being debited or credited in a `Tx`, but not both. In the former case (where the overall state change is a debit to Albert), I would think that your logic that automatically includes all debited accounts in the set of MASP authorizers (here https://github.com/anoma/namada/blob/cf9942a9ec4c9d965b052cedea190abf5d2331d2/wasm/tx_transfer/src/lib.rs#L36) would ensure that `MaspAuthorizer` actions are pushed and verifiers inserted for each transparent input in a `Transaction`. But I might also be missing something here... Also, thanks again for integrating these 2 PRs, let me know if I can assist in any other way.

grarco:
> > Is this still required? I think with the new logic we don't need this: in fact, this might require some additional authorizers which are not needed
> > https://github.com/anoma/namada/blob/17b9fcfa9c85f5a115646120f6595e8cf28c38a7/crates/namada/src/ledger/native_vp/masp.rs#L975-L977
> 
> I am not sure if the VP is as correct if we remove this insertion code. It might be, but I'm not sure yet. How are you reasoning about this? My thinking is as follows: so currently the overall logic is to validate a `Tx` by splitting it into two parts: the `Transaction`, and `Tx - Transaction` and validating those separately/independently - if both parts are correct then so is the whole. The signature verifications in this question are a part of validating the `Transaction` part - essentially we're making sure that all the transparent input addresses (let's say that an Albert is amongst them) in the `Transaction` have signed (directly or indirectly) over the entire `Transaction` object. Are there conditions under which we can avoid doing this check? I guess if the overall state change being validated by the MASP VP resulted in Albert's balance decreasing, then maybe we could avoid checking for signatures since Albert's VP is triggered to check the balance change anyway. But it seems the assumption here would be that Albert's VP (whether it's user, implicit, or something else) would always run logic to check that there's a signature (directly or indirectly) over the `Transaction` object. Am I missing something here?
> 
> What issues come up when integrating this PR with outsourcing? My thinking is that the above insertion is only triggered when a `Transaction` has transparent inputs (say Albert is amongst them). This means that either the overall state change being validated by the MASP VP debits Albert's account, or the `Transaction` has transparent outputs going to Albert, or the overall `Tx` contains some other credits to Albert that cancels out the transparent input. I want to think that the latter two cases are rare and can always be simplified (by netting) so that Albert is exclusively either being debited or credited in a `Tx`, but not both. In the former case (where the overall state change is a debit to Albert), I would think that your logic that automatically includes all debited accounts in the set of MASP authorizers (here
> 
> https://github.com/anoma/namada/blob/cf9942a9ec4c9d965b052cedea190abf5d2331d2/wasm/tx_transfer/src/lib.rs#L36
> 
> ) would ensure that `MaspAuthorizer` actions are pushed and verifiers inserted for each transparent input in a `Transaction`. But I might also be missing something here... Also, thanks again for integrating these 2 PRs, let me know if I can assist in any other way.

I believe your reasoning is correct, never mind. What happened was that in the rebase branch I've removed the commit that returns the debited accounts from `multi_transfer` because I thought we didn't need it since we don't want to push actions for all the debited accounts but only for those that have also been touched by the masp `Transaction` in a way that doesn't match the storage modification. I believe this is always the case in our tests (indeed if I remove the call to `signer.insert` the tests pass). In these cases we don't need the actions because the debit in the masp `Transaction` exactly matches the debit in the storage changes so the source's VP will be triggered anyway (this is what I meant with "additional authorizers" in my original comment). In our transactions we never use intermediaries so the masp actions should never be required.

Thinking twice about it though, as you say, I'm not sure this is always the case. Actually by running a simple calculation on a case where the shielding is redirected (someone shields on behalf of the original sender) the checks don't run for neither the original sender nor the new one if we remove that line. Now I'm not sure this is needed cause I can't see anyone tampering with a transaction to shield tokens on behalf of someone else and also we usually only requires extra validation for debit but I think keeping it is a safe fallback (also, we still want to prevent any sort of tampering)

grarco (approved):

PR #3527, opened 2024-07-18
Murisi/borsh ibc msg transfer

### LINKED ISSUE #3501: Use a tag/discriminant to distinguish between IBC NFT transfers and fungible token transfers
Parsing IBC transfers is a task that was completed in the context of #3494. Right now it seems that the best way for a hardware wallet to parse a `TX_IBC_WASM` transaction is to attempt to parse it as a `MsgTransfer`, and if that fails then to attempt to parse it as a `MsgNftTransfer`. Is this the case? If so, then I guess the assumption here is that there is no byte sequence that can be successfully interpreted as both a `MsgTransfer` and a `MsgNftTransfer`? If so, then it might be simpler/more robust to just use two separate WASMs for these two transaction types, or use the same WASM and embed in the Data section a tagged enumeration instance that can either be inhabited by a `MsgTransfer` or `MsgNftTransfer`. In this way, the errors reported by a parser could also be more precise.

murisi:
## Describe your changes
Attempt to close #3501 and hence simplify the (de)serialization of IBC transactions for third parties. More specifically, the following changes have been made:
* IBC transaction data is now always a tagged serialization of the `IbcMessage` enumeration instead of being an untagged serialization of a `MsgTransfer`, `MsgNftTransfer` or `Any` structure.
* IBC transaction data is now encoded using Borsh serialization instead of Protobufs. This is more consistent with how all the other parts of a `Tx` are serialized using Borsh, and will ease schema generation.
* Updated the test vector generation code (i.e. the generator and the vector printer) to produce transactions in accordance with the above two changes.

Namada 0.40.0


yito88 (commented):
Looks good. Thank you! I left comments for minor stuff.

murisi:
Closed in favour of the rebase in #3560 .

PR #3529, opened 2024-07-18
workaround wasmer leak

tzemanovic:
## Describe your changes

workaround for https://github.com/wasmerio/wasmer/issues/4377 and https://github.com/wasmerio/wasmer/issues/4949

v0.40.0

### DIFF
--- a/crates/namada/src/vm/wasm/compilation_cache/common.rs
+++ b/crates/namada/src/vm/wasm/compilation_cache/common.rs
@@ -36,6 +36,15 @@ pub struct Cache<N: CacheName, A: WasmCacheAccess> {
     name: PhantomData<N>,
     /// Cache access level
     access: PhantomData<A>,
+    /// Wasmer store - The store that's used to compile the modules.
+    // N.B.: This has to be kept alive to avoid segfaults when running them
+    // from cache (otherwise `wasmer_compiler::FRAME_INFO` gets cleared out
+    // when the `Store` is dropped and the next run of this `Module`
+    // doesn't re-instantiate it and crashes when it tries to access it).
+    // Related issues:
+    // - https://github.com/wasmerio/wasmer/issues/4377
+    // - https://github.com/wasmerio/wasmer/issues/4949
+    store: Arc<Store>,
 }
 
 /// This trait is used to give names to different caches
@@ -100,6 +109,7 @@ impl<N: CacheName, A: WasmCacheAccess> Cache<N, A> {
             in_memory,
             name: Default::default(),
             access: Default::default(),
+            store: Arc::new(store()),
         }
     }
 
@@ -112,9 +122,12 @@ impl<N: CacheName, A: WasmCacheAccess> Cache<N, A> {
         code_hash: &Hash,
     ) -> Result<Option<(Module, Store)>, wasm::run::Error> {
         if A::is_read_write() {
-            self.get(code_hash)
+            let module = self.get(code_hash)?;
+            Ok(module.map(|module| (module, store())))
         } else {
-            self.peek(code_hash)
+            let store = store();
+            let module = self.peek(code_hash, &store)?;
+            Ok(module.map(|module| (module, store)))
         }
     }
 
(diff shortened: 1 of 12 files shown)


Fraccaman:
tested on a internal network, can't reproduce the memory leak anymore 👍

crates/namada/src/vm/wasm/compilation_cache/common.rs
```diff
@@ -36,6 +36,12 @@ pub struct Cache<N: CacheName, A: WasmCacheAccess> {
     name: PhantomData<N>,
     /// Cache access level
     access: PhantomData<A>,
+    /// Wasmer store - The store that's used to compile the modules has to be
+    /// kept alive to avoid segfaults when running them from cache (otherwise
+    /// `wasmer_compiler::FRAME_INFO` gets cleared out when the `Store` is
+    /// dropped and the next run of this `Module` doesn't re-instantiate it and
+    /// crashes when it tries to access it)
+    store: Arc<Store>,
```
sug0, line 44: According to https://github.com/wasmerio/wasmer/issues/4949#issuecomment-2245453917, modules from different stores should not be interchanged. I'm not sure what the consequences are... but we should watch out for weird behavior stemming from this.

sug0: also a minor nit is, the majority of this text is comprised of details that don't belong in a docstring IMO. part of it can be sliced out into a regular comment prefixed by `NB`, `NOTE`, or something along these lines

crates/namada/src/vm/wasm/run.rs
```diff
@@ -2117,4 +2195,22 @@ mod tests {
...
+    struct ModuleInfoFrameInfo {
+        _start: usize,
+        _functions: BTreeMap<usize, FunctionInfo>,
+        _module: std::sync::Arc<wasmer_types::ModuleInfo>,
+        _frame_infos: wasmer_compiler::FrameInfosVariant,
+    }
+    struct FunctionInfo {
+        _start: usize,
+        _local_index: wasmer_types::LocalFunctionIndex,
+    }
```
sug0, line 2215: I'm fine with this hack, but be mindful that these aren't part of the public API. in cargo, we set the `wasmer` version to `4.3.5`, however, I think the underlying version may change depending on the interaction with other dependencies. maybe this isn't the case for us since we commit the cargo lock file... anyway, the point is, upstream is free to change the structure of `GlobalFrameInfo`

crates/namada/src/vm/wasm/compilation_cache/common.rs
tzemanovic: yeah, I think we have to live with it until it's fixed upstream as:

- using the same store everywhere leads to a leak
- using a new store without keeping this one in cache leads to invalid memory access segfault

From what I can see the consequence is that the store in cache will never be cleared, but because we only allow a limited subset of wasms to be compiled and cached we should be fine for now. In dry-run txs the read-only cache access doesn't use this store from cache so it should not be possible to grow it from it.

tzemanovic: ah, actually, read-only cache access does use this `store` - let me fix that

crates/namada/src/vm/wasm/run.rs
tzemanovic: dully noted in docstring above :) I think there's no danger in a test - if the api changes it's only going to blow up in a test

crates/namada/src/vm/wasm/compilation_cache/common.rs
tzemanovic: added
- [32b2682](https://github.com/anoma/namada/pull/3529/commits/32b268211c5e2048bcfcde775fa6884cd0c9a7c9) to fix the read-only cache to use a temp store
- [163d393](https://github.com/anoma/namada/pull/3529/commits/163d393266b336726e4595e164b81b876d15cd22) to fix the comment

sug0 (approved):

Issue #2055, opened 2023-10-26
Check & confirm important values


grarco:
**⚠️ REMINDER ⚠️**

Make sure to review and confirm the values of:

- [ ] gas costs (#3554 )
- [ ] constants
- [ ] protocol parameters
- [ ] Rewards and inflation parameters

PR #3554, opened 2024-07-25
Update gas costs

### LINKED ISSUE #3588: Gas metering inconsistency
In #3554 we are seeing some inconsistency in the gas used by transactions. Specifically, it looks like a shielded tx costs less then a transparent transfer, even if we know from benchmarks that this shouldn't be the case.

## The issue

The issue is simple: the gas cost of of a shielded transaction is lower than that of a transparent transfer. From old benchmarks we know this is not the case since the crypto checks done by the masp vp are possibly the slowest checks to run in the entire protocol. By running all the transfer transactions we see that the gas cost end up like this:

`Shielding > Transparent ~= Unshielding > Shielded`

We can see that the transparent transaction is the second most expensive even though we'd expect it to be the cheapest. Note that these are the results of running the entire flow (tx + vps). If we stick to the tx only (avoiding the vps) we see the following:

`Unshielding > Shielded > Shielding > Transparent`

Which is indeed what we expect based on the old benchmarks. So the issue must be somewhere else: in the operations that precede the transaction (compilation), in the vps, a combination of them or  in the balance between the costs of transactions and the other operations.

We can see that something strange is happening with the vps as the costs are the following:

Masp vp: ~ 57 millions
Vp implicit: ~ 800 millions
Tx transfer: ~900 millions

From this we can see that the costs for wasm codes (both txs and vps) are completely out of scale compared with native code (masp vp), which we know should be the most expensive piece of code to run.

We report here the results of different causes taken into consideration.

## Host functions

The first idea was that host functions/ancillary operations (such as compilation) where charging wrong amount of gas. From some tests we can see that:

- Host functions per se have a minimal impact on the total gas costs of txs/vps, exception made for storage writes, but again, the majority of tx costs still comes from the wasm opcodes themselves
- Compilation costs are indeed the majority of the cost for transactions/wasm vps. Unfortunately, from some timings done on a node, we can confirm that these operations are indeed the slowest (most expensive) even when compared to the masp vp. We could think about not charging gas for compilation but this doesn't resolve the underlying issue

The actual gas costs for just the wasm opcodes (no compilation/host fns) are:

Vp implicit: ~ 215 millions
Tx transfer: ~ 190 millions

I've also tried to remove the costs for local function calls and memory growth without any noticeable change.

## Wrong benchmarks

Benchmarks have been reviewed to check for bugs. It doesn't seem like anything is out of place, the results are sensible both between themselves (masp is the most expensive, storage writes cost more than read) and, for wasm opcodes, compared to the expected machine behavior. For example, the benchmark for an `i32.add` results in 0.4 nanoseconds, which for a 2GHz machine translates to ~1 cycle which is indeed the expected behavior on a x86 machine. Also our costs tend to be close to those that can be found here (https://github.com/dfinity/ic/blob/master/rs/execution_environment/benches/wasm_instructions/WASM_BENCHMARKS.md).

Different setups for criterion have been tried, including reducing the warmup time, but the variation in the benchmarks results was minimal.

Also, given the very low timing of wasm opcodes, the mechanism to translate benchmarks timings to gas units involve some ceiling on the lowest value and approximations: I've tried to reduce these approximation with no major changes in the gas outcome, which demonstrates that those approximation cannot be the cause of our issues.

To recap, benchmarks seem to not be the issue. 

## Bugs in the runtime gas injector

Another idea was that there could be bugs in the tool that injects the gas costs in the wat file: more specifically this tool could be injecting more gas than needed leading to the huge costs we are seeing.

By looking at the instrumented wat files the injections look good, they are in the correct places and the costs they charge is correct: specifically the transfer wat counts 7044 injections and vp implicit counts 8365 injections. One strange thing here is that if we use constant gas costs for every opcodes instead of the codes we get from our benchmarks, the number of injections increase to 8589: I don't have an answer for this at the moment.

I was actually able to find one bug: the injector is supposed to divide the execution flow of the program into blocks and charge the gas for a single block when the block is entered. I was able to find a case in which case is done in the wrong way: the injector  doesn't consider function calls as a separate block and would inject gas for the function call and the code following it: this is not correct because the callee could trap leading to the following code being skipped. In our case we had a big cost for the `unreachable` opcode and in our vps/txs it happens that this opcode follows function calls which means that it gets accounted even if it shouldn't. I set it costs to 0 (since unreachable means breaking out of the execution anyway) and could reduce the costs illustrated above to the followings:

Masp vp: ~ 57 millions
Vp implicit: ~ 60 millions
Tx transfer: ~75 millions

These are much better numbers but still not correct, we'd expect masp to be 10x more expensive than vp implicit and tx transfer to be around 60% the cost of the masp vp (+ extra costs for storage write whose cost cannot be tracked down to benchmarks).

I've tried to change the tool from `wasm_instruments` to `wasmer` (which uses a gas metering solution based on a global variable inside the wasm code): I could see the gas costs decreasing to around 43 and 48 millions. Again this is not enough to match what we'd expect from the benchmarks.

A final note, a big component of gas costs (~40 millions) for both tx and implicit vp seems to come from the following line in the macros that we use to generate them:

`let tx_data = BatchedTx::try_from_slice(slice).unwrap();`
 
But again this should be accounted in the benchmarks themselves so it should be the problem.

To summarize this paragraph: there is definitely room to reduce (improve) the gas costs of wasm codes to more reasonable value but we are still not aligned with the expected costs. Moreover, we can see that there's an additional issue: the costs of tx transfer and vp implicit (both wasm codes) are not aligned: the transfer tx should cost 6 times the cost of the implicit vp, but we see that their costs more or less match.  

## Compiler optimizations

Another idea could be that the benchmarks of the single wasm opcodes don't translate well into the benchmark of the entire code (wither vp or tx) because of possible compiler optimizations, so that the resulting code in its entirety runs faster than just the combination of its components.

We use a `Singlepass` compiler so we don't expect very aggressive optimizations but still some optimizations are possible as confirmed by this report stating the improvements of the compiler optimizations over different versions https://wasmer.io/posts/wasmer-2_2.

It's hard to think this is really the case since the gas costs for the transfer tx are in line with the benchmarks (after the updates coming from the previous section) and the issue seems to only affect the implicit vp: if it was really down the the compiler I'd expect to see higher costs for both. It is still possible that the compiler is able to optimize only the implicit vp and not the transfer tx but this is not likely.

By analyzing the output of the compilation we get the following:

Vp implicit: 278599 lines  (WAT: 210560 lines)
Tx transfer: 313887 lines  (WAT: 204594 lines)

We can indeed see that even if the wat file for vp implicit is larger we end up with a compiled file which is shorter that the one of tx transfer: is hard to say though if this is the reason behind our issue, especially since we are just looking at the compiled code statically.

## Multiple gas accounting

I've reviewed the codebase to check if there are places where we account multiple times for the same gas: first of all this would only affect host fns/native vps which is not where we see the problems. In any case I couldn't find any, so errors in calls to the gas meter seem to not be the case.

From some experimentation I found out it's also possible to instrument the same wat file more than once and charge more gas (2x, 3x, etc..), but I looked at the code and it seems to me that instrumentation is correctly done and we only inject gas metering once per file.

## Absence of runtime gas metering for native vps

Another possible cause could be the fact that native vps cannot be instrumented with a runtime gas meter (given the absence of a stable IR) and this could lead to a lower gas consumption. From the benchmarks though, we can see that the crypto checks of the masp vp (for which gas is accounted) represent more than 80% of the timing of the masp vp. A good part of the remaining 20% is covered by the gas charged in host functions (storage reads), so the impact of the opcodes cannot be the cause of the issue we are seeing.

# Final recap

This is a summary of the findings regarding this issue:

- The gas charged for wasm codes is misaligned with the one charged for native vps in different ways: transfer tx seems to be a bit more expensive than it should while implicit vp is even more expensive (~10X)
- Benchmarks seem to be ok
- No wrong calls to the gas meter/middleware were found
- Gas metering for host functions seems ok
- The absence of runtime gas metering for native vp seems to not be the source of this problem
- Compilation costs are by far the biggest component for both txs and vps (we can decide not to charge gas here), but the issue is still there even if we don't account for them
- Some issues were found with the runtime gas injector and some improvements were made but still not enough. The fact that the two tools disagree on the gas for the same code by almost 30% worries me
- Compiler optimizations could play a role in the strange values we see but it's hard to tell

grarco:
## Describe your changes

- Partially addresses #2055.
- Partially addresses #3587  by setting the gas dividers to 1 (they'll be removed in a following PR)
- Closes #3588 by introducing a gas multiplier to align the non-wasm costs to the wasm ones
- Updates the gas costs based on recent benchmarks (`v0.41.0`)
- Reduces the impact of compilation costs by a factor of 100 to account for the (most likely) presence of the cache

`v0.41.0`


Fraccaman (approved):

grarco:
Do not merge this yet as we need to review some things with regards to native vps

brentstone:
Any chance we could also add a test? Perhaps one that checks that transparent transfers cost less gas than shielded transfers or similar?

brentstone (dismissed):
Consider a test.

Fraccaman (approved):

tzemanovic (approved):

Fraccaman:
@mergify requeue

Fraccaman:
@mergify refresh

Fraccaman:
@mergify queue

Fraccaman (approved):

Fraccaman:
@mergify refresh

PR #3557, opened 2024-07-26
CI: use the event head sha to check changelog

tzemanovic:


tzemanovic:
seen it fail here https://github.com/anoma/namada/actions/runs/10116518156/job/27979522547?pr=3402 - it should have used 095f13f1526710a7b8dbb0d481f3f89f02da7b1f but used the commit from base instead

PR #3593, opened 2024-08-07
Add TCP transport support for Ledger wallet

tzemanovic:
## Describe your changes

TODO:
- [x] add instructions on how to setup and use emulator

#3570


murisi (approved):
Looks good to me, thanks! Tested this on both the physical device and Speculos, and everything is working. It would have been nice to avoid needing to use macros, but it seems like the definition of `Exchange` precludes the use of trait objects.

tzemanovic:
> Looks good to me, thanks! Tested this on both the physical device and Speculos, and everything is working. It would have been nice to avoid needing to use macros, but it seems like the definition of `Exchange` precludes the use of trait objects.

Agree, it would be nice to get rid of the macro. I think we could make an enum for the transport layer. I'll add the change

tzemanovic:
@murisi added https://github.com/anoma/namada/commit/8f5cc8bd3b9d934c02469eb9029a85504f62085e

tzemanovic:
squashed fixups in last push

murisi (approved):
Looks good, the device transport construction code is no longer duplicated. Thanks!

tzemanovic:
@mergifyio requeue

PR #3598, opened 2024-08-09
ledger: remove wasm downloader

Fraccaman:
## Describe your changes
We are now shipping wasm artifacts together with the chain setup, no need to let the node download artifacts from an untrusted source.


crates/node/src/lib.rs
```diff
@@ -523,7 +523,7 @@ async fn run_aux_setup(
     scheduled_migration: Option<ScheduledMigration>,
 ) -> RunAuxSetup {
     // Prefetch needed wasm artifacts
```
tzemanovic, line 525:
```suggestion
```

tzemanovic (approved):

PR #3599, opened 2024-08-09
test/state/prefix_iter: avoid generating empty `kvs` list

tzemanovic:
## Describe your changes

Avoids a random test crash where the generated key-value list is empty

0.41.0


crates/state/src/lib.rs
```diff
@@ -1279,15 +1280,22 @@ mod tests {
         )
         .prop_map(|kvs| {
             kvs.into_iter()
-                .filter_map(|(key, val)| {
```
tzemanovic, line 1282: even though the `kvs` is non-empty, the `filter_map` would occasionally make it empty which broke the invariants for the test

Fraccaman (approved):

PR #3611, opened 2024-08-11
Murisi/fix masp vp mint case

murisi:
## Describe your changes
Addressed some of the MASP-IBC issues brought up during an audit. More specifically, the following changes have been made:
* The IBC internal address pre-balance is now only increased when a IbcMsgReceive packet requiring minting occurs
* Modified the MASP-VP to make an early exit if non-allowed keys are modified
* Updated/corrected outdated comments

## Checklist before merging 

### DIFF
--- a/crates/ibc/src/trace.rs
+++ b/crates/ibc/src/trace.rs
@@ -108,13 +108,24 @@ pub fn is_nft_trace(
     }
 }
 
-/// Return true if the source of the given IBC trace is this chain
-pub fn is_sender_chain_source(
+/// Returns true if the denomination originally came from the sender chain, and
+/// false otherwise.
+pub fn is_sender_chain_source_str(
     trace: impl AsRef<str>,
     src_port_id: &PortId,
     src_channel_id: &ChannelId,
 ) -> bool {
-    !trace
+    !is_receiver_chain_source_str(trace, src_port_id, src_channel_id)
+}
+
+/// Returns true if the denomination originally came from the receiving chain,
+/// and false otherwise.
+pub fn is_receiver_chain_source_str(
+    trace: impl AsRef<str>,
+    src_port_id: &PortId,
+    src_channel_id: &ChannelId,
+) -> bool {
+    trace
         .as_ref()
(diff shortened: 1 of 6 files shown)


crates/ibc/src/trace.rs
```diff
@@ -108,13 +108,24 @@ pub fn is_nft_trace(
     }
 }
 
-/// Return true if the source of the given IBC trace is this chain
-pub fn is_sender_chain_source(
+/// Returns true if the denomination originally came from the sender chain, and
+/// false otherwise.
+pub fn is_sender_chain_source_str(
```
yito88, line 113: just curious, why did you add `_str`?

murisi: I added `_str` in order to distinguish https://github.com/anoma/namada/blob/763b5fc0b251144271d6d6ccf134aab76c91eacb/crates/ibc/src/lib.rs#L825 from https://github.com/anoma/namada/blob/763b5fc0b251144271d6d6ccf134aab76c91eacb/crates/ibc/src/lib.rs#L524 since they are both used in the same file. That being said, I should have just used import `as`.

murisi: I've now reverted the naming and use import `as`.

crates/shielded_token/src/vp.rs
```diff
@@ -869,7 +869,7 @@ fn apply_balance_component(
     })
 }
 
-// Verify that the pre balance + the Sapling value balance = the post balance
+// Verify that the pre balance - the Sapling value balance = the post balance
```
cwgoes, line 872: Was logic actually changed here, or was this comment just wrong before?

cwgoes (changes requested):
One question. I also think we should probably write a test-case for the revised functionality?

grarco (approved):
LGTM, thank you!

crates/shielded_token/src/vp.rs
grarco: Wrong comment but the logic is unchanged

cwgoes (approved):

yito88 (approved):

PR #3573, opened 2024-07-31
Add IBC client upgrade test

yito88:
## Describe your changes
closes #3535 

- Added a new E2E test `ibc_upgrade_client`
  - Added a wasm `tx_proposal_ibc_client_upgrade` for testing to store the IBC client state and consensus state to be used in the counterparty chain
- Moved `COMMITMENT_PREFIX`

## Indicate on which release or other PRs this topic is based on
`v0.41.0`

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #3615, opened 2024-08-12
Remove parallel gas

### LINKED ISSUE #3587: Remove gas dividers
It has been decided to remove the gas discounts for parallel executions. We should therefore remove the const `PARALLEL_GAS_DIVIDER` and `MASP_PARALLEL_GAS_DIVIDER` and the logic that handles them. We can do this in two phases:

1. We just set these consts to 1
2. We remove the constants, rework the places were they are used and remove `VpsGas`. We'll also need to rework the way we sum the gas consumed by vps

grarco:
## Describe your changes

Closes #3587.

Depends-On: https://github.com/anoma/namada/pull/3573

This PR removes the parallel gas dividers and the `VpsGas` type which isn't needed anymore. It also removes the gas used field from `VpsResult`.

## Checklist before merging 

### DIFF
--- a/crates/tx/src/data/mod.rs
+++ b/crates/tx/src/data/mod.rs
@@ -26,7 +26,7 @@ use namada_core::ibc::IbcTxDataRefs;
 use namada_core::masp::MaspTxRefs;
 use namada_core::storage;
 use namada_events::Event;
-use namada_gas::{Gas, WholeGas};
+use namada_gas::WholeGas;
 use namada_macros::BorshDeserializer;
 #[cfg(feature = "migrations")]
 use namada_migrations::*;
@@ -481,9 +481,6 @@ pub struct VpsResult {
     pub accepted_vps: BTreeSet<Address>,
     /// The addresses whose VPs rejected the transaction
     pub rejected_vps: BTreeSet<Address>,
-    /// The total gas used by all the VPs
-    // FIXME: can't we just remove it?
-    pub gas_used: Gas,
     /// Errors occurred in any of the VPs, if any
     pub errors: Vec<(Address, String)>,
     /// Validity predicate status flags, containing info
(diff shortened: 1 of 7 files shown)


crates/tx/src/data/mod.rs
```diff
@@ -481,8 +481,6 @@ pub struct VpsResult {
     pub accepted_vps: BTreeSet<Address>,
     /// The addresses whose VPs rejected the transaction
     pub rejected_vps: BTreeSet<Address>,
-    /// The total gas used by all the VPs
-    pub gas_used: VpsGas,
```
grarco, line 485: I've removed this because I believe it carries little information to the final user (since this type ends up being published in the tx result event) given that it only mentions the amount of gas used by vps without specifying the distribution across the different vps and inside each vp (compilation, execution, host Fns,...)

grarco:
Actually my changes to the `VpsResult` are causing the failures to the IBC e2e tests: I suppose it's because this changes the format of `TxResult` so we might need to change something in our Hermes implementation for this to work

tzemanovic:
I rebased onto #3573 to update hermes

tzemanovic (approved):

Fraccaman (approved):

tzemanovic:
ah, looks like there's still some issue with e2e test `ibc_upgrade_client`

grarco:
> ah, looks like there's still some issue with e2e test `ibc_upgrade_client`

@yito88 pushed a commit that fixes it, we should be good now

tzemanovic:
> @yito88 pushed a commit that fixes it, we should be good now
ah, I think we need it in the base PR - will add it there

grarco:
Seems like another test is failing, I'll have a look

PR #3617, opened 2024-08-12
client: print sha256 of proposal wasm

Fraccaman:
## Describe your changes
Printing the sha256 of the wasm associated with a proposal makes it easier to check the its integrity and validatity.

## Checklist before merging 


crates/node/src/shell/governance.rs
```diff
@@ -122,14 +122,6 @@ where
...
                 let proposal_event = match proposal_type {
                     ProposalType::Default => {
-                        let proposal_code =
-                            gov_api::get_proposal_code(&shell.state, id)?
-                                .unwrap_or_default();
-                        let _result = execute_default_proposal(
-                            shell,
-                            id,
-                            proposal_code.clone(),
-                        )?;
```
Fraccaman, line 132: @grarco I think this is not needed in this branch right?

grarco: Correct and I don't know why it is still there. I swear we talked about this thing already a few months ago and we fixed it.

Anyway, I also see that `execute_pgf_funding_proposal` return a boolean that we never use so we might remove that too?

Fraccaman: im now using the result to improve the print statement

grarco: But it looks to me that `execute_pgf_funding_proposal` only ever returns `Ok(true)`

tzemanovic (approved):

PR #3626, opened 2024-08-13
Replace PaymentAddress with MASP when transferring to a shielded address

yito88:
## Checklist before merging 

### DIFF
--- a/crates/sdk/src/tx.rs
+++ b/crates/sdk/src/tx.rs
@@ -4,6 +4,7 @@ use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::fs::File;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use std::time::Duration;
 
 use borsh::BorshSerialize;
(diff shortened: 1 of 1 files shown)


cwgoes (approved):
LGTM at least, @grarco should review as well

yito88:
The test failed because the MASP VP rejected the transaction.
It seems that we can't replace it because the MASP VP checks the receiver address in the IBC packet for replay protection.

grarco:
I've reverted the previous changes. At the moment we can't achieve this for Namada2Namada IBC shielded transfers cause the memo can be used only for either the shielding or the refund and in this case we'd need both. We could update `build_ibc_transfer` after we've changed the refunding to a transparent trasfer.

For now the only thing I did was updating the gaia e2e tests to use the MASP internal address instead of the payment address as the receiver in the IBC packet.

We should also update the docs somewhere to instruct users of other IBC chains to follow this practice when they submit a shielding ibc transaction to Namada (this is probably something for @brentstone)

grarco:
@cwgoes I ran the test locally with a wrong hardcoded address and the test failed. I can confirm that we convert the string that we get in the ibc packet into an `Address` using this function, which indeed casts all payment addresses to the internal MASP address:

https://github.com/anoma/namada/blob/125af64f82edc460847aa5046d911efe322d4558/crates/core/src/address.rs#L396-L412

PR #3629, opened 2024-08-13
Improve governance client validation

Fraccaman:
## Checklist before merging 

### DIFF
--- a/crates/sdk/src/rpc.rs
+++ b/crates/sdk/src/rpc.rs
@@ -936,32 +936,49 @@ pub async fn get_public_key_at<C: crate::queries::Client + Sync>(
 }
 
 /// Query the proposal result
-pub async fn query_proposal_result<C: crate::queries::Client + Sync>(
-    client: &C,
+pub async fn query_proposal_result<N: Namada>(
+    context: &N,
     proposal_id: u64,
 ) -> Result<Option<ProposalResult>, Error> {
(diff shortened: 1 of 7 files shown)


crates/sdk/src/tx.rs
```diff
@@ -2091,7 +2091,8 @@ pub async fn build_vote_proposal(
     if !proposal.can_be_voted(current_epoch, is_validator) {
         edisplay_line!(
             context.io(),
-            "Proposal {} cannot be voted on anymore.",
+            "Proposal {} cannot be voted on anymore, either the voting period \
```
grarco, line 2094:
```suggestion
            "Proposal {} cannot be voted on, either the voting period \
```

crates/sdk/src/rpc.rs
```diff
@@ -936,32 +936,49 @@ pub async fn get_public_key_at<C: crate::queries::Client + Sync>(
...
         proposal
     } else {
         return Ok(None);
     };
-    let stored_proposal_result = convert_response::<C, Option<ProposalResult>>(
-        RPC.vp().gov().proposal_result(client, &proposal_id).await,
-    )?;
+
+    let current_epoch = query_epoch(context.client()).await?;
+    if current_epoch <= proposal.voting_start_epoch {
```
grarco, line 951: `<=` or just `<` here?

Fraccaman: ah right, nice catch!

PR #3646, opened 2024-08-14
Support only a transparent address as a refund target

### LINKED ISSUE #3619: Refund failed when IBC unshielded transfer failed
When transferring an IBC token back to the source chain from a shielded account, the refund failed because the MASP VP rejected it.
The token from the source chain was minted in Namada. So, when transferring back to the source chain, the token is burned on the source Namada. After the IBC transfer has failed by a timeout or a receiving error, it should be refunded by minting it again in Namada.

### LINKED ISSUE #3620: Rework shielded action refund
Currently we handle refunds for MASP over IBC transaction by letting the client construct the refund shielding transaction and embedding it in the IBC packet. This, unfortunately, has a problem when it comes to a failure (or timeout) in the IBC transaction involving an asset which is subject to MASP incentives. In this case we'll need to run the refund shielding action but the MASP VP requires the asset to be correctly epoch (i.e. to refer to the current MASP epoch): this is done so that malicious users cannot backdate their assets and extract more rewards than due.

The issue is that the refund transaction is built at the same time (epoch) of the ibc transaction: if Namada crosses a MASP epoch in between the construction of these transactions and the actual refund than the refund will be invalidated and there will be no way for the original source of the transaction to get their tokens back.

We should apply the logic suggested by @cwgoes and support only transparent refunds: this will sidestep this issue altogether, while simplifying the logic. At the same time, this solution doesn't raise any concern with regards to privacy since we can use a disposable address as the target of the refund.

yito88:
## Describe your changes
Closes #3620 
Closes #3619

## Checklist before merging 


yito88:
hmm, wasm compilation error happened again on my Mac

crates/sdk/src/tx.rs
```diff
@@ -3930,52 +3906,53 @@ async fn target_exists_or_err(
...
-                .to_payment_address(div)
-                .ok_or_else(|| {
-                    Error::Other(
-                        "Converting to a payment address failed".to_string(),
-                    )
-                })?
-                .into();
-            let alias = format!("ibc-refund-target-{}", rng.next_u64());
             let mut wallet = context.wallet_mut().await;
+            if let Some(addr) = wallet.find_address(IBC_REFUND_TARGET_ALIAS) {
```
grarco, line 3931: I believe it's better to generate a new disposable address every time, otherwise we might establish some linkage between the shielded ibc operations and this transparent refund address

crates/ibc/src/context/nft_transfer.rs
```diff
@@ -28,6 +28,7 @@ where
     C: IbcCommonContext,
 {
     inner: Rc<RefCell<C>>,
+    is_shielded: bool,
```
grarco, line 31: Why do we have this new field?

crates/sdk/src/tx.rs
yito88: That's a default one. A user can prepare an address with `--refund-target`.

crates/ibc/src/context/nft_transfer.rs
yito88: Because the sender in the IBC packet is replaced with the refund address. The context doesn't know whether the sender is the actual sender or the refund target.

crates/sdk/src/tx.rs
yito88: But yes, it's better to generate a new one for useability

sug0 (changes requested):

crates/ibc/src/context/nft_transfer.rs
grarco: Right, makes sense, thank you

grarco (approved):

sug0 (approved):

PR #3653, opened 2024-08-15
Add birthdays to MASP keys

batconjurer:
## Describe your changes

Partially addresses Issue #2900. Viewing and spending keys can no be given birthdays in the form of block heights which are loaded into shielded sync. Shielded sync will not try to decrypt a block before a keys birthday with said key. 

## Checklist before merging 


tzemanovic (approved):

sug0 (approved):
These changes look good against `main`. When we integrate them with the phase 3 shielded sync code, we should optimize this function accordingly:

https://github.com/anoma/namada/blob/b6895337d90649abbd509f3416210d60bc9dbae4/crates/sdk/src/masp/shielded_sync/dispatcher.rs#L774-L791

Right now, it's always attemtpting to scan txs without checking the heights of viewing keys.

sug0:
ah benches are broken

PR #3675, opened 2024-08-20
token: fix transfer to self

Fraccaman:
## Describe your changes
To make things easier, lets make transfer fails if `source == dest`

## Checklist before merging 

### DIFF
--- a/crates/trans_token/src/storage.rs
+++ b/crates/trans_token/src/storage.rs
@@ -237,9 +237,10 @@ pub fn transfer<S>(
 where
     S: StorageRead + StorageWrite,
 {
-    if amount.is_zero() {
+    if amount.is_zero() || src == dest {
         return Ok(());
     }
+
     let src_key = balance_key(token, src);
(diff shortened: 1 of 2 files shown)


crates/trans_token/src/storage.rs
```diff
@@ -240,6 +240,13 @@ where
     if amount.is_zero() {
         return Ok(());
     }
+
```
tzemanovic, line 243: let's update the comment for this fn

Fraccaman: done in #[4c5a9f3](https://github.com/anoma/namada/pull/3675/commits/4c5a9f35c879efafcaad87b9e2d580ed834d183b)

tzemanovic (changes requested):
the transfer to self has to be a no-op cause it's used from protocol where it shouldn't fail otherwise a tx might get wrongly rejected

tzemanovic (approved):

PR #3680, opened 2024-08-21
Augment the functionality of `gen_localnet.py`

sug0:
## Describe your changes

This PR augments the functionality of `gen_localnet.py` with the ability to generate full node base directories. Moreover, adds a new `localnet-2` localnet starter kit with 2 validator nodes, and a script to start from this new localnet with either of the validator nodes or a full node.

## Checklist before merging 

### DIFF
--- a/.changelog/unreleased/testing/3680-gen-localnet-fullnodes.md
+++ b/.changelog/unreleased/testing/3680-gen-localnet-fullnodes.md
@@ -1,0 +1,4 @@
+- Augment the functionality of `gen_localnet.py` with the ability
+  to generate full node base directories. Moreover, add a new script
+  to boot up a localnet with two genesis validators or a full node.
+  ([\#3680](https://github.com/anoma/namada/pull/3680))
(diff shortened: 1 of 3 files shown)


tzemanovic:
I think we could use the `init-genesis-validator` and `sign-genesis-txs` to add arbitrary number of validators to `localnet` instead

sug0:
that's what I did. I modified the existing `localnet` dir with those cmds

tzemanovic:
> that's what I did. I modified the existing `localnet` dir with those cmds

ah I see. Can we do that so that the script makes it on-demand in a temp dir so we don't need to track `localnet-2`?

sug0:
> > that's what I did. I modified the existing `localnet` dir with those cmds
> 
> ah I see. Can we do that so that the script makes it on-demand in a temp dir so we don't need to track `localnet-2`?

sure, should be doable

tzemanovic (approved):

PR #3687, opened 2024-08-22
 Apply snapshot for state sync

batconjurer:
## Describe your changes
Addresses the remaining points of Issue #3307 
 - Adds chunking logic to snapshots
 - Implements the `OfferSnapshot` ABCI call
 - Implements the `ApplySnapshotChunk` ABCI call

## Checklist before merging 

### DIFF
--- a/crates/node/src/storage/rocksdb.rs
+++ b/crates/node/src/storage/rocksdb.rs
@@ -821,9 +821,9 @@ pub struct SnapshotMetadata {
 pub struct DbSnapshot<'a>(pub rocksdb::Snapshot<'a>);
 
 impl<'a> DbSnapshot<'a> {
-    /// Write a snapshot of the database out to file. The last line
-    /// of the file contains metadata about how to break the file into
-    /// chunks.
+    /// Write a snapshot of the database out to file.  Also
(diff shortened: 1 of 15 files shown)


crates/node/src/storage/rocksdb.rs
```diff
@@ -737,6 +737,72 @@ impl RocksDB {
             .get_cf(rollback_cf, key)
             .map_err(|e| Error::DBError(e.into_string()))
     }
+
+    /// Writes an entry directly to the db
+    pub fn insert_entry(
```
tzemanovic, line 742: Should we just rm the `height` param since it's not implemented? We could maybe export and use the existing `add_value_bytes_to_batch` instead (also the doc line here is not accurate)

batconjurer: sure

tzemanovic (approved):

batconjurer:
I just want to add one more unit test to get the code coverage up. And it's a good test anyways

sug0:
the test was added in the snapshot improvements pr btw, so this one should be good to go, after the conflict is resolved

PR #3689, opened 2024-08-22
Minor fixes to governance cli/msgs

grarco:
## Describe your changes

Addresses some minor governance issues:

- Removed the unused `PROPOSAL_ETH` cli arg
- Fixes the fact that the required `--proposal-id` arg was not shown in the help message of `proposal-vote`
- Updates the error message in the governance vp that we raise when the voter is not a validator nor delegator to also mention the epoch at which they should have been so (someone might delegate after the beginning of the voting period and struggle to understand why the transaction fails)

## Checklist before merging 


brentstone (approved):
Thx for cleaning all these up, good catches.

Fraccaman (approved):

PR #3691, opened 2024-08-22
CLI queries for total supply info

### LINKED ISSUE #3522: CLI Command Requests
Hello,

Currently testing the PGF, we think that multiple commands could be usefull:
1) Command to query the current total supply
2) Command to retrieve the internal PoS and PGF addresses

Thank you.

brentstone:
## Describe your changes
Closes #3522

## Checklist before merging 

### DIFF
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -1188,7 +1188,10 @@ pub async fn query_total_supply<N: Namada>(
             .total_supply(context.client(), &token)
             .await,
     );
-    display_line!(context.io(), "Total supply of {token}: {supply}");
+    display_line!(
+        context.io(),
+        "Total supply of token {token}: {supply} raw units"
+    );
 }
 
 /// Query the effective total supply of the native token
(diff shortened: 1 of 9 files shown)


crates/apps_lib/src/client/rpc.rs
```diff
@@ -1174,6 +1174,35 @@ pub async fn query_rewards<C: namada_sdk::queries::Client + Sync>(
     )
 }
 
+/// Query token total supply.
+pub async fn query_total_supply<N: Namada>(
+    context: &N,
+    args: args::QueryTotalSupply,
+) {
+    let token = args.token;
+    let supply = unwrap_client_response::<N::Client, token::Amount>(
```
tzemanovic, line 1183: You can use `get_token_total_supply` from sdk here. Let's also call `format_denominated_amount` on it to print it nicer

brentstone: Do you think it is ok to leave the first part as is? Since we get the right error handling from `unwrap_client_response` (and would have to reimplement it if using `get_token_total_supply`).

Formatted denominated amount in 02c79160c923ab417a4812e4a0f17a9f02022f66

tzemanovic:
there's a failing integration test - we need to update the expected string in `crates/tests/src/integration/ledger_tests.rs:1180`

tzemanovic (approved):

PR #3693, opened 2024-08-23
Clean IBC E2E tests

### LINKED ISSUE #3601: Clean up IBC E2E tests
The current IBC E2E tests take a long time.

```
    "e2e::ibc_tests::run_ledger_ibc": 110,
    "e2e::ibc_tests::run_ledger_ibc_with_hermes": 230,
    "e2e::ibc_tests::pgf_over_ibc_with_hermes": 810,
    "e2e::ibc_tests::proposal_ibc_token_inflation": 1100,
    "e2e::ibc_tests::ibc_rate_limit": 430,
    "e2e::ibc_tests::ibc_namada_gaia": 180,
    "e2e::ibc_tests::ibc_upgrade_client": 550,
```

We can fix that by
- Focusing on testing between Namada and Gaia
  - Because we don't have a plan to launch multiple Namada chains
- Unifying some tests to reduce the redundant preparation
- Adjusting proposal parameters to reduce the testing time

So, we could tidy them up as the following five tests.
1. IBC transfer tests
    - Transparent transfers
        - Namada -> Gaia -> Namada
        - Gaia -> Namada -> Gaia
    - Shielding transfers
        - Gaia -> Namada -> (shielded transfer) ->  Namada -> Gaia
        - Namada (from a shielded address) -> Gaia -> Namada (to another shielded address)
    - Refunding when transfer failure
        - Timeout
        - Invalid receiver (Ack with an error)
1. PGF over IBC
1. IBC token inflation
1. IBC client upgrade
1. IBC rate limit

yito88:
## Describe your changes
closes #3601 

- Also refined the IBC client upgrade test with Gaia

## Checklist before merging 


yito88:
The new `ibc_upgrade_client` test requires Gaia v18 or later: heliaxdev/namada-ci#4

crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -2739,112 +1551,17 @@ fn check_gaia_balance(
...
-
-    Ok(())
-}
-
-fn check_funded_balances(
-    dest_port_id: &PortId,
-    dest_channel_id: &ChannelId,
-    test_b: &Test,
+fn check_inflated_balance(
+    test: &Test,
```
grarco, line 1555: Shouldn't we pass here also the initial balance of the viewing key? In this function I see that we check:

`assert!(balance > 0.0);`

but I'm not sure this is completely correct: if we want to check the shielded rewards it means the viewing key will have some balance to produce the rewards. Checking just for the balance to be positive doesn't guarantee we have distributed rewards (or that the specified target has received them): I believe we should pass in the balance from the previous masp epoch and then check that the new balance is greater than this one

yito88: The account's NAM balance should be empty before the inflation.
I added the check. Thanks.

grarco (approved):

sug0 (approved):

sug0:
oh, the upgrade client ibc e2e test failed

yito88:
> oh, the upgrade client ibc e2e test failed
That's expected. It needs a new gaia in CI. https://github.com/heliaxdev/namada-ci/pull/4

PR #3702, opened 2024-08-26
derive Serialize for PosParams, OwnedPosParams, GovernanceParameters

egasimus:
Hi! :bow: This PR enables deriving the `Serialize` trait for three of the structs that we expose through our bleeding-edge client library, [@fadroma/namada](https://github.com/hackbg/fadroma/tree/v2/packages/namada).

Right now we decode these by manually building `js_sys::Object` instances, but this requires us to explicitly name each field, hence it is not future-proof against the struct fields changing. As we migrate towards automatic JSON serialization using `serde_json`, we've found that not all SDK structs derive `serde::Serialize`. Hence, this :pray:


brentstone:
can you run `make fmt` and then add a changelog using `unclog`? See https://github.com/anoma/namada/blob/main/CONTRIBUTING.md

Fraccaman:
ah nice catch about the changelog docs, but it would be better to split them into another PR. Could you drop `c4ef91dea01c06ccda8284b54a190f147aa6b315` ?

Fraccaman (approved):

PR #3715, opened 2024-08-28
improve offline tx handling

### LINKED ISSUE #3417: `dump-tx` invalid serialization
When dumping a tx to via `--dump-tx`, a user can decide to either serialize it to a file or to print the tx in the CLI. The problem is that they are serialized in 2 different ways and only serializing the tx to file let the user sign it later.

Fraccaman:
## Describe your changes
Closes #3417.

## Checklist before merging 

### DIFF
--- a/crates/apps_lib/src/cli.rs
+++ b/crates/apps_lib/src/cli.rs
@@ -2458,6 +2458,7 @@ pub mod cmds {
         InitGenesisEstablishedAccount(InitGenesisEstablishedAccount),
         InitGenesisValidator(InitGenesisValidator),
         PkToTmAddress(PkToTmAddress),
+        SignOffline(SignOffline),
         DefaultBaseDir(DefaultBaseDir),
         EpochSleep(EpochSleep),
         ValidateGenesisTemplates(ValidateGenesisTemplates),
@@ -2486,6 +2487,8 @@ pub mod cmds {
                     SubCmd::parse(matches).map(Self::InitGenesisValidator);
                 let pk_to_tm_address =
                     SubCmd::parse(matches).map(Self::PkToTmAddress);
+                let sign_offline =
+                    SubCmd::parse(matches).map(Self::SignOffline);
                 let default_base_dir =
                     SubCmd::parse(matches).map(Self::DefaultBaseDir);
                 let epoch_sleep = SubCmd::parse(matches).map(Self::EpochSleep);
@@ -2508,6 +2511,7 @@ pub mod cmds {
                     .or(validate_genesis_templates)
                     .or(genesis_tx)
                     .or(parse_migrations_json)
+                    .or(sign_offline)
             })
         }
 
@@ -2522,6 +2526,7 @@ pub mod cmds {
                 .subcommand(InitGenesisEstablishedAccount::def())
                 .subcommand(InitGenesisValidator::def())
                 .subcommand(PkToTmAddress::def())
+                .subcommand(SignOffline::def())
                 .subcommand(DefaultBaseDir::def())
                 .subcommand(EpochSleep::def())
                 .subcommand(ValidateGenesisTemplates::def())
@@ -3190,6 +3195,25 @@ pub mod cmds {
     }
 
     #[derive(Clone, Debug)]
+    pub struct SignOffline(pub args::SignOffline);
+
+    impl SubCmd for SignOffline {
+        const CMD: &'static str = "sign-offline";
+
+        fn parse(matches: &ArgMatches) -> Option<Self> {
+            matches
+                .subcommand_matches(Self::CMD)
+                .map(|matches| Self(args::SignOffline::parse(matches)))
+        }
+
+        fn def() -> App {
+            App::new(Self::CMD)
+                .about(wrap!("Offlne sign a transaction."))
+                .add_args::<args::SignOffline>()
+        }
+    }
+
+    #[derive(Clone, Debug)]
     pub struct DefaultBaseDir(pub args::DefaultBaseDir);
 
     impl SubCmd for DefaultBaseDir {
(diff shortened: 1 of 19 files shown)


crates/apps_lib/src/cli.rs
```diff
@@ -7918,6 +7944,48 @@ pub mod args {
         }
     }
 
+    #[derive(Clone, Debug)]
+    pub struct SignOffline {
+        pub tx_path: PathBuf,
+        pub secret_keys: Vec<common::SecretKey>,
+        pub owner: Address,
```
sug0, line 7951: I think these can be wallet aliases. key/address look-ups are still done offline
```suggestion
        pub secret_keys: Vec<WalletKeypair>,
        pub owner: WalletAddress,
```

crates/core/src/hash.rs
```diff
@@ -51,12 +51,51 @@ pub type HashResult<T> = std::result::Result<T, Error>;
...
+                "Invalid length: expected {} bytes, got {}",
+                HASH_LENGTH,
+                bytes.len()
+            )));
+        }
+
+        let mut array = [0u8; HASH_LENGTH];
+        array.copy_from_slice(&bytes);
+
+        Ok(Hash(array))
```
sug0, line 95: you can avoid allocating an intermediate buffer
```suggestion
        // Ensure the byte array has the correct length
        if hex_string.len() != 2 * HASH_LENGTH {
            return Err(de::Error::custom(format!(
                "Invalid length: expected {} bytes, got {}",
                2 * HASH_LENGTH,
                hex_string.len()
            )));
        }

        // Convert the hex string back to a byte array
        let mut output = [0u8; HASH_LENGTH];
        HEXUPPER
            .decode_mut(hex_string.as_bytes(), &mut output)
            .expect("Hash decoding shouldn't fail after assertion");
            
        Ok(Hash(output))
```

crates/tx/src/lib.rs
```diff
@@ -36,6 +36,88 @@ pub use types::{
...
+                "Invalid length: expected {} bytes, got {}",
+                SALT_LENGTH,
+                bytes.len()
+            )));
+        }
+
+        let mut array = [0u8; SALT_LENGTH];
+        array.copy_from_slice(&bytes);
+
+        Ok(array)
```
sug0, line 86: suggestion similar to the one above here

sug0 (changes requested):
just a couple of suggestions

crates/apps_lib/src/cli.rs
Fraccaman: i think `client utils` cli methods dont have access to the wallet

sug0: you have access to `args::Global` in `sign_offline` (from `crates/apps_lib/src/client/utils.rs`). this struct holds a `base_dir`, and a `is_pre_genesis` arg. you can either load a pre-genesis or post-genesis wallet with this.

tzemanovic: for an example you can see `ClientUtils::EpochSleep` cli handle that constructs a ctx

crates/apps_lib/src/client/tx.rs
```diff
@@ -1069,15 +1072,11 @@ where
                 Some(path) => path.join(filename),
                 None => filename.into(),
             };
-
             let signature_path = File::create(&output_path)
                 .expect("Should be able to create signature file.");
+            serde_json::to_writer_pretty(signature_path, &signature)
+                .expect("Signature should be deserializable.");
```
tzemanovic, line 1078:
```suggestion
                .expect("Signature should be serializable.");
```

crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -123,6 +124,104 @@ fn run_ledger() -> Result<()> {
     Ok(())
 }
 
+#[test]
```
tzemanovic, line 127: I think could maybe be an integration test

crates/tests/src/e2e/helpers.rs
```diff
@@ -727,3 +727,24 @@ pub fn find_gaia_address(
 
     Ok(matched.trim().to_string())
 }
+
+pub fn find_offline_file(
```
tzemanovic, line 731: I'd maybe call this `find_file_with_ext`

crates/apps_lib/src/cli.rs
Fraccaman: Building a context requires a client which requires a connection to a node, I want this to be completly offline

sug0:
```rs
    pub struct Global {
        pub is_pre_genesis: bool,
        pub chain_id: Option<ChainId>,
        pub base_dir: PathBuf,
        pub wasm_dir: Option<PathBuf>,
    }
```
you have access to a value of this type. using `Global`, you can load the wallet by navigating to `$base_dir/$chain_id/wallet.toml` or `$base_dir/pre-genesis/wallet.toml` (depending on whether we're post or pre genesis, respectively). this can be done fully offline. for pre-genesis stuff, we have some examples in the genesis flow code, that is used to sign genesis txs.

Fraccaman: solve with some help from tiago!

sug0 (approved):

tzemanovic (approved):

sug0:
```
    integration::ledger_tests::pgf_governance_proposal
```
failed

tzemanovic:
yeah, seeing this in #3718 and #3735 after rebase too

dan-u410:
there's another (broken) offline signing cmd, `namada client sign-tx` might be nice to deprecate that in favor of `sign-offline` being developed here

PR #3716, opened 2024-08-28
Unsigned MASP `Changes` in SDK

### LINKED ISSUE #3425: Make `Changes` unsigned
#3393 introduces the `Changes` type to track the unutilized amount of a spend note so that we can correctly handle fee payment and actual changes. The type is a map from an extended spending key to a `I128Sum` but these amounts are guaranteed to be non-negative so we should change them to `U128Sum`.

We can also change `from_masp_denominated_i128` to handle unsigned amounts given that this function is used only with changes.

grarco:
## Describe your changes

Closes #3425.

Modifies the `Changes` type in the SDK to wrap `U128Sum` instead of `I128Sum` since changes are guaranteed to be non-negative. Also adjusts `from_masp_denominated_i128` and swap the calls to `from_nonnegative` for `from_pair`.

## Checklist before merging 

### DIFF
--- a/.changelog/unreleased/improvements/3716-unsigned-change-in-sdk.md
+++ b/.changelog/unreleased/improvements/3716-unsigned-change-in-sdk.md
@@ -1,0 +1,2 @@
+- Modified the `Changes` type to use unsigned integers.
+  ([\#3716](https://github.com/anoma/namada/pull/3716))
--- a/crates/core/src/token.rs
+++ b/crates/core/src/token.rs
@@ -232,16 +232,14 @@ impl Amount {
         Self { raw: Uint(raw) }
     }
 
-    /// Given a i128 and [`MaspDigitPos`], construct the corresponding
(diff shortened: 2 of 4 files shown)


murisi (changes requested):
Probably these changes are okay for the most part. The MASP crate uses `I128Sum`s to represent value balances because we needed a signed type that can holds sums and differences of `u64`s without overflowing or underflowing. Given this, while I understand that it is conceptually clearer to hold unsigned amounts in `U128Sum`s. It seems to me that this comes at the cost of introducing another integer type `U128SUm` and sometimes having to do fallible conversions from `U128Sum`s to `I128Sum`s for interoperability with the MASP crate. I guess this is fine if we understand that due to the fallible runtime conversions, the `U128Sum`s (just like the `I128Sums` that they are replacing) can really only hold `U127Sum`s.

grarco:
Closing this as it is no longer necessary

Issue #1497, opened 2023-06-02
Code coverage audit


cwgoes:
A lot of our coverage is provided by two large E2E tests (MASP & ledger transactions), which should be mostly replaced with / supplanted by unit / prop tests with proper mocks so that we can reach sufficient coverage.

tzemanovic:
to my knowledge `make test-unit-coverage` skips e2e

cwgoes:
Hmm - in that case @batconjurer or @mariari can you clarify the concern here?

tzemanovic:
I agree that MASP should have unit / integration / prop tests. There's nothing to catch issues in it besides e2e tests and the lack of these tests is being a serious issue for the more complex integrations

Fraccaman:
I think coverage is still low especially for MASP.

tzemanovic:
I went through coverage report from #3704: https://app.codecov.io/gh/anoma/namada/pull/3704/tree/crates

In our crates, the least coverage comes from sdk (14466 missed lines), node (6593 lines), ibc (2186), apps_lib (2011). I think we can improve the coverage a lot in these if we include integration tests, which should be possible.

Specific section that need attention:
- [ ] account
	- auth & storage mods
	- priority: medium
- [ ] ibc
	- lib - a lot of the ibc functionality is only tested in e2e
	- priority: medium
- [ ] node
	- shell/governance.rs - should also be able to move this into governance crate now
	- priority: medium
- [ ] sdk
	- masp - can we also move it (with shielded-sync sub-mod) into `shielded_token`?
	- priority: high
- [ ] shielded_token
	- vp & validation mods
	- priority: high
- [ ] token
	- `Transfer` methods
	- priority: high
- [ ] tx
	- a lot of functionality inside the `types` mod is not covered 
	- it also looks like the `types` mod contains too much uncategorized logic that should be refactored
	- priority: high

PR #3718, opened 2024-08-28
mv gov sub-module from shell into gov crate

tzemanovic:
## Describe your changes

- related to #1497

## Checklist before merging 


crates/governance/src/finalize_block.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+    Ok(None)
+}
+
+/// Count of passed and rejected proposals.
+#[derive(Default)]
+pub struct ProposalsResult {
+    passed: Vec<u64>,
+    rejected: Vec<u64>,
+}
```
Fraccaman, line 85: maybe we can just return `Result<(), ..>` ? I dont think we need to return to the called which proposal passed and which didn't?

tzemanovic: sounds good

tzemanovic: done in [ae95058](https://github.com/anoma/namada/pull/3718/commits/ae950581613927fff73765ab7ed19b221d78286e)

tzemanovic:
I've added one more refactor to get rid of the tx and vp cache being passed through

Fraccaman (approved):

tzemanovic:
@mergifyio requeue

tzemanovic:
@mergifyio requeue

tzemanovic:
the `integration::ledger_tests::pgf_governance_proposal` keeps failing in CI but locally it passes for me :/

brentstone:
@tzemanovic I was having the same problem with that failing test in #3691 - at some point after a rebasing on `main` it passed but I never figured out why it was failing in CI

tzemanovic:
this one started failing after rebase - same in #3735

PR #3724, opened 2024-08-29
Fixes MASP tx expiration in the SDK

### LINKED ISSUE #3586: MASP txs do not respect tx expiration date set in args
In `gen_shielded_transfer`, MASP txs do not pull in the tx expiration set by users in `args::Tx`. If a user requests no tx expiration, the tx might unexpectedly timeout when it should have been validated by the MASP VP.

grarco:
## Describe your changes

Closes #3586.

- Fixes a bug by which the SDK would not set the correct expiration to MASP `Transaction`s
- Adds an integration test for masp expiration
- Removes some outdated assertions based on the old double-block execution scheme
- Improves assertions on `MockNode`
- Wraps `MockNode` into a single `Arc`

## Checklist before merging 

### DIFF
--- a/crates/sdk/src/tx.rs
+++ b/crates/sdk/src/tx.rs
@@ -2611,6 +2611,7 @@ pub async fn build_ibc_transfer(
         masp_transfer_data,
         masp_fee_data,
         !(args.tx.dry_run || args.tx.dry_run_wrapper),
+        args.tx.expiration.to_datetime(),
     )
     .await?;
     let shielded_tx_epoch = shielded_parts.as_ref().map(|trans| trans.0.epoch);
@@ -3071,6 +3072,7 @@ pub async fn build_shielded_transfer<N: Namada>(
         transfer_data,
         masp_fee_data,
         !(args.tx.dry_run || args.tx.dry_run_wrapper),
+        args.tx.expiration.to_datetime(),
     )
     .await?
     .expect("Shielded transfer must have shielded parts");
@@ -3236,6 +3238,7 @@ pub async fn build_shielding_transfer<N: Namada>(
         transfer_data,
         None,
         !(args.tx.dry_run || args.tx.dry_run_wrapper),
+        args.tx.expiration.to_datetime(),
     )
     .await?
     .expect("Shielding transfer must have shielded parts");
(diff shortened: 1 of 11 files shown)


crates/sdk/src/tx.rs
```diff
@@ -3778,6 +3783,7 @@ pub async fn gen_ibc_shielding_transfer<N: Namada>(
             // Fees are paid from the transparent balance of the relayer
             None,
             true,
+            args.expiration.to_datetime(),
```
grarco, line 3786: @yito88 we might have a failure in Hermes because of this update...

yito88: Tests with Hermes worked well

grarco: Right seems like everything is fine now, thanks!

sug0 (changes requested):
couple of suggestions + comments

crates/tests/src/integration/masp.rs
```diff
@@ -1516,22 +1517,181 @@ fn multiple_unfetched_txs_same_block() -> Result<()> {
...
+        vec![
+            "add",
+            "--alias",
+            "alias_a",
+            "--value",
+            AA_VIEWING_KEY,
+            "--unsafe-dont-encrypt",
+        ],
+    )?;
+    node.assert_success();
```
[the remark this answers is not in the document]
grarco: I've tried your suggestion but the tests fail if we don't add the viewing keys to the wallet

crates/tests/src/integration/masp.rs
```diff
@@ -1516,22 +1517,181 @@ fn multiple_unfetched_txs_same_block() -> Result<()> {
...
+            NAM,
+            "--amount",
+            "50",
+            "--gas-payer",
+            CHRISTEL_KEY,
+            // This is used to set the correct expiration in the masp object,
+            // we don't care about the expiration at the tx level
+            "--expiration",
+            #[allow(clippy::disallowed_methods)]
+            &DateTimeUtc::now().to_string(),
```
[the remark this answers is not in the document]
grarco: I think the comment is misleading, I'll update it. Essentially we want to create an expired masp tx: this will also set the expiration on the header and the tx will fail because of the protocol check. We don't care about this though, cause we are going to override the header

yito88 (approved):

PR #3730, opened 2024-08-30
Enable the signing logic to fall back to the hardware wallet if secret key is not found.

murisi:
## Describe your changes
Currently the transaction signing logic fails fall back to the hardware wallet if the secret key is not found in the wallet store. This is because an error is propagated as soon as the secret key lookup fails. This PR fixes that. Additionally, the signing code now checks that the number of public keys used in signing exceeds the given account's threshold.

## Checklist before merging 

### DIFF
--- a/crates/sdk/src/signing.rs
+++ b/crates/sdk/src/signing.rs
@@ -254,8 +254,8 @@ where
     }
 
     // Then try to sign the raw header using the hardware wallet
-    for pubkey in signing_data.public_keys {
-        if !used_pubkeys.contains(&pubkey) && pubkey != signing_data.fee_payer {
+    for pubkey in &signing_data.public_keys {
+        if !used_pubkeys.contains(pubkey) && *pubkey != signing_data.fee_payer {
             if let Ok(ntx) = sign(
                 tx.clone(),
                 pubkey.clone(),
@@ -282,7 +282,10 @@ where
         Ok(fee_payer_keypair) => {
             tx.sign_wrapper(fee_payer_keypair);
         }
-        Err(_) => {
+        // The case where tge fee payer also signs the inner transaction
+        Err(_)
+            if signing_data.public_keys.contains(&signing_data.fee_payer) =>
+        {
             *tx = sign(
                 tx.clone(),
                 signing_data.fee_payer.clone(),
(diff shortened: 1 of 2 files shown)


crates/sdk/src/signing.rs
```diff
@@ -231,7 +231,13 @@ where
 
         for public_key in &signing_data.public_keys {
             if !used_pubkeys.contains(public_key) {
-                let secret_key = find_key_by_pk(&mut wallet, args, public_key)?;
```
grarco, line 234: I believe this is a modification I made to fix an issue we were having. I understand your changes and they make sense to me but at the same time I believe we need to make sure that the transaction has been signed, otherwise we'd end up with the same issue (unsigned transaction gets submitted). I see that one we call the signing function on the hardware wallet just below we pattern match on:

`if let Ok(ntx) = sign`

but we don't handle the error case. Could we instead return an error in that case?

grarco: Actually here's an addition to what I've previously written. Returning an error if the hardware wallet signature fails could be an option but I don't believe it addresses our issue. So what we want to do is to ensure that all the signatures that the user has requested have been correctly produces and attached to the tx: these must include signatures that are provided directly and those that must be computed by this function (either via the software or hardware wallet).

So perhaps a better solution would be to swap the variable `used_pubkeys` for a `requested_pubkeys` variable. This should be initially populated with all the pub keys that the user requires a signature from. While we go through the function we remove from this list the signers for which we've successfully attached the signature to the tx. At the end, if this list is not empty, it means that we were not able to attach some of the required signatures and we should return an error to prevent submission of the invalid transaction.

murisi: > So what we want to do is to ensure that all the signatures that the user has requested have been correctly produces and attached to the tx: these must include signatures that are provided directly and those that must be computed by this function (either via the software or hardware wallet).

Ah, understood thanks.
> At the end, if this list is not empty, it means that we were not able to attach some of the required signatures and we should return an error to prevent submission of the invalid transaction.

Couldn't there be a problem with this approach since not all of the requested signatures have to be used? We only have to use `SigningTxData::threshold` of the requested signatures, no? Nonetheless, I've essentially implemented your suggestion but with the alteration that it checks that `used_pubkeys` is full enough rather than `requested_pubkeys` being empty. Let me know if I'm misunderstanding...

grarco: Ah right, nice improvement, thanks!

grarco (approved):
LGTM, thank you!

Fraccaman (approved):

PR #3736, opened 2024-08-30
refactor token transfer and add more tests

tzemanovic:
## Describe your changes

Moves the transfer code out of wasm into the token crates and adds more tests

## Checklist before merging 


crates/trans_token/src/tx.rs
```diff
@@ -0,0 +1,548 @@
...
+        // Dest received the amount
+        assert_eq!(read_balance(ctx(), &token, &dest).unwrap(), amount);
+
+        // Src spent the amount
+        assert_eq!(
+            read_balance(ctx(), &token, &src).unwrap(),
+            token::Amount::zero()
+        );
+
+        // Internal token address have to be part of the verifier set
```
grarco, line 292: Can/should we also check that the src and dest addresses are part of the validation set?

grarco: Edit, I see that we do that check in the test in crates/token/src/tx.rs, so I'm not sure my comment is still relevant

tzemanovic: I've added this to be complete and a bit more

grarco (approved):

PR #3745, opened 2024-09-03
use test-dir for network archive to avoid re-compilations after test run

tzemanovic:
## Describe your changes

Without `archive_dir` init-network creates the archive file in the current dir (`crates/tests`) which forces recompilation.


Fraccaman (approved):

Fraccaman:
@mergify requeue

PR #3746, opened 2024-09-03
Update storage gas cost

grarco:
## Describe your changes

Recovers the fix to the storage gas cost that was done in #3510 and went lost in #3554. Adjusts the `DEFAULT_GAS_LIMIT` and the gas limits in the genesis files.

Also improves gas tracking for masp fee payment: the gas meter now does not account for the gas used up to that point (wrapper gas) which could prevent some txs that consume a lot of gas in that phase (e.g. batches) from successfully pay fees via the MASP.

## Checklist before merging 

### DIFF
--- a/crates/node/src/dry_run_tx.rs
+++ b/crates/node/src/dry_run_tx.rs
@@ -37,7 +37,7 @@ where
     let gas_scale = parameters::get_gas_scale(&state)?;
 
     // Wrapper dry run to allow estimating the gas cost of a transaction
-    let (wrapper_hash, extended_tx_result, tx_gas_meter) =
+    let (wrapper_hash, extended_tx_result, tx_gas_meter, gas_used) =
         match tx.header().tx_type {
             TxType::Wrapper(wrapper) => {
                 let gas_limit = wrapper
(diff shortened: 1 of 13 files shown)


grarco:
@yito88 The ibc e2e tests are failing because of gas, the costs are now slightly higher than before

grarco:
> @yito88 The ibc e2e tests are failing because of gas, the costs are now slightly higher than before so we might need to update Hermes

Actually, it seems like the only issue was a bug in the `dry_run` function, we don't need to update Hermes

crates/node/src/dry_run_tx.rs
```diff
@@ -37,7 +37,7 @@ where
     let gas_scale = parameters::get_gas_scale(&state)?;
 
     // Wrapper dry run to allow estimating the gas cost of a transaction
-    let (wrapper_hash, extended_tx_result, tx_gas_meter) =
+    let (wrapper_hash, extended_tx_result, tx_gas_meter, gas_used) =
```
grarco, line 40: The modifications to this file are just temporary to fix a bug for which we were not tracking gas correctly. A better implementation is provided in PR #3758  which is based on this one.

yito88 (approved):

Fraccaman (approved):

PR #3752, opened 2024-09-04
Don't build namadar

brentstone:
## Describe your changes
Just don't build it anymore, can easily incorporate back into `namada_apps` if we decide to use it in the future.

## Checklist before merging 


crates/apps/src/bin/namada/cli.rs
```diff
@@ -58,7 +57,7 @@ fn handle_command(cmd: cli::cmds::Namada, raw_sub_cmd: String) -> Result<()> {
         }
         cli::cmds::Namada::Wallet(_) => handle_subcommand("namadaw", sub_args),
         cli::cmds::Namada::Relayer(_) | cli::cmds::Namada::EthBridgePool(_) => {
```
sug0, line 59: `cli::cmds::Namada::EthBridgePool` is still reachable from `namadac` I think, so we need to handle it separately

crates/apps/Cargo.toml
```diff
@@ -37,12 +37,6 @@ doc = false
 name = "namadaw"
 path = "src/bin/namada-wallet/main.rs"
 
-# Namada relayer
-[[bin]]
-doc = false
-name = "namadar"
-path = "src/bin/namada-relayer/main.rs"
-
```
sug0, line 45: isn't this change alone sufficient to disable the build? without any other changes to rust source files

sug0 (changes requested):

crates/apps/Cargo.toml
brentstone: I tried that at first but then namada still built it, just with the name `namada-relayer` instead of `namadar`

crates/apps/src/bin/namada/cli.rs
brentstone: within this PR or in a different one? Also, how should this be handled?

crates/apps/Cargo.toml
sug0: [0001-Disable-the-Namada-relayer.patch](https://github.com/user-attachments/files/16886236/0001-Disable-the-Namada-relayer.patch)

Try applying these changes over a clean `main`

brentstone: thx this works. Changed

sug0 (approved):

PR #3758, opened 2024-09-05
Improve dry run

### LINKED ISSUE #3360: Refactor `dry_run`
Currently the `dry_run` command has some custom logic to evaluate a transaction. This sometimes leads to problems like #3358 where the logic of dry run doesn't match the one in protocol. We should try to refactor this to at least rely on the `dispatch_tx` function so that changes of a diverging behavior are minimized.

We could also think about mocking a one-transaction block and directly call `finalize_block`: this would enforce the same exact behavior even though it might slow down the simulation given the extra operations performed there. These operations though run only on new epoch so we might be able to side-step them by just mocking a block which is not the first one of a new epoch.

For reference: @Fraccaman, @yito88

grarco:
## Describe your changes

Closes #3360.

Fixes some bugs in the dry-run implementation and refactors it to reuse `dispatch_inner_txs` instead of rebuilding the same logic.

Also some misc updates to comments/error messages.

This PR is based on #3746 (diffs: https://github.com/anoma/namada/pull/3758/files/9aee33a0c44bf0ab70da18185727ba664d6b60d9..999a89c11a908a4a4130c05be4c34c3003dd4e23)

## Checklist before merging 

### DIFF
--- a/.changelog/unreleased/improvements/3530-balance-query-at-height.md
+++ b/.changelog/unreleased/improvements/3530-balance-query-at-height.md
@@ -1,0 +1,2 @@
+- Add optional height parameter to get_token_balance
+  ([\#3530](https://github.com/anoma/namada/pull/3530))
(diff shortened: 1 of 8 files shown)


tzemanovic (approved):

grarco:
@mergifyio requeue

PR #3700, opened 2024-08-23
Propagates error from `is_proposal_accepted`

grarco:
## Describe your changes

Closes #3238.

I've reviewed the code for our native vps and removed the wrong calls to `unwrap_or_default` applied to results (those on options are fine).

⚠️ **WARNING** ⚠️
Pay attention to merging this PR cause it carries breaking changes that span multiple vps, including especially the governance one: so even in the context of the multi-stage deployment of mainnet this could cause problems.

## Checklist before merging 
- [x] If this PR has some consensus breaking changes, I added the corresponding `breaking::` labels
    - This will require 2 reviewers to approve the changes

PR #3804, opened 2024-09-10
Fix transfer triggered vps

grarco:
## Describe your changes

Addresses a bug discovered while attempting shielded actions on campfire. With #2928 we stopped triggering vps for the addresses that do not occupy the first segment of a storage keys. This has an effect on transfers, since the balance keys have as their first address the multitoken account. Our current implementation of the `transfer` and `multi-transfer` prelude functions manually triggers the vp of the source, but we do nothing for the destination(s), so we are not triggering their vps right now.

This PR implements the followings:

- The multi token vp ensures that the vps of all the involved parties (sources and destinations) have been triggered
- The `transfer` and `multi-trasnfer` functions now also trigger the vps of the destinations
- The `execute` method of the ibc context triggers the vp of the involved address for all the packets we support
- The masp balance key is now part of the `is_masp_key` and `is_masp_transfer_key` to ensure correct validation
- Adds two new steps in the ibc test to test rejection of malformed transaction
- New unit test for the masp vp to verify that changing only the balance key is disallowed
- New unit test for the multitoken vp  to verify that the relevant verifiers are triggered

Misc:

- Removes another usage of `unwrap_or_default` in the pgf vp which was not fixed in #3700 
- Adds a clippy lint to prevent the usage of `unwrap_or_default` on errors to avoid the above issue

## Checklist before merging 


crates/governance/src/finalize_block.rs
```diff
@@ -324,6 +324,7 @@ where
         if vote.is_validator() {
             let vote_data = vote.data.clone();
 
+            #[allow(clippy::disallowed_methods)]
```
tzemanovic, line 327: let's propagate the error with `?` instead in here

crates/ibc/src/lib.rs
```diff
@@ -734,6 +778,70 @@ where
     }
 }
 
+// Extract the involved namada address from the packet (either sender or
+// receiver) to trigger its vp. Returns None if an address could not be found
```
tzemanovic, line 782:
```suggestion
/// Extract the involved namada address from the packet (either sender or
/// receiver) to trigger its vp. Returns None if an address could not be found
```

crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -384,6 +387,53 @@ fn ibc_transfers() -> Result<()> {
...
+        100,
+        &port_id_gaia,
+        &channel_id_gaia,
+        None,
+        // MASP VP shall reject it, make it timeout
+        Some(Duration::new(10, 0)),
+    )?;
+    wait_for_packet_relay(&port_id_namada, &channel_id_namada, &test)?;
+    // Check the balance didn't change
+    check_balance(&test, AA_VIEWING_KEY, &ibc_denom_on_namada, 40)?;
```
yito88, line 410: Should we also check the gaia source balance to be refunded? In the actual issue case, the receiver balance wasn't changed even though the source balance was decreased.
```
check_gaia_balance(&test_gaia, GAIA_USER, GAIA_COIN, 810)?;
```

crates/tests/src/e2e/ibc_tests.rs
```diff
@@ -384,6 +387,53 @@ fn ibc_transfers() -> Result<()> {
...
+        101,
+        &port_id_gaia,
+        &channel_id_gaia,
+        Some(shielding_data_path),
+        // MASP VP shall reject it, make it timeout
+        Some(Duration::new(10, 0)),
+    )?;
+    wait_for_packet_relay(&port_id_gaia, &channel_id_gaia, &test_gaia)?;
+    // Check the balances didn't change
+    check_balance(&test, AA_VIEWING_KEY, &ibc_denom_on_namada, 40)?;
```
yito88, line 435: same as above
```
check_gaia_balance(&test_gaia, GAIA_USER, GAIA_COIN, 810)?;
```

crates/tests/src/e2e/ibc_tests.rs
grarco: That would be great yes. I've tried myself but I couldn't make the test pass (I kept seeing a balance of 710): would you have some time to try looking into it? I'm not very familiar with the relayer mechanism and our tests

grarco: Thanks! 🙏🏻

yito88: In my local, the checking the Gaia balance passed. The CI also passed without errors.

yito88 (approved):

tzemanovic (approved):

PR #3835, opened 2024-09-20
add tests & refactor tx crate

### LINKED ISSUE #3200: Refactor inner `Tx`s interface
what if we used something like:
```rs
enum TxCommitmentRef {
    Id(TxCommitments),
    Hash(Hash),
    Index(usize),
}

impl TxCommitmentRef {
    fn get_commitments(
        &self,
        batch: impl IntoIterator<Item = &TxCommitments>,
    ) -> Option<&TxCommitments> {
        match self {
            Self::Id(commitments) => {
                // FIXME: check that `commitments` is present in `batch`?
                commitments
            }
            Self::Hash(commitments_hash) => batch.into_iter().find_map(|commitments| {
                (hash(commitments) == commitments_hash).then_some(commitments)
            }),
            Self::Index(index) => batch.into_iter().nth(index),
        }
    }
}
```

_Originally posted by @sug0 in https://github.com/anoma/namada/pull/3103#discussion_r1589154606_

tzemanovic:
## Describe your changes

related to #1497 

closes #3200

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/tx/src/types.rs
```diff
@@ -2025,8 +1014,493 @@ mod test {
...
+            InnerTxRef::CommitmentHash(Hash::default()),
+            InnerTxRef::Index(1),
+        ] {
+            assert!(tx.code(cmt.clone()).is_none());
+            assert!(tx.data(cmt.clone()).is_none());
+            assert!(tx.memo(cmt).is_none());
+        }
+
+        // Set inner tx data
+        let data_bytes = "bingbong".as_bytes();
```
grarco, line 1355: 🤣

crates/tx/src/types.rs
```diff
@@ -1855,6 +809,17 @@ impl<'tx> Tx {
     }
 }
 
+/// A key to select an inner tx and its sections
+#[derive(Debug, Clone)]
+pub enum InnerTxRef<'a> {
+    /// A direct ref to tx commitments
+    Commitment(&'a TxCommitments),
```
grarco, line 816: If this was an owned value instead of a ref we could use `InnerTx` inside both `BatchedTx` and `BatchedTxRef` in place of `TxCommitments` but this would imply many more changes and those two struct are only used within the protocol, don't know if you have a strong opinion on this

grarco:
Also there are a few functions in here that take as an input a reference to the commitments using `Either`

https://github.com/anoma/namada/blob/760462b434e80e9300ea0352e379708ecb40f757/crates/tx/src/data/mod.rs#L172 

I think we can use the new `InnerTxRef` now

tzemanovic:
closing in favor of #3868

Issue #3828, opened 2024-09-18
Test tx batches


brentstone:
Should prioritize some extra testing for batched txs.

grarco:
Especially atomic batches

PR #3840, opened 2024-09-23
Masp tests

### LINKED ISSUE #3836: Test masp edge cases
We should try to test some masp edge cases:

- Test a block (or even batch) with one shielded action and one MASP tx in both orders and check that the shielded sync command always reconstruct the correct state
- Test a batch containing two masp txs, one successful and one failing (in both orders) and check that the protocol and the shielded sync command work as expected. Also test with an atomic batch

Also test some malicious masp transactions:

- A non-MASP transaction that carries a MASP section (check both the case as-is and the one where the tx manually triggers the MASP VP)
- A valid masp tx that carries more than one masp section (check that both the protocol and the shielded sync command ignore the unused masp sections)
- A valid shielded action tx that also carries a masp section (same checks as above)

grarco:
## Describe your changes

Partially addresses #3836.
Partially addresses #3828.

Adds more integration test for the MASP including some tests with tx batches.
Adds some protests for the MASP vp.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/vp.rs
+++ b/crates/shielded_token/src/vp.rs
@@ -885,8 +885,9 @@ mod shielded_token_tests {
     use std::collections::BTreeSet;
 
     use namada_core::address::testing::nam;
-    use namada_core::address::MASP;
+    use namada_core::address::{IBC, MASP};
     use namada_core::borsh::BorshSerializeExt;
+    use namada_core::masp::TokenMap;
     use namada_gas::{TxGasMeter, VpGasMeter};
     use namada_state::testing::TestState;
     use namada_state::{StateRead, TxIndex};
@@ -898,6 +899,9 @@ mod shielded_token_tests {
     use namada_vm::wasm::VpCache;
     use namada_vm::WasmCacheRwAccess;
     use namada_vp::native_vp::{self, CtxPostStorageRead, CtxPreStorageRead};
+    use namada_vp_env::Error;
+
+    use crate::storage_key::masp_token_map_key;
 
     type CA = WasmCacheRwAccess;
     type Eval<S> = VpEvalWasm<<S as StateRead>::D, <S as StateRead>::H, CA>;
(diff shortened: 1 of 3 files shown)


crates/shielded_token/src/vp.rs
```diff
@@ -978,4 +988,157 @@ mod shielded_token_tests {
...
+                &ctx,
+                &tx.batch_ref_tx(&cmt),
+                &keys_changed,
+                &verifiers
+            )
+            .is_ok());
+        }
+
+        // Changing unknown masp keys is not allowed
+        #[test]
```
batconjurer, line 1097: fmt doesn't work well inside the proptest macro. This will have to be beautified manually

crates/tests/src/integration/masp.rs
```diff
@@ -3758,3 +3726,698 @@ fn identical_output_descriptions() -> Result<()> {
...
+                CHRISTEL_KEY,
+                "--output-folder-path",
+                tempdir.path().to_str().unwrap(),
+                "--dump-tx",
+                "--ledger-address",
+                validator_one_rpc,
+            ],
+        )?;
+        node.assert_success();
+        let file_path = tempdir
```
batconjurer, line 3798: you love to see it

crates/tests/src/integration/masp.rs
```diff
@@ -3758,3 +3726,698 @@ fn identical_output_descriptions() -> Result<()> {
...
+        txs.push(batched_tx.to_bytes());
+    }
+
+    node.clear_results();
+    node.submit_txs(txs);
+
+    // Check the block result
+    {
+        let codes = node.tx_result_codes.lock().unwrap();
+        // If empty than failed in process proposal
```
batconjurer, line 3862:
```suggestion
        // If empty then failed in process proposal
```

crates/tests/src/integration/masp.rs
```diff
@@ -3758,3 +3726,698 @@ fn identical_output_descriptions() -> Result<()> {
...
```
batconjurer, line 4094:
```suggestion
        // If empty then failed in process proposal
```

batconjurer (approved):

Issue #3883, opened 2024-10-03
Improve batching API


grarco:
We currently expose from the SDK the following function to batch transactions together.

https://github.com/anoma/namada/blob/418ef64b9e65bcf381faf4d37391d3a5ce6fdb2a/crates/sdk/src/tx.rs#L2887-L2889

There are some issues with the API of this function. First, we require a collections of `Tx`s but what we need, for every inner tx, is just:

- The `TxCommitments`
- A `Vec<Section>`

The second issue is that the resulting tx will have the `Header` of the first transaction in the provided argument, but this is unclear to the caller. We should instead add another argument to provide a specific `Header` (also because from the previous point we won't have any header anymore).

Finally, the function avoids returning duplicated `SigningTxData` instances to prevent the caller from attaching duplicated signatures. Unfortunately, every `SigningTxData` also carries the wrapper signer, so the caller might still end up signing the wrapper transaction more than once. This is not an issue so long as the intended gas payer has signed the wrapper, but it unnecessarily increase the gas cost of the batch. We should try to improve this even though it's not completely clear to me if there's anything we can do from within this function, especially because the `sign` function explicitly requires `SigningTxData`.

Some changes to `add_inner_tx` will also be required.

The changes proposed here are breaking but we might provide them in a non-breaking fashion by marking as deprecated the current involved functions and providing some new ones.

PR #3896, opened 2024-10-07
Avoid redundant wrapper signatures

grarco:
## Describe your changes

Temporary, non-breaking workaround for #3883.

Adds an SDK function to remove the duplicated wrapper signatures that might come from `build_batch`. Updates `batch_opt_reveal_pk_and_submit` to call such function.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/3896-avoid-redundant-sigs.md
+++ b/.changelog/unreleased/improvements/3896-avoid-redundant-sigs.md
@@ -1,0 +1,3 @@
+- The reveal pk batching process now attaches only one wrapper
+  signature. Added an SDK function to de-bloat wrapper signatures.
+  ([\#3896](https://github.com/anoma/namada/pull/3896))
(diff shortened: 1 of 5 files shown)


grarco:
@tzemanovic seems like we have a failure in the speculos test, could be because of the extra function called now for the signature but I'm not sure

tzemanovic:
> @tzemanovic seems like we have a failure in the speculos test, could be because of the extra function called now for the signature but I'm not sure

Yeah, it is that, but it's kinda odd - the first tx with bond has to be signed twice, but all the other txs (redeleg, unbonding and withdrawing) in that test have not changed

grarco:
@tzemanovic I'm closing this as in hindsight I believe it's better to pursue a cleaner solution such as #3883 (even if SDK breaking) rather than this non-breaking workaround

PR #3899, opened 2024-10-07
feat: Add new rpc query function to check validators' liveness(uptime)

Jeongseup:
## Describe your changes

Hello, I'm Jeongseup from Cosmostation. I hope the Namada mainnet genesis launch is successful. 

I believe it would be beneficial for each validator to monitor their own liveness(uptime) status during the current epoch to prevent slashsing or other penalties. 

Therefore, I have implemented a new RPC query function that retrieves the liveness status of validators who are active in this epoch. 

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/proof_of_stake/src/types/mod.rs
+++ b/crates/proof_of_stake/src/types/mod.rs
@@ -329,6 +329,28 @@ pub struct Redelegation {
     /// Redelegation amount
     pub amount: token::Amount,
 }
+
+/// Some liveness data for a consensus validator
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshDeserializer)]
+pub struct ValidatorLiveness {
+    /// Validator address
+    pub native_address: Address,
+    /// CometBFT address
+    pub comet_address: String,
+    /// Validator missed votes
+    pub missed_votes: u64,
+}
+
+/// Liveness data related to the network and set of consensus validators
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshDeserializer)]
+pub struct LivenessInfo {
+    /// Length of liveness window
+    pub liveness_window_len: u64,
+    /// Liveness threshold
+    pub liveness_threshold: Dec,
+    /// Validators' liveness info
+    pub validators: Vec<ValidatorLiveness>,
+}
 // --------------------------------------------------------------------------------------------
 
 /// A genesis validator definition.
--- a/crates/sdk/src/queries/vp/pos.rs
+++ b/crates/sdk/src/queries/vp/pos.rs
@@ -18,8 +18,9 @@ use namada_proof_of_stake::slashing::{
 };
 use namada_proof_of_stake::storage::{
     bond_handle, get_consensus_key, liveness_sum_missed_votes_handle,
-    read_active_validator_addresses, read_all_validator_addresses,
+    read_all_validator_addresses,
     read_below_capacity_validator_set_addresses_with_stake,
+    read_consensus_validator_set_addresses,
     read_consensus_validator_set_addresses_with_stake, read_pos_params,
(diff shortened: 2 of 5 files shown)


sirouk (commented):
🔥

crates/proof_of_stake/src/types/mod.rs
```diff
@@ -329,6 +329,28 @@ pub struct Redelegation {
     /// Redelegation amount
     pub amount: token::Amount,
 }
+
+/// Some liveness data for a consensus validator
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshDeserializer)]
```
Jeongseup, line 334: Really good Thanks!

Jeongseup:
@brentstone Thank you for checking my PR and improve the codes. I've learned rust from you 🙏

crates/sdk/src/queries/vp/pos.rs
```diff
@@ -249,6 +254,45 @@ where
...
+
+    let mut result = vec![];
+    for validator in consensus_validators {
+        if let Some(pubkey) = get_consensus_key::<_, governance::Store<_>>(
+            ctx.state, &validator, epoch,
+        )? {
+            let comet_address = tm_consensus_key_raw_hash(&pubkey);
+            let sum_liveness_handle = liveness_sum_missed_votes_handle();
+            if let Some(missed_votes) =
+                sum_liveness_handle.get(ctx.state, &validator)?
```
tzemanovic, line 278: nit: this will omit validators that have not missed any votes but the comments say "all consensus validators". Let's either update the comments or default this to 0

Fraccaman: agree, I would prefer to include all validators and default to 0

brentstone: fixed in [16e92a8](https://github.com/anoma/namada/pull/3899/commits/16e92a8e716feca13113d7956082fd41b65043a2)

Jeongseup: Actually I considered the value to default like 0, but it has a posibility to validators about misunderstand. the default 0 means for validator to haven't any missed. So I chose to exclude the value rather than the default value if this command(`sum_liveness_handle.get(ctx.state, &validator)?`) failed for some reason.

crates/sdk/src/queries/vp/pos.rs
```diff
@@ -249,6 +254,44 @@ where
...
+where
+    D: 'static + DB + for<'iter> DBIter<'iter> + Sync,
+    H: 'static + StorageHasher + Sync,
+{
+    let epoch = ctx.state.in_mem().last_epoch;
+    let consensus_validators =
+        read_consensus_validator_set_addresses(ctx.state, epoch)?;
+    let params = read_pos_params::<_, governance::Store<_>>(ctx.state)?;
+
+    let mut result = vec![];
```
tzemanovic, line 270: one more small thing, now that the size of this vec is same as `consensus_validators.len()` we can allocate it to its size

tzemanovic, line 270:
```suggestion
    let mut result = Vec::with_capacity(consensus_validators.len());
```

tzemanovic (approved):

crates/sdk/src/queries/vp/pos.rs
tzemanovic: if the result of this is `None` rather than `Some(missed_votes)` it means that there are no missed votes, not a failure in a command (a possible failure is handled via `?` at the end of the call)

Jeongseup: Ok, understood Thanks @tzemanovic

PR #3926, opened 2024-10-17
write_log: iter batch logs in insertion order

tzemanovic:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/state/src/write_log.rs
+++ b/crates/state/src/write_log.rs
@@ -699,7 +699,6 @@ impl WriteLog {
         for (key, modification) in self.block_write_log.iter().chain(
             self.batch_write_log
                 .iter()
-                .rev()
                 .flat_map(|batch_log| batch_log.write_log.iter()),
         ) {
             if key.split_prefix(prefix).is_some() {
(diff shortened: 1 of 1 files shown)


crates/state/src/write_log.rs
```diff
@@ -1153,49 +1151,134 @@ mod tests {
...
-                // Test for 4.
-                let vp_key = storage::Key::validity_predicate(initialized_account);
-                assert!(changed_keys.contains(&vp_key));
+            // And then commit them to batch
+            write_log.commit_tx_to_batch();
+
+            // Collect all the keys in batch and tx logs
+            let keys: HashSet<storage::Key> = keys
+                .into_iter()
+                .chain(write_log.tx_write_log.write_log.keys().cloned())
```
grarco, line 1263: I'm not sure if we are actually extending the keys here because `write_log.tx_write_log` is taken in `commit_tx_to_batch` to extend `write_log.batch_write_log`. Should we instead read again from `batch_write_log` like we do at line 1233?

grarco: Or maybe we can just clone the content of `tx_write_log` before calling `commit_tx_to_batch`. We could also check that the content of `batch_write_log` after this call is the result of merging the previous value of `keys` with the content of `tx_write_log`

tzemanovic: ah good point, I had this originally before last force push. I'll add it back in

batconjurer (approved):

grarco (approved):

Issue #3927, opened 2024-10-17
`namadac query-protocol-parameters` shows tx_allowlist instead of vp_allowlist


iskay:
Due to a typo, the command `namadac query-protocol-parameters` displays the tx_allowlist for both txs and vps.

`param_storage::get_tx_allowlist_storage_key()` used both times here:
https://github.com/anoma/namada/blob/v0.44.1/crates/apps_lib/src/client/rpc.rs#L729

PR #3939, opened 2024-10-21
Use correct key for `vp_allowlist` query

pysel:
## Describe your changes

Solves: #3927 

A query used a wrong key for a query


brentstone:
Going to close this in favor off #3940, where I've used your commit! Thx

PR #3954, opened 2024-10-28
Murisi/fix masp change

murisi:
## Describe your changes
An attempt to fix the MASP so that rewards can always be withdrawn without restrictions and to also enable the unshielding of untimestamped assets in all circumstances (including those where an asset is subsequently removed from the reward set). More specifically, the following changes have been made:
* Removed the normed asset computations because they were harder to reason about and encouraged mistakes in their maintenance
* Where possible, replaced computations over sums of `AssetType`s with computations over sums of `Address`es to reduce the chances that mismatched epochs cause parts of a balance to be hidden
* Adjusted the MASP fee UX so that users always have to specify which single MASP key they want to use to pay the fees. This is done instead of trying to construct this value from left over change
* Removed the fee computation logic and subsumed it into the logic for constructing ordinary transaction inputs and outputs. The purpose was to simplify code maintenance.
* Added integration tests to verify that MASP rewards can be withdrawn without withdrawing the principle and to also verify that the principal amount can be withdrawn though it is untimestamped

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/cli.rs
+++ b/crates/apps_lib/src/cli.rs
@@ -4661,16 +4661,14 @@ pub mod args {
                 });
             }
 
-            let gas_spending_keys = self
-                .gas_spending_keys
-                .iter()
-                .map(|key| chain_ctx.get_cached(key))
-                .collect();
+            let gas_spending_key = self
+                .gas_spending_key
+                .map(|key| chain_ctx.get_cached(&key));
 
             Ok(TxShieldedTransfer::<SdkTypes> {
                 tx,
                 data,
-                gas_spending_keys,
+                gas_spending_key,
                 disposable_signing_key: self.disposable_signing_key,
                 tx_code_path: self.tx_code_path.to_path_buf(),
             })
@@ -4691,16 +4689,13 @@ pub mod args {
                 token,
                 amount,
             }];
-            let mut gas_spending_keys = vec![];
-            if let Some(key) = GAS_SPENDING_KEY.parse(matches) {
-                gas_spending_keys.push(key);
-            }
+            let gas_spending_key = GAS_SPENDING_KEY.parse(matches);
             let disposable_gas_payer = DISPOSABLE_SIGNING_KEY.parse(matches);
 
             Self {
                 tx,
                 data,
-                gas_spending_keys,
+                gas_spending_key,
                 disposable_signing_key: disposable_gas_payer,
                 tx_code_path,
             }
@@ -4830,16 +4825,14 @@ pub mod args {
                     amount: transfer_data.amount,
                 });
             }
-            let gas_spending_keys = self
-                .gas_spending_keys
-                .iter()
-                .map(|key| chain_ctx.get_cached(key))
-                .collect();
+            let gas_spending_key = self
+                .gas_spending_key
+                .map(|key| chain_ctx.get_cached(&key));
 
             Ok(TxUnshieldingTransfer::<SdkTypes> {
                 tx,
                 data,
-                gas_spending_keys,
+                gas_spending_key,
                 disposable_signing_key: self.disposable_signing_key,
                 source: chain_ctx.get_cached(&self.source),
                 tx_code_path: self.tx_code_path.to_path_buf(),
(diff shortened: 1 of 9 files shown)


crates/core/src/uint.rs
```diff
@@ -614,33 +616,30 @@ impl I256 {
         Self(MAX_SIGNED_VALUE)
     }
 
-    /// Attempt to convert a MASP-denominated integer to an I256
-    /// using the given denomination.
+    /// Given a u128 and [`MaspDigitPos`], construct the corresponding
```
grarco, line 619: We are actually using an `i128`

crates/shielded_token/src/masp.rs
```diff
@@ -118,21 +116,16 @@ struct MaspTargetTransferData {
 pub struct MaspDataLog {
     pub source: Option<TransferSource>,
     pub token: Address,
-    pub amount: token::DenominatedAmount,
+    pub amount: token::Amount,
```
grarco, line 119: This thing is causing issues here #3961. It seems like without the denomination we interpret the amounts wrong leading the huge values

crates/tests/src/integration/masp.rs
```diff
@@ -2369,6 +2369,63 @@ fn dynamic_assets() -> Result<()> {
...
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains(TX_APPLIED_SUCCESS));
+
+    // sync the shielded context
+    run(
+        &node,
+        Bin::Client,
+        vec!["shielded-sync", "--node", validator_one_rpc],
+    )?;
```
grarco, line 2402: Can we also assert the balance is 0?

crates/tests/src/integration/masp.rs
```diff
@@ -2369,6 +2369,63 @@ fn dynamic_assets() -> Result<()> {
...
+                "2",
+                "--gas-payer",
+                BERTHA_KEY,
+                "--ledger-address",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains(TX_APPLIED_SUCCESS));
```
grarco, line 2427: Same here, can we assert the post balance?

crates/tests/src/integration/masp.rs
```diff
@@ -3208,7 +3269,7 @@ fn masp_fee_payment_with_custom_spending_key() -> Result<()> {
         )
     });
     assert!(captured.result.is_ok());
-    assert!(captured.contains("nam: 0"));
+    assert!(captured.contains("nam: 1000"));
```
grarco, line 3272: Could this be due to the `DenominatedAmount` thing I mentioned in the other file?

grarco, line 3272: Are these changes to due to the new way we handle the changes without the `Changes` type?

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -1100,6 +1071,25 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+        token: &Address,
+    ) -> Result<Denomination, TransferErr> {
+        if let Some(denom) = denoms.get(token) {
+            Ok(*denom)
+        } else if let Some(denom) = Self::query_denom(client, &token).await {
+            denoms.insert(token.clone(), denom);
+            Ok(denom)
+        } else {
+            Err(TransferErr::General(format!(
+                "denomination for token {token}"
```
grarco, line 1088: Maybe we can improve this message

crates/apps_lib/src/cli.rs
```diff
@@ -4860,17 +4853,14 @@ pub mod args {
                 token,
                 amount,
             }];
-            let mut gas_spending_keys = vec![];
-            if let Some(key) = GAS_SPENDING_KEY.parse(matches) {
-                gas_spending_keys.push(key);
-            }
+            let gas_spending_key = GAS_SPENDING_KEY.parse(matches);
```
grarco, line 4856: Can we make `DISPOSABLE_SIGNING_KEY` require `GAS_SPENDING_KEY`? This is indeed required with the modifications of this PR but at the moment we don't have a constraint on the cli args. If people forget to manually pass the gas spending key they will stumble into a generic error while contracting the tx

PR #3969, opened 2024-11-04
Removes `disposable-gas-payer` arg for custom txs

### LINKED ISSUE #3963: Remove `disposable-gas-payer` from `TxCustom`
In the cli we allow passing in a `--disposable-gas-payer` when loading a tx from storage for the `tx` command. It doesn't really make sense though to allow for it: if the previously dumped tx (or the first tx of the batch) is indeed doing masp fee payment, then it must be already pointing to a disposable address (generated when the transaction was constructed) that won't match the one we'll generate here and should instead be passed manually. If the tx we are loading does not involve gas payment via the shielded pool instead, generating a disposable address would lead to a failure when submitting it since fees cannot be paid.

grarco:
## Describe your changes

Closes #3963.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


sug0 (approved):
lgtm

PR #4013, opened 2024-11-11
test/e2e: mainnet phases test

tzemanovic:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2808,3 +2819,426 @@ fn test_genesis_manipulation() -> Result<()> {
...
+    // The tokens should have been transferred
+    let balance = find_balance(
+        &test,
+        Who::Validator(0),
+        &gaia_token,
+        AA_VIEWING_KEY,
+        Some(&ibc_denom_on_namada),
+    )?;
+    dbg!("before proposal 4", balance);
+    assert_ne!(balance, 0);
```
tzemanovic, line 3124: this keeps failing and I don't know yet why

crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2808,3 +2820,407 @@ fn test_genesis_manipulation() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_mainnet_phases() -> Result<()> {
+    // Use 1 PGF steward
+    let steward = defaults::albert_address();
```
brentstone, line 2827: Mainnet will have no stewards at genesis, so we should do the same. Perhaps in this test we can elect one in Phase 1 or at some later point?

crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2808,3 +2820,407 @@ fn test_genesis_manipulation() -> Result<()> {
...
+        // There should be no tokens in PGF address
+        let balance = find_balance(
+            &test,
+            Who::Validator(0),
+            NAM,
+            PGF_ADDRESS,
+            None,
+            None,
+        )?;
+        assert_eq!(balance, token::Amount::zero());
```
brentstone, line 2973: can we also test other things in Phase 1 just for completeness? Such as:
- cannot transfer any NAM
- cannot IBC transfer
- can bond, unbond, and withdraw

crates/tests/src/e2e/ledger_tests.rs
```diff
@@ -2808,3 +2820,407 @@ fn test_genesis_manipulation() -> Result<()> {
...
+        &test_gaia,
+        GAIA_USER,
+        &namada_receiver,
+        GAIA_COIN,
+        1,
+        &port_id_gaia,
+        &channel_id_gaia,
+        None,
+        None,
+    )?;
```
brentstone, line 3007: does this  function ensure that the transfer fails?

brentstone:
In general, I think we should be testing lots of actions in each phase to make sure what does and doesn't work as intended. For example, I wrote on the Namada forum some recommendations for users to test during Phase 3 and 4 of the dry-run:
- https://forum.namada.net/t/dry-run-phase-3-gov-proposal/1085/2?u=brentstone
- https://forum.namada.net/t/dry-run-phase-4-gov-proposal/1086/2?u=brentstone

crates/tests/src/e2e/ledger_tests.rs
tzemanovic: this fn doesn't fail - it gets submitted to the relayer where it hangs before timing out (or if the next phase begins first, it goes through). The balance check below that ensures that the transfer doesn't get completed before the next phase is activated

crates/tests/src/e2e/ledger_tests.rs
tzemanovic: yup totally

brentstone:
Hey @tzemanovic, what's the status here of this?

crates/tests/src/e2e/ledger_tests.rs
tzemanovic: applied this suggestion in last version

crates/tests/src/e2e/ledger_tests.rs
tzemanovic: added these too

PR #3578, opened 2024-08-01
Shielded sync phase 3 client

sug0:
## Describe your changes

Closes #3385

## Checklist before merging to `draft`
- [x] I have added a changelog
- [x] Git history is in acceptable state

PR #4016, opened 2024-11-12
Improve the performance of the shielded sync ledger client

sug0:
## Describe your changes

The shielded sync's ledger client continues to perform poorly, even after #3578. This PR introduces some optimizations to the ledger client, and improves its UX a bit.

- Fetches are now logged more frequently on the progress bar. Previously, they weren't being logged because concurrent fetches were starving each other on the first semaphore. To fix this, `fetch_shielded_transfers` now has a single semaphore point, at the top of the method, rather than two.
- Errors returned from RPC servers are no longer printed. They weren't that helpful anyway, and were spamming the screen, obfuscating the output of the progress bars.
- A linear backoff sleep has been added to the ledger client, on errors, to allow the client to recover from potential rate limiting/throttling imposed by the CometBFT RPC server. On successful fetches, the backoff is reduced, returning to the normal/optimal fetch speed.
- The block batch size parameter has been exposed to the CLI and SDK, allowing users to specify the number of blocks that are fetched per concurrent fetch job. This may help performance, in some cases.

In general, the indexer masp client still performs about 10x better than the ledger masp client, while fetching shielded txs.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
+++ b/crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
@@ -702,7 +702,6 @@ where
         }
 
         if self.config.retry_strategy.may_retry() {
-            self.config.fetched_tracker.message(format!("{error}"));
             true
         } else {
             // NB: store last encountered error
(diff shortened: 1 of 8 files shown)


crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -702,7 +702,7 @@ where
         }
 
         if self.config.retry_strategy.may_retry() {
-            self.config.fetched_tracker.message(format!("{error}"));
+            // self.config.fetched_tracker.message(format!("{error}"));
```
batconjurer, line 705: remove?

grarco (approved):

PR #4018, opened 2024-11-13
Avoid updating merkle paths of spent notes

sug0:
## Describe your changes

Avoid updating merkle paths of spent notes. This should optimize the synchronous path of the shielded sync on the ledger client.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4018-update-witness-map-on-spend.md
+++ b/.changelog/unreleased/improvements/4018-update-witness-map-on-spend.md
@@ -1,0 +1,3 @@
+- Avoid updating merkle paths of spent notes. This should optimize
+  the synchronous path of the shielded sync on the ledger client.
+  ([\#4018](https://github.com/anoma/namada/pull/4018))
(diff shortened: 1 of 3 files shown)


grarco (approved):

batconjurer (approved):

sug0:
I think this code might be subtly wrong... The nullifier map (`nf_map`) only gets updated after calling `save_decrypted_shielded_outputs`, but `save_shielded_spends` depends on the `nf_map`to add data to the set of spent notes (`spents`).

The point of reordering the call to `save_shielded_spends` in this PR was to remove entries from the witness map before new witnesses were appended during a call to `update_witness_map`. Might be best to remove the outdated witnesses inside of `update_witness_map` instead.

murisi (approved):
This code looks okay to me if there is no special interaction between this optimization and the speculation code (which I've not looked into before). `save_shielded_spends` being called earlier should be okay since the new notes saved in `save_decrypted_shielded_outputs` can only be nullified by a `Transaction` coming after the one being processed in the current loop iteration.

PR #4030, opened 2024-11-14
Optionally reveal PK for voter if not done so during a proposal vote

### LINKED ISSUE #4029: Voting as a (pre-genesis?) delegator fails
I tried voting with my validator and two delegators keys. One of these delegator keys has only pre-genesis bonds, while the other also has post genesis bonds.

Voting with the delegator key that only has pre-genesis bonds fails with: "Invalid transaction section signature: RuntimeError: The section signature is invalid: signature threshold not met: (0 < 1)".

The rest does succeed.

I have to add that the delegator key with post genesis bonds also failed before I added extra bonds so that's why I suspect this to be the issue.

Command used: `namadac vote-proposal --vote yay --address alias --proposal-id 2`

brentstone:
## Describe your changes
Closes #4029

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/tx.rs
+++ b/crates/apps_lib/src/client/tx.rs
@@ -1053,14 +1053,18 @@ pub async fn submit_vote_proposal<N: Namada>(
 where
     <N::Client as namada_sdk::io::Client>::Error: std::fmt::Display,
 {
-    let (mut tx_builder, signing_data) = args.build(namada).await?;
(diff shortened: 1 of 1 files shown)


tzemanovic:
I think we could also improve the error message for when a PK is not revealed - the current error message is misleading

PR #4039, opened 2024-11-15
Improve tx result message

### LINKED ISSUE #4020: Wrong tx response output
It seems like the `display_batch_resp` function is producing a slightly misleading output.

https://github.com/anoma/namada/blob/cb1e963235665d28d5d82e80004b7b6861877da7/crates/sdk/src/tx.rs#L435-L483

More specifically:

- The `gas units` displayed actually refer to the entire batch and not the single inner tx, we should move this log to the batch level
- In case of `InnerTxResult::OtherFailure` we display the entire result. In case of multiple failures we could end up spamming the error stream of the user

grarco:
## Describe your changes

Closes #4020.

`display_batch_resp` now logs some information at the batch level followed by the results of the single inner txs. In case of `OtherFailure` we only print the error message for that specific failure instead of the entire tx result.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4039-adjust-tx-response-output.md
+++ b/.changelog/unreleased/improvements/4039-adjust-tx-response-output.md
@@ -1,0 +1,2 @@
+- Improved the display of transactions' results.
+  ([\#4039](https://github.com/anoma/namada/pull/4039))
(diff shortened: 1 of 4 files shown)


grarco:
I need to fix some assertions in e2e tests

grarco:
@tzemanovic @brentstone ok this is now ready for review

tzemanovic (approved):

Issue #3901, opened 2024-10-08
Rework wrapper signature workflow


grarco:
We define in the SDK `SigningTxData`, which is a support type to sign transactions, as follows:

https://github.com/anoma/namada/blob/7ac415a68791245e15c22ba551b1fddb301dfbcb/crates/sdk/src/signing.rs#L65-L76

After #3883 and #3900 it seems like the `fee_payer` field is too limited. We don't always want to sign the wrapper transaction or sometimes we might have an offline signature that we'd want to use instead of producing a new one. This seems to lead to some workarounds in the codebase to allow these different workflows: we should think about reworking this piece of data to support these cases (no signature or load a signature).

Also, as a minor note, there are a couple of function (`tx_signers` and `aux_signing_data`) that specifically handle the MASP case: we should see if we can collapse the MASP case to some other case to avoid this specific handling (like not passing an owner when the source of a tx is the MASP)

grarco:
In addition, we should also review the `signatures` and `wrapper_signature` fields of `args::Tx`. We should be using these only when loading a tx from storage (with the `tx` command), so maybe it would be better to take them out of the general `Tx` args and move them into the specific args for `TxCustom`

PR #4055, opened 2024-11-20
wrap tx by elsewho

### LINKED ISSUE #4034: test a raw tx and wrapped tx signed by different account holder
Add a test case for the following use case:

- Have one address with no tokens left for gas fees for txs
- build and sign a raw tx without a wrapper
- give the tx to the generous-address holder to pay for the gas fee to wrap, sign and submit tx

tzemanovic:
## Describe your changes

closes #4034

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/tests/src/integration/ledger_tests.rs
+++ b/crates/tests/src/integration/ledger_tests.rs
@@ -2534,7 +2534,7 @@ fn wrap_tx_by_elsewho() -> Result<()> {
                 &tx,
                 "--signatures",
                 &offline_sig,
-                "--signing-keys",
+                "--gas-payer",
                 CHRISTEL_KEY,
             ]),
         )
--- a/crates/tests/src/integration/ledger_tests.rs
+++ b/crates/tests/src/integration/ledger_tests.rs
@@ -24,6 +24,7 @@ use namada_node::shell::SnapshotSync;
 use namada_node::storage::DbSnapshot;
 use namada_sdk::account::AccountPublicKeysMap;
 use namada_sdk::collections::HashMap;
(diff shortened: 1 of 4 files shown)


crates/tests/src/integration/ledger_tests.rs
```diff
@@ -2353,6 +2354,217 @@ fn scheduled_migration() -> Result<()> {
...
+        run(
+            &node,
+            Bin::Client,
+            apply_use_device(vec![
+                "tx",
+                "--tx-path",
+                &tx,
+                "--signatures",
+                &offline_sig,
+                "--signing-keys",
```
grarco, line 2537: This is actually attaching an extra raw signature coming from the wrapper signer (together with the signature on the wrapper itself). I think it's fine for now, it shouldn't represent a problem but it would be nice to investigate further the behavior of `tx_signers` which seems like it's preventing the correct functioning of the `--gas-payer` argument (that I believe should be the correct way to achieve this flow).

https://github.com/anoma/namada/blob/3e405ce6c630ef82aaa0f3c71423d3a2fcb6fedf/crates/sdk/src/signing.rs#L168-L191

grarco (approved):
Left a minor comment but this PR looks good to me, thanks for testing this flow!

crates/tests/src/integration/ledger_tests.rs
grarco: Ah this might be related to #3901

grarco: Nevermind, your changes in d69279f solve this!

PR #4061, opened 2024-11-20
dump tx output for reveal_pk transactions when requested

dan-u410:
## Describe your changes
It is currently not possible to generate the `reveal_pk` transaction offline. This PR will dump the transaction when requested. 

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/tx.rs
+++ b/crates/apps_lib/src/client/tx.rs
@@ -1153,12 +1153,19 @@ pub async fn submit_reveal_pk<N: Namada>(
 where
     <N::Client as namada_sdk::io::Client>::Error: std::fmt::Display,
 {
-    let tx_data =
-        submit_reveal_aux(namada, &args.tx, &(&args.public_key).into()).await?;
+    if args.tx.dump_tx || args.tx.dump_wrapper_tx {
+        let tx_data =
+            tx::build_reveal_pk(namada, &args.tx, &args.public_key).await?;
+        tx::dump_tx(namada.io(), &args.tx, tx_data.0.clone())?;
+    } else {
+        let tx_data =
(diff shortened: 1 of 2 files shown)


dan-u410:
there is some complexity to the implementation:
- to construct the `reveal_pk` transaction body, the code calls out to a helper `submit_reveal_aux`
- this method has some internal complexity such as checking network state and possibly returning `none` in cases where the transaction isnt needed
- as an advanced user, id prefer to always get an output when requested
- so I call out to `build_reveal_pk` directly to ensure we get a tx when the user requests to dump the transaction bytes

there are a few ways to implement this so wanted to add color to my choice. happy to change.

crates/apps_lib/src/client/tx.rs
```diff
@@ -1153,12 +1153,31 @@ pub async fn submit_reveal_pk<N: Namada>(
 where
     <N::Client as namada_sdk::io::Client>::Error: std::fmt::Display,
 {
-    let tx_data =
-        submit_reveal_aux(namada, &args.tx, &(&args.public_key).into()).await?;
+    if args.tx.dump_tx || args.tx.dump_wrapper_tx {
+        if let Address::Implicit(ImplicitAddress(pkh)) =
+            &(&args.public_key).into()
+        {
+            let public_key = &namada
```
tzemanovic, line 1160: we don't need to look-up the PK as we already have it in `args.public_key`

tzemanovic: fixed in aaf76793f1521cdf3a34f166d166226df1229f64

tzemanovic (approved):
thx for the fix!

PR #4083, opened 2024-11-22
Updates `query-protocol-parameters` to display some missing params

grarco:
## Describe your changes

Updates `query-protocol-parameters` to display some missing params. Also refactored it to reconstruct the `Parameters` type to ensure we don't miss any.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


brentstone:
Should we also do the same for `vp_user` if we're going to do it for `vp_implicit`?

grarco:
> Should we also do the same for `vp_user` if we're going to do it for `vp_implicit`?

No that one is not a protocol parameter

brentstone (approved):

PR #4120, opened 2024-11-29
Fix offline wrapper signature by elsewho

grarco:
## Describe your changes

Fixes offline wrapper signatures and adds two new tests for this. The bug was only affecting offline wrapper signing when the signer did not own the keys to also sign the inner transaction:

- Once the inner tx had been dumped and signed offline the user had to reload the tx, wrap it and dump the wrapper
- At this point they needed to produce the signature for the wrapper offline. To do so, they needed to pass to the offline signing command both the serialized wrapper and the inner signatures (that were not attached to the tx when wrapped), but the command does not accept serialized signatures, just signing keys. So this flow was only possible if the wrapper signer also owned the keys to sign the inner transactions (effectively producing the inner signatures a second time)

The `tx` command now attaches the raw signatures to the tx when a `--dump-wrapper-tx` is requested to allow for a different wrapper signer.

The offline sign command and the custom tx one now return errors if a wrapper singer/signature is provided but the tx is not a wrapper.

Also removes a redundant `sign-tx` client command.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/apps_lib/src/client/tx.rs
```diff
@@ -283,9 +283,28 @@ pub async fn submit_custom<N: Namada>(
...
+                let sig = tx::SignatureIndex::try_from_json_bytes(bytes)
+                    .map_err(|err| {
+                        error::Error::Encode(error::EncodingError::Serde(
+                            err.to_string(),
+                        ))
+                    })?;
+                acc.push(sig);
+                Ok(acc)
+            },
+        )?;
```
tzemanovic, line 305: you can `collect` with a `Result`
```suggestion
        let signatures = args
            .tx
            .signatures
            .iter()
            .map(|bytes| {
                tx::SignatureIndex::try_from_json_bytes(bytes).map_err(|err| {
                    error::Error::Encode(error::EncodingError::Serde(
                        err.to_string(),
                    ))
                })
            })
            .collect::<error::Result<Vec<_>>>()?;
```

tzemanovic (approved):

crates/apps_lib/src/client/tx.rs
grarco: Done in 25c2564e893d76dcd2832d0b540f83d9fe4c0f3e

PR #4135, opened 2024-12-02
migration fixes

### LINKED ISSUE #4112: Invalid MerkleTreeStoresRead after restarting a 0.45 to 0.46 upgraded validator
After successfully upgrading a validator from v0.45 to v0.46 with the [dry-run upgrade instructions](https://forum.namada.net/t/coordinating-the-dry-run-upgrade-to-v0-46-0/1189) (while not making the mistake to pass a [HTML instead of JSON](https://github.com/anoma/namada/issues/4108)), restarting namadan results in a crash with the error Invalid MerkleTreeStoresRead.

```2024-11-29T01:01:31.068320Z  INFO namada_node: Done loading MASP verifying keys.
2024-11-29T01:01:31.068898Z  INFO namada_node::storage::rocksdb: Using 2 compactions threads for RocksDB.
2024-11-29T01:01:31.086428Z  INFO namada_node::broadcaster: Starting broadcaster.
Message:  Merkle tree should be restored: Custom(CustomError(MerkleTree("Invalid MerkleTreeStoresRead")))
Location: /home/runner/work/namada/namada/crates/state/src/wl_state.rs:586```

For context, this validator ran v0.45 until block 182000, then was upgraded to v0.46 successfully and ran until it timed out at 182002, where there was not enough VP online to reach consensus.

### PR DESCRIPTION
**Pull request by tzemanovic**
## Describe your changes

- check that migration file can be decoded on node start-up
- validate migration before starting to commit block to avoid leaving node in corrupt state on failure
- skips writing diffs for migration to close #4112 (extended e2e migration test to cover this)

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### CODE REVIEW

**Comment by tzemanovic** [comment]
ah, an integration test is failing

File: crates/node/src/storage/rocksdb.rs [hunk 1]
```diff
@@ -1817,113 +1816,130 @@ impl DB for RocksDB {
...
-        self.db
-            .overwrite_entry(&mut self.batch, cf, key, value)
+    fn write(
+        &mut self,
+        db: &Self::DB,
+        key: &Key,
+        cf: &DbColFam,
+        value: impl AsRef<[u8]>,
+    ) {
+        db.overwrite_entry(&mut self.batch, cf, key, value)
```
yito88, line 1902: It seems that it could be written to the diff storage even if the key should be for no diff.
Can we use `db.batch_write_subspace_val()` for the subspace keys in `overwrite_entry`?

tzemanovic: good question - unfortunately we cannot pass a column family to `batch_write_subspace_val` as it's only for subspaces, but I think we can extend `overwrite_entry` to pass through a flag to decide. We also need to do the same for `fn delete` below this one (currently it's hard-coded to `persist_diffs = true`)

tzemanovic: added [d59a87b](https://github.com/anoma/namada/pull/4135/commits/d59a87b1341a1e98b9437d1b703b3ebac3983dba)

crates/storage/src/mockdb.rs
```diff
@@ -641,9 +642,46 @@ impl DB for MockDB {
...
+        let epoch_key = format!("{prefix}/{EPOCH_KEY_SEGMENT}");
+        let epoch: Epoch = self.read_value(epoch_key)?.unwrap();
+
+        for st in StoreType::iter() {
+            if st.is_stored_every_block() {
+                let key_prefix = if st.is_stored_every_block() {
+                    tree_key_prefix_with_height(st, height)
+                } else {
+                    tree_key_prefix_with_epoch(st, epoch)
+                };
```
yito88, line 672: In this case, we can use always `tree_key_prefix_with_height`

tzemanovic: What about the case when the migration happens to run on a block during epoch change? Do we maybe also need to use `is_full_commit` for `update_last_block_merkle_tree` to correctly update the stores with `tree_key_prefix_with_epoch`?

yito88: ah, yes, we need to update the other trees when the epoch change.

tzemanovic: added in [ffefc11](https://github.com/anoma/namada/pull/4135/commits/ffefc11da05b9977856e1c5cd7f8e8194904cf7c)

grarco (approved):

yito88 (approved):

batconjurer (approved):

brentstone:
@tzemanovic what's the latest with this branch?

tzemanovic:
> @tzemanovic what's the latest with this branch?

it's rdy to merge whenever we want to include it

PR #4146, opened 2024-12-06
Fix masp rewards estimate

grarco:
## Describe your changes

Various adjustments to the masp rewards estimation logic:

- Fixed the logic to correctly account for rewards accrued by the native token and by assets that have been shielded in the current epoch (for which no conversions are available yet but the conversion from the previous epoch can be used as an estimate)
- Fixed the logic to allow estimating rewards for assets that have been shielded before the last epoch
- Changed the API of the estimate function to require a `ValueSum` instead of a viewing key (this is to support a UI view with the possible rewards for different tokens, cc @ChrisHoltDesign, @mateuszjasiuk)
- The function now returns a `DenominatedAmount` instead of an `i128`
- Changed the name of the cli command to `estimate-shielding-rewards`
- Updated and improved tests

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -407,11 +407,42 @@ pub async fn query_rewards_estimate(
 ) {
     let mut shielded = context.shielded_mut().await;
     let _ = shielded.load().await;
-    let rewards_estimate = shielded
-        .estimate_next_epoch_rewards(context, &args.owner.as_viewing_key())
+    let raw_balance = match shielded
+        .compute_shielded_balance(&args.owner.as_viewing_key())
         .await
-        .unwrap()
-        .unsigned_abs();
+    {
+        Ok(balance) => balance,
+        Err(e) => {
+            edisplay_line!(
+                context.io(),
+                "Failed to query shielded balance: {}",
+                e
+            );
+            cli::safe_exit(1);
+        }
+    };
+
+    let rewards_estimate = match raw_balance {
+        Some(balance) => {
+            match shielded
+                .estimate_next_epoch_rewards(context, &balance)
+                .await
+            {
+                Ok(estimate) => estimate.unsigned_abs(),
+                Err(e) => {
+                    edisplay_line!(
+                        context.io(),
+                        "Failed to estimate rewards for the next MASP epoch: \
+                         {}",
+                        e
+                    );
+                    cli::safe_exit(1);
+                }
+            }
+        }
+        None => 0,
+    };
+
     let rewards_estimate = DenominatedAmount::new(
         Amount::from_u128(rewards_estimate),
         NATIVE_MAX_DECIMAL_PLACES.into(),
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -50,9 +50,7 @@ use namada_sdk::rpc::{
 use namada_sdk::storage::BlockResults;
 use namada_sdk::tendermint_rpc::endpoint::status;
 use namada_sdk::time::DateTimeUtc;
-use namada_sdk::token::{
-    DenominatedAmount, MaspDigitPos, NATIVE_MAX_DECIMAL_PLACES,
-};
+use namada_sdk::token::{DenominatedAmount, MaspDigitPos};
 use namada_sdk::tx::display_batch_resp;
 use namada_sdk::wallet::AddressVpType;
 use namada_sdk::{error, state as storage, token, Namada};
@@ -428,7 +426,7 @@ pub async fn query_rewards_estimate(
                 .estimate_next_epoch_rewards(context, &balance)
                 .await
             {
-                Ok(estimate) => estimate.unsigned_abs(),
+                Ok(estimate) => estimate,
                 Err(e) => {
                     edisplay_line!(
                         context.io(),
@@ -440,13 +438,9 @@ pub async fn query_rewards_estimate(
                 }
             }
         }
-        None => 0,
+        None => DenominatedAmount::native(Amount::zero()),
     };
 
-    let rewards_estimate = DenominatedAmount::new(
-        Amount::from_u128(rewards_estimate),
-        NATIVE_MAX_DECIMAL_PLACES.into(),
-    );
     display_line!(
         context.io(),
         "Estimated native token rewards for the next MASP epoch: {}",
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -403,7 +403,7 @@ pub async fn query_proposal_by_id<C: Client + Sync>(
 /// Estimate MASP rewards for next MASP epoch
 pub async fn query_rewards_estimate(
     context: &impl Namada,
-    args: args::QueryRewardsEstimate,
+    args: args::QueryShieldingRewardsEstimate,
 ) {
     let mut shielded = context.shielded_mut().await;
     let _ = shielded.load().await;
--- a/crates/shielded_token/src/masp/shielded_wallet.rs
+++ b/crates/shielded_token/src/masp/shielded_wallet.rs
@@ -741,9 +741,9 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
         let next_epoch = current_epoch
             .next()
             .ok_or_else(|| eyre!("Overflowed MASP epoch"))?;
-        // Get the current amount including conversions, this serves as the
-        // reference point to estimate future rewards
-        let (current_exchanged_balance, mut conversions) = self
+        // Get the current amount, this serves as the reference point to
+        // estimate future rewards
+        let current_exchanged_balance = self
             .compute_exchanged_amount(
                 context.client(),
                 context.io(),
@@ -751,12 +751,15 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
                 current_epoch,
                 Conversions::new(),
             )
-            .await?;
+            .await?
+            .0;
 
-        // Query missing conversions first, i.e. those for assets that carry
-        // the current epoch. There are no conversions for these yet so we
-        // need to see if there are conversions for the previous epoch
-        for (asset_type, _) in raw_balance.components() {
+        let mut latest_conversions = Conversions::new();
+
+        // We need to query conversions for the latest assets in the current
+        // exchanged balance. For these assets there are no conversions yet so
+        // we need to see if there are conversions for the previous epoch
+        for (asset_type, _) in current_exchanged_balance.components() {
             let mut asset = match self
                 .decode_asset_type(context.client(), *asset_type)
                 .await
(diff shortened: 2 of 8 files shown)


crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -713,95 +709,228 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+                .0;
+            let mut est_conv = I128Sum::zero();
+            for ((_, mut asset_data), val) in decoded_conv.into_components() {
+                // We must upconvert the components of the conversion unless
+                // this conversion is for a non-native token and the component
+                // itself is the native token. In this case we should avoid
+                // upconverting it since conversions for non-native tokens must
+                // always refer to a specific epoch (MaspEpoch(0)) native token
+                if asset.token == native_token
+                    || asset_data.token != native_token
```
batconjurer, line 809:
```suggestion
                if !(asset.token != native_token
                    && asset_data.token == native_token)
```

batconjurer: This matches the comment more closely. I was confused until I wrote it out like this.

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -713,95 +709,228 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+
+        // Extract only the native token balance for rewards. Depending on
+        // rewards being enabled for the native token or not, we'll find the
+        // balance we look for at either epoch 0 or the current epoch. If
+        // rewards are not enabled for the native token we might have
+        // balances at any epochs in between these two but these would be
+        // guaranteed to be the same in both the exchanged amounts we've
+        // computed here above, meaning they are irrelevant for the
+        // estimation of the next epoch rewards since they would cancel out
+        // anyway
```
batconjurer, line 842:
```suggestion
        // Extract only the native token balance for rewards. Depending on
        // rewards being enabled for the native token or not, we'll find the
        // balance we look for at either epoch 0 or the current epoch. If
        // rewards are not enabled for the native token we might have
        // balances at any epochs in between these two. These would be
        // guaranteed to be the same in both the exchanged amounts we've
        // computed here above. This means they are irrelevant for the
        // estimation of the next epoch rewards since they would cancel out
        // anyway.
```

batconjurer: nit: run on sentence

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -713,95 +709,228 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+                        return Err(eyre!(
+                            "Found reward asset other than the native token"
+                        ));
+                    }
+                    match epoch {
+                        Some(ep) if ep == epoch_zero => (),
+                        _ => {
+                            return Err(eyre!(
+                                "Found reward asset with an epoch different \
+                                 than the current one"
```
batconjurer, line 920: This error message doesn't seem correct.

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2007,21 +2245,37 @@ mod test_shielded_wallet {
...
 
+                // If this flag is clear we test that the estimation logic can
+                // hanlde rewards for assets coming only from the current epoch,
+                // i.e. assets for which there are no conversions yet.
+                if shield_at_previous_epoch {
+                    let asset_data = AssetData {
+                        token: incentivized_token.clone(),
+                        denom: 0.into(),
+                        position: MaspDigitPos::Zero,
+                        epoch: Some(MaspEpoch::new(1)),
```
batconjurer, line 2270: isn't the current epoch 2?

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2032,11 +2286,184 @@ mod test_shielded_wallet {
...
+                    .await
+                    .unwrap()
+                    .unwrap_or_else(ValueSum::zero);
+                let rewards_est = wallet.estimate_next_epoch_rewards(&context, &balance)
+                    .await.expect("Test failed");
+                assert_eq!(
+                    rewards_est,
+                    DenominatedAmount::native(
+                        Amount::from_masp_denominated_i128(
+                            (1 + i128::from(shield_at_previous_epoch)) * reward_rate * i128::from(principal),
```
batconjurer, line 2303: I'm confused about the `1+` here.

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2032,11 +2286,184 @@ mod test_shielded_wallet {
...
+
+        /// In this test, we have a single incentivized token, the native one,
+        /// shielded at MaspEpoch(2), owned by the shielded wallet.
+        /// The amount of owned token is the parameter `principal`.
+        ///
+        /// We add a conversion from MaspEpoch(1) to MaspEpoch(2)
+        /// which issues `reward_rate` nam tokens for our incentivized token.
+        ///
+        /// Optionally, the test also shields the same `principal` amount at
+        /// MaspEpoch(1).
```
batconjurer, line 2318: Again, maybe I'm confused about what's being tested, but shouldn't you shield at epoch 2 (the current epoch)?

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2032,11 +2286,184 @@ mod test_shielded_wallet {
...
+                    .await
+                    .unwrap()
+                    .unwrap_or_else(ValueSum::zero);
+                let rewards_est = wallet.estimate_next_epoch_rewards(&context, &balance).await.expect("Test failed");
+                if shield_at_previous_epoch {
+                    assert_eq!(
+                        rewards_est,
+                        DenominatedAmount::native(
+                            Amount::from_masp_denominated_i128(
+                                (2 + reward_rate) * reward_rate * i128::from(principal),
```
batconjurer, line 2451: Another mysterious `2+`. Maybe write out the formula's meaning in a comment.

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2272,22 +2782,27 @@ mod test_shielded_wallet {
...
-                let principal2 = i128::from(principal2);
-                // reward from epoch 1->2 + reward from epoch 2->3 + reward from
-                // epoch 2->3
-                let expected_tok1_rewards =
-                    principal1 * tok1_reward_rate + principal1 + principal1;
-                // reward from epoch 2->3 + reward from epoch 2->3
-                let expected_tok2_rewards =
-                    principal2 * tok2_reward_rate + principal2 * tok2_reward_rate;
+                // The native token balances at epoch 3 and 4 are given by the shielded amounts times the rewards times the conversion for the native asset
+                let native_balance_at_3 = (i128::from(principal1) * (tok1_reward_rate + 1) + i128::from(principal2) * tok2_reward_rate) * 2;
```
batconjurer, line 2795: Wait, maybe I've got it. Is this supposed to return the balance, i.e. the principal + rewards? If so, then these mysterious `x+` make sense.

batconjurer (changes requested):

crates/apps_lib/src/client/rpc.rs
```diff
@@ -371,19 +369,46 @@ pub async fn query_proposal_by_id<C: Client + Sync>(
...
 pub async fn query_rewards_estimate(
     context: &impl Namada,
-    args: args::QueryRewardsEstimate,
+    args: args::QueryShieldingRewardsEstimate,
 ) {
     let mut shielded = context.shielded_mut().await;
     let _ = shielded.load().await;
-    let rewards_estimate = shielded
-        .estimate_next_epoch_rewards(context, &args.owner.as_viewing_key())
+    let raw_balance = match shielded
```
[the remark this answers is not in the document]
grarco: To allow passing arbitrary balances to that function so that we can use it in the interface to present the user with an estimate of the rewards they can expect

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: You're right the comment is a bit misleading, I changed that though instead of modifying the if condition

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Yes but if `shielded_at_previous_epoch` is **clear** then we only shield at the current epoch (2), if it is set instead we also shield at the previous one, let me rephrase the comment

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: So the expected rewards are just `principal * reward_rate` but if we also shielded at the previous epoch then we need to multiply this value by 2, hence the first addition. You think it would be cleaner with two different assertions selected by an `if` on `shield_at_previous_epoch`?

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Yes but optionally I also want to shield an additional `principal` amount at the previous epoch to check that the rewards estimate logic holds

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Same answer as before, only difference is that rewards for the native token compound, so we need to add 2 instead of 1 to the multiplier this time

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Only in this case, for the previous two cases no, it's supposed to return the estimated rewards

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Same as before, but in this case we should also account for the compounding of rewards when the native token itself accrues rewards, I've added a comment to the formula

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -713,95 +709,228 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+            asset.redate_to_next_epoch();
+            let decoded_conv = self
+                .decode_sum(context.client(), conv.clone().into())
+                .await
+                .0;
+            let mut est_conv = I128Sum::zero();
+            for ((_, mut asset_data), val) in decoded_conv.into_components() {
+                // We must upconvert the components of the conversion unless
+                // this conversion is for a non-native token and the component
+                // itself is the native token. In this case we should avoid
```
batconjurer:
```suggestion
                //  the component itself is the native token and this conversion
                //  is for a non-native token. In this case we should avoid
```

crates/shielded_token/src/masp/shielded_wallet.rs
batconjurer: I feel like you left me hanging here. You checked if the epoch is zero and error otherwise.

crates/shielded_token/src/masp/shielded_wallet.rs
batconjurer: 🙈

batconjurer (changes requested):
Just some nits left.

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: So we extract the native token components from the result of the subtraction of the two balances. Then, to be able to sum all these components and return  a `DenominatedAmount`, we normalize them at `MaspEpoch(0)`. But we could pick any other epoch so long as it's a fixed one. The new error message is:

`Found reward asset with an epoch different than the 0th`

Should I change it to something else?

batconjurer: I'm good with that

brentstone:
@grarco what is the status here? Any chance this can be included in the next release?

grarco:
> @grarco what is the status here? Any chance this can be included in the next release?

I've just rebased it to solve a conflict with main, I guess I need approval from @batconjurer

batconjurer (approved):

PR #4181, opened 2024-12-12
docs: remove repetitive words

ChengenH:


tzemanovic:
we are very fond of our repetitions, thank you very much

PR #4205, opened 2024-12-25
Improve `namadac next-epoch-info`

brentstone:
## Describe your changes
For example, currently on mainnet with `v1.0.0`:
```
Last committed block height: 269035, time: 2024-12-25T09:35:13.291756821+00:00
Current epoch: 87.

First block height of this epoch 87: 268949.
Minimum number of blocks in an epoch: 2700.
Minimum amount of time for an epoch: 21600 seconds.

Earliest height at which epoch 88 can begin is block 271649.
```

And with the changes:
```
Last committed block height: 269044, time: 2024-12-25T09:36:15.592519345+00:00

Current epoch: 87.
First block height of epoch 87: 268949.

Minimum number of blocks in an epoch: 2700.
Minimum amount of time for an epoch: 6h-0m-0.

Next epoch (88) begins in 5h-49m-0s or at block height 271649, whichever occurs later.

```

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -49,6 +49,7 @@ use namada_sdk::rpc::{
 };
 use namada_sdk::storage::BlockResults;
 use namada_sdk::tendermint_rpc::endpoint::status;
+use namada_sdk::time::DateTimeUtc;
 use namada_sdk::token::{
     DenominatedAmount, MaspDigitPos, NATIVE_MAX_DECIMAL_PLACES,
 };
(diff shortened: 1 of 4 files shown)


crates/core/src/time.rs
```diff
@@ -206,6 +206,16 @@ impl DateTimeUtc {
     pub fn next_second(&self) -> Self {
         *self + DurationSecs(1)
     }
+
+    /// Returns the number of seconds in between two `DateTimeUtc` instances.
+    /// Ass
```
tzemanovic, line 211: :donkey:

tzemanovic, line 211:
```suggestion
```

brentstone: lol whoops

brentstone: fixed in 6ce70ce70223f7e650fe4412373c23bd8bc6e9fb

crates/apps_lib/src/client/rpc.rs
```diff
@@ -90,17 +91,30 @@ pub async fn query_and_print_masp_epoch(context: &impl Namada) -> MaspEpoch {
 /// Query and print some information to help discern when the next epoch will
 /// begin.
 pub async fn query_and_print_next_epoch_info(context: &impl Namada) {
-    query_block(context).await;
+    println!();
+    let current_time = query_block(context).await.unwrap();
```
tzemanovic, line 95: this is not exactly current time, but it's probably close enough. Still, maybe it would be better to just use `DateTimeUtc::now()` here

crates/apps_lib/src/client/rpc.rs
```diff
@@ -138,7 +168,7 @@ pub async fn query_and_print_status(
 }
 
 /// Query the last committed block
-pub async fn query_block(context: &impl Namada) {
```
tzemanovic, line 141: I think it would probably be clearer and more flexible to return the whole `block` here

tzemanovic (approved):

Fraccaman (approved):

crates/apps_lib/src/client/rpc.rs
brentstone: Blocks are quick enough that I don't think there would be that much of a difference with `DateTimeUtc::now()`, especially considering the latency between reading the data and when it gets output to the user at CLI.

Looks like this got merged though, so I'm open to another PR that improves this.

crates/apps_lib/src/client/rpc.rs
brentstone: Yeah that makes sense. Should we do it in a follow-up PR since this was merged?

PR #4230, opened 2025-01-10
Avoid duplicated signatures when building transactions

### LINKED ISSUE #4203: Duplicated signatures in transactions
Some of the transactions observed in mainnet are carrying duplicated inner signatures with the only difference in the two signature sections being the `signer`: in one of the two it is expressed as an address while in the other one as a public key. This behavior seems to happen on bundles containing a `reveal_pk` and a `bond` transaction.

When we construct the bundle we try to avoid duplicated `SigningTxData` structs to avoid this kind of situations but unfortunately it is still possible for two different instances of such type to lead to the same signature being produced. We should try to improve this check further or improve the logic that we use to discard duplicated sections

grarco:
## Describe your changes

Closes #4203.

This PR modifies adds a new method on the `Tx` type to remove duplicated sections. This function is then called in the signing function before producing the wrapper signature (which is computed on all the sections) and the `dump_tx` function (in case the dumped tx needed to be signed offline).

To support this logic the `PartialEq` implementation for the `Authorization` type is now manually provided and ignores the way the signer is specified, focusing instead on the targets and the actual signature.

As a side note, also the `PartialEq` implementation for `SigningTxData` has been updated to account for the shielded hash.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/tx/src/section.rs
+++ b/crates/tx/src/section.rs
@@ -443,8 +443,7 @@ pub struct Authorization {
 
 impl PartialEq for Authorization {
     fn eq(&self, other: &Self) -> bool {
-        // Deconstruct the two instances to ensure we don't forget any new
-        // field
+        // Deconstruct the two instances to ensure we don't forget any new field
         let Authorization {
             targets,
             signer: _,
@@ -457,15 +456,18 @@ impl PartialEq for Authorization {
         } = other;
 
         // Two authorizations are equal when they are computed over the same
-        // target(s) and the signatures match ,regardless of how the signer is
+        // target(s) and the signatures match, regardless of how the signer is
         // expressed
 
-        // Check equivalence of the targets ignoring the specific ordering
-        if targets.len() != other_targets.len() {
-            return false;
-        }
+        // Check equivalence of the targets ignoring the specific ordering and
+        // duplicates (the PartialEq implementation of IndexSet ignores the
+        // order)
+        let unique_targets =
+            HashSet::<&namada_account::Hash>::from_iter(targets.iter());
+        let unique_other_targets =
+            HashSet::<&namada_account::Hash>::from_iter(other_targets.iter());
 
-        if !targets.iter().all(|pubkey| other_targets.contains(pubkey)) {
+        if unique_targets != unique_other_targets {
             return false;
         }
 
--- a/crates/tx/src/section.rs
+++ b/crates/tx/src/section.rs
@@ -431,7 +431,6 @@ pub enum Signer {
     BorshSchema,
     Serialize,
     Deserialize,
-    PartialEq,
 )]
 pub struct Authorization {
     /// The hash of the section being signed
@@ -440,6 +439,39 @@ pub struct Authorization {
     pub signer: Signer,
     /// The signature over the above hash
     pub signatures: BTreeMap<u8, common::Signature>,
+}
+
+impl PartialEq for Authorization {
+    fn eq(&self, other: &Self) -> bool {
+        // Deconstruct the two instances to ensure we don't forget any new
+        // field
+        let Authorization {
+            targets,
+            signer: _,
+            signatures,
+        } = self;
+        let Authorization {
+            targets: other_targets,
+            signer: _,
+            signatures: other_signatures,
+        } = other;
+
+        // Two authorizations are equal when they are computed over the same
+        // target(s) and the signatures match ,regardless of how the signer is
+        // expressed
+
+        // Check equivalence of the targets ignoring the specific ordering
+        if targets.len() != other_targets.len() {
+            return false;
+        }
+
+        if !targets.iter().all(|pubkey| other_targets.contains(pubkey)) {
+            return false;
+        }
+
+        // Check equivalence of the signatures
+        signatures == other_signatures
+    }
 }
 
 impl Authorization {
--- a/crates/tx/src/section.rs
+++ b/crates/tx/src/section.rs
@@ -411,13 +411,45 @@ pub type Memo = Vec<u8>;
     BorshSchema,
     Serialize,
     Deserialize,
-    PartialEq,
 )]
 pub enum Signer {
     /// The address of a multisignature account
     Address(Address),
     /// The public keys that constitute a signer
     PubKeys(Vec<common::PublicKey>),
+}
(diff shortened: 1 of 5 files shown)


tzemanovic (approved):

crates/tx/src/section.rs
```diff
@@ -442,6 +441,39 @@ pub struct Authorization {
...
+            signer: _,
+            signatures,
+        } = self;
+        let Authorization {
+            targets: other_targets,
+            signer: _,
+            signatures: other_signatures,
+        } = other;
+
+        // Two authorizations are equal when they are computed over the same
```
murisi, line 459: This definition could work. But it has the implication that if `Authorization::signer` contains the variant `Signer::PubKeys` and someone tampers with one of the public keys in that list so that `Authorization::verify_signature` can no longer pass, this `PartialEq` implementation would still say that the tampered instance is still equal to the original. No? If so, is this desirable? Given that this function defines a partial equivalence relation, I guess `false` should be returned in those cases where there is insufficient information to check/conclude equality of the signers? No?

crates/tx/src/section.rs
```diff
@@ -442,6 +441,39 @@ pub struct Authorization {
     pub signatures: BTreeMap<u8, common::Signature>,
 }
 
+impl PartialEq for Authorization {
+    fn eq(&self, other: &Self) -> bool {
```
murisi, line 445: Does it matter that partially equivalent `Authorization`s might not have the same cryptographic hash? Or is this impossible? Could this sort of definition of partial equality confuse people in the future? (This definition seems to be calculated specifically to deal with our duplicate signatures issue (no?), but could be unintuitive in other cases/situations...)

murisi (changes requested):
While the overall approach taken here to avoid duplicate signatures makes sense, checking whether `Authorization` sections are duplicates after they have been produced (by checking structural equality) seems less robust than not producing duplicates in the first place. But perhaps this (situation where a signing process is redundantly being done twice) is unavoidable...

crates/tx/src/section.rs
```diff
@@ -442,6 +441,39 @@ pub struct Authorization {
...
+            targets: other_targets,
+            signer: _,
+            signatures: other_signatures,
+        } = other;
+
+        // Two authorizations are equal when they are computed over the same
+        // target(s) and the signatures match ,regardless of how the signer is
+        // expressed
+
+        // Check equivalence of the targets ignoring the specific ordering
```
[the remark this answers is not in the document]
grarco: Same as above, thanks!

crates/tx/src/section.rs
grarco: That's correct. My reasoning was that once we establish the equality of the signature's targets and of the signatures themselves we can safely consider the two `Authorization`s to be the same even if the signers are expressed differently. In case of the signers being straight different (which we cause one of the two instances to fail the signature verification check) I believe we are looking at something else than just equality between the two instances: I guess one should check the integrity of the sections at the byte level using a digest or similar techniques but I don't have a super strong opinion on this (see also my answer to your next question since it's related to this one)

grarco:
> While the overall approach taken here to avoid duplicate signatures makes sense, checking whether `Authorization` sections are duplicates after they have been produced (by checking structural equality) seems less robust than not producing duplicates in the first place. But perhaps this (situation where a signing process is redundantly being done twice) is unavoidable...

We already try to filter duplicated instanced of `SigningTxData` but there are cases where this logic fails (such as in case of different wrapper signers, even though we should only produce a single wrapper signature, see #3901, or the case of an offline procedure where we could get duplicated signatures from the members of a multisig account): hence the need to integrate this logic with a pruning mechanism for duplicated sections. I do agree with you though and we should try to improve the signing procedure over time to avoid producing duplicated signatures as much as possible

grarco:
@murisi I've pushed another commit to reintroduce the signer as part of the equivalence logic for `Authorization`. To do so I've implemented (yet again) a custom version of `PartialEq` for `Signer`, let me know if this addresses your concerns

crates/tx/src/section.rs
murisi: > I don't believe PartialEq should guarantee a byte-level match between the two instances but rather a logical equivalence (as I said in the previous question, when safety is critical one should explicitly rely on the digest of the expected instance).

Good point. My expectations of `PartialEq` were too restrictive.
> I could try to reintroduce the check on the signer with a more refined logic.

Thanks! This would be nice if it doesn't break the PR...
> My only concern is the case when a multisig signer is expressed as a set of keys in one instance and as an address in the other one: in this case there's no way I can resolve the equivalence without a query to storage (but there's a chance we always express multisig signers as a set of keys which would solve this issue).

Understood, I see the dilemma here. In this vein of thought, the comparison function could even be parametrized by a map from multisig addresses to public keys (given that the client must have known this in order to produce the signature sections in the first place). But this is probably overkill.

murisi:
> > While the overall approach taken here to avoid duplicate signatures makes sense, checking whether `Authorization` sections are duplicates after they have been produced (by checking structural equality) seems less robust than not producing duplicates in the first place. But perhaps this (situation where a signing process is redundantly being done twice) is unavoidable...
> 
> We already try to filter duplicated instanced of `SigningTxData` but there are cases where this logic fails (such as in case of different wrapper signers, even though we should only produce a single wrapper signature, see #3901, or the case of an offline procedure where we could get duplicated signatures from the members of a multisig account): hence the need to integrate this logic with a pruning mechanism for duplicated sections. I do agree with you though and we should try to improve the signing procedure over time to avoid producing duplicated signatures as much as possible

Understood, I see your point: there are a lot of ways in which duplicate signatures can occur. And trying to catch duplicates during signing might actually be more complicated (or even more confusing) than removing them after signing.

murisi (approved):
Apologies for the delay. This PR looks good to me, thanks!

murisi:
> @murisi I've pushed another commit to reintroduce the signer as part of the equivalence logic for `Authorization`. To do so I've implemented (yet again) a custom version of `PartialEq` for `Signer`, let me know if this addresses your concerns

I have looked at the adjustments - they do address my concerns. Thanks!

PR #4245, opened 2025-01-17
avoid running cometbft as a validator when tendermint_node is set to non-validator

### LINKED ISSUE #4139: configuration `tendermint_mode=Full`  has unexpected behavior may lead to double sign
#### Description:
Namada V1.0.0    

I need clarification on the tendermint_mode configuration in config.toml. Specifically, the behavior when setting it to "Full".

#### Steps to Reproduce:
Start a node in Full mode (tendermint_mode = "Full").
Observe the "This node is not validator" message at startup.
If the node has the correct consensus keys in the cometbft directory and enough voting power (VP), it starts signing blocks regardless of the mode.
#### Observed Behavior:
When set to Validator, if the validator keys are missing from wallet.toml, the node crashes with an error.
When set to Full, the node runs without issue but may start signing if it has consensus keys and enough VP.
#### Expected Behavior:
Clear and predictable behavior depending on the mode to prevent potential double signing or misconfiguration.

#### Questions:
1.Are there any additional side effects of this configuration?
2. What this mode should be in multi-signer setups?
For instance, should full nodes connected to a cosigner cluster be set to "Validator" or "Full"?
Setting to "Validator" requires keys in wallet.toml (which could be any validator key, not the actual one used in the multi-signer), while "Full" works without them. Are there any risks or hidden side effects in doing this?
### Expected Impact:
Clarifying this behavior is critical for avoiding misconfigurations, especially in high-risk setups like multi-signing where improper modes could lead to double signing or node crashes.
I suggest removing this configuration entirely, if possible, similar to how other CometBFT-based chains operate. To my knowledge, these chains do not use this setting—the node simply begins signing blocks as long as it has sufficient voting power (VP).

tzemanovic:
## Describe your changes

closes #4139

The PR adds a new field in config `last_tendermint_mode` (non-breaking as it's optional) which will be persisted on node start-up. If the mode changes from validator to non-validator, the node will backup and replace the validator key and state in the CometBFT dir.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -284,10 +284,12 @@ fn emit_warning_on_non_64bit_cpu() {
 
 /// Run the ledger with an async runtime
 pub fn run(
-    config: config::Ledger,
+    config: config::Config,
     wasm_dir: PathBuf,
     scheduled_migration: Option<ScheduledMigration>,
 ) {
+    handle_tendermint_mode_change(&config);
+
     emit_warning_on_non_64bit_cpu();
 
     let logical_cores = num_cpus::get();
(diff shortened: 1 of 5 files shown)


crates/node/src/lib.rs
```diff
@@ -323,7 +325,54 @@ pub fn run(
...
+    ) && matches!(
+        config.ledger.shell.last_tendermint_mode,
+        Some(TendermintMode::Validator)
+    ) {
+        // Backup and replace CometBFT validator key and state
+        let cometbft_dir = config.ledger.cometbft_dir();
+        namada_apps_lib::tendermint_node::backup_validator_key_and_state(
+            &cometbft_dir,
+        );
+        namada_apps_lib::tendermint_node::write_dummy_validator_key_and_state(
```
grarco, line 349: It looks to me like we could enter this branch even if the node is transitioning from a full node to a validator, is it ok to write this dummy data in this case?

tzemanovic: I think the second part of the condition should be `false` in that case (`last_tendermint_mode == Full`)

grarco: My bad, I didn't see the second `matches!`, thought there was only one

grarco (approved):
lgtm, thanks!

PR #4246, opened 2025-01-17
IBC rate limit query

grarco:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/sdk/src/rpc.rs
```diff
@@ -1518,3 +1519,25 @@ pub async fn query_ibc_denom<N: Namada>(
 
     token.as_ref().to_string()
 }
+
+/// Query the IBC rate limit for the provided token
+pub async fn query_ibc_rate_limit<C: Client + Sync>(
```
grarco, line 1524: An alternative to this function could be to add another path to the `TOKEN` https://github.com/anoma/namada/blob/9eb747ce145d622fda5296475c1d35eb4c3e86a4/crates/sdk/src/queries/vp/token.rs#L15

grarco:
@brentstone I've also added a cli command for the query. Let me know if we only need the sdk function so that I can revert that change

crates/sdk/src/rpc.rs
```diff
@@ -1518,3 +1519,25 @@ pub async fn query_ibc_denom<N: Namada>(
...
+    client: &C,
+    token: &Address,
+) -> Result<Amount, error::Error> {
```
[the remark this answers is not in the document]
brentstone: yes definitely

crates/sdk/src/rpc.rs
brentstone: don't think we can add a router path now that mainnet has been launched, perhaps @tzemanovic can comment here

brentstone:
@grarco I've fixed up some things and I think the PR is rdy now

crates/sdk/src/rpc.rs
tzemanovic: You can add more paths, but then the query only works against a node that is updated to contain that (additions are not breaking changes). Still, if it can be implemented without extra path then that's preferable. You can also implement it as a client-only query if you prefer to have it as a method on router though it's not much different from this (see `mod client_only_methods` in e.g. `crates/sdk/src/queries/vp/token.rs` )

crates/apps_lib/src/client/rpc.rs
```diff
@@ -2446,3 +2466,68 @@ pub async fn compute_proposal_votes<C: Client + Sync>(
...
+            display_line!(
+                context.io(),
+                "IBC rate limits for token {}:\nGlobal mint limit: \
+                 {}\nThroughput limit: {} per epoch",
+                token_alias,
+                mint_str,
+                throughput_str
+            );
+        }
+        Err(_) => {
```
grarco, line 2503: Actually, given the way the ibc vp validates transactions, this logic is wrong. The sdk `query_ibc_rate_limits` functions returns an error if even one of the two keys is missing and in here we fallback to the default values for both, but the vp does actually support setting only one of the two and then using the default param for the other one. I'll try to rework this

grarco: Done in 9a1b9b3

brentstone: thx, gotcha!

crates/ibc/src/parameters.rs
```diff
@@ -13,6 +13,9 @@ pub struct IbcParameters {
     pub default_per_epoch_throughput_limit: Amount,
 }
 
+/// Alias for a specific token's IBC limits
+pub type IbcTokenRateLimits = IbcParameters;
```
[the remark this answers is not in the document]
brentstone: done in 96d24aa. Mind checking over again @tzemanovic? We're positive this doesn't break state? Had to make changes in multiple places to accommodate this as you suggested.

PR #4251, opened 2025-01-19
Support IBC signing on hardware wallets not supporting MASP

murisi:
## Describe your changes
Pure IBC transaction signing currently fails on the Ledger Nano S. This is because it tries to use the hardware wallet to generate MASP build parameters even when they are not needed. This PR modifies the code to ignore MASP build parameter generation failures if they turn out to not be needed for the construction of an IBC transaction.

An alternative approach to this PR could be to determine the precise conditions under which MASP build parameters would be required, and only do generation in those circumstances. This also would work fine, but would involve making sure that the determined conditions do not fall out of sync with the transaction building code that follows its usage.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


grarco (approved):
LGTM, thanks! Should we also open an issue tracking the alternative approach you proposed in the description?

tzemanovic (approved):

PR #4272, opened 2025-01-24
wallet: disallow deriving ed25519 keys from default HD path

tzemanovic:
## Describe your changes

Follow-up to #4252. We disallow deriving ed25519 keys from the default HD path that's used for deriving shielded keys with the newly default modified ZIP32 to prevent accidental leak of these keys.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/wallet/src/lib.rs
+++ b/crates/wallet/src/lib.rs
@@ -646,6 +646,11 @@ impl<U: WalletIo> Wallet<U> {
     /// provided, will prompt for password from stdin.
     /// Stores the key in decrypted key cache and returns the alias of the key
     /// and a reference-counting pointer to the key.
+    ///
+    /// Any usage of this function should be careful not expose a shielded key
+    /// that may be derived via modified ZIP32 from this key (specifically when
+    /// the scheme is Ed25519 with the default HD derivation path and no
+    /// encryption password is being used).
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


murisi (approved):
Thanks! This will work. Just wanted to note two things that we have discussed already:
* Users who derive a private key using `namadaw gen --alias ...`, note the displayed mnemonic, and use this mnemonic to derive a shielded spending/viewing key will end up in the undesirable state where both the seed and the derived shielded spending/viewing key are both stored in the software wallet. This situation is worse if the user did key generation using the flag `--unsafe-dont-encrypt`, since that would mean that the seed is stored unencrypted.
* The functions `Wallet::derive_store_hd_secret_key` and `Wallet::derive_store_key_from_mnemonic_code` allow for the combined derivation and (potentially unencrypted) storage of modified ZIP 32 seed. So future changes to the codebase that make use of these functions might inadvertently bypass the checks made in this PR.

So while this PR blocks certain ways of reaching a state where a (potentially unencrypted) seed and its derivative shielded spending/viewing key are stored in the software wallet, there remain straightforward ways to enter this state from both the CLI and the internal APIs. And probably this situation is unavoidable due to the dual use of `m/44'/877'/0'/0'/0'` in `namadaw gen`'s Ed25519 flow and also in the hardware wallet's  modified ZIP 32 flow.

PR #4283, opened 2025-01-29
PoS claim rewards - add token transfer event

tzemanovic:
## Describe your changes

Emit a token transfer emit with `"pos-claim-rewards"` descriptor when claiming non-zero staking rewards.

An update to a live chain with this change can be applied via a gov proposal replacing the `tx_claim_rewards` wasm

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/proof_of_stake/src/lib.rs
+++ b/crates/proof_of_stake/src/lib.rs
@@ -115,6 +115,9 @@ use crate::validator_set_update::{
     update_validator_set,
 };
 
+/// Event descriptor for claim-rewards token transfer
+pub const CLAIM_REWARDS_EVENT_DESC: &str = "pos-claim-rewards";
+
 /// PoS storage `Keys/Read/Write` implementation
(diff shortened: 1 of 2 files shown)


crates/proof_of_stake/src/lib.rs
```diff
@@ -2826,12 +2829,26 @@ where
         take_rewards_from_counter(storage, &source, validator)?;
     checked!(reward_tokens += counter_rewards)?;
 
+    // Return early if there's nothing to claim
+    if counter_rewards.is_zero() {
+        return Ok(counter_rewards);
+    }
```
Fraccaman, line 2835: i think we should emit the event also in this case

Fraccaman (changes requested):

Fraccaman:
Not sure why a unit test would fail here

Fraccaman (approved):

PR #4288, opened 2025-01-31
Murisi/reverse conversion at source

murisi:
## Describe your changes
Based on the bug fix in #4287. An alternative to #4290 that simplifies the exchanging an amount to some past epoch. If the current epoch is 10 and it is desired to convert an asset from epoch 2 to epoch 6 when only conversions to the latest epoch, 10, are available. Then the following process is now done:
1) Query the conversion from epoch 2 to epoch 10, call it `A`
2) Query the conversion from epoch 6 to epoch 10, call it `B`. (If we were converting to epoch 10, then `B` would equal 0 since that is the conversion from an epoch to itself.)
3) Then return `C=A-B`. Because of the way the conversion tree is constructed, `C` is the conversion from epoch 2 to epoch 6 that would have been returned during epoch 6.

So more generally `query_allowed_conversion` now returns the result of a subtraction `A-B`. And by definition, the second operand of the subtraction is 0 when converting to the latest epoch, which ultimately means the result is simply `A` in that case. Additionally, `compute_exchanged_amount` has been simplified removing the forward and backward conversions.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/bug-fixes/4287-consolidate-dust.md
+++ b/.changelog/unreleased/bug-fixes/4287-consolidate-dust.md
@@ -1,0 +1,2 @@
+- Fix underestimation of MASP balances
+  ([\#4287](https://github.com/anoma/namada/pull/4287))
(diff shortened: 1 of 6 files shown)


grarco (approved):
lgtm even though personally I prefer the approach of #4290

PR #4290, opened 2025-01-31
Murisi/remove reverse conversions

murisi:
## Describe your changes
An alternative to #4288 where the ability to choose the epoch to which amounts are exchanged to is simply removed. This functionality became redundant once pinned transactions were removed in https://github.com/anoma/namada/pull/3142 . More generally, this functionality is only needed for properly implementing functionality like account statements and pinned transactions where the amounts  that were received need to be printed out exactly as they were in the past without any influence from conversions in the current tree.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/masp/shielded_wallet.rs
+++ b/crates/shielded_token/src/masp/shielded_wallet.rs
@@ -507,7 +507,6 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
         &'a mut self,
         client: &C,
         asset_type: AssetType,
-        target_epoch: Option<MaspEpoch>,
         conversions: &'a mut Conversions,
     ) -> Result<(), eyre::Error> {
         let btree_map::Entry::Vacant(conv_entry) =
@@ -516,7 +515,7 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
             return Ok(());
         };
         // Get the conversion for the given asset type, otherwise fail
-        let Some((token, denom, position, ep, mut conv, path)) =
+        let Some((token, denom, position, ep, conv, path)) =
             Self::query_conversion(client, asset_type).await
         else {
             return Ok(());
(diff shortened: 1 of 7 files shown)


crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -2076,13 +2035,12 @@ mod test_shielded_wallet {
         // to construct conversions properly, just like the protocol does, i.e.
         // with conversions for non-consecutive epochs
 
-        // Query the balance with conversions at epoch 4
+        // Query the balance with conversions at epoch 5
```
grarco, line 2038: Probably this test is not needed anymore since it's testing the functionality that this pr removes?

grarco (approved):

crates/shielded_token/src/masp/shielded_wallet.rs
murisi: Thanks, you're right. But maybe this test can just be repurposed since the specific checks that it does are still useful in the absence of inverse conversions...

grarco: Right but I think for this test to make sense we should shield at epoch 4 instead of 5. Right now we shield at 5 and query at 5 so we just get back the same amount. We should instead shield at 4, query at epoch 4 expecting the same amount and then query at epoch 5 expecting the rewards

batconjurer (approved):

crates/shielded_token/src/masp/shielded_wallet.rs
murisi: Understood, thanks. The shielding of 10 NAM now happens at epoch 1, and computing the exchanged amount yields 160 NAM of epoch 5. This seems to line up with the conversion -2NAM@ep1 + 32NAM@ep5 being applied 5 times. That being said, `compute_exchanged_amount` should not be affected by the value of the current MASP epoch (since I don't think it queries this value). It should really only be affected by what's currently in the allowed conversion Merkle tree that's constructed on lines 1999 to 2014, I think.

PR #4303, opened 2025-02-03
Murisi/track undated assets

murisi:
## Describe your changes
This PR attempts to address issue #4304 by distributing rewards only among the dated assets. Doing this involved tracking the balance of dated/undated assets in the MASP at any point in time. More specifically, the following changes have been made:
* A new storage key has been introduced to track the amount of undated tokens of each token address in the MASP
* The transfer WASMs have been modified to update the storage key that tracks undated token amounts
* The MASP VP has been modified to check that the undated token tracking storage keys are updated correctly
* Rewards are now only distributed to the total undated assets of some token rather than to all assets of some token
  * Hence undated tokens that may have been locked into the MASP pool prior to enabling the rewards program for the token are no longer "rewarded"

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/ibc/src/lib.rs
+++ b/crates/ibc/src/lib.rs
@@ -693,26 +693,22 @@ where
                         })?,
                 );
                 // Record the tokens credited/debited in this NFT transfer
-                let ibc_traces: Vec<_> = msg
+                let tokens = msg
                     .message
                     .packet_data
                     .token_ids
                     .0
                     .iter()
                     .map(|token_id| {
-                        ibc_trace_for_nft(
+                        convert_to_address(ibc_trace_for_nft(
                             &msg.message.packet_data.class_id,
                             token_id,
-                        )
-                    })
-                    .collect();
-                let mut tokens = BTreeSet::new();
-                for ibc_trace in ibc_traces {
-                    let token = convert_to_address(ibc_trace)
+                        ))
                         .into_storage_result()
-                        .map_err(Error::Storage)?;
-                    tokens.insert(token);
-                }
+                        .map_err(Error::Storage)
+                    })
+                    .collect::<Result<_, _>>()?;
+
                 send_nft_transfer_execute(
                     &mut self.ctx,
                     &mut nft_transfer_ctx,
@@ -747,21 +743,14 @@ where
                             .is_receiving_success(msg)?
                             .is_some_and(|ack_succ| ack_succ) =>
                     {
-                        let tokens = if msg.packet.port_id_on_b.as_str()
-                            == PORT_ID_STR
-                        {
-                            // Record the token credited/debited in this
-                            // transfer
-                            let packet_data =
-                                serde_json::from_slice::<PacketData>(
-                                    &msg.packet.data,
-                                )
-                                .map_err(StorageError::new)
-                                .map_err(Error::Storage)?;
-                            let ibc_denom = packet_data.token.denom.to_string();
+                        let ibc_traces = extract_traces_from_recv_msg(msg)
+                            .map_err(StorageError::new)
+                            .map_err(Error::Storage)?;
+                        let mut tokens = BTreeSet::new();
+                        for ibc_trace in ibc_traces {
                             // Get the received token
                             let token = received_ibc_token(
-                                ibc_denom,
+                                ibc_trace,
                                 &msg.packet.port_id_on_a,
                                 &msg.packet.chan_id_on_a,
                                 &msg.packet.port_id_on_b,
@@ -769,47 +758,8 @@ where
                             )
                             .into_storage_result()
                             .map_err(Error::Storage)?;
-                            [token].into()
-                        } else if msg.packet.port_id_on_b.as_str()
-                            == NFT_PORT_ID_STR
-                        {
-                            // Record the tokenS credited/debited in this NFT
-                            // transfer
-                            let packet_data =
-                                serde_json::from_slice::<NftPacketData>(
-                                    &msg.packet.data,
-                                )
-                                .map_err(StorageError::new)
-                                .map_err(Error::Storage)?;
-                            let ibc_traces: Vec<_> = packet_data
-                                .token_ids
-                                .0
-                                .iter()
-                                .map(|token_id| {
-                                    ibc_trace_for_nft(
-                                        &packet_data.class_id,
-                                        token_id,
-                                    )
-                                })
-                                .collect();
-                            let mut tokens = BTreeSet::new();
-                            for ibc_trace in ibc_traces {
-                                // Get the received token
-                                let token = received_ibc_token(
-                                    ibc_trace,
-                                    &msg.packet.port_id_on_a,
-                                    &msg.packet.chan_id_on_a,
-                                    &msg.packet.port_id_on_b,
-                                    &msg.packet.chan_id_on_b,
-                                )
-                                .into_storage_result()
-                                .map_err(Error::Storage)?;
-                                tokens.insert(token);
-                            }
-                            tokens
-                        } else {
-                            BTreeSet::new()
-                        };
+                            tokens.insert(token);
+                        }
                         (extract_masp_tx_from_packet(&msg.packet), tokens)
                     }
                     #[cfg(is_apple_silicon)]
--- a/crates/ibc/src/msg.rs
+++ b/crates/ibc/src/msg.rs
@@ -11,7 +11,9 @@ use ibc::apps::nft_transfer::types::PORT_ID_STR as NFT_PORT_ID_STR;
 use ibc::apps::transfer::types::msgs::transfer::MsgTransfer as IbcMsgTransfer;
 use ibc::apps::transfer::types::packet::PacketData;
 use ibc::apps::transfer::types::PORT_ID_STR as FT_PORT_ID_STR;
-use ibc::core::channel::types::msgs::PacketMsg;
+use ibc::core::channel::types::msgs::{
+    MsgRecvPacket as IbcMsgRecvPacket, PacketMsg,
+};
 use ibc::core::channel::types::packet::Packet;
 use ibc::core::handler::types::msgs::MsgEnvelope;
 use ibc::core::host::types::identifiers::PortId;
@@ -20,6 +22,8 @@ use masp_primitives::transaction::Transaction as MaspTransaction;
 use namada_core::borsh::BorshSerializeExt;
 use namada_core::string_encoding::StringEncoded;
 use serde::{Deserialize, Serialize};
+
+use crate::trace;
 
 trait Sealed {}
 
@@ -324,6 +328,38 @@ fn extract_memo_from_packet(
             );
             None
         }
+    }
+}
+
+/// Extract the traces used in a receive packet message
+pub fn extract_traces_from_recv_msg(
+    msg: &IbcMsgRecvPacket,
+) -> Result<Vec<String>, serde_json::Error> {
+    match msg.packet.port_id_on_b.as_str() {
+        FT_PORT_ID_STR => {
+            // Record the token credited/debited in this
+            // transfer
+            let packet_data =
+                serde_json::from_slice::<PacketData>(&msg.packet.data)?;
+            let ibc_denom = packet_data.token.denom.to_string();
+            Ok([ibc_denom].into())
+        }
+        NFT_PORT_ID_STR => {
+            // Record the tokens credited/debited in this NFT
+            // transfer
+            let packet_data =
+                serde_json::from_slice::<NftPacketData>(&msg.packet.data)?;
+            let ibc_traces: Vec<_> = packet_data
+                .token_ids
+                .0
+                .iter()
+                .map(|token_id| {
+                    trace::ibc_trace_for_nft(&packet_data.class_id, token_id)
+                })
+                .collect();
+            Ok(ibc_traces)
+        }
+        _ => Ok(vec![]),
     }
 }
 


crates/shielded_token/src/conversion.rs
```diff
@@ -85,6 +85,28 @@ where
...
+{
+    use crate::read_undated_balance;
+    let masp_addr = MASP;
+
+    // total locked amount in the Shielded pool
+    let total_tokens_in_masp =
+        TransToken::read_balance(storage, token, &masp_addr)?;
+    // Since dated and undated tokens are stored together in the pool, subtract
+    // the latter to get the dated balance
+    let masp_dated_balance = read_undated_balance(storage, token)?;
```
grarco, line 106: Wrong variable name here

grarco, line 106:
```suggestion
    let masp_undated_balance = read_undated_balance(storage, token)?;
```

crates/shielded_token/src/conversion.rs
```diff
@@ -155,17 +177,21 @@ where
...
 
+    // total locked amount in the Shielded pool
+    let rewardable_tokens_in_masp =
+        get_masp_dated_balance::<S, TransToken>(storage, token)?;
+
     // inflation-per-token = inflation / locked tokens = n/PRECISION
     // ∴ n = (inflation * PRECISION) / locked tokens
     // Since we must put the notes in a compatible format with the
     // note format, we must make the inflation amount discrete.
-    let noterized_inflation = if total_tokens_in_masp.is_zero() {
```
grarco, line 162: Do we still need to read the `total_tokens_in_masp` variable in the prelude of this function?

crates/token/src/tx.rs
```diff
@@ -94,6 +102,67 @@ where
...
+    for token in tokens {
+        let Some(denom) = read_denom(env, &token)? else {
+            // Ignore asset type if denomination cannot be read
+            continue;
+        };
+        // Read the undated balance of the current token
+        let undated_balance = read_undated_balance(env, &token)?;
+        // Save the undated balance in a map for updating
+        undated_balances.insert(token.clone(), I320::from(undated_balance));
+        let asset_type_err = |err| {
```
grarco, line 118: Can we move this outside of the `for token in tokens` loop

crates/trans_token/src/tx.rs
```diff
@@ -56,7 +56,7 @@ pub fn multi_transfer<ENV>(
     sources: impl CreditOrDebit,
     targets: impl CreditOrDebit,
     event_desc: Cow<'static, str>,
-) -> Result<HashSet<Address>>
+) -> Result<(HashSet<Address>, HashSet<Address>)>
```
grarco, line 59: We should update the docstring with the new return type

crates/token/src/tx.rs
```diff
@@ -94,6 +102,67 @@ where
     env.push_action(Action::Masp(MaspAction::MaspSectionRef(
         masp_section_ref,
     )))?;
+    // Record undated balance changes
+    let mut undated_balances = BTreeMap::new();
+    let mut asset_types = BTreeMap::new();
```
grarco, line 107:
```suggestion
    let mut unepoched_asset_types = BTreeMap::new();
```

grarco (commented):
Net of the failing tests and some minor things this PR looks good to me

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -437,7 +437,7 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
     async fn precompute_asset_types<C: Client + Sync>(
         &mut self,
         client: &C,
-        tokens: Vec<&Address>,
+        tokens: BTreeSet<&Address>,
```
batconjurer, line 440: this could just be an `impl Iterator<Item=&Address>`

crates/shielded_token/src/vp.rs
```diff
@@ -44,16 +45,35 @@ pub struct MaspVp<'ctx, CTX, Params, Gov, Ibc, TransToken, Transfer> {
 }
 
 // Balances changed by a transaction
-#[derive(Default, Debug, Clone)]
+#[derive(Debug, Clone)]
 struct ChangedBalances {
-    // Maps asset types to their decodings
-    tokens: BTreeMap<AssetType, (Address, token::Denomination, MaspDigitPos)>,
+    // Maps unepoched asset types to their decodings
+    unepoched_tokens:
```
batconjurer, line 51: It would be nice to either use `undated` or `unepoched` consistently in var and field names

crates/shielded_token/src/vp.rs
```diff
@@ -44,16 +45,35 @@ pub struct MaspVp<'ctx, CTX, Params, Gov, Ibc, TransToken, Transfer> {
 }
 
 // Balances changed by a transaction
-#[derive(Default, Debug, Clone)]
+#[derive(Debug, Clone)]
 struct ChangedBalances {
-    // Maps asset types to their decodings
-    tokens: BTreeMap<AssetType, (Address, token::Denomination, MaspDigitPos)>,
```
batconjurer, line 50: maybe we should call these `epoched_tokens` now. Similarly for `pre` and `post`

batconjurer (changes requested):
Just some nits

crates/core/src/masp.rs
```diff
@@ -238,6 +238,39 @@ pub fn encode_asset_type(
     .encode()
 }
 
+/// Encode the assets that are used for masp rewards
+pub fn encode_reward_asset_types(
+    native_token: &Address,
+) -> Result<[AssetType; 4], std::io::Error> {
```
sug0, line 244: Rename to `encode_native_token_reward_asset_types`? Or make it explicit in the docstring that this function is only supposed to be called with the native token address. This will improve the docs published to `docs.rs`.

crates/core/src/masp.rs
```diff
@@ -238,6 +238,39 @@ pub fn encode_asset_type(
...
+    use crate::token::{MaspDigitPos, NATIVE_MAX_DECIMAL_PLACES};
+    Ok([
+        encode_asset_type(
+            native_token.clone(),
+            NATIVE_MAX_DECIMAL_PLACES.into(),
+            MaspDigitPos::Zero,
+            Some(MaspEpoch::zero()),
+        )?,
```
sug0, line 252: all of these will be infallible, anyway, no?

crates/shielded_token/src/conversion.rs
```diff
@@ -85,6 +85,28 @@ where
...
+    use crate::read_undated_balance;
+    let masp_addr = MASP;
+
+    // total locked amount in the Shielded pool
+    let total_tokens_in_masp =
+        TransToken::read_balance(storage, token, &masp_addr)?;
+    // Since dated and undated tokens are stored together in the pool, subtract
+    // the latter to get the dated balance
+    let masp_dated_balance = read_undated_balance(storage, token)?;
+    Ok(checked!(total_tokens_in_masp - masp_dated_balance)?)
```
sug0, line 107: It should always hold that `masp_undated_balance <= total_tokens_in_masp`. If this checked subtraction fails, we are in trouble. I'd prefer unwrapping `checked!()`.
```suggestion
    Ok(checked!(total_tokens_in_masp - masp_dated_balance).unwrap())
```

crates/shielded_token/src/conversion.rs
```diff
@@ -298,36 +323,8 @@ where
     // reward tokens with the zeroth epoch to minimize the number of convert
     // notes clients have to use. This trick works under the assumption that
     // reward tokens will then be reinflated back to the current epoch.
```
sug0, line 325:
```
    // Construct MASP asset type for rewards. Always deflate and timestamp
    // reward tokens with the zeroth epoch to minimize the number of convert
    // notes clients have to use. This trick works under the assumption that
    // reward tokens will then be reinflated back to the current epoch.
```
this comment should be moved inside the def of `encode_reward_asset_types`

crates/shielded_token/src/storage_key.rs
```diff
@@ -85,6 +89,29 @@ pub fn masp_last_inflation_key<TransToken: trans_token::Keys>(
...
+    }
+}
+
+/// Obtain the storage key for the undated balance of a token
+pub fn masp_undated_balance_key(token_address: &Address) -> storage::Key {
+    storage::Key::from(address::MASP.to_db_key())
+        .push(&MASP_UNDATED_BALANCE_KEY.to_owned())
+        .expect("Cannot obtain a storage key")
+        .push(&token_address.to_string().to_db_key())
+        .expect("Cannot obtain a storage key")
```
sug0, line 112: replace the `.push(...).expect(...)` pattern with `.with_segment(...)`

crates/shielded_token/src/vp.rs
sug0: I would go with `undated`/`dated`, since that seems like the most consistently used terminology

crates/tests/src/integration/masp.rs
```diff
@@ -3564,7 +3564,7 @@ fn dynamic_assets() -> Result<()> {
         )
     });
     assert!(captured.result.is_ok());
-    assert!(captured.contains("nam: 0.06262"));
+    assert!(captured.contains("nam: 0.063"));
```
sug0, line 3567: I believe some of the MASP integration tests had one or two cases where we were in fact testing the `undated`/`dated` rewards mechanism. Could you please quickly do a pass over all the tests in this file (`crates/tests/src/integration/masp.rs`) to check whether they still make sense, given the changes from this PR?

It seems like you just updated the balances in the failing tests :p

sug0 (changes requested):
I have left a few suggestions, nothing too annoying to fix, I hope. Overall, this is a fine solution. The only thing I would point out is that it partitions the privacy set into dated and undated assets. After a shielding transaction is applied, you can tell whether the shielded asset was dated or not, based on the keys changed in storage. Whether this is relevant information or not isn't really clear to me.

EDIT: I guess this doesn't matter, because we could already tell before this PR if the asset was dated or not, given its presence or absence in the token map...

grarco:
> I have left a few suggestions, nothing too annoying to fix, I hope. Overall, this is a fine solution. The only thing I would point out is that it partitions the privacy set into dated and undated assets. After a shielding transaction is applied, you can tell whether the shielded asset was dated or not, based on the keys changed in storage. Whether this is relevant information or not isn't really clear to me.
> 
> EDIT: I guess this doesn't matter, because we could already tell before this PR if the asset was dated or not, given its presence or absence in the token map...

I think it doesn't really matter since that information is taken from the `sapling_value_balance` which is already publicly available and displays the asset types and the amounts. That value is only populated with assets coming in or out of the pool: the assets and amounts that are instead circulating inside the pool won't show up there and therefore won't leak any data

crates/shielded_token/src/conversion.rs
murisi: I believe so. `total_tokens_in_masp` is still used in the inflation computation at https://github.com/anoma/namada/blob/cb6794c91f6aa4de2ae94232bd6cf744288a89e9/crates/shielded_token/src/conversion.rs#L133 . More broadly, the purpose of the shielded rewards system is to incentivise end users to lock their tokens until the target amount is reached, and both dated and undated amounts count as locked because they both contribute to the anonymity set.

crates/shielded_token/src/conversion.rs
murisi: Done, thanks!

crates/shielded_token/src/masp/shielded_wallet.rs
murisi: The motivation for this change was to try and avoid doing the same precomputation twice (with the redundant querying and encoding that comes with it) if the input contains duplicates, not that this is actually a serious problem.

crates/shielded_token/src/vp.rs
```diff
@@ -195,6 +215,41 @@ where
...
+    fn apply_undated_balances(
+        ctx: &'ctx CTX,
+        keys_changed: &BTreeSet<Key>,
+        mut result: ChangedBalances,
+    ) -> Result<ChangedBalances> {
+        // Record the undated balances of the keys that changed
+        for token in keys_changed.iter().filter_map(is_masp_undated_balance_key)
+        {
+            // Read and store the undated balance before this tx is applied
+            let pre_reward_balance: Amount = ctx
```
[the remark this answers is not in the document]
murisi: Fixed, thanks!

crates/core/src/masp.rs
murisi: The code for encoding is actually non-trivial involving some rejection sampling in the MASP crate, so it's not immediately clear that this cannot happen. That being said, I was trying to make the signature of `encode_reward_asset_types` consistent with `encode_asset_type`, and follow the precedent already set of propagating asset type encoding errors.

crates/shielded_token/src/conversion.rs
murisi: True, an underflow here would most likely mean that `masp_dated_balance` is in a very bad state. That being said, the policy followed for protocol code in this codebase seems to be to use checked arithmetic and propagate/handle arithmetic errors even if we believe or can prove that there will never be underflow. In the case that I have made a logical error in this PR, it would probably be better to just have the reward distribution mechanism fail and not distribute anything than to crash the whole node.

crates/shielded_token/src/conversion.rs
murisi: Done, thanks!

crates/shielded_token/src/storage_key.rs
murisi: Oops, did not realize this function could be used here. Fixed, thanks!

crates/tests/src/integration/masp.rs
murisi: @sug0 Thanks for highlighting that the integration tests might somehow be invalidated by these tracking changes, much appreciated. The logic of the tests still makes sense to me: everything about undated tokens never receiving rewards and dated tokens receiving rewards only after the rewards rate is raised above zero remains valid. The only exception to this is https://github.com/anoma/namada/blob/3558ccfb163850b7fca62e4d32223b53c503c2d3/crates/tests/src/integration/masp.rs#L3761 . I do not believe that the shielded rewards program can handle rewards for a token being disabled and then being re-enabled in the future. This is because allowed conversions are essentially telescoping series, and a disable followed by an enable interrupts the telescope.

I was a little surprised that the expected balances in `dynamic_assets` had to be raised to make the tests pass. But this is plausible in a hypothetical situation where `inflation = 10`, `total_tokens_in_masp = 6`, and `rewardable_tokens_in_masp = 5`. If, as before, all dated and undated assets were being rewarded, then each note would get a reward of `floor(10/6)=1`, which would imply an inflation of `6*floor(10/6)=6`. Whereas if only dated tokens are being rewarded, then each note would get a reward of `floor(10/5)=2`, which would imply an inflation of `5*floor(10/5)=10` which is higher than before.

crates/ibc/src/lib.rs
```diff
@@ -695,24 +740,91 @@ where
...
                 // Extract MASP tx from the memo in the packet if needed
-                let masp_tx = match &*envelope {
+                let (masp_tx, tokens) = match &*envelope {
                     MsgEnvelope::Packet(PacketMsg::Recv(msg))
                         if self
                             .is_receiving_success(msg)?
                             .is_some_and(|ack_succ| ack_succ) =>
                     {
-                        extract_masp_tx_from_packet(&msg.packet)
+                        let tokens = if msg.packet.port_id_on_b.as_str()
```
yito88, line 749: I know the `execute` function is already big. Can we add a helper function to retrieve the tokens like `extract_masp_tx_from_packet()`?
It would be better to make the helper to get the received `vec![ibc_denom]` or `ibc_traces` from the packet.

murisi: Done thanks. Made a separate function `msg_recv_packet_tokens` to extract the tokens used in an IBC receive packet message.

crates/ibc/src/lib.rs
```diff
@@ -746,69 +802,7 @@ where
...
-                                )
-                                .into_storage_result()
-                                .map_err(Error::Storage)?;
-                                tokens.insert(token);
-                            }
-                            tokens
-                        } else {
-                            BTreeSet::new()
-                        };
+                        let tokens = msg_recv_packet_tokens(msg)?;
```
yito88, line 805: Can we move the helper function to `crates/ibc/src/msg.rs` and make it return `Vec<String>` as a list of IBC traces?
`lib.rs` is also big. `received_ibc_token` can be used for a RecvPacket message of both the FT and NFT packets.
e.g.
```rust
            let ibc_traces = extract_traces_from_recv_msg(msg)?;
            let mut tokens = BTreeSet::new();
            for ibc_trace in ibc_traces {
                // Get the received token
                let token = received_ibc_token(
                    ibc_trace,
                    &msg.packet.port_id_on_a,
                    &msg.packet.chan_id_on_a,
                    &msg.packet.port_id_on_b,
                    &msg.packet.chan_id_on_b,
                )
                .into_storage_result()
                .map_err(Error::Storage)?;
                tokens.insert(token);
            }
```

batconjurer (approved):

sug0 (approved):

PR #4311, opened 2025-02-04
Murisi/refactor update allowed conversions

murisi:
## Describe your changes
Based on #4308 . This PR refactors the reward distribution code. First the digit loop and the conditional that checks whether the current token is native have been inverted to yield a conditional with a digit loop in both its consequent and alternate. Then the logic for the native conversion computation was moved into a separate function as was the logic for non-native conversion computation.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/conversion.rs
+++ b/crates/shielded_token/src/conversion.rs
@@ -359,60 +359,54 @@ where
         masp_reward_denoms.insert(token.clone(), denom);
         // Dispense a transparent reward in parallel to the shielded rewards
         let addr_bal = TransToken::read_balance(storage, token, &masp_addr)?;
-
         // Get the last rewarded amount of the native token
         let normed_inflation = *storage
             .conversion_state_mut()
             .normed_inflation
             .get_or_insert(ref_inflation);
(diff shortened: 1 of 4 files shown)


crates/shielded_token/src/conversion.rs
```diff
@@ -226,18 +245,236 @@ where
...
+            encode_asset_type(token.clone(), denom, digit, Some(masp_epoch))
+                .into_storage_result()?;
+
+        current_convs.insert(
+            (token.clone(), denom, digit),
+            checked!(
+                MaspAmount::from_pair(old_asset, -precision_i128)
+                    + &MaspAmount::from_pair(new_asset, precision_i128)
+                    + &MaspAmount::from_pair(
+                        reward_assets[digit as usize],
```
grarco, line 444: This assumes that a reward for, e.g., digit 2 will automatically be given out as an amount of digit 2 native token, is that right?

murisi: Yes, exactly. Each allowed conversion put into the Merkle tree deals with exactly one digit. No mixing of digits is currently happening.

grarco (approved):

PR #4358, opened 2025-02-10
wallet: check min ledger app version for shielded key derivation

### LINKED ISSUE #4327: check and enforce Ledger app version with a matching modified ZIP32 derivation path
In #4307 we've changed the derivation path used for ZIP32 to a non-default value (`m/44'/877'/0'/0'/2147483647'`). Once the same change is applied and released in the Ledger app, we should check and enforce that this or greater version is being used to avoid mismatch between the two.

tzemanovic:
## Describe your changes

closes #4327

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/tx.rs
+++ b/crates/apps_lib/src/client/tx.rs
@@ -943,10 +943,34 @@ async fn augment_masp_hardware_keys(
         let transport = WalletTransport::from_arg(args.device_transport);
         let app = NamadaApp::new(transport);
         let wallet = namada.wallet().await;
+        let mut checked_app_version = false;
         // Augment the pseudo spending key with a proof authorization key
         for source in sources {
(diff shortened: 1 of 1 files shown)


murisi (approved):
Looks good to me, thanks! Maybe we could also consider a general warning to upgrade the Ledger app to a version >= v3.0.0 (and move their funds to a key derived with the new modified ZIP 32 algorithm) whenever using an old version of the app to sign a shielded transaction... As it stands, `augment_masp_hardware_keys` will silently derive viewing keys using whatever ZIP 32 derivation algorithm the hardware wallet provides in the process of checking that at a given path, the software wallet key matches that of the hardware wallet.

murisi (approved):
Thanks for adding warning during transaction construction. Everything looks good to me!

Issue #4422, opened 2025-02-27
Degenerate Shielded Rewards Precision for IBC Tokens


murisi:
Inflation is often a small fraction of the total amount of an asset that is locked. So if inflation were 5 and the locked amount were 100, then the reward would `5/100=0.05 NAM` for every unit of an asset. Unfortunately decimals are not supported by the rewards circuit and the `0.05` would get truncated/floored to 0, so a workaround is applied where `0.05*precision NAM` are rewarded for every `precision T` units of an asset. How is `precision` chosen for a given asset `T`? The shielded rewards algorithm currently uses a precision of `10^(max(denom,3)-3)` under the assumption that `1 T` is equivalent to a raw amount of `10^denom`. Unfortunately, `denom` always equals 0 for IBC tokens (see https://github.com/anoma/namada/blob/75059e7aab2c96f2d8f867770c382c9e07da18bc/crates/trans_token/src/storage.rs#L185) meaning that `precision = 1` meaning that the reward fraction is floored to `0` just like the `0.05` example above. So probably no reward will be distributed for IBC tokens even if its parameters were set to be non-zero.

PR #4423, opened 2025-02-27
Allow the shielded reward precision for tokens to be explicitly set.

murisi:
## Describe your changes
An attempt to address https://github.com/anoma/namada/issues/4422 by introducing new storage keys to hold the shielded reward precision for each token and reading that instead of deducting it. More precisely, the changes are as follows:
* Made a new storage key to hold the shielded rewards precision for each token
* The function for obtaining a token's reward precision now reads this new storage key
  * And if the storage key is uninitialized, it initializes it using the old algorithm: `10^(max(denom,3)-3)`
* Implements as an example a migration that sets the precision for existing IBC tokens
  * Alternatively, for an example of a governance upgrade taking advantage of this PR, see: https://github.com/anoma/namada-governance-upgrades/pull/38 .

This PR should be backwards compatible in the sense it can read/support current mainnet state without any migrations. And simultaneously it allows the shielded rewards precision to be set for new tokens as they are introduced by governance proposals. But to fix the precision of IBC tokens that have already been introduced into the system, purging of its `ConversionLeaf::conversion` entries (by setting them to `MaspAmount::zero()`) would also have to be done in addition to setting its desired shielded reward precision in storage. (More generally, once the shielded reward precision for a token is set, it cannot be changed without doing a similar purging process.)

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4423-configurable-masp-precision.md
+++ b/.changelog/unreleased/improvements/4423-configurable-masp-precision.md
@@ -1,0 +1,2 @@
+- Allow the shielded reward precision for tokens to be explicitly set via
+  governance proposals ([\#4423](https://github.com/anoma/namada/pull/4423))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5088,9 +5088,9 @@ dependencies = [
 
 [[package]]
 name = "masp_note_encryption"
-version = "1.2.0"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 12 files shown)


grarco (approved):

yito88:
@murisi Do we need to modify the e2e test [`ibc_token_inflation`](https://github.com/anoma/namada/blob/395b2ba31f89173190e45edc7fd54f8a6febb4b4/crates/tests/src/e2e/ibc_tests.rs#L900) to check this change is effective?

yito88 (approved):

PR #4442, opened 2025-03-03
Murisi/client anchor patch optimized

murisi:
## Describe your changes
Implemented a patch on the MASP client to search for the `Transaction` ordering used by the protocol. Specifically, for a given block, the client goes through different permutations of its `Transaction`s and checks if applying them would result in a note commitment tree anchor in agreement with the protocol. When this condition is met, the client then moves onto the next block doing the same thing until the client is fully synchronized.

Improvements from original patch in https://github.com/anoma/namada/tree/murisi/check-acknowledgement :
* Restricted the set of permutations tried to only those that can actually happen
* Reduced the amount of commitment anchor existence queries by only doing so on block boundaries
* Deduplicated calls to `update_witness_map` by not destroying `ShieldedContext` state after finding correct ordering
* Now respect `last_witnessed_tx` so that previously applied `Transaction`s are no longer reapplied
* Now respect the `needs_witness_map_update` capability by not updating witness map if it is false
* Replaced error-prone index arithmetic with iterator usage and factored out hack into separate function
* Adjusted the client to not reorder (to block ordering) the `Transaction`s it obtains from events or the indexer

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/sdk/src/masp/utilities.rs
+++ b/crates/sdk/src/masp/utilities.rs
@@ -111,6 +111,7 @@ impl<C: Client + Send + Sync> LedgerMaspClient<C> {
                     .data
             };
 
+            let mut masp_index = 0;
             for IndexedMaspData {
                 tx_index,
                 masp_refs,
@@ -126,6 +127,7 @@ impl<C: Client + Send + Sync> LedgerMaspClient<C> {
                     &mut txs,
                     extracted_masp_txs,
                     height.into(),
+                    &mut masp_index,
                     tx_index,
                 )?;
             }
@@ -517,17 +519,23 @@ impl MaspClient for IndexerMaspClient {
             }
         }
 
-        let mut stream_of_fetches = stream::iter(fetches)
-            .buffer_unordered(self.shared.max_concurrent_fetches);
+        let mut stream_of_fetches =
+            stream::iter(fetches).buffered(self.shared.max_concurrent_fetches);
         let mut txs = vec![];
 
         while let Some(result) = stream_of_fetches.next().await {
+            let mut prev_block_height = None;
+            let mut masp_index = 0;
             for Transaction {
                 batch,
                 block_index,
                 block_height,
             } in result?
             {
+                if Some(block_height) != prev_block_height {
+                    masp_index = 0;
+                    prev_block_height = Some(block_height);
+                }
                 let mut extracted_masp_txs = Vec::with_capacity(batch.len());
 
                 for TransactionSlot { bytes } in batch {
@@ -546,6 +554,7 @@ impl MaspClient for IndexerMaspClient {
                     &mut txs,
                     extracted_masp_txs,
                     block_height.into(),
+                    &mut masp_index,
                     block_index.into(),
                 )?;
             }
@@ -657,17 +666,22 @@ impl MaspClient for IndexerMaspClient {
         Ok(payload
             .notes_index
             .into_iter()
+            .enumerate()
             .map(
-                |Note {
-                     block_index,
-                     batch_index,
-                     block_height,
-                     note_position,
-                 }| {
+                |(
+                    masp_index,
+                    Note {
+                        block_index,
+                        batch_index,
+                        block_height,
+                        note_position,
+                    },
+                )| {
                     (
                         IndexedTx {
-                            index: TxIndex(block_index),
-                            height: BlockHeight(block_height),
+                            block_index: TxIndex(block_index),
+                            masp_index: masp_index as u32,
+                            block_height: BlockHeight(block_height),
                             batch_index: Some(batch_index),
                         },
                         note_position,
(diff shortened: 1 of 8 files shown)


crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -381,7 +382,101 @@ where
         }
     }
 
-    fn apply_cache_to_shielded_context(
+    // Search for the Transaction order usedby the protocol using trial and
```
batconjurer, line 385:
```suggestion
    // Search for the Transaction order used by the protocol using trial and
```

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -381,7 +382,101 @@ where
...
+    ) -> Result<Vec<(IndexedTx, Transaction)>, eyre::Error> {
+        let mut ordered_txs = vec![];
+        // Get the unordered Transactions with their indices
+        let mut indexed_txs = self.cache.fetched.take().into_iter().peekable();
+        let needs_witness_map_update =
+            self.client.capabilities().needs_witness_map_update();
+        while let Some(first_tx) = indexed_txs.next() {
+            // Take only Transactions belonging to one block
+            let mut block_txs = vec![first_tx];
+            while let Some(next_tx) = indexed_txs
```
batconjurer, line 399: This can be simplified using `take_while` and `collect`. Then we don't need the iterator to be peekable. You might need to combine the `take_while` with the iterator take `by_ref`. Not sure

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -381,7 +382,101 @@ where
...
+                .next_if(|next_tx| next_tx.0.height == block_txs[0].0.height)
+            {
+                block_txs.push(next_tx);
+            }
+            // Do not apply these Transactions if they have been applied before
+            if !needs_witness_map_update
```
sug0, line 405: This method will return an empty vec of `ordered_txs` if `needs_witness_map_update = true`, since we are skipping all updates of that vec from the outer-most loop (we hit the `continue` branch). This means that the indexer client is broken with this patch.

I think we should skip calling `transaction_order_search` from the indexer client, anyway, because the commitment tree isn't built locally, but rather fetched from the indexer. That also means we don't need to modify the `IndexerMaspClient` with an additional tendermint client field, because in practice it will never be needed.

sug0 (changes requested):

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -381,7 +382,101 @@ where
...
+                        tracing::debug!("Transaction Index: {:?}", indexed_tx);
+                        self.ctx
+                            .update_witness_map(*indexed_tx, stx_batch)
+                            .await?;
+                    }
+                }
+                // Compute what the note commitment tree root would look like
+                // after applying the above order and check if it's recognized
+                // by the protocol
+                let root = self.ctx.tree.root();
```
batconjurer, line 448: for this search, we only need to update the tree, not the witness maps and note indexes. Once we have found the order, we can update those. This should save a lot of computation and cloning here.

batconjurer (changes requested):

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -143,7 +143,7 @@ impl<U: ShieldedUtils + MaybeSend + MaybeSync> ShieldedWallet<U> {
 
     /// Update the merkle tree of witnesses the first time we
     /// scan new MASP transactions.
-    pub(crate) fn update_witness_map(
+    pub(crate) async fn update_witness_map(
```
sug0, line 146: this doesn't need to by async

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
sug0: After employing a similar hack to the masp indexer, the `/tx` endpoint should be returning the correct execution order, so I think we can restrict the ordering shenanigens to the ledger client (i.e. the client that solely utilizes the cometbft rpc).

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
batconjurer: Well shucks.

crates/sdk/src/masp/utilities.rs
```diff
@@ -508,17 +519,23 @@ impl MaspClient for IndexerMaspClient {
             }
         }
 
-        let mut stream_of_fetches = stream::iter(fetches)
-            .buffer_unordered(self.shared.max_concurrent_fetches);
+        let mut stream_of_fetches =
+            stream::iter(fetches).buffered(self.shared.max_concurrent_fetches);
```
sug0, line 523: This change won't matter, because we're using a `BTreeMap` to sort the fetched txs based on their `IndexedTx` key. Until a fix on the node software is available, I think we should change the inner type of the `Fetched` struct to a `Vec`, so that we preserve the order of the txs returned by the indexer's webserver.

sug0 (changes requested):

crates/sdk/src/masp/utilities.rs
sug0: It doesn't create extra problems, it's just that the order returned by the client (in your PR) is not respected, because all fetched masp txs end up in the `Fetched` cache, which sorts txs based on the returned `IndexedTx`.

PR #4463, opened 2025-03-12
Murisi/load storage conversions

murisi:
## Describe your changes
Based on #4423 . Implemented loading conversion leaf updates from storage. This feature is required because the asset conversion table is currently stored in memory making it difficult to modify using governance proposals. Furthermore, the fact that a given epoch's conversion state is a function of the previous epoch's conversion state makes it difficult to correctly alter it without more advanced migration capabilities (that can inspect the current state).

So in order to address the above problems, this PR does the following:
* It introduces a new key to store conversion leaf updates: `<MASP address>/conversions/<asset type>`. These keys store `AllowedConversion`s instead of `I128Sum` to give the creator a opportunities to optimize their construction.
* The reward distribution algorithm now scans through all keys of this form and sets the corresponding entry in the conversion table (in memory) to the value currently in storage
* Afterwards the reward distribution algorithm deletes all keys of the above form to ensure that they are not reapplied in future epochs. I.e. these keys are not always initialized and are generally intended to be write-only.
* Introduces a new integration test called `reset_conversions` that zeros all the rewards that have been distributed for BTC and checks that BTC holders' balances go down. The conversion leafs are not simply zeroed because that would make them malformed.
* Implements as an example a migration that clears all existing rewards for specified IBC tokens using the mechanism introduced by this PR. The generated migration is about 27 megabytes. Alternatively, existing rewards can be cleared using a governance proposal like https://github.com/anoma/namada-governance-upgrades/pull/39 .

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/conversion.rs
+++ b/crates/shielded_token/src/conversion.rs
@@ -107,11 +107,11 @@ where
         storage.read(&reward_precision_key)?.map_or_else(
             || -> Result<u128> {
                 // Since reading reward precision has failed, choose a
-                // thousandth of the given token
+                // thousandth of the given token. But clamp the precision above
+                // by 10^38, the maximum power of 10 that can be contained by a
+                // u128.
                 let precision_denom =
-                    std::cmp::max(u32::from(denomination.0), 3)
-                        .checked_sub(3)
-                        .expect("Cannot underflow");
+                    u32::from(denomination.0).saturating_sub(3).clamp(0, 38);
                 let reward_precision = checked!(10u128 ^ precision_denom)?;
                 // Record the precision that is now being used so that it does
                 // not have to be recomputed each time, and to
--- a/crates/shielded_token/src/conversion.rs
+++ b/crates/shielded_token/src/conversion.rs
@@ -27,17 +27,20 @@ use namada_core::masp::MaspEpoch;
 use namada_core::token::MaspDigitPos;
 use namada_core::token::{Amount, DenominatedAmount, Denomination};
 use namada_core::uint::Uint;
+#[cfg(any(feature = "multicore", test))]
 use namada_state::iter_prefix_with_filter_map;
 use namada_systems::{parameters, trans_token};
 
-use crate::storage_key::{
-    is_masp_conversion_key, masp_conversion_key_prefix, masp_kd_gain_key,
-    masp_kp_gain_key, masp_last_inflation_key, masp_last_locked_amount_key,
-    masp_locked_amount_target_key, masp_max_reward_rate_key,
-    masp_reward_precision_key,
-};
 #[cfg(any(feature = "multicore", test))]
-use crate::storage_key::{masp_assets_hash_key, masp_token_map_key};
+use crate::storage_key::{
+    is_masp_conversion_key, masp_assets_hash_key, masp_conversion_key_prefix,
+    masp_token_map_key,
+};
+use crate::storage_key::{
+    masp_kd_gain_key, masp_kp_gain_key, masp_last_inflation_key,
+    masp_last_locked_amount_key, masp_locked_amount_target_key,
+    masp_max_reward_rate_key, masp_reward_precision_key,
+};
 #[cfg(any(feature = "multicore", test))]
 use crate::{ConversionLeaf, Error, OptionExt, ResultExt};
 use crate::{Result, StorageRead, StorageWrite, WithConversionState};
@@ -525,8 +528,6 @@ fn apply_stored_conversion_updates<S>(storage: &mut S) -> Result<()>
 where
     S: StorageWrite + StorageRead + WithConversionState,
 {
-    use masp_primitives::transaction::components::I128Sum;
-
     let conversion_key_prefix = masp_conversion_key_prefix();
     let mut conversion_updates = BTreeMap::new();
     // Read conversion updates from storage and store them in a map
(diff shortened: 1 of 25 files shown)


crates/shielded_token/src/conversion.rs
```diff
@@ -503,6 +573,7 @@ where
 
     use crate::mint_rewards;
 
+    //apply_stored_conversion_updates(storage)?;
```
batconjurer, line 576: what's going on here?

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1447,1054 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+                NAM,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 0.18963"));
+
+    /*// Get the necessary information to construct BTC asset types
```
batconjurer, line 1738: why is all this commented out?

batconjurer (approved):

crates/shielded_token/src/conversion.rs
```diff
@@ -490,6 +495,48 @@ where
...
+        storage,
+        &conversion_key_prefix,
+        is_masp_conversion_key,
+    )? {
+        match conv_result {
+            Ok((asset_type, conv)) => {
+                conversion_updates.insert(asset_type, conv);
+            }
+            Err(err) => {
+                tracing::warn!("Encountered malformed conversion: {}", err);
```
grarco, line 518: Would it be better to crash in this case? Not sure about the consequences of missing an intended conversion update

crates/shielded_token/src/conversion.rs
```diff
@@ -490,6 +495,48 @@ where
...
+    }
+    // Apply the conversion updates to the in memory structure
+    let assets = &mut storage.conversion_state_mut().assets;
+    for (asset_type, conv) in conversion_updates {
+        let Some(leaf) = assets.get_mut(&asset_type) else {
+            tracing::warn!(
+                "Encountered non-existent asset type: {}",
+                asset_type
+            );
+            continue;
```
grarco, line 531: Same question here

crates/shielded_token/src/storage_key.rs
```diff
@@ -160,6 +175,20 @@ pub fn is_masp_token_map_key(key: &storage::Key) -> bool {
         ] if *addr == address::MASP && prefix == MASP_TOKEN_MAP_KEY)
 }
 
+/// Check if the given storage key is a masp nullifier key
```
grarco, line 178: Wrong docstring here

crates/shielded_token/src/conversion.rs
murisi: Good question thanks. I think the answer depends on the context in which we use this PR. If we are using this PR to purge existing conversions that should not be allowed (i.e. 10^-8 BTC giving 10^71 NAM rewards), then perhaps crashing would be better. But if we are using this PR under the assumption that existing conversions should be allowed and that we only just want to change their precisions, then perhaps printing out a warning might be okay. The worst thing that I think can happen here is that after the deltas are added, the conversion for the target asset type becomes malformed (i.e. no longer of the form `-PRECISION TOK[older epoch] + PRECISION TOK[current epoch] + REWARD NAM[epoch 0]`) whilst still being conceptually allowed (since it's the sum of two allowed conversions). And here I believe that malformed conversions are simply unusable by clients, not exploitable.

This PR is written assuming the latter context since the pre-phase 4 hard fork is just about adjusting precisions rather than fixing past malfunctions. My thought process was that is if the migration somehow generates and writes to storage incorrect `AllowedConversion` serializations, then this shouldn't bring everything down (or break anything) and we can always try again before the next MASP epoch. But hopefully this worst case scenario will never actually happen since it is easy to check that the serializations are valid. Are there any other considerations here?

crates/shielded_token/src/conversion.rs
murisi: I think roughly the same considerations as above apply here too, though in this case the problem is that a conversion is made for an asset type that has not been previously rewarded by the shielded rewards program. I think this could only be caused by a fencepost error in the migration or an application of the migration in the incorrect MASP epoch. I think these errors should probably be avoidable with enough care and testing. But should they happen in the latter context (described in the response above), it should probably also be okay for the protocol to proceed as if nothing happened.

crates/shielded_token/src/conversion.rs
grarco: I think your point makes sense, thanks!

crates/shielded_token/src/storage_key.rs
murisi: Fixed, thanks!

crates/shielded_token/src/conversion.rs
murisi: Actually, this warning should be triggered when applying migrations to the mainnet, campfire, and housefire. This is because the migrations generate conversions for tokens starting from the very epoch when in actuality there are no conversions for asset types with earlier timestamps. And this in turn is because some of the tokens did not have shielded rewards enabled from the beginning. Ultimately, all this means that `storage.conversion_state_mut().assets` will not have keys for the earlier asset types of a token. However this should not be problematic for the reasons above.

PR #4489, opened 2025-03-21
Sorted tx results in events

### LINKED ISSUE #3873: `BatchedTxResult`s order
The `TxResult` type that carries results for all the inner transactions of a given batch internally relies on a `BTreeMap`, meaning that the results are ordered based on the hash of the inner tx and not the order in which the transactions were actually executed.

https://github.com/anoma/namada/blob/810c258af3ddf2483af4d1306644db09bb2987e3/crates/tx/src/data/mod.rs#L222

This type then gets serialized and emitted as part of the `tx/applied` event under the `batch` attribute, so tools cannot rely on the order of the entries of this map. This is not an issue since they should use the hash of the inner tx to retrieve its result, but we could also guarantee a correct ordering nonetheless.

This is related to #3856 and shares the same considerations about being breaking.

grarco:
## Describe your changes

Closes #3873.

Changes the `BTreeSet` of `TxResult` for a `IndexSet` to preserve the order of inner txs' results in the events.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4489-sorted-tx-results.md
+++ b/.changelog/unreleased/improvements/4489-sorted-tx-results.md
@@ -1,0 +1,2 @@
+- The orderd of results of the inner transactions of a batch is now preserved in
+  the events. ([\#4489](https://github.com/anoma/namada/pull/4489))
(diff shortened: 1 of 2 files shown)


brentstone:
@grarco can this be merged?

Issue #4372, opened 2025-02-12
MASP Rewards Distribution Algorithm Mints (Slightly) Too Much NAM


murisi:
The current MASP rewards distribution algorithm mints NAM in order to be able to service claims for shielded rewards. The problem is that it mints too much NAM for a few reasons including users choosing to unshield without claiming their rewards and the token being distributed across shielded accounts in such a way that the majority of token owners never meet the conversion threshold.

The specific case this issue is concerned with is overminting due to rounding errors occurring in the execution of the rewards distribution algorithm. This overminting can be observed at https://github.com/anoma/namada/blob/63665af03fb2e67d42c9e4e77de4402b95137e8e/crates/tests/src/integration/masp.rs#L2285 where the MASP address' NAM balance is slightly more than zero even though all assets have been unshielded. This amount though it is probably unclaimable still counts as "locked" and will impact future inflation/rewards computations.

The cause of this problem is that for non-native tokens NAM is currently minted based on the inflated reward instead of the deflated reward. (See: https://github.com/anoma/namada/blob/63665af03fb2e67d42c9e4e77de4402b95137e8e/crates/shielded_token/src/conversion.rs#L470). This is problematic because the `AllowedConversion`s that are actually put into the conversion Merkle tree use the deflated reward. A potential solution to this is to first compute the total deflated reward in NAM[0] and then inflate that amount to the current epoch using a NAM conversion. This approach is attempted in https://github.com/anoma/namada/pull/4369 .

But even https://github.com/anoma/namada/pull/4369 (which mints NAM based on deflated rewards) still overmints NAM. See: https://github.com/anoma/namada/blob/11a2498222c4ce07077b1d636e8f420e454dc0ee/crates/tests/src/integration/masp.rs#L2285 . The reason is that in order to have sufficient NAM for the case where users combine their pre-existing NAM[0] with NAM[0] rewards from their non-native tokens, this algorithm avoids rounding down/throwing out amounts not meeting the conversion threshold too early.

The PR https://github.com/anoma/namada/pull/4285 manages to avoid overminting of NAM, but this comes at the cost of compounding rewards. See: https://github.com/anoma/namada/blob/817bc6e2fc212ce12bdd8bab5c0a3ef4ebd99630/crates/tests/src/integration/masp.rs#L2287 . In this PR overminting does not happen because rewards for non-native tokens are no longer deflated into `NAM[0]` terms meaning that the same value that's put into the conversion tree also determines minting and because early rounding can happen.

PR #4491, opened 2025-03-21
Added integration test for auto-compounding.

murisi:
## Describe your changes
Added an integration test to check that the automatic compounding of a shielded account balance is the same as manually compounding a shielded account balance. Here, manual compounding refers to a shielded pool user immediately unshielding shielded rewards they receive (thereby applying/realizing rewards at that point in time) and then reshielding them (with the timestamp of the current MASP epoch). Though this means that different allowed conversions (from the protocol) are now applied onto the automatically and manually compounded accounts, the results of shielded balance queries are expected to be the approximately the same for all MASP epochs.

See the results of compounding a principal amount of 0.1 BTC below:
| Ep.| Auto Compounding | Manual Compounding | Claimable | MASP Balance |
|---|---|---|---|---|
|0| 0.1 BTC | 0.1 BTC | 0.2 BTC |
|1| 0.1 BTC + 0.0317 NAM | 0.1 BTC + 0.0317 NAM | 0.2 BTC + 0.0634 NAM | 0.2 BTC + 0.0634 NAM |
|2| 0.1 BTC + 0.09534 NAM | 0.1 BTC + 0.09533 NAM | 0.2 BTC + 0.19067 NAM | 0.2 BTC + 0.190688 NAM |
|3| 0.1 BTC + 0.191008 NAM | 0.1 BTC + 0.190982 NAM | 0.2 BTC + 0.38199 NAM | 0.2 BTC + 0.382016 NAM |
|4| 0.1 BTC + 0.31854 NAM | 0.1 BTC + 0.31851 NAM | 0.2 BTC + 0.63705 NAM | 0.2 BTC + 0.637092 NAM |

Note that auto-compounding gives slightly higher balances than manual compounding. This is probably because NAM  conversions to the latest epoch have ever higher thresholds which means that the dust that doesn't get converted will also get higher. Also note that the MASP balance is slightly higher than the claimable amount, which is the sum of the shielded balances held by all the shielded key holders. This is due to rounding errors in the deflation process as explained in https://github.com/anoma/namada/issues/4372 . The growth of the gap between what is claimable and the MASP balance will depend on the inflation rate of NAM, and the relative precisions of NAM and the token being rewarded.
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/tests/src/integration/masp.rs
+++ b/crates/tests/src/integration/masp.rs
@@ -1447,6 +1447,1118 @@ fn enable_rewards_after_shielding() -> Result<()> {
     Ok(())
 }
 
+/// In this test we verify that the results of auto-compounding are
+/// approximately equal to what is obtained by manually unshielding and
+/// reshielding each time.
+#[test]
+fn auto_compounding() -> Result<()> {
+    // This address doesn't matter for tests. But an argument is required.
+    let validator_one_rpc = "http://127.0.0.1:26567";
+    // Download the shielded pool parameters before starting node
+    let _ = FsShieldedUtils::new(PathBuf::new());
+    let (mut node, _services) = setup::setup()?;
+    // Wait till epoch boundary
+    node.next_masp_epoch();
+    // Send 0.1 BTC from Albert to Albert's payment address
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            apply_use_device(vec![
+                "shield",
+                "--source",
+                ALBERT,
+                "--target",
+                AA_PAYMENT_ADDRESS,
+                "--token",
+                BTC,
+                "--amount",
+                "0.1",
+                "--signing-keys",
+                ALBERT_KEY,
+                "--node",
+                validator_one_rpc,
+            ]),
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains(TX_APPLIED_SUCCESS));
+
+    // Send 0.1 BTC from Albert to Bertha's payment address
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            apply_use_device(vec![
+                "shield",
+                "--source",
+                ALBERT,
+                "--target",
+                AB_PAYMENT_ADDRESS,
+                "--token",
+                BTC,
+                "--amount",
+                "0.1",
+                "--signing-keys",
+                ALBERT_KEY,
+                "--node",
+                validator_one_rpc,
+            ]),
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains(TX_APPLIED_SUCCESS));
+
+    // sync the shielded context
+    run(
+        &node,
+        Bin::Client,
+        vec![
+            "shielded-sync",
+            "--viewing-keys",
+            AA_VIEWING_KEY,
+            AB_VIEWING_KEY,
+            "--node",
+            validator_one_rpc,
+        ],
+    )?;
+
+    // Assert BTC balance at Albert's shielded key is 0.1
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            vec![
+                "balance",
+                "--owner",
+                AA_VIEWING_KEY,
+                "--token",
+                BTC,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("btc: 0.1"));
+
+    // Assert BTC balance at Bertha's shielded key is 0.1
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            vec![
+                "balance",
+                "--owner",
+                AB_VIEWING_KEY,
+                "--token",
+                BTC,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("btc: 0.1"));
+
+    // Assert NAM balance at Albert's shielded key is 0
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            vec![
+                "balance",
+                "--owner",
+                AA_VIEWING_KEY,
+                "--token",
+                NAM,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 0"));
+
+    // Assert NAM balance at Bertha's shielded key is 0
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
+            vec![
+                "balance",
+                "--owner",
+                AB_VIEWING_KEY,
+                "--token",
+                NAM,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 0"));
+
+    // Assert the rewards estimate for Albert's shielded key is also zero
+    let captured = CapturedOutput::of(|| {
+        run(
+            &node,
+            Bin::Client,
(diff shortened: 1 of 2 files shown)


crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+                ALBERT_KEY,
+                "--node",
+                validator_one_rpc,
+            ]),
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains(TX_APPLIED_SUCCESS));
+
+    // Send 0.1 BTC from Albert to PA
```
batconjurer, line 1482: I've forgotten who the `AB_PAYMENT_ADDRESS` belongs to. Is it Albert's diversified address? Either way, a little clarification in the comment here seems apropos.

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+                NAM,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("nam: 0"));
+
+    // Assert NAM balance at VK(A) is 0
```
batconjurer, line 1578: Same here

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+                BTC,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(captured.contains("btc: 0.1"));
+
+    // Assert BTC balance at VK(B) is still 0.1
```
batconjurer, line 1668: ditto

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+            "shielded-sync",
+            "--viewing-keys",
+            AA_VIEWING_KEY,
+            AB_VIEWING_KEY,
+            "--node",
+            validator_one_rpc,
+        ],
+    )?;
+
+    // Send 0.0317 NAM from Albert to PA
```
batconjurer, line 1824: ditto

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+    node.next_masp_epoch();
+
+    // sync the shielded context
+    run(
+        &node,
+        Bin::Client,
+        vec!["shielded-sync", "--node", validator_one_rpc],
+    )?;
+
+    // Assert BTC balance at VK(A) is still 0.1
```
batconjurer, line 1873: ditto

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+    let captured = CapturedOutput::of(|| {
```
batconjurer, line 1874: Should we also have a balance check for `AB_VIEWING_KEY`?

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+                AB_VIEWING_KEY,
+                "--node",
+                validator_one_rpc,
+            ],
+        )
+    });
+    assert!(captured.result.is_ok());
+    assert!(
+        captured.contains(
+            "Estimated native token rewards for the next MASP epoch: 0"
```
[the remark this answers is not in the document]
batconjurer: Since we are manually compounding for `AB`, I think 0 is correct(?). But for `AA`, which is tested above, I think you have a point.

crates/tests/src/integration/masp.rs
```diff
@@ -1442,6 +1442,1069 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+
+    // sync the shielded context
+    run(
+        &node,
+        Bin::Client,
+        vec!["shielded-sync", "--node", validator_one_rpc],
+    )?;
+
+    // Assert BTC balance at VK(A) is still 0.1
+    let captured = CapturedOutput::of(|| {
```
batconjurer, line 2087: Again, should we check `AB` balance as well?

batconjurer: Hell, it seems a few of these checks can be made into a closure to DRY up the code a bit.

batconjurer (changes requested):

crates/tests/src/integration/masp.rs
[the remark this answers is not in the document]
grarco: It could also be due to this https://github.com/anoma/namada/pull/4285#issuecomment-2724892598

crates/tests/src/integration/masp.rs
murisi: I've clarified the target payment addresses, thanks!

crates/tests/src/integration/masp.rs
murisi: Ah, yes you're right. Done. Thanks!

crates/tests/src/integration/masp.rs
murisi: Done, thanks! I had briefly thought about this and then got carried away with making sure that the balance numbers make sense and that all the rounding works as expected.

crates/tests/src/integration/masp.rs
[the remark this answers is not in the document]
murisi: > Since the original BTCs are still in the pool, shouldn't we see the extra rewards coming from the new epoch?

Thanks for bringing this up. I've just checked, and there are actually rewards. The call to `captured.contains` here is misleading because the string `"Estimated ... next MASP epoch: 0"` is technically contained by strings of the form `"Estimated ... next MASP epoch: 0.xyz..."`.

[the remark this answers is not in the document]
murisi: > Since we are manually compounding for AB, I think 0 is correct(?). But for AA, which is tested above, I think you have a point.

Once the substring issue above is fixed, it seems that the estimate is not actually zero for `AA` or for `AB`. But is this what we expect? For `AA` (automatic compounding), I think this is what we all expected. For `AB` (manual compounding) this is what I would expect given my superficial understanding of the estimates algorithm. It takes the manually shielded balance after it is exchanged to the current epoch, and essentially reapplies/duplicates the conversions from the previous epoch to the current (after relabelling them to make it look like they convert from the current epoch to the next one in the future). @batconjurer Please let me know if we should still be concerned about the expected estimates given the latest push? Thanks!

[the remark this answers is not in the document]
batconjurer: Yes, that's how it works, so there should be rewards for both cases

murisi:
> Should this still be worked on if #4285 is closed?

I have attempted to address all the review comments, and am awaiting feedback. So the work here might already be complete.

> Should this still be merged if #4285 is closed?

I think there is value on merging this PR. Probably this PR does three things: 1) it verifies the invariant that manual compounding is equivalent to automatic compounding is satisfied, 2) it documents the existence of such an invariant on the shielded reward system (on the level of the `namadac` client), 3) it prevents future regressions with respect to this invariant. The only drawback of this PR I can think of is that it adds around 40 seconds (on my machine) to the time it takes to execute the integration tests. But, ultimately, merging this PR is optional in the sense that it has already served its purpose of allowing us to choose between manual and automatic compounding.

> ... if #4285 is closed?

This PR is only applicable when #4285 is closed. This is because this PR checks that automatic compounding is equivalent to manual compounding while #4285 which removes automatic compounding. Concretely, if #4285 and #4491 were to be combined, the integration test in this PR would fail due to the unmet expectation that shielded balances automatically increase on each epoch.

grarco (approved):

PR #4542, opened 2025-04-07
Make the control coefficient the inverse of the epochs per year.

murisi:
## Describe your changes
Currently the MASP inflation control coefficient is `max_reward_rate/epochs_per_year`. This is not problematic for the running of the rewards system, but it potentially complicates the semantics of the formula and unnecessarily makes the numerical bounds for desirable nominal proportional gain values dependent upon the selection of the maximum reward rate. This PR removes the factor `max_reward_rate` from the control value and adjusts the nominal proportional gain and nominal derivative gain values in tests so that the same result is achieved.

In more detail, the minor problems with the including the maximum reward rate in the control coefficient are:
1) it's somewhat redundant because the maximum reward rate is already being used to clamp the control value since it's a factor of the other argument to the `min` function. Arguably people might expect the maximum reward rate to behave more like a clamp on reward rates (or the control value specifically) rather than a general scaling factor.
2) it's harder to understand what's being computed because the control value computation involves the maximum reward rate in a way that does not eventually multiply it by the current supply of NAM - had this been the case, then their product could have been understood to be the maximum reward
3) it makes the numerical bounds on the set of nominal proportional gain values that make the PD-controller yield non-zero inflation dependent upon the chosen maximum reward rate value (which too has its own selection constraints). It might be simpler if these values can be set independently. See: https://hackmd.io/oqLHjyvwRMaxSD5JJNrB2w?both#Notes

Applying this PR would require the nominal proportional gain and nominal derivative gain parameters in the target network to be modified. So it may not be worth trying to address this issue at this point.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/tests/src/integration/masp.rs
+++ b/crates/tests/src/integration/masp.rs
@@ -623,8 +623,10 @@ fn values_spanning_multiple_masp_digits() -> Result<()> {
             // the inflation being computed by the pd controller to
             // exceed the amount of test tokens in the masp
             max_reward_rate: Dec::from_str("999999999999999.0").unwrap(),
-            kp_gain_nom: Dec::from_str("99999999999999999999").unwrap(),
-            kd_gain_nom: Dec::from_str("99999999999999999999").unwrap(),
+            kp_gain_nom: Dec::from_str("99999999999999899999000000000000001")
+                .unwrap(),
+            kd_gain_nom: Dec::from_str("99999999999999899999000000000000001")
+                .unwrap(),
             locked_amount_target: u64::MAX,
         }),
         &mut node.shell.lock().unwrap().state,
--- a/.changelog/unreleased/improvements/4542-simplify-masp-inflation.md
+++ b/.changelog/unreleased/improvements/4542-simplify-masp-inflation.md
@@ -1,0 +1,2 @@
+- Simply the MASP inflattion PD-controller formula to increase clarity and
+  reduce redundancy. ([\#4542](https://github.com/anoma/namada/pull/4542))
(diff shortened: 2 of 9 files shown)


crates/tests/src/integration/masp.rs
```diff
@@ -623,8 +623,10 @@ fn values_spanning_multiple_masp_digits() -> Result<()> {
             // the inflation being computed by the pd controller to
             // exceed the amount of test tokens in the masp
             max_reward_rate: Dec::from_str("999999999999999.0").unwrap(),
-            kp_gain_nom: Dec::from_str("99999999999999999999").unwrap(),
-            kd_gain_nom: Dec::from_str("99999999999999999999").unwrap(),
+            kp_gain_nom: Dec::from_str("99999999999999899999000000000000001")
```
batconjurer, line 626: I find it exceptionally odd that these numbers needed to be changed. What's up with that?

batconjurer (approved):
This looks fine, although I have one question. Left a comment

murisi:
This PR is being closed because it might disrupt the testing that has already begun. It should be noted that without this PR, nominal proportional gain and nominal derivative gain values are likely larger (maybe 100x) than would be the case if this PR were to be applied. This is because these two values need to compensate for the fact that they are multiplied by `max_reward_rate`, a value that is usually around 0.01.

brentstone:
Thx @murisi, maybe we reopen this somewhere down the line but I think best to keep on the backburner for now.

PR #4543, opened 2025-04-07
Murisi/native token migrations

murisi:
## Describe your changes
Based on https://github.com/anoma/namada/pull/4540 . Extended the migration code to allow the conversions for arbitrary Namada addresses to be reset. If the token being reset happens to be the native token, then the scheduled base native precision key will also be set. This functionality might be useful if we determine that we need to change the native token's precision in order to actually realize rewards for ordinary end-users.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/examples/make-db-migration.rs
+++ b/examples/make-db-migration.rs
@@ -9,8 +9,10 @@ use masp_primitives::transaction::components::I128Sum;
 use namada_core::borsh::BorshSerializeExt;
 use namada_core::hash::Hash;
 use namada_core::masp::{Precision, encode_asset_type};
+use namada_core::storage::Key;
 use namada_macros::BorshDeserializer;
 use namada_migrations::REGISTER_DESERIALIZERS;
+use namada_parameters::storage::get_tx_allowlist_storage_key;
 use namada_sdk::address::Address;
 use namada_sdk::ibc::trace::ibc_token;
 use namada_sdk::masp_primitives::asset_type::AssetType;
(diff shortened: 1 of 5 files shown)


grarco (approved):

examples/make-db-migration.rs
```diff
@@ -122,45 +120,53 @@ fn shielded_reward_precision_migration() {
...
-fn shielded_reward_reset_migration() {
-    pub type ChannelId = &'static str;
-    pub type BaseToken = &'static str;
-
+fn shielded_reward_reset_migration(
+    updates: &mut Vec<migrations::DbUpdateType>,
+) {
+    // The address of the native token. This is what rewards are denominated in.
+    let native_token =
+        Address::from_str("tnam1qxgfw7myv4dh0qna4hq0xdg6lx77fzl7dcem8h7e")
```
batconjurer, line 142: Can we make the string here a const up above?

examples/make-db-migration.rs
```diff
@@ -122,45 +120,53 @@ fn shielded_reward_precision_migration() {
...
 
-    let changes = migrations::DbChanges {
-        changes: updates.into_iter().collect(),
-    };
-    std::fs::write(
-        "shielded_reward_precision_migration.json",
-        serde_json::to_string(&changes).unwrap(),
-    )
-    .unwrap();
+// A convenience data structure to allow token addresses to be more readably
```
batconjurer, line 125: Hyper nit, but top level doc strings should be ///

batconjurer: Oh dear, I see all the doc strings on the top level functions throughout this code uses // instead. I guess just leave it and I'll be mad in private 😅

batconjurer (changes requested):
Nits

examples/make-db-migration.rs
brentstone: lol honestly we have time to fix this docstring thing up and should...

brentstone: ask an LLM to do it for you

murisi: Fixed the top level doc strings, thanks!

examples/make-db-migration.rs
murisi: Put the string into a `const` above, thanks!

batconjurer (approved):

PR #4544, opened 2025-04-08
Improve message for failing atomic batch

### LINKED ISSUE #4462: Improve display of tx result
In `display_batch_resp` we should slightly improve the message we produce.

https://github.com/anoma/namada/blob/427a57cab8c93a18f293b00e9f54e6e9da5ac147/crates/sdk/src/tx.rs#L440

More specifically, in case of a failed atomic batch that still ran masp fee payment, we should make it clear that the first inner transaction was indeed committed and only the rest of the batch has been discarded.

grarco:
## Describe your changes

Closes #4462.
Extends `display_batch_resp` to print information about masp fee payment in case of a failed atomic batch.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/sdk/src/tx.rs
+++ b/crates/sdk/src/tx.rs
@@ -480,14 +480,38 @@ pub fn display_batch_resp(context: &impl Namada, resp: &TxResponse) {
         false
     };
     let batch_results = resp.batch_result();
-    if !batch_results.is_empty() {
+    // If the batch contains at least one transaction
+    if let Some((first_inner_hash, first_result)) = batch_results.first() {
         if !wrapper_successful {
-            display_line!(
-                context.io(),
-                "Since the batch in its entirety failed, none of the \
-                 transactions listed below have been committed. Their results \
-                 are provided for completeness.",
-            );
+            // Check if fees were paid via the shielded pool, in this case the
+            // first transaction of the batch was committed regardless of the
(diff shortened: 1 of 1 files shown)


crates/sdk/src/tx.rs
```diff
@@ -480,14 +480,38 @@ pub fn display_batch_resp(context: &impl Namada, resp: &TxResponse) {
...
+            // batch failure (even for atomic batches)
+            let masp_fee_payment = if let InnerTxResult::Success(res) =
+                first_result
+            {
+                res.events.iter().any(|event| {
+                    event.kind() == &namada_tx::event::masp_types::FEE_PAYMENT
+                })
+            } else {
+                false
+            };
```
sug0, line 497: suggestion: this is more compact and imo more readable
```suggestion
            let masp_fee_payment = matches!(
                first_result,
                InnerTxResult::Success(res)
                    if res.events.iter().any(|event| {
                        event.kind() == &namada_tx::event::masp_types::FEE_PAYMENT
                    }),
            );

```

sug0 (approved):

crates/sdk/src/tx.rs
grarco: Done in a456b1a3fb36de72304c30ff2408a3be47b50058

PR #4566, opened 2025-04-11
Test masp events

grarco:
## Describe your changes

Adds an integration test to check the new masp events: it verifies that the protocol emits these events correctly and that the client can reconstruct the correct shielded state (specifically, the correct order in which the output notes were committed).

Adds an integration test to verify that even in a failing atomic batch the masp fee paying transaction is guaranteed to be committed.

Fixes a minor bug in the integration tests for which arbitrary events were casted to `tx/applied` events.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/tests/src/integration/masp.rs
+++ b/crates/tests/src/integration/masp.rs
@@ -6818,7 +6818,6 @@ fn get_shielded_hash(tx: &namada_sdk::tx::Tx) -> Option<MaspTxId> {
 // that both the protocol and the shielded sync command behave correctly. Since
 // the batches are not atomic check that the valid transactions get committed
 // and the balances are correctly updated
-// FIXME: should test masp fee payment for failing atomic batches? Yes
 #[test]
 fn masp_batch() -> Result<()> {
     // This address doesn't matter for tests. But an argument is required.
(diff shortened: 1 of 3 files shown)


crates/tests/src/integration/masp.rs
```diff
@@ -7830,3 +8138,521 @@ fn speculative_context() -> Result<()> {
...
+        .unwrap()
+        .path();
+    let batch_tx1_bytes = std::fs::read(&file_path).unwrap();
+    std::fs::remove_file(&file_path).unwrap();
+
+    // Create the batch
+    let tx0: namada_sdk::tx::Tx =
+        serde_json::from_slice(&batch_tx0_bytes).unwrap();
+    let tx1: namada_sdk::tx::Tx =
+        serde_json::from_slice(&batch_tx1_bytes).unwrap();
```
sug0, line 8367: you can reconstruct the expected commitment tree from these txs, by getting the `MaspTxId` within. from the rebuilt commitment tree, after execution, you can check that the tree root is in namada's storage

grarco: Good one, I've done it in 0444309cac

sug0 (approved):

PR #4573, opened 2025-04-14
Added an example program to derive constraints on shielded reward parameters

murisi:
## Describe your changes
Wrote a Cargo example that derives constraints on shielded reward parameters in order to ensure that end-users get non-zero rewards. Essentially this example takes the native token supply, incentivisation threshold, target locked amount, and locked amount tolerance and computes bounds on inflation, token precision, and nominal proportional gain sufficient to yield non-zero rewards. See the formulas upon which this example program is based at: https://hackmd.io/oqLHjyvwRMaxSD5JJNrB2w#Setting-parameters-to-ensure-non-zero-rewards . This program can be run with the following command: `cargo run --example shielded-rewards`.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/core/src/token.rs
+++ b/crates/core/src/token.rs
@@ -512,8 +512,8 @@ impl DenominatedAmount {
         }
     }
 
-    /// Attempt to increase the precision of an amount. Can fail
-    /// if the resulting amount does not fit into 256 bits.
+    /// Return an equivalent denominated amount with the given denomination. Can
+    /// fail if the resulting amount does not fit into 256 bits.
     pub fn increase_precision(
         self,
         denom: Denomination,
@@ -534,8 +534,43 @@ impl DenominatedAmount {
             .ok_or(AmountParseError::PrecisionOverflow)
     }
 
+    /// Return the closest denominated amount with the given denomination and
+    /// the error.
+    pub fn approximate(
+        self,
+        denom: Denomination,
+    ) -> Result<(Self, Self), AmountParseError> {
+        if denom.0 < self.denom.0 {
+            // Divide numerator and denominator by a power of 10
+            let amount = self.amount.raw_amount();
+            #[allow(clippy::arithmetic_side_effects)]
+            let (quot, rem) = Uint::from(10)
+                .checked_pow(Uint::from(self.denom.0 - denom.0))
+                .and_then(|scaling| {
+                    amount.checked_div(scaling).zip(amount.checked_rem(scaling))
+                })
+                .ok_or(AmountParseError::PrecisionOverflow)?;
+            let approx = Self {
+                amount: quot.into(),
+                denom,
+            };
+            let error = Self {
+                amount: rem.into(),
+                denom: self.denom,
+            };
+            Ok((approx, error))
+        } else {
+            // Multiply numerator and denominator by a power of 10
+            let error = Self {
+                amount: 0.into(),
+                denom: self.denom,
+            };
+            self.increase_precision(denom).map(|x| (x, error))
+        }
+    }
+
     /// Create a new [`DenominatedAmount`] with the same underlying
-    /// amout but a new denomination.
+    /// amount but a new denomination.
     pub fn redenominate(self, new_denom: u8) -> Self {
         Self {
             amount: self.amount,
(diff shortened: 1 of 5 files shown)


crates/core/src/token.rs
```diff
@@ -590,6 +625,38 @@ impl DenominatedAmount {
         })
     }
 
+    /// Checked division computed to the given precission. Returns `None` on
```
batconjurer, line 628:
```suggestion
    /// Checked division computed to the given precision. Returns `None` on
```

examples/shielded_rewards.rs
```diff
@@ -0,0 +1,289 @@
...
+    // A necessary condition for users holding more than the threshold to get
+    // rewards is that P<=M.
+    let max_precision = incent_threshold.amount().raw_amount();
+    println!(
+        "At an inflation of {} {}, the precision must be between {} and {} in \
+         order to realize non-zero rewards.",
+        inflation, native_code, min_precision, max_precision
+    );
+
+    // Let P be the precision of the token TOK
```
batconjurer, line 184: The doc string should be moved up. `P` was referenced above already

examples/shielded_rewards.rs
```diff
@@ -0,0 +1,289 @@
...
+        .trim()
+        .parse::<Dec>()
+        .map_err(std::io::Error::other)?;
+    println!(
+        "Users holding at least {} {} in the shielded pool will receive \
+         rewards.",
+        precision, incent_code
+    );
+
+    // It must be that S*C/Y >= I, which implies that the reward rate
```
batconjurer, line 199: `C` is defined below. Consider moving doc string up. Also, `S` seems to be missing a definition in the doc strings entirely.

batconjurer (changes requested):
Some nits in the doc strings. Also, I know that this is just an example, but it might actually end up being fairly useful. It will be annoying to have to restart the entire flow because of malformed input. I would suggest two UX improvements: One allow parsing all inputs from a file optionally (this is once users have a rough idea of the inputs and just want to tweak things) and also, instead of eagerly erroring out on bad input, just loop until correct input is given.

crates/core/src/token.rs
murisi: Fixed, thanks.

examples/shielded_rewards.rs
murisi: Thanks. I think `S` is defined at https://github.com/anoma/namada/blob/1cb9fccc857d2c0608f01a01fb72518bda9e5b2b/examples/shielded_rewards.rs#L81 .

brentstone:
@murisi we should merge this if we can, is it ok?

murisi:
> @murisi we should merge this if we can, is it ok?

@brentstone This PR touches a file outside the `examples` directory in order to have greater control over decimal arithmetic, namely `crates/core/src/token.rs`. Even if the mathematics and computations in the `examples` are assumed to be correct (because they have been used successfully in the upgrades), we should try to get at least one approval to the changes in `token.rs` because an error there would be very problematic.

brentstone:
@tzemanovic or @batconjurer would you mind reviewing?

batconjurer:
This is the third time I've been asked to review this with my comments from the first review not being addressed. I'll say it again: 
`
 Also, I know that this is just an example, but it might actually end up being fairly useful. It will be annoying to have to restart the entire flow because of malformed input. I would suggest two UX improvements: One allow parsing all inputs from a file optionally (this is once users have a rough idea of the inputs and just want to tweak things) and also, instead of eagerly erroring out on bad input, just loop until correct input is given.
`

murisi:
> This is the third time I've been asked to review this with my comments from the first review not being addressed. I'll say it again: `Also, I know that this is just an example, but it might actually end up being fairly useful. It will be annoying to have to restart the entire flow because of malformed input. I would suggest two UX improvements: One allow parsing all inputs from a file optionally (this is once users have a rough idea of the inputs and just want to tweak things) and also, instead of eagerly erroring out on bad input, just loop until correct input is given.`

I really don't know how I missed that comment. My bad. Both ideas (file input and error recovery) are very good. The former is a little awkward to implement because the input range of certain variables depends on the value selected for earlier variables, but I think this can be achieved. Indeed file inputs (and outputs) would also allow us to keep (and share) a record of how the precision parameters were derived. And the latter idea would go with increased input validation (like making sure that numbers are within certain ranges). Thanks for bringing up these points - will work on them today or tomorrow.

batconjurer (approved):

PR #4589, opened 2025-04-18
Murisi/set masp params migration

murisi:
## Describe your changes
Implemented a hard-fork migration that allows shielded reward parameters to be set. This is useful when testing shielded rewards on mainnet clones.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/examples/make-db-migration.rs
+++ b/examples/make-db-migration.rs
@@ -36,6 +36,10 @@ pub type ChannelId = &'static str;
 pub type BaseToken = &'static str;
 /// Represents a Namada address in Bech32m encoding
 pub type AddressBech32m = &'static str;
+/// Represents the hash of a WASM binary
+pub type WasmHash = &'static str;
+/// Represents the bytes of a WASM binary
+pub type WasmBytes = &'static [u8];
 /// The type hash of the conversion state structure in v0.31.9
 pub const OLD_CONVERSION_STATE_TYPE_HASH: &str =
     "05E2FD0BEBD54A05AAE349BBDE61F90893F09A72850EFD4F69060821EC5DE65F";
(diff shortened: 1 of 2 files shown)


examples/make-db-migration.rs
```diff
@@ -125,51 +158,63 @@ pub fn shielded_reward_precision_migration(
...
-        (TokenAddress::Ibc("channel-1", "uosmo"), 1000u128),
-        (TokenAddress::Ibc("channel-2", "uatom"), 1000u128),
-        (TokenAddress::Ibc("channel-3", "utia"), 1000u128),
-        (TokenAddress::Ibc("channel-0", "stuosmo"), 1000u128),
-        (TokenAddress::Ibc("channel-0", "stuatom"), 1000u128),
-        (TokenAddress::Ibc("channel-0", "stutia"), 1000u128),
+    const TOKENS: [(TokenAddress, Denomination, Precision); 6] = [
+        (
+            TokenAddress::Ibc("channel-1", "uosmo"),
+            Denomination(0u8),
```
grarco, line 178: Do we really need this extra denomination element in the tuple? It seems like it's always set to 0 and we are not sharing code with the similar loop we run down here in `conversion_state_migration`

examples/make-db-migration.rs
```diff
@@ -450,25 +508,28 @@ pub fn wasm_migration(updates: &mut Vec<migrations::DbUpdateType>) {
...
     // wasm_updates[x].1) The name of WASM being updated
     // wasm_updates[x].2) The bytes of the new WASM code
-    let wasm_updates: Vec<(&str, &str, &[u8])> =
-        vec![
+    const WASM_UPDATES: [(WasmHash, &str, WasmBytes); 2] = [
         (
             "83afcbf97c35188991ae2e73db2f48cb8d019c4295fe5323d9c3dfebcd5dbec0",
             "tx_transfer.wasm",
-            //include_bytes!("tx_transfer.5c7e44e61c00df351fa7c497cd2e186d71909f1a18db0c8d362dff36057e0fbf.wasm"),
+            // include_bytes!("tx_transfer.
```
grarco, line 515: Can we remove these commented lines?

examples/make-db-migration.rs
```diff
@@ -563,6 +624,118 @@ pub fn wasm_migration(updates: &mut Vec<migrations::DbUpdateType>) {
...
+    const TOKENS: [(
+        Denomination,
+        TokenAddress,
+        MaxRewardRate,
+        TargetLockedAmount,
+        KpGain,
+        KdGain,
+    ); 6] = [
+        (
+            Denomination(0),
```
grarco, line 641: Same here, to my understanding the denomination of ibc assets should always be 0

grarco (approved):
Minor comments but looks good to me

examples/make-db-migration.rs
murisi: Good point. It's not necessary to have the denomination element in the tuple if we're considering only the mainnet. But when testing and using this script on localnets with tokens having different denominations (to observe infllation changes), I found it handy to be able to specify non-zero denominations and non-IBC tokens. My thinking in writing this code generally (beyond the pre-phase 4 upgrade) was to provide templates/guides that can be specialized for modifying the state of arbitrary localnets, testnets, and mainnets.

examples/make-db-migration.rs
murisi: While any byte sequence can be used as the third element of this tuple, I suspect the most common expression here will be of the form `include_bytes("... .wasm")`. Maybe it might be better to replace/expand this line with a remark indicating this expected usage to ensure that users of this example have no doubts about whether they are applying it correctly...

murisi: Just rewrote the comments, thanks.

PR #4593, opened 2025-04-21
Forbid targeting MASP payment addresses in IBC transfers

### LINKED ISSUE #4591: Error out if IBC signer is a payment address
Currently, we allow IBC transfers to target `znam` addresses. This allows some leakage of information about the identity of the target.

https://github.com/anoma/namada/blob/733d6b00674bce7926041835f7b20a7c91389187/crates/core/src/address.rs#L409-L425

We should modify the `TryFrom` to error out if the receiver is a `znam1...` address. All MASP IBC shieldings should target the MASP transparent address, instead. The error message should reflect this, suggesting the usage of the MASP address (as opposed to a payment address).

sug0:
## Describe your changes

Closes #4591

Targeting MASP payment addresses leaks some information about the receiver. With the changes in this PR, the MASP transparent address is used as an alternative target for shieldings over IBC.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/core/src/address.rs
+++ b/crates/core/src/address.rs
@@ -410,17 +410,25 @@ impl TryFrom<&Signer> for Address {
     type Error = DecodeError;
 
     fn try_from(signer: &Signer) -> Result<Self> {
-        // The given address should be an address or payment address. When
-        // sending a token from a spending key, it has been already
-        // replaced with the MASP address.
-        Address::decode(signer.as_ref()).or(
-            match crate::masp::PaymentAddress::from_str(signer.as_ref()) {
-                Ok(_) => Ok(MASP),
-                Err(_) => Err(DecodeError::InvalidInnerEncoding(format!(
-                    "Invalid address for IBC transfer: {signer}"
-                ))),
-            },
-        )
+        let signer: &str = signer.as_ref();
+
+        #[cold]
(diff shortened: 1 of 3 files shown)


yito88 (approved):

crates/core/src/address.rs
```diff
@@ -410,17 +410,25 @@ impl TryFrom<&Signer> for Address {
...
-            },
-        )
+        let signer: &str = signer.as_ref();
+
+        #[cold]
+        fn is_znam(signer: &str) -> bool {
+            crate::masp::PaymentAddress::from_str(signer).is_ok()
+        }
+
+        Address::decode(signer).map_err(|err| {
```
grarco, line 420: Should we also adjust/check the `receiver` argument in the sdk that eventually ends up in the `PacketData`? It seems to me these trait implementation is used in the vp but by the time this check runs we have already leaked information

sug0: hm, the receiver of the ibc transfer is a namada shielded account. I think the best we can do is validate the tx before it enters the mempool, which might be a bit overkill. we can't validate the receiver in the packet sent to namada. it's gonna be set from the ui/cli of whatever chain sends the packet.

PR #4619, opened 2025-04-29
Improve shielded fee payment UX

### LINKED ISSUE #3964: Improve MASP fee payment UX
Some suggestions coming from https://github.com/anoma/namada/issues/3961#issuecomment-2447810438 (thanks @Rigorously).

The implementation of masp fee payment in the client (as of #3959) has some flaws:

- The client always require to pass in a `gas-spending-key` when using `disposable-gas-payer`. We could actually imply the usage of the `--source` and allow the second arg to be optionally used when we want to use a different key for the fees
- Using `--gas-payer` when the source of a transfer is the MASP is still allowed but it's unsafe because it leaks information. In this case we should always use a new disposable address and unshield the funds from the pool. We should prevent the usage of that argument in these cases or at least limit it to dev/testing scenarios
- Because of the previous two points we might be able to just remove the `disposable-gas-payer` argument altogether

grarco:
## Describe your changes

Closes #3964.

- the `gas-spending-key` argument is now implied to be the first `source` of the tx when not provided
- removed the `--disposable-gas-payer` argument which is now consider the default (the user can always override it by providing a transparent `--gas-payer` or `--signing-keys`)
- the saving of the disposable gas payer keys in the wallet has been deferred to right before the tx's submission (in case of previous errors during the build process saving the key in the wallet is useless)

## Checklist before merging 
    - Relevant PR if applies: https://github.com/anoma/namada-docs/pull/443
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/cli.rs
+++ b/crates/apps_lib/src/cli.rs
@@ -7755,7 +7755,9 @@ pub mod args {
             )
             .arg(SIGNING_KEYS.def().help(wrap!(
                 "Sign the transaction with the key for the given public key, \
-                 public key hash or alias from your wallet."
+                 public key hash or alias from your wallet. Do not provide \
+                 this argument if the source of the transaction is a shielded \
+                 address as it isn't needed and it would leak information."
             )))
             .arg(SIGNATURES.def().help(wrap!(
                 "List of file paths containing a serialized signature to be \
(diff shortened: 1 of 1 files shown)


sug0 (approved):

brentstone (approved):
Thx, I like these changes

grarco:
@mergifyio requeue

Issue #3255, opened 2024-05-16
Increase precision of locked amounts in shielded token params


sug0:
In the shielded token param struct,
```
pub struct ShieldedParams {
    pub max_reward_rate: Dec,
    pub kd_gain_nom: Dec,
    pub kp_gain_nom: Dec,
    pub locked_amount_target: u64,
}
```
locked amounts are represented as `u64` values. We should increase their precision to 256 bits, by switching to a `Uint` here.

This change would break the genesis files, since `Uint` values are parsed from strings instead of numbers.

cwgoes:
@sug0 Is there a concrete case where we would need to do this?

sug0:
@cwgoes kind of, but it won't be that useful for mainnet. tokens that have ridiculously large denominations won't be able to be stored in a `u64`. on namada mainnet, the only genesis token should be nam, which isn't affected.

PR #4628, opened 2025-05-07
Allow shielded reward precisions and `Uint` target locked amounts in genesis.

murisi:
## Describe your changes
The shielded reward precision parameter was recently introduced into Namada to fix the shortcomings of inferring this value from denominations. This PR allows this parameter to be set in the genesis files in a backwards compatible way by falling back to inferring its value in the absence of the parameter. Additionally this PR also allows genesis to specify target locked amounts greater than what's representable using `u64`s (potentially addressing https://github.com/anoma/namada/issues/3255 ). This is done by also allowing the target to be specified as a string containing a valid amount. This too is backwards compatible because raw unquoted `u64`s are still supported. This PR is not strictly necessary, but may make it easier to configure localnets for testing.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/shielded_token/src/lib.rs
```diff
@@ -42,6 +45,56 @@ pub use storage::*;
...
+    Serialize,
+)]
+// This object can be deserialized from either form
+#[serde(untagged)]
+pub enum Poly<X, Y> {
+    /// The primary way this object can be represented
+    Left(X),
+    /// The secondry way this object can be represented
+    Right(Y),
+}
```
sug0, line 71: this seems unnecessary, the `either` crate is imported into the workspace. perhaps the only problem is not having borsh serialization code

crates/shielded_token/src/lib.rs
```diff
@@ -42,6 +45,56 @@ pub use storage::*;
...
+    }
+
+    /// Convert to the given type without regard for the representaation
+    pub fn into<Z: From<X> + From<Y>>(self) -> Z {
+        match self {
+            Self::Left(x) => x.into(),
+            Self::Right(y) => y.into(),
+        }
+    }
+}
```
sug0, line 96: https://docs.rs/either/latest/either/enum.Either.html#method.either_into

crates/shielded_token/src/lib.rs
```diff
@@ -42,6 +45,56 @@ pub use storage::*;
...
+
+impl<X, Y> Poly<X, Y> {
+    /// Switch the primary and secondry representations
+    pub fn switch(self) -> Poly<Y, X> {
+        match self {
+            Self::Left(x) => Poly::Right(x),
+            Self::Right(y) => Poly::Left(y),
+        }
+    }
```
sug0, line 87: https://docs.rs/either/latest/either/enum.Either.html#method.flip

murisi:
> something's not quite right with integration tests failing

Just checked, the failures happened for two reasons:
* The gas used to execute MASP transactions exceeded the limits which are there in the integration tests. Slightly increasing the gas limit fixes the error, but I'm not sure why this PR would cause increased gas usage even when using WASMs from its ancestor 1e428c9f8a87834ce315ab5c78d4446deb1649cc . When the transaction WASMs are held constant, it seems that the increased gas usage purely comes from the gas meter injection. Moreover, it seems that this PR causes there to be more calls to the gas injector/consumption function, suggesting that some block of code in the WASM is executed more than before (like a loop iterating more or some `if` condition now being satisfied).
* The other cause of the failure is that even for completely unchanged genesis files, this PR causes the generated native token address to change. It seems that changes to the `token::ShieldedParams` structure itself impact address generation. Given that this would probably break new nodes syncing from genesis, I'm closing this PR. But feel free to try and resuscitate it if time allows.

PR #4643, opened 2025-05-19
Fix to receive IBC denom when the base denom has slash

yito88:
## Describe your changes
Namada couldn't receive some tokes including `/` in the base denom, e.g. `factory/stars16da2uus9zrsy83h23ur42v3lglg5rmyrpqnju4/dust` or Eureka token `transfer/08-wasm-1369/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2`.
- Fixed to relax the condition in `is_ibc_denom`
- Remove `/` from the storage key of IBC trace

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/ibc/src/storage.rs
+++ b/crates/ibc/src/storage.rs
@@ -500,7 +500,8 @@ pub fn ibc_trace_key(
     addr: impl AsRef<str>,
     token_hash: impl AsRef<str>,
 ) -> Key {
-    ibc_trace_key_prefix(Some(addr.as_ref().to_string()))
+    // Remove '/' because IBC denom could have '/'
+    ibc_trace_key_prefix(Some(addr.as_ref().replace('/', "")))
         .push(&token_hash.as_ref().to_string().to_db_key())
         .expect("Cannot obtain a storage key")
 }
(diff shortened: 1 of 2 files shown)


crates/ibc/src/storage.rs
```diff
@@ -500,7 +500,8 @@ pub fn ibc_trace_key(
     addr: impl AsRef<str>,
     token_hash: impl AsRef<str>,
 ) -> Key {
-    ibc_trace_key_prefix(Some(addr.as_ref().to_string()))
+    // Remove '/' because IBC denom could have '/'
+    ibc_trace_key_prefix(Some(addr.as_ref().replace('/', "")))
```
sug0, line 504: don't we want to preserve the entire trace? maybe we should replace `/` with some other character that is guaranteed to not appear in ICS-20 trace paths (`\x00`, for instance). when recovering the trace path from storage, we would replace `\x00` with `/` again, perhaps.

sug0: ah but wait, how come `addr` has a slash at all? isn't this the account submitting tokens on the counterparty? for instance, `osmo1...` from osmosis

yito88: Thanks, that's a good question.
The `addr` as a token address was for scanning all of the related tokens, but it isn't used anymore because we need to specify the token when querying balances.
(For example, `namadac balance --token uatom` showed balances of `uatom` related IBC tokens like `transfer/channel-12/uatom` and `transfer/channel-23/uatom` and so on.)
We don't have to store IBC traces for each token now. So, I removed these stores. (Existing traces are left on the current chain, but they're not used.)

sug0 (approved):

grarco (approved):

tzemanovic:
I'll rebase on v101.1.0

PR #4644, opened 2025-05-20
Fee foreign reserve

### LINKED ISSUE #4496: Add "foreign reserve" account and change fee processing logic
I propose three (hopefully simple) changes to non-native fee processing:
1. Add a "foreign reserve" account (owned by the protocol) which can hold non-native tokens.
2. Send non-native token base fees to that account.
3. Burn native token base fees.

For now, the foreign reserve account does not need any ability to _spend_ the funds -- that can be added later.

[Reference discussion on the forums](https://forum.namada.net/t/idea-buy-and-burn-with-non-native-fee-tokens/1519/13?u=cwgoes)

grarco:
## Describe your changes

Closes #4496.

Changes the base fee logic: burns it if the fee token is the native one and sends it to PGF otherwise. For now the base fees for non-native tokens is sent to the PGF internal account. An alternative would be to create a new internal account for this specific purpose, here's a list of possible pros of both solutions:

- PGF account
    - less code required
    - pgf cannot spend non-native tokens, so for now this is safe
- New internal account
    - better logical separation of funds (not really user-related, more for devs), especially useful if we wanted to keep track of the original block proposers from whom these fees are coming

## Checklist before merging 
    - Relevant PR if applies: https://github.com/anoma/namada-docs/pull/445
- [x] If this PR affects services such as `namada-indexer` or `namada-masp-indexer`, a corresponding PR is opened in that repo
    - Relevant PR if applies: https://github.com/anoma/namada-indexer/pull/380

### DIFF
--- a/crates/node/src/protocol.rs
+++ b/crates/node/src/protocol.rs
@@ -469,6 +469,29 @@ pub struct MaspTxResult {
     masp_section_ref: MaspTxRef,
 }
 
+pub(crate) struct ProtocolGasPrice(pub(crate) token::Amount);
+
+/// A trait to handle the available gas prices for a given token, the protocol
+/// one and the optional override coming from a block proposer local
+/// configuration
+pub(crate) trait MinimumGasPrice {
+    /// The gas price, potentially coming from the local configuration of the
+    /// block proposer (only for prepare_proposal)
+    fn price(&self) -> token::Amount;
+    /// The protocol imposed gas price
+    fn protocol_price(&self) -> token::Amount;
+}
+
+impl MinimumGasPrice for ProtocolGasPrice {
+    fn price(&self) -> token::Amount {
+        self.0
+    }
+
+    fn protocol_price(&self) -> token::Amount {
+        self.0
+    }
+}
+
 /// Performs the required operation on a wrapper transaction:
 ///  - replay protection
 ///  - fee payment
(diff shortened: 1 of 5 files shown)


crates/node/src/protocol.rs
```diff
@@ -584,149 +611,169 @@ where
...
+            UserAccount::Internal(namada_sdk::address::PGF),
+            wrapper.fee.token.clone(),
+            fee_components.base.into(),
+            post_bal.into(),
+            pgf_post_balance,
+        )
+    };
+    shell_params.state.write_log_mut().emit_event(
+        TokenEvent {
+            descriptor: FEE_PAYMENT_DESCRIPTOR,
```
sug0, line 768:
```suggestion
            descriptor: FEE_PAYMENT_DESCRIPTOR,
```
should we use a different descriptor here? `wrapper-fee-treasury` or smth, idk

cwgoes:
Can the internal PGF account receive ICS20 transfers? This is part of the requirements for https://github.com/anoma/namada/issues/4496. Even if it should already work, I think it would be good to add a test-case for this.

sug0:
> Can the internal PGF account receive ICS20 transfers? This is part of the requirements for #4496. Even if it should already work, I think it would be good to add a test-case for this.

@cwgoes  the pgf vp is quite permissive with respect to balance changes. it doesn't validate any balance key changes, so anyone is free to send money to the PGF account, whether it's through an ICS-20 transfer, or any other mechanism. the only checks to balances are performed by the multitoken vp, that verify if the changes balance out to 0

crates/node/src/shell/mod.rs
```diff
@@ -1518,49 +1518,79 @@ where
...
+        Error::TxApply(protocol::Error::FeeError(e.to_string()))
+    })?;
+    let fees = token::denom_to_amount(denom_amt, &wrapper.fee.token, storage)?;
+    let minimum_gas_price =
+        parameters::read_gas_cost(storage, &wrapper.fee.token)
+            .expect("Must be able to read gas cost parameter")
+            .ok_or(Error::TxApply(protocol::Error::FeeError(format!(
+                "The provided {} token is not allowed for fee payment",
+                wrapper.fee.token
+            ))))?;
```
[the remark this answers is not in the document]
grarco: The issue is that `fee_data_check` is also called by `prepare_proposal` where the minimum gas price could be the one set locally by the block proposer and not the protocol one. We could pass the minimum gas price to `fee_data_check` as an `Option` and make sure that `prepare_proposal` passes in a `None` (or maybe a `None` only when a custom value overrides the protocol one), but it would make it slightly easier to misuse this function.

Actually, the protocol minimum is red even in prepare proposal to compare it with the custom local value, I'll try to see if there's a clean way to propagate that

crates/node/src/protocol.rs
grarco: We can do that yes, should we also create a new one for when we burn the base fee (native token)? Maybe we can just create two descriptors, one for the tip and one for the base fee. It looks like no external tool is using these events so even changing them shouldn't require modifications anywhere else

sug0: yeah, I think one for tips and another for base fees is enough

grarco: I've split the event descriptor in 93de5ad128. I've also changed the logic to emit the tip event only if a tip is present

brentstone:
BTW, I think we should hold off on merging this until a governance proposal approving these kinds of changes is passed by the mainnet community @grarco @cwgoes

PR #4664, opened 2025-06-02
Add versioning to the shielded wallet

### LINKED ISSUE #4658: Loading the shielded wallet is stupid
Currently, the shielded wallet attempts to find a file and deserialize it, but of which are fallible. Upon failure, a default (empty) wallet is created. This is fine if the file is missing, but causes stupid shit when we've changed the wallet format and deserialize fails. Thus to unshittify the situation, we should do the following

 - If the file is missing, use a default
 - If deserialization fails, try to run a migration (this requires versioning shielded wallets)
 - If deserializing / migration fails, error out explicitly

batconjurer:
## Describe your changes
Closes #4658

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4664-add-versioning-to-shielded-wallet.md
+++ b/.changelog/unreleased/improvements/4664-add-versioning-to-shielded-wallet.md
@@ -1,4 +1,4 @@
-- Currently, the shielded wallet attempts to find a file and deserialize it, but of which are fallible. Upon failure, 
+- Currently, the shielded wallet attempts to find a file and deserialize it, both of which are fallible. Upon failure, 
   a default (empty) wallet is created. This is fine if the file is missing, but causes unexpeceted behavior when we've changed
   the wallet format and deserialize fails. This PR changes the logic to the following:
 


grarco (approved):

crates/node/src/bench_utils.rs
```diff
@@ -803,25 +804,33 @@ impl ShieldedUtils for BenchShieldedUtils {
...
         };
-        let mut ctx_file = File::open(
-            self.context_dir.0.path().to_path_buf().join(file_name),
-        )?;
+        let Ok(mut ctx_file) =
+            File::open(self.context_dir.0.path().to_path_buf().join(file_name))
+        else {
+            // a missing file means there is nothing to load.
+            return Ok(());
+        };
```
sug0, line 812: the exact error you'll get is [`NotFound`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.NotFound)
```suggestion
        let mut ctx_file =
            File::open(self.context_dir.0.path().to_path_buf().join(file_name)).or_else(|err| {
                if err.kind() == std::io::ErrorKind::NotFound {
                    Ok(())
                } else {
                    Err(err)
                }
            })?;
```

crates/shielded_token/src/masp.rs
```diff
@@ -1125,24 +1129,36 @@ pub mod fs {
                     ContextSyncStatus::Speculative => SPECULATIVE_FILE_NAME,
                 }
             };
-            let mut ctx_file = File::open(self.context_dir.join(file_name))?;
+            let Ok(mut ctx_file) = File::open(self.context_dir.join(file_name))
+            else {
+                // a missing file means there is nothing to load.
+                return Ok(());
+            };
```
sug0, line 1136: similar change here

sug0 (changes requested):

sug0 (approved):

PR #4667, opened 2025-06-02
pgf/governance disallow non-native transfers

Fraccaman:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/governance/src/vp/mod.rs
+++ b/crates/governance/src/vp/mod.rs
@@ -1177,7 +1177,7 @@ mod test {
     use assert_matches::assert_matches;
     use namada_core::address::Address;
     use namada_core::address::testing::{
-        established_address_1, established_address_3, nam,
+        btc, established_address_1, established_address_3, nam
     };
     use namada_core::borsh::BorshSerializeExt;
     use namada_core::chain::testing::get_dummy_header;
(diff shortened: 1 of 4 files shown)


crates/governance/src/vp/mod.rs
```diff
@@ -3113,4 +3158,76 @@ mod test {
...
+        );
+
+        assert_matches!(
+            GovernanceVp::validate_tx(
+                &ctx,
+                &batched_tx,
+                &keys_changed,
+                &verifiers
+            ),
+            Err(_)
```
grarco, line 3230: Can we also check the error message here? Just to be sure that we fail for the expected reason

crates/governance/src/vp/pgf.rs
```diff
@@ -207,33 +212,193 @@ where
             },
         )
     }
+
+    /// Validate a pgf balance change
+    pub fn is_valid_balance_change(
+        _ctx: &'ctx CTX,
+        _token: &Address,
+    ) -> Result<()> {
+        Err(Error::new_alloc(
```
grarco, line 221: Same here, could be a `new_const`. Maybe we don't even need this function, could just return the error directly when matching on the key but it's just a personal preference

crates/governance/src/vp/mod.rs
```diff
@@ -3158,4 +3158,76 @@ mod test {
...
+        initialize_account_balance(
+            &mut state,
+            &ADDRESS,
+            token::Amount::native_whole(0),
+            &nam(),
+        );
+        state.commit_block().unwrap();
+
+        let balance_key = balance_key(&btc(), &ADDRESS);
+        let keys_changed = [balance_key].into();
```
sug0, line 3195: pre/post are the same here, no? there were no writes to storage after the block was committed. is the vp rejecting null balance changes?

Fraccaman:
@cwgoes we should always allow credit to pgf and governance addresses right?

brentstone:
> @cwgoes we should always allow credit to pgf and governance addresses right?

Anyone should always be allowed to send tokens TO these addresses, yes. @Fraccaman

sug0 (changes requested):
I think the code might be fine now, but if possible here are some tests that would be nice to have:
1. making sure the pgf/gov balance can decrease/increase with governance proposals (with native or non-native tokens)
2. making sure the pgf/gov balance can increase (with native or non-native tokens)
3. making sure the pgf/gov balance cannot decrease (with native or non-native tokens)

only a subset of these tests is implemented. it should be relatively easy to impl them, should be a copy/paste of some test and changing a few things

brentstone:
@sug0 let's leave the remaining tests for another PR. We need to merge this and cut a release ASAP to at least start testing on Housefire

Fraccaman:
> making sure the pgf/gov balance can decrease/increase with governance proposals (with native or non-native tokens)

We have an e2e test that does this with IBC

> making sure the pgf/gov balance can increase (with native or non-native tokens)

I think i've added `test_pgf_non_native_credit` and `test_governance_non_native_credit` to check if balance can increase with non-native tokens 

> making sure the pgf/gov balance cannot decrease

Same here with `test_governance_non_native_debit` and `test_pgf_non_native_debit`

Native token credit is also tested by e2e test and unit test where we check that a proposal was initialized correctly (the proposal author has to deposit some nam). Maybe I can add a test for native token debit.

grarco (approved):

PR #4671, opened 2025-06-03
Add conversions caching

batconjurer:
## Describe your changes
Conversions are cached when performing Masp balance queries. Futhermore, the decoded asset type cache is used more effectively. Care must be taken to clear the masp conversions cache on load if the masp epoch has changed. If the masp epoch changes during a balance query, it may produce invalid output. However, repeating the query after the epoch has completed will yield the correct result. The new cache resides on the shielded wallet, so a migration has been added.
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -60,6 +60,26 @@ use crate::masp::{
 #[cfg(any(test, feature = "testing"))]
 use crate::masp::{ENV_VAR_MASP_TEST_SEED, testing};
 
+/// Type alias for the respose we get from
```
sug0, line 63:
```suggestion
/// Type alias for the response we get from
```

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -60,6 +60,26 @@ use crate::masp::{
...
+/// querying conversions
+pub type ConversionData = (
+    Address,
+    Denomination,
+    MaspDigitPos,
+    MaspEpoch,
+    I128Sum,
+    MerklePath<Node>,
+);
```
sug0, line 72: why not add a proper struct instead of a type alias for a tuple

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -438,6 +460,31 @@ pub trait ShieldedQueries<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
+        asset_type: AssetType,
+        cache: &mut EpochedConversions,
+    ) -> Option<(
+        Address,
+        Denomination,
+        MaspDigitPos,
+        MaspEpoch,
+        I128Sum,
+        MerklePath<Node>,
+    )> {
```
sug0, line 477: `Option<ConversionData>`

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -462,6 +509,24 @@ pub trait ShieldedQueries<U: ShieldedUtils + MaybeSend + MaybeSync>:
 pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
     ShieldedQueries<U>
 {
+    /// A method that loads the wallet and empties the conversion cache if the
+    /// masp epoch has advanced. This method is only called by functions
+    /// which are relatively fast so the chances of crossing a masp boundary
+    /// are minimal. However, if it happens, the function may produce
+    /// invalid output and need to be run again.
+    #[allow(async_fn_in_trait)]
+    async fn load_with_caching<C: Client + Sync>(
```
sug0, line 518: if you key conversions by address and masp epoch, I think you don't need to worry about clearing the cache. it will grow out of control unless it is trimmed occasionally, though. the trimming could be done when saving or smth

crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -472,6 +537,10 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
     ) -> Result<(), eyre::Error> {
         // To facilitate lookups of human-readable token names
         for token in tokens {
+            // check if token is already in the cache
+            if self.asset_types.values().any(|ass| ass.token == *token) {
```
sug0, line 541: 🫏 ?

PR #4685, opened 2025-06-05
Meter wasm gas using mutable global

sug0:
## Describe your changes

Supersedes #4647

The likelihood of this being a consensus breaking upgrade is nearly null, but it's still safer to upgrade to a software release containing this PR in a hardfork.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4685-faster-gas.md
+++ b/.changelog/unreleased/improvements/4685-faster-gas.md
@@ -1,0 +1,2 @@
+- Perform wasm gas accounting using mutable global, to avoid frequent host/guest
+  vm context switches. ([\#4685](https://github.com/anoma/namada/pull/4685))


sug0:
```
failures:
    wasm::run::tests::test_tx_stack_limiter
    wasm::run::tests::test_vp_stack_limiter
```
the stack limiter tests are now failing... hmmm. the changes we've made to wasm affect the stack limiter code that's been injected, after all, so these are probably consensus breaking changes 

**EDIT:** fixed the test by forking [`wasm-instrument`](https://github.com/anoma/nam-wasm-instrument), changing the behavior of the stack limiter instrumenting code to allow specifying a list of functions exempt of instrumentation

crates/gas/src/lib.rs
```diff
@@ -373,56 +411,64 @@ pub trait GasMetering {
...
+                )
+                .ok_or(Error::GasOverflow)?
+                .into(),
+        )
     }
 }
 
 /// Gas metering in a transaction
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
grarco, line 449: I don't think it is safe to provide a default implementation for the gas meters given that the `gas_scale` is a protocol parameter

crates/vp/src/native_vp.rs
```diff
@@ -68,6 +68,8 @@ where
     pub vp_wasm_cache: CA,
     /// VP evaluator type
     pub eval: PhantomData<EVAL>,
+    /// WASM instructions gas meter kind
+    pub gas_meter_kind: GasMeterKind,
```
grarco, line 72: Do we need this? It looks like the `new` function is always called with `GasMeterKind::MutGlobal`

crates/vm/src/wasm/host_env.rs
```diff
@@ -3,13 +3,208 @@
...
+
+        let current_tx_gas = self.read_wasm_gas(None);
+
+        if current_tx_gas == u64::MAX.into() {
+            return Err(namada_gas::Error::TransactionGasExceededError(
+                self.tx_gas_limit.get_whole_gas_units(self.gas_scale),
+            ));
+        }
+
+        self.initial_gas
```
grarco, line 74: Doesn't this return the gas left to be used instead of the gas used so far? Shouldn't the latter just be `current_tx_gas`

grarco: Ah ok it looks like the wasm global is tracking the gas left, not the one used so far, so you need to run this subtraction

crates/vm/src/wasm/compilation_cache/common.rs
```diff
@@ -53,8 +54,15 @@ pub trait CacheName: Clone + std::fmt::Debug {
     fn name() -> &'static str;
 }
 
+/// WASM cache key consists of the code hash and gas meter kind.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct CacheKey {
+    code_hash: Hash,
+    gas_meter_kind: GasMeterKind,
```
grarco, line 61: Aren't wasm modules expected to be always compiled with `GasMeterKind::MutGlobal`? Or do we use the native meter at times? It looks like even for governance proposals we use the global variable

crates/vp/src/native_vp.rs
sug0: there is a case where we use the other gas meter kind (host fn), when we call the eval vp host fn

crates/vm/src/wasm/compilation_cache/common.rs
sug0: the only use so far is in the eval vp host fn, I think. originally I discussed with tomas using it for dry runs, as well, but I think ultimately we decided against it. the reasoning behind keeping the host fn gas meter was to allow running arbitrary untrusted wasms, which could mess with the gas global value, to achieve infinite execution times

crates/tests/src/vm_host_env/tx.rs
```diff
@@ -226,17 +231,30 @@ impl TestTxEnv {
 
     /// Apply the tx changes to the write log.
     pub fn execute_tx(&mut self) -> Result<(), Error> {
-        wasm::run::tx(
+        let gas_meter = RefCell::new(
+            if let GasMeter::Native(meter) = &mut *self.gas_meter.borrow_mut() {
+                std::mem::take(meter)
+            } else {
+                unreachable!()
```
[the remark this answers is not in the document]
sug0: hm, without adding a new generic parameter to the wasm context, this was the easiest solution to fix these tests

PR #4691, opened 2025-06-09
Wasm code events

batconjurer:
## Describe your changes
For each inner tx, an event is emitted giving the human readable wasm name of the code contained in said tx.
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/node/src/shell/finalize_block.rs
```diff
@@ -470,6 +476,41 @@ where
...
+            .map(|c| c.code.hash())
+        {
+            None => "none".to_string(),
+            Some(hash) => self
+                .state
+                .read(&Key::wasm_code_name(&hash))
+                .ok()
+                .flatten()
+                .unwrap_or("unknown".to_string()),
+        };
```
sug0, line 504: is it better to give up emitting an event in case of `none` or `unknown`? essentially, early return from this fn instead

crates/tx/src/event.rs
```diff
@@ -204,6 +206,33 @@ impl From<MaspEvent> for Event {
...
+impl From<TxWasmEvent> for Event {
+    fn from(tx_wasm_event: TxWasmEvent) -> Self {
+        let composite =
+            Self::new(EventType::new("tx-wasm-name"), EventLevel::Tx)
+                .with(CodeName(tx_wasm_event.name))
+                .with(InnerTxHash(tx_wasm_event.inner_tx_hash));
+        match tx_wasm_event.hash {
+            None => composite.into(),
+            Some(hash) => composite.with(TxHash(hash)).into(),
+        }
```
sug0, line 232: use `.unwrap_or_default()` to get a zero hash of the wrapper, if it doesn't exist

crates/node/src/shell/finalize_block.rs
batconjurer: I'm not sure, is it better if the indexer can't find an event for a tx or get one that says it's not clear how to deserialize it?

sug0: hmm, either way it should be easy to handle in the indexer codebase. I have a slight preference either for not emitting an event at all, or replacing the string with the empty string, to signal absence

sug0 (approved):

grarco (approved):

PR #4704, opened 2025-06-19
Optimize RocksDB configuration for improved performance

aso20455:
## Describe your changes

I improved how our database works by fine-tuning RocksDB settings. The changes make searches faster with bloom filters, save disk space with smart compression, and reduce memory usage with better caching. Each database section (column family) now has settings matched to how it's actually used - some optimized for reading, others for writing. These improvements should make the chain run smoother without requiring any code changes.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/node/src/storage/rocksdb.rs
+++ b/crates/node/src/storage/rocksdb.rs
@@ -157,6 +157,23 @@ pub fn open(
 
     db_opts.set_bytes_per_sync(1048576);
     set_max_open_files(&mut db_opts);
+    
+    // Memory usage and performance optimization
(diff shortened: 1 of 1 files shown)


yito88:
@aso20455 Thank you for the proposal! Do you have the result of the optimization?
Currently, I don't believe that RocksDB read/write is the bottleneck for each commit. Additionally, performance varies by node. I mean that not all operators use the same machine, and someone might prioritize the throughput/latency over the memory usage.
However, general optimization should be considered for the future. I'm really interested in your measurement result.

PR #4712, opened 2025-06-25
Make CometBFT `max_tx_bytes` configurable

yito88:
## Describe your changes
Currently, the `max_tx_bytes` is hardcoded, and we couldn't make a proposal with a wasm code over 1MB.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/miscellaneous/4712-comet-max-tx-size.md
+++ b/.changelog/unreleased/miscellaneous/4712-comet-max-tx-size.md
@@ -1,0 +1,2 @@
+- Make CometBFT max_tx_bytes configurable
+  ([\#4712](https://github.com/anoma/namada/issues/4712))


crates/node/src/tendermint_node.rs
```diff
@@ -281,17 +281,6 @@ async fn update_tendermint_config(
...
 
-        // Drop txs from the mempool that are larger than 1 MiB
-        //
-        // The application (Namada) can assign arbitrary max tx sizes,
-        // which are subject to consensus. Either way, nodes are able to
-        // configure their local mempool however they please.
-        //
-        // 1 MiB is a reasonable value that allows governance proposal txs
-        // containing wasm code to be proposed by a leading validator
-        // during some round's start
```
sug0, line 292: Could you please keep these docs? I think they're useful for future changes. Simply delete the line that hardcodes `MaxTxBytes` to 1 MiB, and instead point to the CometBFT code that sets the default parameter to 1 MiB.

https://github.com/cometbft/cometbft/blob/c653eee08d7f2c44bd87fcca0b8a16666cd08fee/config/config.go#L1016

sug0 (changes requested):

Issue #4721, opened 2025-07-01
Write IBC ack error on failures


sug0:
## Problem statement

Right now, if an IBC transaction fails on Namada, especially due to VP errors, we do not emit any acknowledgement events to inform the counterparty. This leads to the funds only being recoverable on the source chain by waiting for the transfer to timeout, which in some cases (e.g. with Packet Forward Middleware retries) can take 10 minutes.

To considerably improve the UX, transfers (ideally) should fail immediately with an ack error being written to storage, accompanied by an event.

### Define transaction epilogue

A generic solution could involve a tx epilogue, where additional logic can be executed at the end of VP execution.

yito88:
Essentially, this issue is that a VP execution can't submit an event in the current transaction execution model. We can't add other changes, including events, since any change is handled in the transaction execution, and event submission should also be checked by the VP.

Some errors can be caught in the execution by validating the IBC transfer in the transaction execution because ibc-rs can call the validation functions (just return Ok now) in the execution function. However, it can't resolve this issue when other VPs, such as the Multitoken VP and MASP VP, reject the transaction without incorporating the same validations of these VPs into their validation functions.

https://github.com/informalsystems/ibc-rs/blob/be82d123448a59ccea305c9e918f07aca5ec1a6f/ibc-apps/ics20-transfer/src/handler/on_recv_packet.rs#L54

https://github.com/informalsystems/ibc-rs/blob/be82d123448a59ccea305c9e918f07aca5ec1a6f/ibc-apps/ics20-transfer/src/handler/on_recv_packet.rs#L102

https://github.com/namada-net/namada/blob/a71fb62cef2f9dedd7e1de7d15e5144a4e93cdb1/crates/ibc/src/context/token_transfer.rs#L220

https://github.com/namada-net/namada/blob/a71fb62cef2f9dedd7e1de7d15e5144a4e93cdb1/crates/ibc/src/context/token_transfer.rs#L231

PR #4722, opened 2025-07-01
Allow bech32 encoding of MASP address in shielded receive middleware

sug0:
## Describe your changes

This fix is required by #4723.

Rather than encoding the MASP address to `bech32m` to compare it with the ICS-20 receiver, we parse a transparent address from the ICS-20 receiver string (which can be either `bech32`/`bech32m`), and compare it against the MASP address.

This PR also fixes a minor bug where the overflow receiver address wouldn't be added as a verifier, causing the Multitoken VP to reject txs in case of ack errors, which would result in timeouts (the latter which are a symptom of #4721).

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4722-validate-masp-shielded-swaps.md
+++ b/.changelog/unreleased/improvements/4722-validate-masp-shielded-swaps.md
@@ -1,0 +1,2 @@
+- Allow bech32 encoding of MASP address in shielded receive middleware.
+  ([\#4722](https://github.com/anoma/namada/pull/4722))


sug0:
gonna add a wee test to this PR, expect a f-push

brentstone:
Can this be merged? How critical is it?

sug0:
@brentstone it should be in the next release. it fixes shielded swaps with stride (as the output denom)

PR #4725, opened 2025-07-07
Fix masp height overflow in SDK

grarco:
## Describe your changes

Fixes a bug for which the desired expiration height of a masp transaction could overflow during the building process

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/shielded_token/src/masp/shielded_wallet.rs
+++ b/crates/shielded_token/src/masp/shielded_wallet.rs
@@ -1156,10 +1156,13 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
             Some(expiration) => {
                 // Try to match a DateTime expiration with a plausible
                 // corresponding block height
-                let last_block_height = Self::query_block(context.client())
-                    .await
-                    .map_err(|e| TransferErr::General(e.to_string()))?
-                    .unwrap_or(1);
+                let last_block_height = u32::try_from(
+                    Self::query_block(context.client())
+                        .await
+                        .map_err(|e| TransferErr::General(e.to_string()))?
+                        .unwrap_or(1),
+                )
+                .map_err(|e| TransferErr::General(e.to_string()))?;
                 let max_block_time =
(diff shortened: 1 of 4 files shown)


crates/shielded_token/src/masp/shielded_wallet.rs
```diff
@@ -1156,10 +1156,13 @@ pub trait ShieldedApi<U: ShieldedUtils + MaybeSend + MaybeSync>:
...
                 // corresponding block height
-                let last_block_height = Self::query_block(context.client())
-                    .await
-                    .map_err(|e| TransferErr::General(e.to_string()))?
-                    .unwrap_or(1);
+                let last_block_height = u32::try_from(
+                    Self::query_block(context.client())
+                        .await
+                        .map_err(|e| TransferErr::General(e.to_string()))?
+                        .unwrap_or(1),
```
sug0, line 1163: this was already an issue before, but I'll comment on it anyway: the rpc error might not have been a missing block, is it fine to proceed with 1 as the block height in those cases? guess the tx will fail be be applied, since it'll reach the expiration height...

sug0 (commented):
just a wee comment

crates/shielded_token/src/masp/shielded_wallet.rs
grarco: Yes it will fail because of a very early expiration. Still even if this is safe I don't think it's what we should do: I'll just propagate the error instead, thanks for pointing this out!

grarco: We are actually propagating the error and unwrapping_or on the inner `Option` so this should be safe (as in, only return `None` when no block has been produced yet). Actually the rpc query gets the last block from the `InMemory` struct. This is an `Option` but we wrap it inside an `Ok` all the times without anything being actually fallible. We might actually want to adjust that instead...

brentstone:
Can this be merged?

grarco:
> Can this be merged?

Yes It's already scheduled for the upcoming release

PR #4726, opened 2025-07-08
Adding a shielding fee for normal shielding txs. Still need to check IBC

### LINKED ISSUE #4716: Add fees for shielding to the MASP
When shielding to the MASP (i.e. zero shielded inputs), an additional flat fee should be paid. In the first implementation, this will fee will be sent to PGF. The fee amount will be a protocol parameter.

The current task list consists of:
- [ ] Add fee data to the MASP tx. This can just be an authorizing signature. In theory, anyone should be able to pay this fee. This must be hardened against replay attempts however.
- [ ] Change the tx generation functions and cli to allow signing over this fee.
- [ ] Change the tx generation functions and cli to check that the user can pay the fee before broadcasting.
- [ ] Probably need to change mempool checks. 
- [ ] Change the user vp so that it correctly authorizes the fee payment. I think this can be implemented by adding another `MaspAction::MaspAuthorizer` action to be fed into the vp.
- [ ] Add a parameter for the fee. This should be changeable via governance.
- [ ] Change the MASP vp to enfore the payment of a fee when shielding
- [ ] The protocol needs to be changed to actually affect the fee payment.

batconjurer:
## Describe your changes
Closes #4716 
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/cli.rs
+++ b/crates/apps_lib/src/cli.rs
@@ -500,6 +500,7 @@ pub mod cmds {
     }
 
     #[derive(Clone, Debug)]
+    #[allow(clippy::large_enum_variant)]
     pub enum NamadaClientWithContext {
         // Ledger cmds
         TxCustom(TxCustom),
@@ -3672,6 +3673,10 @@ pub mod args {
         arg("self-bond-amount");
     pub const SENDER: Arg<String> = arg("sender");
     pub const SHIELDED: ArgFlag = flag("shielded");
+    pub const SHIELDING_FEE_PAYER: Arg<WalletPublicKey> =
+        arg("shielding-fee-payer");
+    pub const SHIELDING_FEE_TOKEN: Arg<WalletAddress> =
+        arg("shielding-fee-token");
     pub const SHOW_IBC_TOKENS: ArgFlag = flag("show-ibc-tokens");
     pub const SLIPPAGE: ArgOpt<f64> = arg_opt("slippage-percentage");
     pub const SIGNING_KEYS: ArgMulti<WalletPublicKey, GlobStar> =
@@ -4946,12 +4951,15 @@ pub mod args {
                     amount: transfer_data.amount,
                 });
             }
+            let shielding_fee_payer = chain_ctx.get(&self.shielding_fee_payer);
 
             Ok(TxShieldingTransfer::<SdkTypes> {
                 tx,
                 sources: data,
                 targets,
                 tx_code_path: self.tx_code_path.to_path_buf(),
+                shielding_fee_payer,
+                shielding_fee_token: chain_ctx.get(&self.shielding_fee_token),
             })
         }
     }
@@ -4964,6 +4972,8 @@ pub mod args {
             let token = TOKEN.parse(matches);
             let amount = InputAmount::Unvalidated(AMOUNT.parse(matches));
             let tx_code_path = PathBuf::from(TX_TRANSFER_WASM);
+            let shielding_fee_payer = SHIELDING_FEE_PAYER.parse(matches);
+            let shielding_fee_token = SHIELDING_FEE_TOKEN.parse(matches);
             let data = vec![TxTransparentSource {
                 source,
                 token: token.clone(),
@@ -4980,6 +4990,8 @@ pub mod args {
                 sources: data,
                 targets,
                 tx_code_path,
+                shielding_fee_payer,
+                shielding_fee_token,
             }
         }
 
@@ -5000,6 +5012,13 @@ pub mod args {
                         .def()
                         .help(wrap!("The amount to transfer in decimal.")),
                 )
+                .arg(SHIELDING_FEE_PAYER.def().help(wrap!(
+                    "The implicit account which will pay the fee for \
+                     shielding tokens"
+                )))
+                .arg(SHIELDING_FEE_TOKEN.def().help(wrap!(
+                    "The token which will be used to pay the shielding fee"
+                )))
         }
     }
 
@@ -7243,6 +7262,8 @@ pub mod args {
                         IbcShieldingTransferAsset::Address(chain_ctx.get(&addr))
                     }
                 },
+                shielding_fee_payer: chain_ctx.get(&self.shielding_fee_payer),
+                shielding_fee_token: chain_ctx.get(&self.shielding_fee_token),
             })
         }
     }
@@ -7266,6 +7287,8 @@ pub mod args {
                     None => TxExpiration::Default,
                 }
             };
+            let shielding_fee_payer = SHIELDING_FEE_PAYER.parse(matches);
+            let shielding_fee_token = SHIELDING_FEE_TOKEN.parse(matches);
 
             Self {
                 query,
@@ -7278,6 +7301,8 @@ pub mod args {
                     channel_id,
                     token,
                 },
+                shielding_fee_payer,
+                shielding_fee_token,
             }
         }
 
@@ -7319,6 +7344,14 @@ pub mod args {
                 .arg(CHANNEL_ID.def().help(wrap!(
                     "The channel ID via which the token is received."
                 )))
+                .arg(SHIELDING_FEE_PAYER.def().help(wrap!(
+                    "The implicit account paying the fee for shielding tokens"
+                )))
+                .arg(
+                    SHIELDING_FEE_TOKEN
+                        .def()
+                        .help(wrap!("The token used to pay the shielding fee")),
+                )
         }
     }
 
--- a/crates/sdk/src/rpc.rs
+++ b/crates/sdk/src/rpc.rs
@@ -264,6 +264,18 @@ pub async fn get_staking_rewards_rate<C: Client + Sync>(
     )
 }
 
+/// Query the effective total supply of the native token
+pub async fn get_shielding_fee_amount<C: Client + Sync>(
+    client: &C,
+    token: &Address,
+) -> Result<DenominatedAmount, error::Error> {
+    query_storage_value(
+        client,
+        &namada_token::storage_key::masp_shielding_fee_amount(token),
+    )
+    .await
+}
+
 /// Check if the given address is a known validator.
 pub async fn is_validator<C: namada_io::Client + Sync>(
     client: &C,
--- a/crates/shielded_token/src/vp.rs
+++ b/crates/shielded_token/src/vp.rs
@@ -625,8 +625,13 @@ where
             }
         }
         // The transaction shall not push masp authorizer actions that are not
-        // needed cause this might lead vps to run a wrong validation logic
-        if !actions_authorizers.is_empty() {
+        // needed because this might lead vps to run a wrong validation logic
+        if !actions_authorizers.is_empty()
+            && shielded_tx
+                .transparent_bundle()
+                .map(|b| b.vin.is_empty())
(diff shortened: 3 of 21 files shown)


crates/apps_lib/src/client/rpc.rs
```diff
@@ -874,6 +882,7 @@ pub async fn query_protocol_parameters(
         epochs_per_year,
         masp_epoch_multiplier,
         masp_fee_payment_gas_limit,
+        masp_nam_shielding_fee: masp_nam_shielding_fee.to_string_precise(),
```
grarco, line 885: Would it be possible to also return the price for the other supported tokens?

crates/apps_lib/src/config/genesis/templates.rs
```diff
@@ -290,6 +292,8 @@ pub struct ChainParams<T: TemplateValidation> {
     pub max_block_gas: u64,
     /// Gas limit of a masp transaction paying fees
     pub masp_fee_payment_gas_limit: u64,
+    /// The amount of NAM the MASP shielding fee costs
+    pub masp_nam_shielding_fee: String,
```
grarco, line 296: Don't we need a map here to track the costs for the non-native tokens?

crates/apps_lib/src/client/rpc.rs
batconjurer: In theory, but at genesis I think only supporting nam is fine

crates/apps_lib/src/config/genesis/templates.rs
batconjurer: Since, we're arent going to support non-native tokens at genesis, I'm resolving this comment

crates/apps_lib/src/cli.rs
```diff
@@ -5000,6 +5016,13 @@ pub mod args {
                         .def()
                         .help(wrap!("The amount to transfer in decimal.")),
                 )
+                .arg(SHIELDING_FEE_PAYER.def().help(wrap!(
+                    "The implicit account which will pay the fee for \
+                     shielding tokens"
+                )))
```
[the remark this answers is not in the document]
grarco: It makes no sense to pay fees via the MASP when shielding since the source of the transfer is public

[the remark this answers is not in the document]
sug0: ok, good point

crates/apps_lib/src/client/rpc.rs
```diff
@@ -843,6 +843,14 @@ pub async fn query_protocol_parameters(
...
             .await
             .expect("Parameter should be defined.");
+    let native_token = rpc::query_native_token(context.client())
+        .await
+        .expect("Native token should be defined");
+    let key = param_storage::masp_shielding_fee_amount(&native_token);
+    let masp_nam_shielding_fee: DenominatedAmount =
+        query_storage_value(context.client(), &key)
+            .await
+            .expect("Parameter should be defined.");
```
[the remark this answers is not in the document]
batconjurer: at genesis, only nam is supported for paying shielded fees. Others can be added post-genesis

crates/core/src/parameters.rs
```diff
@@ -52,6 +52,8 @@ pub struct Parameters {
     pub masp_epoch_multiplier: u64,
     /// The gas limit for a masp transaction paying fees
     pub masp_fee_payment_gas_limit: u64,
+    /// The amount of NAM the MASP shielding fee costs
+    pub masp_nam_shielding_fee: String,
```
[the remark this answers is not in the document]
batconjurer: This will have to change to Dec in a later pr since its supposed to be a percent of the amount. The best way to handle the decimal places is with strings IMO.

crates/sdk/src/rpc.rs
```diff
@@ -264,6 +264,18 @@ pub async fn get_staking_rewards_rate<C: Client + Sync>(
     )
 }
 
+/// Query the effective total supply of the native token
+pub async fn get_shielding_fee_amount<C: Client + Sync>(
+    client: &C,
+    token: &Address,
+) -> Result<DenominatedAmount, error::Error> {
```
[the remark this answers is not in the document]
batconjurer: ha, yep

crates/node/src/bench_utils.rs
```diff
@@ -1308,7 +1317,11 @@ impl BenchShieldedCtx {
                     )
                     .unwrap(),
                 Some(shielded),
-                None,
+                Some(vec![Section::ShieldingFee {
+                    payer: shielding_fee_key,
+                    token: native_token,
+                    cmt: shielded_section_hash,
+                }]),
```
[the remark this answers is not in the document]
batconjurer: I feel like this is more proper, but we can revisit it later if need be.

crates/token/src/tx.rs
```diff
@@ -207,10 +209,68 @@ where
...
+            ));
+        }
+    }
 
     Ok(())
 }
 
+/// The key for getting the shielding fee amount of the provided
+/// token.
+fn masp_shielding_fee_amount(token: &Address) -> Key {
```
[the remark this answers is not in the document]
batconjurer: Yeah, so this was annoying. I tried to define it in one place, but everywhere sensible I put I got into trouble with the unit tests that enforces no cross deps. I'm open to suggestions for a canonical place to put it

crates/shielded_token/src/vp.rs
```diff
@@ -625,8 +625,13 @@ where
...
         // The transaction shall not push masp authorizer actions that are not
-        // needed cause this might lead vps to run a wrong validation logic
-        if !actions_authorizers.is_empty() {
+        // needed because this might lead vps to run a wrong validation logic
+        if !actions_authorizers.is_empty()
+            && shielded_tx
+                .transparent_bundle()
+                .map(|b| b.vin.is_empty())
+                .unwrap_or(true)
+        {
```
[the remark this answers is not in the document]
batconjurer: Yes, that should be added. Not relevant to the highlighted code block, but I'll add it.

crates/tests/src/integration/masp.rs
```diff
@@ -1451,6 +1482,540 @@ fn enable_rewards_after_shielding() -> Result<()> {
...
+/// This tests that shielding fees are enforced. We test the following
+/// cases
+///
+/// 1. Test the happy flow.
+/// 2. Test that if the shielding fee section is missing from the tx, it is
+///    rejected
+/// 3. Test that if the shielding fee holds the wrong Masp Tx id, it is
+///    rejected.
+/// 4. Test that if the account listed in the shielding fee does not sign the
+///    tx, it is rejected
```
grarco, line 1494: Can we also test that if the shielding fee section is duplicated the protocol still gets the fee payment only once? I believe an attacker (or a malicious wrapper signer) could duplicate the fee section and rewrap the tx (or before wrapping it) to cause damage to the inner signer. I believe this shouldn't work given your logic but it would be nice to have a test covering that case

crates/shielded_token/src/vp.rs
```diff
@@ -624,15 +624,33 @@ where
...
+        let masp_tx_id = shielded_tx.txid().into();
+        let has_trans_inputs = shielded_tx
+            .transparent_bundle()
+            .map(|b| !b.vin.is_empty())
+            .unwrap_or(false);
         // The transaction shall not push masp authorizer actions that are not
-        // needed cause this might lead vps to run a wrong validation logic
-        if !actions_authorizers.is_empty() {
+        // needed because this might lead vps to run a wrong validation logic
+        if !actions_authorizers.is_empty() && !has_trans_inputs {
```
grarco, line 634: Here we might have an issue: if the tx has transparent inputs then we might miss the fact that some of the required authorizers were not actually needed

crates/shielded_token/src/vp.rs
```diff
@@ -624,15 +624,33 @@ where
...
             let error = Error::new_const(
                 "Found masp authorizer actions that are not required",
             );
             tracing::debug!("{error}");
             return Err(error);
         }
+        // check that the tx has a shielding fee section
```
grarco, line 641: Don't we also need to check that the right amount of fees has been paid with a supported token and to the expected recipient?

crates/tx/src/section.rs
```diff
@@ -61,6 +61,15 @@ pub enum Section {
     MaspBuilder(MaspBuilder),
     /// Wrap a header with a section for the purposes of computing hashes
     Header(Header),
+    /// The account paying the fee for a shielding transaction
+    ShieldingFee {
```
grarco, line 65: Quick idea I haven't reasoned too deeply about (especially in terms of safety): would it be possible to avoid this new section type and encode all the necessary data directly inside the `Transfer` data section? Maybe we could just add two entires: the fee payer as  a source and the PGF account as the target. The `multitransfer` function will just update the balances accordingly. The masp vp should have all the necessary data to check if the PGF account has been credited the fees when the vector of inputs of the transparent bundle is non-empty. It should also be able to verify that the amount credited to the PGF account is in a supported token and matches the expected amount for that token

batconjurer: Tiago was proposing the same thing. It would indeed work to put this as an extra data section. It seems really hacky to me, but we can discuss at standup.

batconjurer: The difficult part would be knowing which masp tx was paid for it multiple shields are in a batch. We need a commitment.

crates/tx/src/section.rs
```diff
@@ -61,6 +61,15 @@ pub enum Section {
...
+        /// The key for the implicit account
+        payer: common::PublicKey,
+        /// The address of the token paying the fee
+        token: Address,
+        /// The inner tx this fee is for
+        cmt: MaspTxId,
+    },
```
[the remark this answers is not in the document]
batconjurer: I don't thing extra data is signed. The comments say that this is never sent to the hardware wallet. This is fine for memos but not for this.

crates/shielded_token/src/vp.rs
batconjurer: I think you're right. I will try to fix the above block. I

crates/token/src/tx.rs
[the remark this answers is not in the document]
grarco: Ah right, ok I think it's ok to have multiple implementations

crates/shielded_token/src/vp.rs
```diff
@@ -624,20 +624,31 @@ where
                 return Err(error);
             }
         }
+        let masp_tx_id = shielded_tx.txid().into();
+        let has_trans_inputs = shielded_tx
+            .transparent_bundle()
+            .map(|b| !b.vin.empty())
+            .unwrap_or(false);
```
sug0: nit: `.map(|...| <bool>).unwrap_or(false)` is a pattern that can be replaced by the more readable fn `.is_some_and(|...| <bool>)`

sug0 (approved):
overall this is correct, in terms of impl alone. we should check in with @cwgoes that this meets the desired design principles. also, should pitch in some opinion from @murisi about the hw wallet changes required

grarco (approved):

brentstone:
Let's get some new reviews here

brentstone:
Just want to note that the fee must only be taken if the associated tx is successful @sug0

PR #4729, opened 2025-07-11
fix wasm host-in-guest alloc

tzemanovic:
## Describe your changes

This fixes an issue with tx and VP WASM inputs that are larger than the initial WASM memory and hence require to grow memory. It turned out that simply calling `memory.grow` for this (as performed by `memory::write_memory_bytes` from host didn't update the guest properly which was then overriding the already used memory ranges from the initial data and hence corrupting the memory, leading to WASM runtime memory errors.

To fix this, we're now injecting and exporting a simple alloc fn into WASMs and calling it from host to write the initial data.

Additionally, the max WASM memory size is increased as with the current limit it was not possible to decode a tx with 2 MiB  of data (the newly added `wasm::run::tests::test_tx_alloc` test would fail with unreachable when it runs out of memory).

For a tx with data that fits in the initial memory size there is no change in WASM execution (the alloc call will return early).

The Rust src for the alloc fn that's being injected is here https://github.com/namada-net/namada-wasm-min-alloc

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/bug-fixes/4729-fix-wasm-alloc.md
+++ b/.changelog/unreleased/bug-fixes/4729-fix-wasm-alloc.md
@@ -1,0 +1,2 @@
+- Allow to execute WASM with a tx payload that doesn't fit into the initial WASM
+  memory. ([\#4729](https://github.com/anoma/namada/pull/4729))


yito88 (approved):

sug0:
quick review of `namada-wasm-min-alloc`:
https://github.com/namada-net/namada-wasm-min-alloc/blob/5e654ca43c93bf74dfa5ce6628fae35e82dbb3b4/src/lib.rs#L43-L46
shouldn't the panic handler simply be `wasm32::unreachable()`?

crates/vm/src/wasm/memory.rs
```diff
@@ -192,6 +199,26 @@ pub fn write_vp_inputs(
...
+fn wasm_alloc(
+    instance: &wasmer::Instance,
+    store: &mut impl wasmer::AsStoreMut,
+    len: u64,
+) -> Result<u64> {
+    let alloc_fn = instance
+        .exports
+        .get_function(super::run::ALLOC_FN_NAME)
+        .unwrap()
+        .typed::<i32, i32>(&*store)
```
sug0, line 213: it doesn't work with `u32` instead of `i32`? not that we'll have the need to allocate 4 GiB, but...

crates/vm/src/wasm/memory.rs
```diff
@@ -192,6 +199,26 @@ pub fn write_vp_inputs(
...
+        .unwrap();
+
+    let len: i32 = len.try_into().map_err(Error::TryFromInt)?;
```
sug0, line 216: perhaps we can convert len to a `u32`, then cast (using `as`) from `u32` to `i32`, upon calling the guest fn, in order to retain the bit repr. the wasm fn is interpreting the bits as unsigned, so it should be safe

tzemanovic:
> quick review of `namada-wasm-min-alloc`: https://github.com/namada-net/namada-wasm-min-alloc/blob/5e654ca43c93bf74dfa5ce6628fae35e82dbb3b4/src/lib.rs#L43-L46 shouldn't the panic handler simply be `wasm32::unreachable()`?

it doesn't matter, it doesn't affect the generated code as it's panic-free

PR #4776, opened 2025-07-25
The MASP conversions query now requires a MASP epoch to be specified.

murisi:
## Describe your changes
Querying the conversion state becomes more difficult as nodes progress to higher and higher epochs. For instance, on a mainnet clone running on accelerated epochs, client queries for conversions always timeout (even after modifying `timeout_broadcast_tx_commit`) when the node goes beyond a certain MASP epoch. This happens because the conversion state grows linearly with the MASP epoch counter. This PR attempts to address this problem by making the conversions RPC endpoint require that clients specify the MASP epoch they want conversions from. This implies two things: first the size of the response from the conversions RPC endpoint should now be constant (equal to the number of tokens in the shielded rewards program), and second a client requiring all conversions now has to do a separate query for each MASP epoch.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -2143,9 +2143,20 @@ pub async fn query_conversions(
         .wallet()
         .await
         .get_addresses_with_vp_type(AddressVpType::Token);
-    let conversions = rpc::query_conversions(context.client())
+
+    // Download conversions from all epochs to facilitate decoding asset types
+    let mut conversions = BTreeMap::new();
+    let from = MaspEpoch::zero();
+    let to = rpc::query_masp_epoch(context.client())
         .await
-        .expect("Conversions should be defined");
+        .expect("Unable to query current MASP epoch");
+    for epoch in MaspEpoch::iter_bounds_inclusive(from, to) {
+        conversions.append(
+            &mut rpc::query_conversions(context.client(), &epoch)
+                .await
+                .expect("Conversions should be defined"),
+        );
+    }
 
     if args.dump_tree {
         display_line!(context.io(), "Conversions: {conversions:?}");
@@ -2153,6 +2164,7 @@ pub async fn query_conversions(
 
     // Track whether any non-sentinel conversions are found
     let mut conversions_found = false;
+
     for (addr, _denom, digit, epoch, amt) in conversions.values() {
         // If the user has specified any targets, then meet them
         // If we have a sentinel conversion, then skip printing
--- a/crates/apps_lib/src/client/rpc.rs
+++ b/crates/apps_lib/src/client/rpc.rs
@@ -6,6 +6,7 @@ use std::io;
 use color_eyre::owo_colors::OwoColorize;
 use data_encoding::HEXLOWER;
 use either::Either;
+use futures::StreamExt;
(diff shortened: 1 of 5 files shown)


brentstone:
@murisi - it was mentioned to me from @Fraccaman that we desire this work. Is this mergeable (I see the label that might suggest otherwise). Does it need review?

murisi:
> @murisi - it was mentioned to me from @Fraccaman that we desire this work. Is this mergeable (I see the label that might suggest otherwise). Does it need review?

@brentstone It's been a while since I've looked at this. I believe this PR did help when the mainnet clone's epochs got very high. I probably put the `do-not-merge` tag because these changes are not purely client-side; they change an RPC function. So there could be client-node version/compatibility issues arising. Yes, it needs a review and some testing to ensure that the PR is compatible with the latest releases of Namada, and that it has no adverse effects (due to having lots of smaller queries). The PR might also need to be touched up with a changelog entry.

brentstone:
@batconjurer would you mind peeking at this?

batconjurer (approved):

batconjurer:
This looks good to me. In the future, it would also be possible to speed up the server side processing by keying the conversions by epoch since right now the server still needs to do a linear scan over the whole conversion state. That is a much more invasive change though.

crates/apps_lib/src/client/rpc.rs
```diff
@@ -2143,16 +2143,28 @@ pub async fn query_conversions(
...
         .await
-        .expect("Conversions should be defined");
+        .expect("Unable to query current MASP epoch");
+    for epoch in MaspEpoch::iter_bounds_inclusive(from, to) {
+        conversions.append(
+            &mut rpc::query_conversions(context.client(), &epoch)
+                .await
+                .expect("Conversions should be defined"),
+        );
+    }
```
sug0, line 2159: there's no reason why we cannot run these queries concurrently (e.g. with [this](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered)). the `Stream` can be collected into a `BTreeMap` (see [`StreamExt::collect`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.collect))

sug0 (changes requested):
we can optimize conversion queries. right now, we need to wait for the full round-trip time of the conversions at a specific masp epoch. say we had 100 masp epochs to go through, and the RTT was 100 ms. we would need to wait 10 seconds for each query to be resolved. with 100 concurrent requests, ideally we would complete all queries in 100 ms. of course in practice they'll be slower, but there should be a major speed-up from launching these queries concurrently.

crates/apps_lib/src/client/rpc.rs
grarco: I've attempted this in 46ea724dccb9c87597bbfc50836064fb0dba7c1d even though I had to `fold` the maps instead of `collect` because of the limitations of the `Extend` trait on maps. Should we also add a couple of args for the query command to allow the user to specify the window boundary so that we can avoid querying from epoch 0 all the times?

grarco:
Force-pushed to rebase on main

sug0:
lgtm, just a small question

PR #4785, opened 2025-08-07
Improve shielded sync

sug0:
## Describe your changes

- Build commitment tree and witness map locally
   - Uses [`leanbridgetree`](https://github.com/namada-net/leanbridgetree) crate for this purpose
   - If the number of blocks to sync exceeds a certain threshold, we race with the local approach, by fetching the commitment tree, witness map and notes index in the background
     - Whatever finishes first is the data we store in the shielded wallet (which should have been identical, regardless of being built locally or fetched from the MASP indexer)
- Fix block index not being properly used
   - This is the major source of the speed-up in this PR
- Optimize the layout of the shielded wallet
   - Removed unnecessary fields
- Only fetch block index after a threshold of $N$ blocks (where it makes sense latency wise to wait for that network query vs. fetching $N$ potentially empty blocks)

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4785-bridge-tree-shielded-wallet.md
+++ b/.changelog/unreleased/improvements/4785-bridge-tree-shielded-wallet.md
@@ -1,0 +1,2 @@
+- Optimize shielded sync and shielded wallet layout.
+  ([\#4785](https://github.com/anoma/namada/pull/4785))


sug0:
~~this is currently crashing when calling `remove_mark` in integration tests~~

batconjurer (approved):

sug0:
https://github.com/namada-net/namada-masp-indexer/pull/94 and https://github.com/namada-net/namada-masp-indexer/pull/95 made this PR somewhat obsolete, in its current form.

benchmarks revealed that when syncing from scratch, performance is actually better using the older shielded sync client code; however, when syncing a smaller number of blocks, the latency of fetching the witness map tends to be greater than the compute workload of computing the merkle tree witnesses locally

therefore, the idea is to create a hybrid approach, that switches to the older strategy after a certain number of blocks, and keeps the new strategy for smaller wallet updates (to avoid incurring extra network latency)

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -492,25 +508,37 @@ where
...
 
-        // the latest block height which has been added to the witness Merkle
-        // tree
-        let last_witnessed_tx = self.ctx.note_index.keys().max().cloned();
+        if self.ctx.synced_height == BlockHeight(0)
+            && self.client.capabilities().may_fetch_pre_built_tree()
+            && !self.birthdays.is_empty()
+        {
+            self.ctx.synced_height = min_birthday;
+        }
```
sug0, line 541: this change is kinda sketchy. if for whatever reason we finish first with the local approach, we could end up with an invalid commitment tree that's missing a bunch of data (every note committed before `min_birthday`).

might need to think of an approach to safely keep this optimization, or remove it altogether

sug0:
rebased to (hopefully) fix CI timing out, and the flakiness in e2e tests

sug0:
> It seems that you are trying to fetch commitment trees and witness maps in the background, and if they don't finish in time you just rebuild them manually and cancel the fetch jobs.

Yeah pretty much. Not gonna lie, the documentation is a bit lacking. 😂

sug0:
adding back `do-not-merge` because the MASP dependency needs to point to `crates.io`, and the PR will probably need to be rebased again (CI is still broken)

brentstone:
> > It seems that you are trying to fetch commitment trees and witness maps in the background, and if they don't finish in time you just rebuild them manually and cancel the fetch jobs.
> 
> Yeah pretty much. Not gonna lie, the documentation is a bit lacking. 😂

Can you pls improve the documentation then @sug0? Thx

crates/shielded_token/src/masp/shielded_sync/dispatcher.rs
```diff
@@ -733,17 +818,122 @@ where
         }
     }
 
+    fn vk_synced_height(&self, vk: &ViewingKey) -> BlockHeight {
+        self.birthdays
```
grarco, line 822: Shouldn't we be looking at `ctx.synced_height` before the birthdays?

sug0:
@mergifyio requeue

PR #4790, opened 2025-08-14
Frontend MASP fee

grarco:
## Describe your changes

Possible implementation of masp frontend sustainability fees. The sdk has been extended to allow specifying an extra recipient for this fee.

For operations operating originating from the Namada chain this fee is pretty much free, any token can be used.

For incoming ibc shielding operations, the actual fee payment has to be encoded in the transparent vout of the Masp bundle (since the ibc recv packet limits us to a single recipient). This is just a commitment to the actual namada address so the format of the IBC packets has been changed to include not only the masp transaction but also this ancillary data. The WASM ibc transaction has been changed to execute this extra transparent transfer. On top of this, since the ibc recv packet also limits us to a single token, the fee has to be paid in the same token which gets shielded and nft transfers do not support this frontend fee.

Alternative approach:
- Encode the fee as a shielded output of the bundle. This would still limit us to a single asset but would require no change in the protocol/wasm transaction/ibc packets. On the downside, it generates notes with possibly little amounts that might be a burden to spend from the perspective of the frontend providers (due to the gas limits)

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/cli.rs
+++ b/crates/apps_lib/src/cli.rs
@@ -3548,6 +3548,9 @@ pub mod args {
     pub const FEE_PAYER_OPT: ArgOpt<WalletPublicKey> = arg_opt("gas-payer");
     pub const FILE_PATH: Arg<String> = arg("file");
     pub const FORCE: ArgFlag = flag("force");
+    #[cfg(any(test, feature = "testing"))]
+    pub const FRONTEND_SUS_FEE: ArgOpt<WalletAddress> =
+        arg_opt("frontend-sus-fee");
     pub const FULL_RESET: ArgFlag = flag("full-reset");
     pub const GAS_LIMIT: ArgDefault<GasLimit> = arg_default(
         "gas-limit",
@@ -4930,11 +4933,11 @@ pub mod args {
             ctx: &mut Context,
         ) -> Result<TxShieldingTransfer<SdkTypes>, Self::Error> {
             let tx = self.tx.to_sdk(ctx)?;
-            let mut data = vec![];
+            let mut sources = vec![];
             let chain_ctx = ctx.borrow_mut_chain_or_exit();
 
             for transfer_data in self.sources {
-                data.push(TxTransparentSource {
+                sources.push(TxTransparentSource {
                     source: chain_ctx.get(&transfer_data.source),
                     token: chain_ctx.get(&transfer_data.token),
                     amount: transfer_data.amount,
@@ -4949,13 +4952,19 @@ pub mod args {
                     amount: transfer_data.amount,
                 });
             }
+            let frontend_sus_fee =
+                self.frontend_sus_fee.map(|fee| TxTransparentTarget {
+                    target: chain_ctx.get(&fee.target),
+                    token: chain_ctx.get(&fee.token),
+                    amount: fee.amount,
+                });
 
             Ok(TxShieldingTransfer::<SdkTypes> {
                 tx,
-                sources: data,
+                sources,
                 targets,
                 tx_code_path: self.tx_code_path.to_path_buf(),
-                frontend_sus_fee: None,
+                frontend_sus_fee,
             })
         }
     }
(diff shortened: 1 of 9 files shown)


crates/sdk/src/tx.rs
```diff
@@ -2827,6 +2828,63 @@ pub async fn build_ibc_transfer(
...
+
+                transfer = transfer
+                    .transfer(
+                        source.to_owned(),
+                        target.to_owned(),
+                        token.to_owned(),
+                        validated_amount,
+                    )
+                    .ok_or(Error::Other(
+                        "Combined transfer overflows".to_string(),
```
batconjurer, line 2866: Will this catch the case that both the sus fee + raw transfer amount together exceed source's balance?

crates/sdk/src/tx.rs
```diff
@@ -3524,6 +3582,38 @@ pub async fn build_shielding_transfer<N: Namada>(
...
+            context,
+            sus_fee_amt.to_owned(),
+            sus_fee_token,
+            args.tx.force,
+        )
+        .await?;
+
+        // Decrease the amount to shield to account for this frontend fee, take
+        // it from the first shielding input that matches the fee token
+        for (_, token, amount) in &mut transfer_data.sources {
```
batconjurer, line 3602: hold up. So we are diverting some of the shielding funds to the fee? How did it get bundled in there when it is explicitly fed in as a separate amount. Does it simplify balance checks? Does that mean the input shielded amount is supposed to `intended_amount + fee`? with the understanding that the fee gets subtracted off?

crates/sdk/src/tx.rs
```diff
@@ -4149,6 +4239,22 @@ pub async fn gen_ibc_shielding_transfer<N: Namada>(
...
         .await;
 
+    let extra_recipient = match &args.frontend_sus_fee {
+        Some((target, amount)) => {
+            // Validate the amount given
+            let validated_fee_amount =
+                validate_amount(context, amount.to_owned(), &token, false)
+                    .await?;
+
+            Some(MaspFrontendSusFee {
```
batconjurer, line 4249: do we do a balance check for this?

crates/tests/src/integration/masp.rs
```diff
@@ -9365,3 +9365,180 @@ fn multiple_inputs_from_single_note() -> Result<()> {
 
     Ok(())
 }
+
+// Test that shielding and unshielding transactions can pay a small fee to a
+// transparent address as a form of sustainability fee for frontend providers
+#[test]
+fn frontend_sus_fee() -> Result<()> {
```
batconjurer, line 9372: I don't think the code is adequately checking that all of the debits added are less than the user's balance. That check should be implemented by the client and tested here as well.

crates/apps_lib/src/cli.rs
```diff
@@ -5001,6 +5055,10 @@ pub mod args {
                         .def()
                         .help(wrap!("The amount to transfer in decimal.")),
                 )
+                .arg(__TEST_FRONTEND_SUS_FEE.def().help(wrap!(
```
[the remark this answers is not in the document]
grarco: Yes we still need to expose them to run the integration and e2e tests

crates/tests/src/integration/masp.rs
grarco: Done in 56418fe8dc0dd1668fcd9c17b31216b7ff7e65a1

crates/apps_lib/src/cli.rs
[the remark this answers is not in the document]
batconjurer: Is there a clever way to keep this from showing up in the cli help menu?

crates/sdk/src/tx.rs
```diff
@@ -2827,6 +2827,84 @@ pub async fn build_ibc_transfer(
         masp_fee_data
     };
 
+    if let Some((target, percentage)) = &args.frontend_sus_fee {
+        match (&source, &args.ibc_shielding_data) {
+            (&MASP, None) => {
+                // NOTE: The frontend fee should NOT account for the masp fee
+                // payment amount
+                let sus_fee_amt = namada_token::Amount::from_uint(
+                    validated_amount
```
batconjurer, line 2836: this block of code for the percentage calculation isn't nice looking and appears in a few spots. Can we hide it in the amount module?

batconjurer (changes requested):
nits

crates/sdk/src/tx.rs
grarco: Done in 3fe86a7b5e3de149a84b21ab83fa475161d237e2, but the new function does also the validation part and lives in the same module

crates/apps_lib/src/cli.rs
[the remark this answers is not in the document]
grarco: Hid them in e90c61992b0166596e18323c1f6fb88886969c17

batconjurer (approved):

PR #4799, opened 2025-09-06
safer wasm cache key

tzemanovic:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


sug0 (approved):
lgtm

here's an alternative approach that works even if no `.git` repo is present: you compute a hash digest of the entire project directory's contents. this way, you always get a unique identifier for your cache keys, if the source code ever changes

tzemanovic:
> lgtm
> 
> here's an alternative approach that works even if no `.git` repo is present: you compute a hash digest of the entire project directory's contents. this way, you always get a unique identifier for your cache keys, if the source code ever changes

I think only time `.git` is not present is when we're publishing to crates.io, for which we'll have a unique pkg version so the fallback should handle it

PR #4816, opened 2025-09-10
Refactor SDK args and sigs

grarco:
## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:


crates/sdk/src/signing.rs
```diff
@@ -405,45 +461,101 @@ pub async fn aux_signing_data(
...
+        },
+        _ => {
+            if let Some(pubkey) = public_keys.first() {
+                FeeAuthorization::Signer {
+                    pubkey: pubkey.to_owned(),
+                    disposable_fee_payer: false,
+                }
+                // NOTE: the disposable gas_payer is always overridden by
+                // signing inputs
+            } else if disposable_gas_payer {
```
grarco, line 490: Technically if a disposable gas payer is required no signer should be provided since it's not needed, grows the size of the tx (and the gas required) and might leak information. Still, I've maintained the same logic we've had in place for a while for the wrapper signer: if the user specifically passes a signing key we still produce a signature for that (i.e. the user is still in control even though we try to guide them)

grarco:
@tzemanovic I also fixed a bug for which we were attaching empty auth sections to the transactions (you can see it in de7eda50523ab0f17d4ee16fb57af9d1d7fcd559). That's what was making the e2e tests fail (unless something else comes up in this CI run)

tzemanovic:
> @tzemanovic I also fixed a bug for which we were attaching empty auth sections to the transactions (you can see it in [de7eda5](https://github.com/namada-net/namada/commit/de7eda50523ab0f17d4ee16fb57af9d1d7fcd559)). That's what was making the e2e tests fail (unless something else comes up in this CI run)

thx! I'm just merging https://github.com/namada-net/namada/pull/4832 which I think will introduce some conflicts if you don't mind rebasing this after

grarco:
Yes go ahead, I need to fix some more broken tests in the meantime

grarco:
@tzemanovic I think this is read for a final re-review. Mind that I changed the following:

- the logic of `inner_tx_signer` has been reverted to the previous one where the public keys take precedence over the `default_signer` instead of the two being merged together
- `build_batch` has been updated to allow batching txs that are already batch (it was previously limited to a single inner transaction)

tzemanovic:
I've just added a tiny commit to fix the CI changelog check

PR #4817, opened 2025-09-10
Change MASP zk backend to bellperson

batconjurer:
## Describe your changes
The masp now uses the bellperson backend to synthesize circuits and build zk proofs. Benchmarking this indicates substantial speed ups in proving time.

### DIFF
--- a/.changelog/unreleased/improvements/4817-change-masp-zk-backend.md
+++ b/.changelog/unreleased/improvements/4817-change-masp-zk-backend.md
@@ -1,0 +1,3 @@
+- The masp now uses the bellperson backend to synthesize circuits and build zk
+  proofs. Benchmarking this indicates substantial speed ups in proving time.
+  ([\#4817](https://github.com/namada-net/namada/pull/4817))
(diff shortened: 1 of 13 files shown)


sug0:
since this PR changes the backend globally (i.e. not just client-side, but also validator-side), it would be nice to test it by validating proofs produced with `bellman`, e.g. by running a fullnode with the new `bellperson` backend against some network that had committed masp txs with proofs still produced with `bellman`

batconjurer:
> since this PR changes the backend globally (i.e. not just client-side, but also validator-side), it would be nice to test it by validating proofs produced with `bellman`, e.g. by running a fullnode with the new `bellperson` backend against some network that had committed masp txs with proofs still produced with `bellman`

So I checked locally that shielding, shielded transfer, and unshielding all works when a devnet uses bellperson but the client is built from main

brentstone:
@tzemanovic @batconjurer I think it might be worth cutting a path release v201.0.6 that also just has this alone on top of v201.0.5. Thoughts?

Fraccaman:
since this PR changes CI with a new image version I think we could first upgrade CI, rebase this one and then merge it. Im going to open a PR shortly.

PR #4832, opened 2025-09-23
Dry run with mock sigs

### LINKED ISSUE #4714: allow to dry-run txs with fake signatures
Add a new dry-run endpoint or option which allows to estimate gas cost without proper signatures. This can be done by patching the source accounts in the temp storage used for dry-run with VPs that accept fake signatures

batconjurer:
## Describe your changes
Closes #4714 
 - When dry-running, dummy signatures are used in txs
 - Validation does not perform signature checks when dry-running

I have tested that the gas estimation hasn't changed between dry-running with signatures and with dummies. It is curious however that in both cases there is a small discrepancy between dry-running and the actual tx. 

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/improvements/4832-dry-run-mock-sigs.md
+++ b/.changelog/unreleased/improvements/4832-dry-run-mock-sigs.md
@@ -1,0 +1,8 @@
+- Closes #4714
+  - When dry-running, dummy signatures are used in
+    txs
+  - Validation does not perform signature checks when dry-running
+  
+I have tested that the gas estimation hasn't changed between dry-running
+with signatures and with dummies. It is curious however that in both cases there is a small discrepancy between dry-running and the actual tx.
+([\#4714](https://github.com/namada-net/namada/issues/4714))


tzemanovic:
The changes look great but I think we'll need to find where the discrepancy comes from as we need this feature for more accurate gas estimation

batconjurer:
> The changes look great but I think we'll need to find where the discrepancy comes from as we need this feature for more accurate gas estimation

I'm not sure how this feature is supposed to give more accurate gas estimation unless you mean that we can now dry-run txs that require a hardware wallet to sign. In any case, this discrepancy existed before these changes and it seems from my limited testing that these changes preserve the values for both the dry-run and actual run. The difference is like 100 - 200 gas units, and the dry-runs are higher than the actual.

tzemanovic:
> > The changes look great but I think we'll need to find where the discrepancy comes from as we need this feature for more accurate gas estimation
> 
> I'm not sure how this feature is supposed to give more accurate gas estimation unless you mean that we can now dry-run txs that require a hardware wallet to sign. In any case, this discrepancy existed before these changes and it seems from my limited testing that these changes preserve the values for both the dry-run and actual run. The difference is like 100 - 200 gas units, and the dry-runs are higher than the actual.

Ah, it's aimed for the interface where we need to estimate gas cost before signing, mostly for the txs that have very variable gas cost depending on the execution path (e.g. PoS redelegation). But if the dry-run estimates are a bit higher I think that's acceptable

tzemanovic:
@mergifyio requeue

tzemanovic:
@mergifyio requeue

tzemanovic:
@mergifyio requeue

Fraccaman:
@mergifyio requeue

tzemanovic:
@mergifyio requeue

Fraccaman:
@Mergifyio requeue

tzemanovic:
I'm backporting this to libs as https://github.com/namada-net/namada/pull/4816 depends on it for some mocking fns

PR #4883, opened 2025-10-01
Skip conversions with no rewards

### LINKED ISSUE #4789: Avoid 0 rewards conversions in MASP transactions
When building MASP txs we convert the assets of the user to their latest available counterpart. There's a chance though, that the applied conversion doesn't bring in any rewards but just converts the asset to the latest masp epoch (this could happen if no rewards had been produced for that specific asset).

In this case, the  builder will include the conversion but this is technically useless, as it brings nothing to the involved amounts and only takes up resources, namely space in the transaction (which might also turn into possible issues with the hardware wallet) and gas to verify these conversions (both to verify their anchors and for the convert circuit).

We could try to modify `apply_conversion` to remove such conversion from `Conversions` if that's the case

cc: @mateuszjasiuk

grarco:
## Describe your changes

Closes #4789.

Modifies `apply_conversion` to skip conversions that don't carry any rewards (i.e. conversions that only update the epoch of the asset). 

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/SDK/4789-no-zero-reward-convs.md
+++ b/.changelog/unreleased/SDK/4789-no-zero-reward-convs.md
@@ -1,0 +1,2 @@
+- Updated the transaction building process to avoid using MASP conversions with
+  no rewards. ([\#4789](https://github.com/namada-net/namada/issues/4789))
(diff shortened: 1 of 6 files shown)


grarco:
Putting this back to draft as it needs some more work

grarco:
@sug0 I've fixed a bug and improved the tests, this is now ready for review

PR #4939, opened 2025-10-22
Handle IBC deposits/withdraws to/from payment addresses

### LINKED ISSUE #4168: Shielded refund design
Currently, Namada does not support the shielded refunding of some tx. However, it may be possible to do so, with the following:

1. The user optionally specifies a shielded refund MASP transaction option via the CLI/SDK.
2. The CLI/SDK generates a shielding transaction associated with the funds unshielded via IBC.
3. The protocol intercepts the ICS-20 `send_transfer_execute` handler from `ibc-rs`, and stores the generated shielding transaction associated with this `port`, `channel` and `sequence-number` triplet.
4. Upon receiving an ICS-20 acknowledgement, the protocol can intercept the `on_acknowledgement_packet_execute` handler from `ibc-rs`, and retrieve the shielding tx associated with the `port`, `channel` and `sequence-number` triplet retrieved from the packet.
  1. In case the ack is a `success`, the shielding transfer is deleted from storage with no further action.
  2. Otherwise, the shielding tx's state changes are applied, to reshield the transferred funds.

### LINKED ISSUE #4830: Blind IBC to znam MASP depositing (memoless)
I was speaking with @sug0, who believes that there's a path to send an IBC asset directly to a znam without needing proofs in the memo field. If I understand correctly, this feature would be a very powerful feature for Borderless Private USDC user experience because it would eliminate the need to generate proofs on the deposit side of the app. Currently the user must deposit USDC to a tnam and then they must shield afterwards, and it must be a new tnam each time.

sug0:
## Describe your changes

Closes #4830
Closes #4168

Allow IBC deposits into payment addresses (without generating MASP transactions), and IBC withdraws from payment addresses (to automatically get refunded, if an IBC unshielding transfer fails on the counter-party).

These changes should substantially improve the UX of MASP over IBC.

## Checklist before merging 
    - Relevant PR if applies: 
- [x] If this PR affects services such as `namada-indexer` or `namada-masp-indexer`, a corresponding PR is opened in that repo
    - Relevant PR if applies: https://github.com/namada-net/namada-masp-indexer/pull/111

### DIFF
--- a/.changelog/unreleased/improvements/4939-znam-ibc-shielding.md
+++ b/.changelog/unreleased/improvements/4939-znam-ibc-shielding.md
@@ -1,0 +1,4 @@
+- Allow IBC deposits into payment addresses (without generating MASP
+  transactions), and IBC withdraws from payment addresses (to automatically
+  get refunded, if an IBC unshielding transfer fails on the counter-party).
+  ([\#4939](https://github.com/namada-net/namada/pull/4939))


crates/vm/src/wasm/host_env.rs
```diff
@@ -251,6 +251,7 @@ where
             "namada_tx_write" => Function::new_typed_with_env(wasm_store, &env, wrap_tx::_4(host_env::tx_write)),
             "namada_tx_write_temp" => Function::new_typed_with_env(wasm_store, &env, wrap_tx::_4(host_env::tx_write_temp)),
             "namada_tx_yield_value" => Function::new_typed_with_env(wasm_store, &env, wrap_tx::_2(host_env::tx_yield_value)),
+            "namada_tx_has_conversion" => Function::new_typed_with_env(wasm_store, &env, wrap_tx::_2(host_env::tx_has_conversion)),
```
tzemanovic: there's probably no way around this, but this will necessitate update of all tx and vp wasms (i.e. migration)

yito88 (approved):
Great work! Thank you!

sug0:
sorry about the f-push, had to fix shielded swaps

crates/state/src/in_memory.rs
```diff
@@ -334,3 +335,173 @@ where
...
+
+                Ok((value, checked!(gas * MEMORY_ACCESS_GAS_PER_BYTE)?.into()))
+            }
+            // invalid request
+            _ => Err(invalid_key()),
+        }
+    }
+
+    /// Key used to query the conversion state of a given [`AssetType`].
+    pub fn has_conversion_key(asset_type: &AssetType) -> namada_storage::Key {
```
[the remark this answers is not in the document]
sug0: the naming might be a bit odd, but the idea behind `has_conversion_key` is that the retrieved value is a boolean, not a conversion

crates/shielded_token/src/vp.rs
```diff
@@ -128,8 +226,12 @@ where
     fn valid_nullifiers_reveal(
         ctx: &'ctx CTX,
         keys_changed: &BTreeSet<Key>,
-        transaction: &Transaction,
+        masp_source: &MaspSource,
     ) -> Result<()> {
+        let MaspSource::UserTx(transaction) = masp_source else {
```
grarco: I believe here we are making an assumption on the correct behavior of the wasm transaction. As it stands right now this early exit is correct but in case of a buggy or malicious tx that reveals unnecessary nullifiers we would skip the check at line 274 and accept an invalid tx

sug0: The masp vp was already assuming some collaboration from wasm txs, by means of the actions wasm protocol, which is used to inform the masp vp of shielded tx section references, and bundle signers. I definitely agree with you, though, just pointing out this kind of flaw isn't new. In any case, the chances of a nullifier collision due to this are pretty low, $1/(2^{32 \times 8} - 1)$. I'd say relying on wasm whitelisting for some degree of security is fine.

crates/ibc/src/storage.rs
```diff
@@ -677,3 +678,41 @@ pub fn inflight_packet_key(inflight_packet_key: &InFlightPacketKey) -> Key {
...
+pub fn write_shielding_counter<S: StorageWrite>(
+    storage: &mut S,
+    counter: u64,
+) -> Result<()> {
+    storage.write(&masp_shielding_counter_key(), counter)
+}
+
+/// Increment the IBC shielding counter in storage,
+/// and return its previous value.
+pub fn fetch_and_increment_shielding_counter<S: StorageRead + StorageWrite>(
```
grarco: This seems to be unused

crates/ibc/src/context/token_transfer.rs
```diff
@@ -170,34 +183,252 @@ where
...
+                position: digit,
+                epoch: epoched_asset,
+            };
+            let asset_type = asset.encode().map_err(|_| HostError::Other {
+                description: format!(
+                    "Failed to create asset type of IBC shielding: {asset:?}"
+                ),
+            })?;
+
+            let rseed = namada_core::hash::Hash::sha256(format!(
```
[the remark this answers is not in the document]
sug0: this is basically the note's source of randomness, when fed to the zk proof. it's important `rseed` is unique, to yield unique nullifers every time

sug0:
oops, I broke something between the last two f-pushes

murisi (approved):
Thanks for clarifying the context of this code. I see nothing obviously wrong or improvable with the changes to the MASP VP or the transfer context. I just wonder if we could utilize existing code and execution paths more in order to minimize risk. For instance, the transfer context has code to update the note commitment tree and the undated balances though `apply_shielded_transfer` does something similar. In the MASP VP, the new logic skips the anchor check execution paths when the `ProtocolIbcShielding` condition is met.

I guess I just wonder if there could be a single function to create a dummy transaction based off `(owner_pa, token, amount, ...)` (potentially having correct anchors) that can be:
* constructed (from the event) and used by clients to update their shielded state
* constructed and used by the transfer context to update undated balances and note commitment tree (somehow using the `apply_shielded_transfer`  function)
* constructed by the MASP VP (from the event) and fed into the existing verification logic in a way that minimizes the number of checks that are disabled/the number of paths through the verification logic.

A unified dummy transaction logic would ensure that IBC transactions are synchronized, executed, and verified in the exactly same way. Overall the code is likely correct, but it would be nice to leverage existing mainnet code as much as possible to minimize risk of unexpected behaviour. Of course, this hypothetical approach might not be practical!

grarco (approved):

sug0:
@mergifyio requeue

PR #4953, opened 2025-11-03
Remove speculative context

### LINKED ISSUE #4074: Evaluate removal of the speculative shielded context
Follows #4019.

We could consider removing the pre-caching mechanism of the shielded wallet (a.k.a. the speculative shielded context) and go back to calling `shielded-sync` in the client automatically before any masp-related command (balance query, shielded tx, unshielding tx). The reasons are:

- Poor UX: in general both being in a speculative context or on an old confirmed context leads to a bad experience for the user who could see outdated balances tor find themselves unable to produce valid transactions. The current solution based on warning messages and the need to manually call `shielded-sync` all the times is a bit tedious
- Possible information leaks: being on a speculative (or in general, old state) can leak some information when spending a key
- Performance of the `shielded-sync` command should be good now and should not be a concern in terms of UX

There's only one possible catch:

- In the docs (https://docs.namada.net/users/fees#masp-fee-payment-gas-limit) we suggest a way t circumvent a limitation when paying fees via the MASP that would require building a batch of masp transactions. This is not currently supported by the client (since we only support batching from the SDK), but if we wanted it we'd need to apply some logic to invalidate spent notes just like we do in the speculative context: this could be done in a separate command though, without the need to have a special type of context that applies to all commands

grarco:
## Describe your changes

Closes #4074.

Removes the speculative shielded context and the `shielded-sync` cli command. The shielded-sync operations is now embedded directly in the cli commands that need it:

- Shielded transfer
- Unshielding transfer
- IBC transfer (when the source is shielded)
- Balance query (for shielded owners)
- Shielded rewards estimate

## Checklist before merging 
    - Relevant PR if applies: https://github.com/namada-net/namada-docs/pull/459 
    - Relevant PR if applies:

### DIFF
--- a/crates/apps_lib/src/cli/client.rs
+++ b/crates/apps_lib/src/cli/client.rs
@@ -3,8 +3,10 @@ use std::io::Read;
 use color_eyre::eyre::Result;
 use namada_sdk::io::{Io, NamadaIo, display_line};
 use namada_sdk::masp::ShieldedContext;
-use namada_sdk::{Namada, NamadaImpl};
+use namada_sdk::masp::fs::FsShieldedUtils;
+use namada_sdk::{Namada, NamadaImpl, ShieldedWallet};
 use namada_wallet::DatedViewingKey;
+use tokio::signal::unix::SignalKind;
 
 use crate::cli;
 use crate::cli::api::{CliApi, CliClient};
@@ -97,8 +99,8 @@ impl CliApi {
                         )
                         .expect("Missing required shielded-sync arguments");
                         sync_args.viewing_keys.extend(extra_sync_vks);
-                        let shielded = crate::client::masp::syncing(
-                            ShieldedContext::new(shielded),
+                        let shielded = Self::scoped_shielded_sync(
+                            shielded,
                             client.clone(),
                             sync_args,
                             &io,
@@ -155,8 +157,8 @@ impl CliApi {
                         )
                         .expect("Missing required shielded-sync arguments");
                         sync_args.viewing_keys.extend(extra_sync_vks);
-                        let shielded = crate::client::masp::syncing(
-                            ShieldedContext::new(shielded),
+                        let shielded = Self::scoped_shielded_sync(
+                            shielded,
                             client.clone(),
                             sync_args,
                             &io,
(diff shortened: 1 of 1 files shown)


crates/apps_lib/src/cli.rs
```diff
@@ -5164,6 +5148,8 @@ pub mod args {
             ctx: &mut Context,
         ) -> Result<TxIbcTransfer<SdkTypes>, Self::Error> {
             let tx = self.tx.to_sdk(ctx)?;
+            let shielded_sync =
+                self.shielded_sync.map(|sync| sync.to_sdk(ctx).unwrap());
```
sug0: I think these shielded sync args are missing in `TxOsmosisSwap` and `GenIbcShieldingTransfer`

sug0:
oh one thing this might "break" is ctrl+c after running the implicit shielded sync, because it installs a SIGINT handler that overrides the default handler. if an rpc is hanging, we might not be able to ctrl+c out of namadac

grarco:
> oh one thing this might "break" is ctrl+c after running the implicit shielded sync, because it installs a SIGINT handler that overrides the default handler. if an rpc is hanging, we might not be able to ctrl+c out of namadac

Right, I think we could fix this in two ways:

1. We could launch the implicit shielded-sync function in a child process and wait on it. The custom handler for SIGINT should be relative to the child only and rpcs shouldn't block the `submit_*` functions anymore
2. We just uninstall the custom handler after we are done with the sync process

I've added the `do-not-merge` label for now until we have a fix for this

sug0:
@grarco I kinda like the first option, despite it being a bit hacky; from `namada_apps_lib`, we can assume we are executing one of our binaries, so the first value of `std::env::args()` will give you the executable path (either `namada` or `namadac`), from which we should be able to invoke `shielded-sync`

grarco:
@sug0 attempted the change in d72523dc85bdb5f1c44cf2e660b29483d8027d6e

grarco:
> @sug0 attempted the change in [d72523d](https://github.com/namada-net/namada/commit/d72523dc85bdb5f1c44cf2e660b29483d8027d6e)

Never mind, the previous solution with the child process was giving me problems when it came to tests. I've now tried to override the custom signal handlers after we are done with the shielded-sync command. It might be slightly hacky but it's easier and faster to implement

PR #4994, opened 2025-12-05
Public keys when mocking signatures

### LINKED ISSUE #4989: No private keys for mock signatures
We have introduced mock signature generation (and verification) for dry runs in #4832. To minimize code duplication in the SDK we used the same interface of `sign_wrapper` for `mock_sign_wrapper` and the same interface of `sign_raw` for `mock`. Both these mock functions still require a private key as a parameter, which makes little sense given that they won't produce an actual signature. This causes problems with Namadillo since it needs to dry-run transactions (with mock signatures) without interacting with the extension. We should change these mock functions to require public keys instead of private ones.

cc: @mateuszjasiuk

grarco:
## Describe your changes

Closes #4989.

Changes the signature functions to require a public key instead of a private one when mocking a signature. Introduces a new `SigOrPubKey` trait to share the behavior of public and private keys.

## Checklist before merging 
    - Relevant PR if applies: 
    - Relevant PR if applies:

### DIFF
--- a/.changelog/unreleased/SDK/4994-pubkeys-for-mock-sigs.md
+++ b/.changelog/unreleased/SDK/4994-pubkeys-for-mock-sigs.md
@@ -1,0 +1,3 @@
+- Changed the SDK signing functions to accept a public key instead of a
+  private one when mocking a signature. ([\#4994](https://github.com/namada-
+  net/namada/pull/4994))


grarco:
@mergifyio requeue

grarco:
@mergifyio requeue

PR #5035, opened 2026-09-04
Add bounded inactive UltraHonk profile codecs

### LINKED ISSUE #5034: Add staged consensus-native UltraHonk verifier profile support
## Problem

ZKA bridge validity predicates need a consensus-native way to verify the exact UltraHonk/BN254/Keccak proofs produced by their pinned Noir/Barretenberg pipeline. Namada currently has no corresponding verifier host import, authenticated verifier-profile registry, deterministic resource schedule, or activation lifecycle, so the production path must remain fail-closed.

ZKA ADR-0009 specifies the intended trust boundary and wire formats. Its reserved `zka-ultrahonk-bn254-keccak-zk-bb5-v1` feasibility profile is permanently non-admissible because the production trusted-setup artifacts and review dispositions were never assigned. Current export and transfer artifacts also produce 7,616-byte proofs, above ABI v1's fixed 7,232-byte ceiling, so they require a separately versioned successor ABI/profile rather than changing v1 in place.

Reference: https://github.com/pyramidal-io/ZKA.dev/blob/main/docs/adr/0009-namada-consensus-native-ultrahonk-verification-bou.md

## Proposed staged work

1. Add total, bounded codecs for the ADR-0009 profile-binding, three-row allowlist, trusted-setup provenance, and inactive-selection records. This stage is deliberately non-activating and retains v1's exact 7,232-byte ceiling.
2. Extract and audit a verifier-only UltraHonk BN254/Keccak core with a panic-free canonical parser and reproducible dependency/source lock.
3. Add the root-frame-authenticated VP host ABI and bind calls to host-derived chain, VP code, protocol-version, and statement-domain context.
4. Add an immutable governance registry and fail-closed activation/rotation lifecycle over complete, profile-selected rows and inline authenticated verifier CRS bytes.
5. Prove deterministic gas and native memory bounds, then publish cross-tool positive and adversarial KATs across pinned `bb -t evm`, the Rust host verifier, and generated Solidity verifiers.
6. Wire the real ZKA bridge VP path, remove the transfer fallback only after all positive paths are real, and use a two-phase protocol activation.

Each stage should be an atomic, reviewable PR. The first implementation stage is ready locally with an authoritative byte-for-byte ADR golden vector, exhaustive truncation/trailing/mutation coverage, fixed allocation, and no storage, governance, verifier, or activation API.

## Security invariants

- The reserved v1 profile remains inactive and non-admissible.
- No caller bytes, local file, cache, download, implementation default, or ambient network source may substitute for protocol-published, authenticated verifier artifacts.
- A selected profile binds complete circuit rows, exact lengths/encodings, all artifact and schema digests, inline canonical G2 CRS bytes, named provenance, and resource/gas versions.
- Caller proof/input parse failures and false equations are transaction-invalid; malformed referenced authenticated verifier state is validator-fatal before vote or commit.
- Parsing, allocation, gas, and results are deterministic across supported architectures.
- Activation occurs only after exact-artifact cryptographic, trusted-setup, codec, runtime, gas, governance, and state-machine reviews.

## Maintainer input requested

- Is `crates/parameters` the preferred home for the non-activating wire codecs and inactive protocol selection record?
- Which future Namada protocol version/release should own the successor host ABI and activation gate?
- Should the verifier core live in-tree or as a separately audited, pinned crate?

beddows:
## Summary

- adds total, fixed-layout codecs for the ADR-0009 UltraHonk v1
  profile-binding, allowlist, trusted-setup provenance, and inactive-selection
  records in `crates/parameters`
- binds the implementation to the authoritative byte-for-byte golden vector
  and exercises malformed, truncated, trailing, and mutated inputs
- adds the required unreleased changelog fragment

This is deliberately non-activating plumbing only. It does **not** add a
verifier, verifier host import, storage or governance API, protocol selection,
gas schedule, CRS distribution, bridge integration, or an activation path.
The reserved v1 profile remains inactive and retains its exact 7,232-byte proof
ceiling.

Closes #5034

Refs: req_namada_ultrahonk_publish_an_admissible_s_471cf9c0

## Verification

- `env PROTOC=/private/tmp/zka-protoc-35.0.LhW9Ic/bin/protoc rustup run 1.85.0 cargo test -p namada_parameters --lib --offline` (25 passed)
- `cargo fmt --all -- --check`
- `SHA=$(git rev-parse HEAD) bash .github/workflows/scripts/check-changelog.sh`

### DIFF
--- a/.changelog/unreleased/features/5034-ultrahonk-profile-codecs.md
+++ b/.changelog/unreleased/features/5034-ultrahonk-profile-codecs.md
@@ -1,0 +1,4 @@
+- Added bounded, inactive protocol-profile codecs to support a future audited
+  UltraHonk verifier integration, while keeping verifier execution and
+  activation unavailable.
+  ([\#5034](https://github.com/namada-net/namada/issues/5034))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6258,6 +6258,7 @@ dependencies = [
  "namada_systems",
  "namada_tx",
  "namada_vp_env",
+ "sha2 0.10.8",
  "smooth-operator",
  "thiserror 2.0.12",
 ]
--- a/crates/parameters/Cargo.toml
+++ b/crates/parameters/Cargo.toml
@@ -25,6 +25,7 @@ namada_systems.workspace = true
 namada_tx.workspace = true
 namada_vp_env.workspace = true
 
+sha2.workspace = true
 smooth-operator.workspace = true
 thiserror.workspace = true
 
--- a/crates/parameters/src/lib.rs
+++ b/crates/parameters/src/lib.rs
@@ -18,6 +18,7 @@
 )]
 
 pub mod storage;
+pub mod ultrahonk_profile;
 pub mod vp;
 mod wasm_allowlist;
 use std::collections::BTreeMap;
--- a/crates/parameters/src/ultrahonk_profile.rs
+++ b/crates/parameters/src/ultrahonk_profile.rs
@@ -1,0 +1,1666 @@
+//! Bounded wire codecs for ZKA UltraHonk v1 circuit profiles.
+//!
+//! This module implements the fixed-layout profile records specified by ZKA
+//! ADR-0009. It is deliberately non-activating: it does not define a canonical
+//! production profile identifier, discriminator, allowlist, CRS, gas schedule,
+//! resource schedule, storage key, or governance/activation API. A protocol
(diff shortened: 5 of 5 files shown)


beddows:
Withdrawing this PR. The specification and golden vector it implements are not publicly available yet, and the codecs are specific to the ZKA profile rather than a general Namada capability, so this work belongs downstream for now. Thanks for your time.

